Issue #69, opened 2020-11-27
Deprecate shop/skate & shop/skateboards in favor of shop/sports with sport/skateboard


tas50:
I've noticed a good number of nodes tagged as shop/skate & shop/skateboards in the US. Since these are just specialty sport stores it seems like they should be tagged as shop/sports with sport/skateboard instead.

matkoniecz:
Note that https://wiki.openstreetmap.org/wiki/Tag:shop%3Dskate is not describe as deprecated. Have you discussed it anywhere (tagging mailing list or some similar forum?)

tiptoptom:
Looks like it's not often used:
https://taghistory.raifer.tech/#***/shop/skate&***/shop/skateboards

See also:
https://github.com/openstreetmap/id-tagging-schema/blob/main/GUIDELINES.md#tag-updates-and-additions

tordans:
OK, lets close this for now…

PR #70, opened 2020-11-27
Deprecate shop/skate & shop/skateboard for shop/sports with skateboard

tas50:
Correctly tag these as specialty sports shops.

Fixes https://github.com/openstreetmap/id-tagging-schema/issues/69

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1391,6 +1391,14 @@
     "replace": {"office": "estate_agent"}
   },
   {
+    "old": {"shop": "skate"},
+    "replace": {"shop": "sports", "sport": "skateboard"}
+  },
+  {
+    "old": {"shop": "skateboard"},
+    "replace": {"shop": "sports", "sport": "skateboard"}
+  },
+  {
     "old": {"shop": "telecommunications"},
     "replace": {"shop": "telecommunication"}
   },


tas50:
This is rebased now to resolve the conflict

matkoniecz:
Note that https://wiki.openstreetmap.org/wiki/Tag:shop%3Dskate is not describe as deprecated. Have you discussed it anywhere (tagging mailing list or some similar forum?)

TurnrDev:
It only had 24 uses 😮

Adamant36:
Unfortunately there isn't really a good tag for skateboard shops right now. If I remember correctly like 5 years ago there same uses of shop=skateboard that were mass edited into shop=board + skateboard:sales=yes or something like that. If you could even call it a mass edit since there wasn't that many instances of shop=skateboard to start with from what I remember. Later there was some brewhaha about shop=board being to ambiguous and people didn't like the mass editing involved in it. In the interim and unware of the other factors at the time I started using shop=skate. I think there was like 5 uses and no Wikipedia article at that point. I seem to remember looking into it and there either wasn't any uses of shop=sport + sport=skateboard or the usage was extremely low. 

Just as a side to that, I personally have zero opinion on what is the better tag outside of that shop=board + skateboard:sales=yes is just inane and stupid. Although post hock dumping every single thing even slightly related to sports into some kind of shop=sports + sports=whatever tagging scheme isn't much better. Especially in cases like this where that was never the tagging scheme being used. Otherwise you could justify really lame tagging nonsense like using shop=sports + sports=baking for baking shops or whatever just because there's baking competitions. More on topic, places that sell bicycles are tagged as shop=bicycle, not shop=sports + sports=bicycle for a reason. The same reason that, IMO, would apply here for why skateboard shops shouldn't be tagged as sports shops "just because."

Adamant36:
I don't really care about it one or another at thos point, but there really should have a wider discussion about then was allowed for here if shop=skate is going to be automatically replaced with sport=skateboard since at least of the usage is for places that are "skate" shops, not "skateboard" shops and that was a big benefit to shop=skate in the first place. Otherwise whats the option now sport=skateboard;rollerblade;blah;blah;blah or some sales name space nonsense? At the end of the day shop=sports only makes sense when it's a general sporting goods store that doesn't need sports=whatever tag added it or your just losing important information. The change should have been discussed on the mailing list or forum before it was implemented regardless though. Don't be supprised if whatever "upgrading" this does gets reverted or new uses of it are retagged to shop=skate.

Issue #39, opened 2020-11-09
Update shop/mattress to be shop/bed


tas50:
A lot of people have incorrectly tagged bedding stores as shop/mattress. We should update these to be shop/bed.

PR #71, opened 2020-11-27
Deprecate shop/mattress for shop/bed

tas50:
Mattress stores should be tagged as bed stores.

Fixes https://github.com/openstreetmap/id-tagging-schema/issues/39


quincylvania:
@tas50 Thanks! There are only 64 uses of `shop=mattress` currently, so it's likely these can be treated as synonymous.

PR #73, opened 2020-11-29
Add parking=street_side

jdhoek:
Adds a preset for [street-side parking](https://wiki.openstreetmap.org/wiki/Proposed_features/parking%3Dstreet_side).

The proposal for `parking=street_side` has been approved, and the tag is in use. This PR adds the preset for the explicitly mapped variant. The implicitly mapped variant uses the `parking:lane` tag on the highway itself instead of mapping the areas separately and will be offered via a separate PR later.

---

This seems to work, but when I preview this in iD I see this:

tags-street_side

* The wiki-link for the preset seems to be `Key:parking` rather than `Tag:parking=street_side`. Any idea why?

* The label in the preset for `parking:street_side` is just `parking/street_side`. How do I set the label to the value of `label` in `data/fields/parking/street_side.json`?

* No wiki documentation for the field `parking:street_side` either, but that may because the [page](https://wiki.openstreetmap.org/wiki/Key:parking:street_side) has only recently been created.

Are there more files I should edit?

### DIFF
--- a/data/fields/parking.json
+++ b/data/fields/parking.json
@@ -11,7 +11,8 @@
             "carports": "Carports",
             "garage_boxes": "Garage Boxes",
             "rooftop": "Rooftop",
-            "sheds": "Sheds"
+            "sheds": "Sheds",
+            "street_side": "Street-side"
         }
     },
     "autoSuggestions" : false,
--- a/data/fields/parking/orientation.json
+++ b/data/fields/parking/orientation.json
@@ -1,0 +1,12 @@
+{
+    "key": "parking:orientation",
+    "type": "combo",
+    "label": "Orientation",
+    "strings": {
+        "options": {
+            "parallel": "Parallel to the Carriage Way",
+            "diagonal": "Diagonal in Relation to the Carriage Way",
+            "perpendicular": "Meets the Carriage Way at a Straight Angle"
+        }
+    }
+}
--- a/data/presets/amenity/parking/street-side.json
+++ b/data/presets/amenity/parking/street-side.json
@@ -1,0 +1,35 @@
+{
+    "icon": "temaki-car_parked",
+    "fields": [
+        "{amenity/parking}",
+        "parking/orientation"
+    ],
+    "geometry": [
+        "area",
+        "point",
+        "vertex"
+    ],
+    "tags": {
+        "amenity": "parking",
+        "parking": "street_side"
+    },
+    "addTags": {
+        "amenity": "parking",
+        "parking": "street_side"
+    },
+    "reference": {
+        "key": "parking",
+        "value": "street_side"
+    },
+    "terms": [
+        "automobile parking",
+        "car parking",
+        "vehicle parking",
+        "street parking",
+        "street-side parking",
+        "streetside parking",
+        "road-side parking",
+        "roadside parking"
+    ],
+    "name": "Street-side Parking"
+}
--- a/interim/source_strings.yaml
+++ b/interim/source_strings.yaml
@@ -1772,10 +1772,22 @@ en:
           rooftop: Rooftop
           # parking=sheds
           sheds: Sheds
+          # parking=street_side
+          street_side: Street-side
           # parking=surface
           surface: Surface
           # parking=underground
           underground: Underground
+      parking/orientation:
+        # 'parking:orientation=*'
+        label: Orientation
+        options:
+          # 'parking:orientation=diagonal'
+          diagonal: Diagonal in Relation to the Carriage Way
+          # 'parking:orientation=parallel'
+          parallel: Parallel to the Carriage Way
+          # 'parking:orientation=perpendicular'
+          perpendicular: Meets the Carriage Way at a Straight Angle
       parking_entrance:
         # parking=*
         label: Type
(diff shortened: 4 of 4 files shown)


SupaplexOSM:
Note: "parking:street_side" should be changed to "parking:orientation" after it has been renamed during the implementation of the proposal (see footnote 1 on proposal page: https://wiki.openstreetmap.org/wiki/Proposed_features/parking%3Dstreet_side).

jdhoek:
@SupaplexOSM Oh that's right! Commit amended.

TheAdventurer64:
Wow! I was just going to add this. Nice work!

maro-21:
I fully support `parking=street_side` tag and I like it but I think we shouldn't have a separate preset for it, just an item on the list like we have for other values would be enough.

We already have separate presets for Multilevel Parking Garage and Underground Parking (for me not necessary either), but we don't have it for e.g. parking=surface - the most common value.

One preset, "Parking Lot" would be enough. Then a user can choose the type of the parking.

The reason for this is that after typing "parking" (at least in my language) I would like to see the main presets:
1. `amenity=parking`
2. `amenity=bicycle_parking`
3. and `amenity=motorcycle_parking`.

But what I see is:
1. Underground Parking
2. Bicycle Parking
3. Parking Lot
4. Motorcycle Parking
5. Multilevel Parking Garage
6. Park and Ride Lot.
If we add another one, the main preset, "Parking Lot" will be even lower. But it should be on the first position becasue it's one of the most popular and basic preset/tag on OSM.

> The wiki-link for the preset seems to be Key:parking rather than Tag:parking=street_side. Any idea why?
> No wiki documentation for the field parking:street_side either, but that may because the page has only recently been created.

The answer to both question is: because there is no data item, like this: https://wiki.openstreetmap.org/wiki/Item:Q7598

jdhoek:
> I fully support parking=street_side tag and I like it but I think we shouldn't have a separate preset for it, just an item on the list like we have for other values would be enough.

`parking=street_side` is unique amongst the parking-values in having the `parking:orientation` tag. I don't see how that can be easily supported without a preset.

> If we add another one, the main preset, "Parking Lot" will be even lower. But it should be on the first position becasue it's one of the most popular and basic preset/tag on OSM.

That sounds like an issue with priorities in sorting. I agree that the generic parking preset should show up on top. I think there is some support in the tagging language that allows for some results to be prioritised, but I'm not knowledgeable enough about that. It's better to open a separate issue for the sorting problem to keep this PR topical though.

jdhoek:
@quincylvania Any idea how I can address the three bullet-points in the OP?

imagoiq:
Could it be addressed in some ways by using a `preset_category`?

ireun:
+1

data/fields/parking/orientation.json
```diff
@@ -0,0 +1,12 @@
+{
+    "key": "parking:orientation",
+    "type": "combo",
+    "label": "Parking bay orientation",
```
mbrzakovic, line 4: I'd recommend simplifying to just: 'Orientation' (its perhaps self-evident that it is indeed 'parking' orientation)

data/fields/parking/orientation.json
```diff
@@ -0,0 +1,12 @@
...
+    "strings": {
+        "options": {
+            "parallel": "Parallel to the carriage way",
```
mbrzakovic, line 7: Don't have a strong opinion on this, but I would recommend simplifying this part as well too.
(Parallel, diagonal, perpendicular).
Or we could do short versions: Parallel to street, etc.. (using 'street' here as it is more inline with the 'Street-side' parking)

mbrzakovic (approved):
Solid change, street-side parking are starting to be used intensively.

I just double checked locally (built this code and pushed to the iD) and I am confirming that [appropriate wiki page](https://wiki.openstreetmap.org/wiki/Tag:parking=street_side) is being linked in the iD for amenity/parking/street-side preset.

source_strings.yaml should be updated, thought I don't consider this a pr blocker (usually this is done via: npm run build)

data/fields/parking.json
```diff
@@ -11,7 +11,8 @@
             "carports": "Carports",
             "garage_boxes": "Garage Boxes",
             "rooftop": "Rooftop",
-            "sheds": "Sheds"
+            "sheds": "Sheds",
+            "street_side": "Street-side"
```
1ec5, line 15:
```suggestion
            "street_side": "Streetside"
```

data/fields/parking/orientation.json
1ec5: Yes, field options should be as brief as possible, because the inspector sidebar can be resized to a rather narrow width and some translations may require more characters than in English.

For consistency with other field options, use Title Case instead of Sentence case. Sentence case is suitable for tooltips, which to my knowledge aren’t supported by schema-builder yet.

data/presets/amenity/parking/street-side.json
```diff
@@ -0,0 +1,35 @@
...
+        "automobile parking",
+        "car parking",
+        "vehicle parking",
+        "street parking",
+        "street-side parking",
+        "streetside parking",
+        "road-side parking",
+        "roadside parking"
+    ],
+    "name": "Street-side Parking"
```
1ec5, line 34:
```suggestion
    "name": "Streetside Parking"
```

1ec5 (commented):
A bit of copyediting:

data/fields/parking.json
jdhoek: See my comments regarding the spelling of street-side above.

data/fields/parking/orientation.json
jdhoek: I think I'll stick with the current descriptions, as these match with the proposal and wiki documentation.

data/fields/parking/orientation.json
jdhoek: Done.

jdhoek:
> source_strings.yaml should be updated, thought I don't consider this a pr blocker (usually this is done via: npm run build)

Done.

jdhoek:
Note for a follow-up PR: in the year since I've submitted this PR `parking=lane` has gotten more strictly defined alongside `parking=street_side`. `parking:orientation` can be used with `parking=lane` as well, so it might benefit from the same approach.

`parking=lane` is not as common as `parking=street_side` because it competes more directly with the attribute-based tagging via `parking:lane` on the `highway=*`, but it is well used.

data/fields/parking/orientation.json
1ec5: The wiki documentation isn’t intended to dictate the precise UI for an editor field; we do have some leeway to tighten up the descriptions for a UI as long as it doesn’t contradict the documentation and actual usage. At the very least, the American English locale should use correct English spelling such as “[streetside](https://en.wiktionary.org/wiki/streetside)” and “[carriageway](https://en.wiktionary.org/wiki/carriageway#English)”. (But note that the term “carriageway” sounds incredibly archaic to a lay American English speaker.)

Moreover, the wiki doesn’t have a problem with other presets and fields using title case.

PR #80, opened 2020-12-09
Add a codespell check

peternewman:


peternewman:
Runs cleanly: https://github.com/peternewman/id-tagging-schema/actions/runs/411576759

PR #81, opened 2020-12-09
Add shop=insurance to deprecations

westnordost:
See https://wiki.openstreetmap.org/wiki/Tag:shop%3Dinsurance
It has never been approved. There was an abandoned proposal.

`shop=insurance` is used 430 times. `office=insurance` is used 40440 times.

It is documented to be synonymous to `office=insurance`. The wiki recommends since 14 July 2013‎ to use the more widespread `office=insurance` instead.

taghistory

https://taghistory.raifer.tech/#***/shop/insurance&***/office/insurance

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1335,6 +1335,10 @@
     "replace": {"shop": "art"}
   },
   {
+    "old": {"shop": "insurance"},
+    "replace": {"office": "insurance"}
+  },
+  {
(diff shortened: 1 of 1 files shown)


quincylvania:
@westnordost Thanks for including the background info on these, makes them easy to evaluate!

westnordost:
...and later to understand why a decision was made in case someone complains... :-)

PR #85, opened 2020-12-10
deprecate craft=car_repair -> shop=car_repair

westnordost:
`craft=car_repair` has been suggested to be deleted basically immediately after the wiki page had been created in 2013. See https://wiki.openstreetmap.org/wiki/Tag%3Acraft%3Dcar_repair , it's marked as deprecated now. It has about 100 usages.


quincylvania:
🚗👩‍🔧

PR #87, opened 2020-12-10
Add entrance=main

westnordost:
`entrance=main` (main entrance) is currently used 524321 times.

In JOSM, the icon for a main entrance is different from the icon for other entrances.

Maybe it makes sense to also add `entrance=service` (back entrance) ~40k usages. There is also `entrance=staircase` (entrance to staircase) ~300k usages.


quincylvania:
🚪🚶

PR #88, opened 2020-12-10
Add oil power plant

westnordost:
Only mapped about 300 times, but I think it makes sense to have the "normal" power plants (coal, gas, oil, nuclear, waste) complete as presets.

See https://wiki.openstreetmap.org/wiki/Tag%3Aplant%3Asource%3Doil

It uses the same icon as gas power plants (a flame), I think that also fits for oil.


quincylvania:
👍

PR #97, opened 2020-12-17
Update defibrillator.json

chnuessli:
I was thinking a change of the preset fields to this tag would be nice, often are the opening_hours and the access tags not filled out. Maybe that will help to prevent this.


mxdanger:
What is your reasoning for removing the ref tag?

chnuessli:
I made an overpass query to check if it was used often, but in the area it wasn't. If you wish, i could add it again.

chnuessli:
Is there something that could be added?

PR #103, opened 2020-12-29
Deprecate office/tax for office/tax_advisor

tas50:
This is a common mis-tagging around OSM.

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1066,6 +1066,10 @@
     "replace": {"office": "estate_agent"}
   },
   {
+    "old": {"office": "tax"},
+    "replace": {"office": "tax_advisor"}
+  },
+  {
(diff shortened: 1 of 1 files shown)


quincylvania:
The [wiki mentions](https://wiki.openstreetmap.org/wiki/Tag%3Aoffice%3Dtax) that another possible upgrade path for this is `office=government` + `government=tax`. We usually don't like to deprecate tags in iD if there are competing options.

It's not a big deal either way since there are only 210 uses. Maybe someone can go through and manually review them all?

tas50:
I think this can actually get closed. I zoomed out in overpass and realized that this tag is heavily used outside the US for government tax offices. I cleaned up all the US/Canada tax_advisor offices by hand so we're probably good to go now.

PR #113, opened 2021-01-16
Add presets for marsh, swamp and reedbed

Bertware:
Add presets for marsh, swamp and reedbed. This makes it significantly easier to map areas with wetlands as they become easier to find (=easier for new mappers) and they'll be suggested as individual types in the iD editor, meaning you can set them with one click instead of having to define the wetland type over and over. Reedbed is a lesser-used wetland type (5%, 52k tags) but is common, and making it easier to find could boost correct usage as well. 

https://wiki.openstreetmap.org/wiki/Tag:wetland%3Dmarsh
https://wiki.openstreetmap.org/wiki/Tag:wetland%3Dswamp
https://wiki.openstreetmap.org/wiki/Tag:wetland%3Dreedbed

### DIFF
--- a/data/presets/natural/marsh.json
+++ b/data/presets/natural/marsh.json
@@ -1,0 +1,18 @@
+{
+    "icon": "maki-wetland",
+    "fields": [
+        "wetland",
+        "salt",
+        "tidal"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "natural": "wetland",
+        "wetland": "marsh"
+    },
+    "terms": [],
+    "name": "Marsh"
+}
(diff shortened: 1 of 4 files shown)


TurnrDev (approved):

mbrzakovic (changes requested):
This change is justified given that all 3 types are approved and have substantial usage in osm.

I noticed that the list of allowed geometries for these types (and wetland) includes points, while the [wiki ](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dwetland)discourages such usage, and [josm only allows areas](https://josm.openstreetmap.de/browser/trunk/resources/data/defaultpresets.xml#L2207).
Curious on the reasoning behind this.

tyrasd:
merged as f47186efba8f9b3103f031e7410bf4d0117449cc

Bertware:
Sorry, I completely forgot to fix the geometry, agree with @mbrzakovic that points did not have to be included. Thanks for fixing and merging @tyrasd!

PR #122, opened 2021-02-05
Update Defibrillator presets

chnuessli:
In this PR, i want to change the presets in emergency=defibrillator. I think it would be better tagging on this type of node, when this presets will be showed. thank you


data/presets/emergency/defibrillator.json
```diff
@@ -2,13 +2,14 @@
     "icon": "maki-defibrillator",
     "fields": [
         "indoor",
-        "ref",
```
peternewman, line 5: Ref should be kept, it's used e.g. here:
https://osm.mathmos.net/defib/no-coords.html

chnuessli: It still can be used, it's just not on the frond-page.
But either way, i can work this in again. Should i?

peternewman: Personally I would have said so as it's used by at least one data consumer. Both post boxes and bus stops (probably all transport) seem to have some form of ref in their field list.

chnuessli: Hi, i added the ref again to the editor and made some code corrections in the json format. this commit is added to the pr. https://github.com/chnuessli/id-tagging-schema/commit/b038c43c3a8976d364ccea1005be77dd8e1bc3fa

peternewman: Thanks @chnuessli . FWIW, for me at least this PR seems to have lots of other random changed files too, I don't know if you just need to update or something.

chnuessli:
Hi @peternewman yes, but i reverted the wrong commit... it was a problem with changes between windows/mac on the coding device 👎🏼 . But it was reverted, anyway the PR is untouched by the repo owner, i hope it will be merged soon.

TurnrDev:
@chnuessli Can you rebase? It looks like your PR is deleting tests and touching the readme which is not ideal

chnuessli:
@TurnrDev I will set up a new PR and close this one

TurnrDev:
Moved to #135

PR #126, opened 2021-02-08
Office Movers is outdated

### LINKED ISSUE #124: Alias "Movers" to "Moving Company Office"
I typed too many letters (i.e. `mov` works, but `movers` doesn't), and eventually consulted the Wiki to find the appropriate tagging. Can you add a match for `movers?` to point people at `office=moving_company`?

TheAdventurer64:
Fixes #124

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1058,6 +1058,9 @@
     "old": {"office": "consultant"},
     "replace": {"office": "consulting"}
   },
+   {
+    "old": {"office": "movers"},
+    "replace": {"office": "moving_company"}
   {
     "old": {"office": "real_estate"},
     "replace": {"office": "estate_agent"}
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1058,9 +1058,10 @@
     "old": {"office": "consultant"},
     "replace": {"office": "consulting"}
   },
-   {
+  {
     "old": {"office": "movers"},
     "replace": {"office": "moving_company"}
+  },
   {
     "old": {"office": "real_estate"},
     "replace": {"office": "estate_agent"}


data/deprecated.json
```diff
@@ -1058,6 +1058,9 @@
     "old": {"office": "consultant"},
     "replace": {"office": "consulting"}
   },
+   {
+    "old": {"office": "movers"},
+    "replace": {"office": "moving_company"}
```
TheAdventurer64, line 1063: Forgot to add `},` here.

matkoniecz:
This tag has 2 uses worldwide: https://taginfo.openstreetmap.org/tags/?key=office&value=movers

matkoniecz (changes requested):
It should be handled via alias like in https://github.com/openstreetmap/id-tagging-schema/commit/942e4ce508b02ba3ea5ab516620181b5326e8a6a - not via deprecation

data/deprecated.json
TurnrDev: A comma on L1063 would in fact be invalid JSON, this is fine.

TurnrDev (approved):

TurnrDev:
> 
> 
> It should be handled via alias like in [942e4ce](https://github.com/openstreetmap/id-tagging-schema/commit/942e4ce508b02ba3ea5ab516620181b5326e8a6a) - not via deprecation

Actually, yes, you're right. I retract my approval of this change.

TurnrDev (approved):

mbrzakovic:
Unfortunately, I don't think the recommended change is resolving the issue.
I recommended a more appropriate solution in https://github.com/openstreetmap/id-tagging-schema/issues/124#issuecomment-965094972

PR #129, opened 2021-02-14
remove unsafe deprecation

### LINKED ISSUE #128: sloped_curb=yes - is it really safe to assume kerb=lowered?
See https://github.com/openstreetmap/id-tagging-schema/blob/main/data/deprecated.json#L1451

Note that at least some values of `sloped_curb` are mapping to rather `flush`.

Is it safe to assume that `sloped_curb=yes` means `kerb=lowered` rather than for example `flush`?

matkoniecz:
Sadly. sloped_curb=yes is unspecific and cannot be safely transformed into a single tag

fixes  #128

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1448,10 +1448,6 @@
     "replace": {"kerb": "raised"}
   },
   {
-    "old": {"sloped_curb": "yes"},
-    "replace": {"kerb": "lowered"}
-  },
-  {
     "old": {"speed_limit": "*"},
     "replace": {"maxspeed": "$1"}
   },


Marc-marc-marc:
change it to kerb=yes isn't a better fix ?

tyrasd:
@matkoniecz did you have a reason to keep this open, given @quincylvania's [reply in #128](https://github.com/openstreetmap/id-tagging-schema/issues/128#issuecomment-779328245). To me the argument seemed quite reasonable.

matkoniecz:
I will close it but rather based on fact that sloped curb lowering to flush is to be tagged as `=lowered` according to current tagging proposal.

PR #132, opened 2021-02-20
Fix platform tags not following PTv2 tagging scheme correctly

riiga:
The PTv2 tagging scheme says tags such as `bus=yes` goes on the stop_position only, not on any station or platform. This pull request fixes that.

See [the Wiki](https://wiki.openstreetmap.org/w/index.php?oldid=625726) for reference.

### DIFF
--- a/data/presets/public_transport/platform/_aerialway_point.json
+++ b/data/presets/public_transport/platform/_aerialway_point.json
@@ -10,8 +10,7 @@
         "point"
     ],
     "tags": {
-        "public_transport": "platform",
-        "aerialway": "yes"
+        "public_transport": "platform"
     },
     "reference": {
         "key": "public_transport",
--- a/data/presets/public_transport/platform/_ferry_point.json
+++ b/data/presets/public_transport/platform/_ferry_point.json
@@ -10,8 +10,7 @@
         "point"
     ],
     "tags": {
-        "public_transport": "platform",
-        "ferry": "yes"
+        "public_transport": "platform"
     },
     "reference": {
         "key": "public_transport",
--- a/data/presets/public_transport/platform/_light_rail_point.json
+++ b/data/presets/public_transport/platform/_light_rail_point.json
@@ -10,8 +10,7 @@
         "point"
     ],
     "tags": {
-        "public_transport": "platform",
-        "light_rail": "yes"
+        "public_transport": "platform"
     },
     "reference": {
         "key": "railway",
--- a/data/presets/public_transport/platform/_monorail_point.json
+++ b/data/presets/public_transport/platform/_monorail_point.json
@@ -10,8 +10,7 @@
         "point"
     ],
     "tags": {
-        "public_transport": "platform",
-        "monorail": "yes"
+        "public_transport": "platform"
     },
     "reference": {
         "key": "railway",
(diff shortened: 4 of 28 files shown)


bhousel:
Sorry to say, this won't work.  The tags in the `tags` object need to uniquely identify the feature.

riiga:
I see. I'll try to fix that.

bhousel:
Also this relates to some previous discussion on the iD repo here: https://github.com/openstreetmap/iD/issues/7524

At the time when I implemented the PTv2 presets back in 2017, the osm wiki said that these tags like `bus=yes` should be added https://github.com/openstreetmap/iD/pull/4585 

The wiki was changed sometime last summer to remove the tags, but the software does still need _something_ to be able to tell a Bus Stop from a Rail Stop, etc, so I think will need to keep them around.

riiga:
I updated with a new proposal.

The wiki never said those tags should be used if you check the actual proposal that was approved which I linked. It has been the same since 2011.

The software can tell the difference by matching multiple presets if there is a strict need to tell one platform from another, not by introducing tagging which is for the software only. JOSM does this without problem for example.

TurnrDev:
Bus Stops: `highway=bus_stop` + `public_transport=platform`
Bus Station: `amenity=bus_station` + `public_transport=station`
Railway Platforms: `railway=platform` + `public_transport=platform`
Railway Stations: `railway=station` + `public_transport=station`

Aerialway Stations: `aerialway=station` + `public_transport=station` ??

Ferry point: `amenity=ferry_terminal` + `public_transport=platform` 

Light Rail Station: `railway=station` + `station=light_rail` + `public_transport=platform`
Light Rail Platform is the same as Railway Platform. Before PTv2, these were both tagged the same anyway.

Monorail Station: `railway=station` + `station=monorail` + `public_transport=platform`
Monorail Platform is the same as Railway Platform. Before PTv2, these were both tagged the same anyway.


There's more but I lost track. It seems most the railway based ones share the same tagging.

bhousel:
> I updated with a new proposal.

All you did was just delete all the presets and replace them with a single one that says "Transit Platform".

riiga:
> All you did was just delete all the presets and replace them with a single one that says "Transit Platform".

Yes, that's the PTv2 tagging. A platform is a platform. The mode of transportation goes on the `stop_position`.

TurnrDev:
> 
> 
> > I updated with a new proposal.
>
> All you did was just delete all the presets and replace them with a single one that says "Transit Platform".

Yeah, that won't be ideal. I can take a crack at it if we like. The problem I'm facing is when we get to the railway-related tagging.

TurnrDev:
> 
> 
> > All you did was just delete all the presets and replace them with a single one that says "Transit Platform".
> 
> Yes, that's the PTv2 tagging. A platform is a platform. The mode of transportation goes on the `stop_position`.

We still need the other PTv1 tags as the two systems are supposed to work alongside each other. PTv2 was designed as an extension to PTv1, but a replacement of.

TurnrDev:
>  These original tags are still more commonly used and were not invalidated by the new tag, and should be kept in order to ensure compatibility with legacy software. Additionally, at least some prefer more specific tags rather than single tag supposed to cover objects as diverse as railway platforms, bus stops and airport infrastructure. 

[the wiki as of right now](https://wiki.openstreetmap.org/w/index.php?title=Tag:public_transport%3Dplatform&oldid=2042000)

PR #134, opened 2021-02-20
Add leisure=schoolyard

TurnrDev:
Add schoolyard preset based on wiki page for [leisure=schoolyard](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dschoolyard). 

Icon might need changing but I don't know 

Requirement for gravitystorm/openstreetmap-carto#4324


TurnrDev:
Is this repo maintained?

tordans:
I think not, as it is part of the ID ecosystem. 

https://blog.openstreetmap.org/2021/04/26/joint-statement-from-quincy-morgan-and-the-osmf-board/

https://twitter.com/quincylvania/status/1387072971773206533?s=21

https://twitter.com/osm_tech/status/1389546368197349376?s=21

bryceco:
It’s not abandoned, it’s just waiting for Quincy’s replacement to come along. Or somebody could adopt it. But nothing here will be available to iD until iD has a new maintainer so there’s no hurry to merge here.

TurnrDev:
> https://twitter.com/osm_tech/status/1389546368197349376?s=21

Ah shame, if I had a better knowledge of JavaScript I would have applied to the position.

bhousel:
> It’s not abandoned, it’s just waiting for Quincy’s replacement to come along. Or somebody could adopt it. But nothing here will be available to iD until iD has a new maintainer so there’s no hurry to merge here.

Projects shouldn't really depend on a single person to be the maintainer.  Is anyone from the openstreetmap org willing to work with me to transfer this repo over to either the osmlab or ideditor organization so that the community can maintain it?   I could just move the code, but it's nicer to have the whole repo transferred so we can keep the issues and history.

mbrzakovic (dismissed):
I don't see much traction on the usage of this tag, however [wiki ](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dschoolyard) clearly states that this tagging is appropriate for mapping schoolyards and [this comment](https://github.com/gravitystorm/openstreetmap-carto/issues/4324#issuecomment-778519405) does have a good point,

data/presets/leisure/schoolyard.json
```diff
@@ -0,0 +1,23 @@
...
+        "access_simple"
+    ],
+    "moreFields": [
+      "surface"
+    ],
+    "geometry": [
+        "area"
+    ],
+    "terms": [
+        "school playground",
```
1ec5, line 15: By the way, this is also known as a “blacktop” in American English (even when the asphalt pavement has faded to a light gray) to distinguish from the other meaning of “schoolyard” that refers to the whole grounds.

tyrasd (dismissed):
While the usage of this tag has increased quite a bit over the past year or so, it still is orders of magnitudes lower than what would expect from a really well established tag. Given the the tag also didn't undergo a proposal stage as far as I can see, I think its best to not yet merge this at the moment.

Also, the playground icon is really not ideal, since the schoolyard is supposed to be a mapped alongside a school's actual playground.

1ec5:
> it still is orders of magnitudes lower than what would expect from a really well established tag

“Orders of magnitude”, plural, might be overstating it a bit. 🙂 It’s pretty unlikely that the community will ever find the same order of magnitude of schoolyards as playgrounds (which are two orders of magnitude more common in the database). A correctly micromapped schoolground might have just one `leisure=schoolyard` but a dozen `leisure=playground`s corresponding to the mulched area around each playground set.

> Also, the playground icon is really not ideal, since the schoolyard is supposed to be a mapped alongside a school's actual playground.

Agreed, the existing swing icon would be misleading for this preset. I’m unsure what would be a better preset though. These things are characterized by being quite plain, often little more than a big paved area. Maybe something like the existing Pedestrian Area icon, but with a figure of a running child inside instead of a generic walking person? Or playing with a jumprope (skipping rope)?

tyrasd:
I was [comparing it](https://taghistory.raifer.tech/#***/leisure/schoolyard&***/building/school) to `building=school`, which one could consider to be around at the same micromapping-level (going with the example diagram given in the wiki for `leisure=schoolyard`), and I assumed that most schools in reality consist of a similar number yards as of buildings. But perhaps its that's a bit "unfair" to the poor non-building tag. :wink:

The main problem I have with accepting this relatively new tag (again, compared to its school related map feature siblings) is that it didn't undergo a proposal stage is that it could always be the case that it turns out to be "problematic". Maybe the `leisure` is suboptimal, and something like `landuse=schoolyard` would have been the better choice. Or maybe these places would be better mapped as a special kind of playground (e.g. `leisure=playground` + `playground=schoolyard`) or even as a pedestrian area (thinking about it… something like `highway=pedestrian` + `area=yes` + `pedestrian=schoolyard` + `access=…` would even work to bad from a routing point of view). I just don't know. :shrug: 

FWIW, JOSM also doesn't have a preset for `leisure=schoolyard` at the moment.

data/presets/leisure/schoolyard.json
TurnrDev, line 15:
```suggestion
        "blacktop",
        "school playground",
```

matkoniecz:
> The main problem I have with accepting this relatively new tag (again, compared to its school related map feature siblings) is that it didn't undergo a proposal stage is that it could always be the case that it turns out to be "problematic". Maybe the leisure is suboptimal, and something like landuse=schoolyard would have been the better choice. Or maybe these places would be better mapped as a special kind of playground (e.g. leisure=playground + playground=schoolyard) or even as a pedestrian area (thinking about it… something like highway=pedestrian + area=yes + pedestrian=schoolyard + access=… would even work to bad from a routing point of view). I just don't know. 🤷

I think it is perfectly fine for this tag to be far less used, it is kind of micromapping - in the same way as typical school will not have footway going to mapped or trees next to it will not be mapped

what you think about making a thread at community forums to check is there any real opposition to this tag or concerns? And accept is as valid if there are no real complaints

tyrasd:
> what you think about making a thread at community forums to check is there any real opposition to this tag or concerns? And accept is as valid if there are no real complaints

Something like a forum thread would be the minimum to be expected for a tag that's meant to be used in the micromapping of a thing as prominent as schools.

See also our new [general guidelines](https://github.com/openstreetmap/id-tagging-schema/blob/main/GUIDELINES.md#general-guidelines).

@TurnrDev I know it has been a long time since you opened this PR. Are you still interested in bringing this effort forward? If yes, it would be very much appreciated if you could start a discussion about this and after a suitable time report back here with a summary. :pray:

TurnrDev:
I appreciate I was the one who initially created this PR, and thus it makes sense for me to lead that discussion, however, leading discussions is not really one of my strengths and my investment to this is not that high. (I'm always that person with the mic off and camera off in any meeting above 3 people.) I'm going to kindly decline, but with that said, I'm going to leave the PR open and am happy to make any changes necessary if someone else does pick up that discussion.

matkoniecz:
hmm, I think I can create thread like this om community forums:

title: leisure=schoolyard - is it a good tag worth promoting?

body: 

There is https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dschoolyard tag that seems reasonable to me. But, partially because it is kind of micromapping it is used relatively rarely. 

At https://github.com/openstreetmap/id-tagging-schema/pull/134 it was proposed to add it to presets but maintainers are worried that there are some possible problems with it as it is relatively rare and have not went through proposal process.

Are you aware about any issues with this tag that should result in it not being promoted by presets? Or is it a fine tag worth supporting for example in iD presets?

tyrasd:
@TurnrDev fair enough. That's totally ok. 

@matkoniecz thanks for the offer to take this over! :+1:  If you want, you can quote my concerns from my comment above: https://github.com/openstreetmap/id-tagging-schema/pull/134#issuecomment-996027141

matkoniecz:
posted on https://community.openstreetmap.org/t/leisure-schoolyard-is-it-a-good-tag-worth-promoting/117170

TurnrDev:
Seems the overall consensus on your post if positive - how long do we usually leave these things? @matkoniecz

matkoniecz:
I would say that anyone who wanted to comment probably did and there is no point in waiting more for comments.

I have somewhere on my todo list to go through comments/reactions and check what is the general sentiment and whether any serious issues were raised.

TurnrDev:
## Test-Documentation

https://pr-134--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=19.92/52.98011/-1.15071&disable_features=boundaries&background=Bing&id=w1485346385

### Search
image

### Info-`i`

image

### Wording

- [x] American English

data/presets/leisure/schoolyard.json
```diff
@@ -0,0 +1,24 @@
+{
+    "icon": "maki-playground",
+    "fields": [
+        "name",
+        "operator",
```
andrewharvey: What would be the point of `operator`? Wouldn't it almost always be the `amenity=school` that the `leisure=schoolyard` is within?

TurnrDev: Most things seem to have this by default, I didn't much think about it.

matkoniecz:
`pinhead-person_running_after_person` icon would work, I think

https://pinhead.ink/#person_running_after_person

TurnrDev:
I've updated that now @matkoniecz

matkoniecz:
```

🏗   Validating and building dist files...

expected terms

[
  "blacktop",
  "recess yard",
  "school playground",
  "schoolyard"
]

differ from actual ones

blacktop,school playground,schoolyard,recess yard
Expected terms in preset "leisure/schoolyard" to be lowercase and sorted alphabetically.
Error: Process completed with exit code 1.

```

data/presets/leisure/schoolyard.json
```diff
@@ -0,0 +1,23 @@
...
+    ],
+    "geometry": [
+        "area"
+    ],
+    "terms": [
+        "blacktop",
+        "school playground",
+        "schoolyard",
+        "recess yard"
+    ],
```
matkoniecz, line 15:
```suggestion
    "terms": [
        "blacktop",
        "recess yard",
        "school playground",
        "schoolyard"
    ],
```

matkoniecz (approved):

matkoniecz:
Thanks for the PR! It is now merged (finally!) but note that before you will see it in iD few things need to happen.

PR #135, opened 2021-02-21
Update defibrillator.json

chnuessli:
so let's try this pr. just simply add more fields to the presets.

### DIFF
--- a/data/presets/emergency/defibrillator.json
+++ b/data/presets/emergency/defibrillator.json
@@ -4,11 +4,13 @@
         "indoor",
         "ref",
         "operator",
-        "defibrillator/location"
+        "defibrillator/location",
+        "opening_hours",
+        "access",
+        "description"
     ],
     "moreFields": [
         "level",
-        "opening_hours",
         "access_simple"
     ],
     "geometry": [


chnuessli:
@Turnrdev can you check this PR?

TurnrDev (approved):
Looks much better!

chnuessli:
Is there something that i could do? Or any reason why the pull request is open?

TurnrDev (approved):

TurnrDev:
Hopefully a mod sees it

chnuessli:
@quincylvania can i add something more?

data/presets/emergency/defibrillator.json
```diff
@@ -4,11 +4,13 @@
         "indoor",
         "ref",
         "operator",
-        "defibrillator/location"
+        "defibrillator/location",
+        "opening_hours",
+        "access",
```
mbrzakovic, line 9: The `access` field is generally used to specify access per type of transportation vehicle (foot, bicycles, motor etc.)
Given this [wiki statement](https://wiki.openstreetmap.org/wiki/Tag:emergency%3Ddefibrillator): _access=*, if the access to the device is restricted_

I think that access_simple in 'moreFields' is sufficient to cover this scenario.

data/presets/emergency/defibrillator.json
```diff
@@ -4,11 +4,13 @@
...
+        "description"
```
mbrzakovic, line 10: I don't have a strong opinion on this, but given that `description` is more of a situation-specific type of information (it should be provided only when there is something special to be disclosed), I would recommend keeping it in 'moreField'. Overall I agree that this field is helpful.

mbrzakovic (changes requested):

data/presets/emergency/defibrillator.json
chnuessli: Hi there.
I don't mind to change it. The PR is never accepted anyway and i think, never will.

mbrzakovic: Please address the comments and I will work on accepting the change :)

data/presets/emergency/defibrillator.json
chnuessli: This field is often not set on defibrillators. I would highly recomment to put it on front, because then, it's not forgotten and more helpful. Thank you

chnuessli:
I changed `description`on the last commit.

data/presets/emergency/defibrillator.json
mbrzakovic: Just to clarify, could you please provide what is the expected tag key-value pair that would be included if the access field is populated? Could you provide link/id to osm entity that would serve as an example.

chnuessli: Here's an example of the tags that would be used on a defi: https://www.openstreetmap.org/node/9285955621

mbrzakovic: Please look into [access ](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/access.json) field and reconsider this. This field offers more complex specification of access, as mentioned above, it allows specification of access per type of transportation vehicle.
If the intent is for users to be more clearly asked to specify access=yes/no, then I suggest you move `access_simple` field from moreFields to fields.

chnuessli:
I've reconsidered it. access_simple is not what i wanted. I wanted a set of meaningful and clear changes. If this is not possible, i am happily closing this PR.

PR #136, opened 2021-02-21
Add capacity for childcare and kindergarten

bagage:


TurnrDev (approved):

mbrzakovic:
[Wiki page](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dkindergarten) is suggesting this is a valid combination however [taginfo](https://taginfo.openstreetmap.org/tags/amenity=kindergarten#combinations) shows that kindergarten + capacity is very rarely used (same for childcare).
This seems like an item that shouldn't belong in the 'fields' but somewhere in between 'moreFields' and not including it at all as a suggestion.
@TurnrDev opinion?

tyrasd (changes requested):
can you please include `amenity=school` here as well and move the field to the `moreFields` section?

bagage:
@tyrasd  should be good :)

tyrasd (approved):

PR #144, opened 2021-02-27
add reservoir_type=tailings

### LINKED ISSUE #142: Suggest tag upgrade for tailings ponds
The tag `man_made=tailings_pond` has been [approved](https://wiki.openstreetmap.org/wiki/Tag:man_made=tailings_pond) via unanimous vote.

This replaces existing tagging for tailings ponds, which used:

`natural=water`+`water=reservoir`+`reservoir_type=tailings`
OR:
`landuse=reservoir`+`reservoir_type=tailings`

For any feature tagged `reservoir_type=tailings`, iD should suggest an upgrade which would:
* Set `man_made=tailings_pond`
* Remove `natural=water`, `water=reservoir`, and/or `landuse=reservoir` tags if present

Note that this tagging upgrade would remove tailings ponds from rendering as water, converting them (as of 2/26/21) to a tag which does not currently render (in the Carto style).  However, it is better that these features do not render than to have them render incorrectly as a water feature.

Marc-marc-marc:
add several combinaison with the depreciated tag reservoir_type=tailings

closes #142


tyrasd (changes requested):
For this to work well, we also need a preset for the new `man_made=tailings_pond` tag. otherwise users would be presented with the rather generic preset for a _"man made feature"_ after upgrading the tags. @Marc-marc-marc can you please add that here?

matkoniecz:
> can you please add that here?

I guess that it would be also OK with separate PR? If someone else would want to do this or @Marc-marc-marc is stuck with git/github interface?

tyrasd:
:+1: sure, for me it would be also fine if someone else took over (if it's ok with @Marc-marc-marc of course). Just base your PR/branch of from this one.

PR #166, opened 2021-04-25
Added a preset for amenity=ticket_validator

Lukas458:
amenity=ticket_validator (https://wiki.openstreetmap.org/wiki/Tag%3Aamenity%3Dticket_validator) has increased by over 2300 uses by this time, but it din't have a preset yet. I did not open an issue for that, but I can if necessary. So as the usage for this tag seems very clear to me, I thought it would be worth adding preset for it. I used the icon from `vending_machine,` maybe when you're not okay with it, you can remove it. I hope everything else fits.

### DIFF
--- a/data/presets/amenity/ticket_validator
+++ b/data/presets/amenity/ticket_validator
@@ -17,4 +17,10 @@
     "tags": {
         "amenity": "ticket_validator"
     },
+    "terms": [
+        "ticket",
+        "validator",
+        "puncher",
+    ],
+    "name": "Ticket Validator"
 }
--- a/data/presets/amenity/ticket_validator
+++ b/data/presets/amenity/ticket_validator
@@ -1,0 +1,20 @@
+{
+    "icon": "temaki-vending_machine",
+    "fields": [
+        "ref",
+        "operator",
+    ],
+    "moreFields": [
+        "colour",
+        "covered",
+        "indoor",
+        "level",
+        "support"
+    ],
+    "geometry": [
+        "point"
+    ],
+    "tags": {
+        "amenity": "ticket_validator"
+    },
+}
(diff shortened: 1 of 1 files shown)


data/presets/amenity/ticket_validator
```diff
@@ -0,0 +1,20 @@
...
+        "indoor",
+        "level",
+        "support"
+    ],
+    "geometry": [
+        "point"
+    ],
+    "tags": {
+        "amenity": "ticket_validator"
+    },
```
mbrzakovic, line 19: I recommend we also provide Name and some search information.
e.g.:
Name: Ticket Validator
Aliases: ticket puncher, punching machine or ticket marker
Terms: ticket, validator, puncher

mbrzakovic (changes requested):

data/presets/amenity/ticket_validator
Lukas458: Oh yes, sure, you're right. I forgot the name. Thank you, I added the terms you mentioned and the name now. Do I have to add the aliases as themselves, too?

mbrzakovic: Thanks for doing this promptly. Adding aliases as well, would be a slight improvement, please read more [here](https://github.com/ideditor/schema-builder/issues/3)

mbrzakovic (approved):

data/presets/amenity/ticket_validator
Lukas458: Yes, thank you. Added them in https://github.com/openstreetmap/id-tagging-schema/pull/166/commits/14f66c248c82af30b7b5195749a00d73d4b2811d

Lukas458:
Thank you!

Issue #170, opened 2021-05-07
Add `opening_date` & `check_date` to `construction=building` preset


Pengor:
I think it would be helpful when mapping buildings that are under construction to have the `opening_date` and `check_date` tags available in the preset. This is most useful when the date of completion for the building's construction is known, but I don't think that's uncommon.

Pengor:
Fixed by #171

PR #171, opened 2021-05-07
Add fields to building=construction preset

Pengor:
Addresses #170
Adds opening_date and check_date fields to the building=construction preset


data/presets/building/construction.json
```diff
@@ -1,5 +1,11 @@
 {
     "icon": "maki-building",
+    "fields": [
+        "name",
```
mbrzakovic, line 4: I'd recommend adding "construction" field as well as I see it [already being used very frequently](https://taginfo.openstreetmap.org/tags/building=construction#combinations) and the [wiki itself](https://wiki.openstreetmap.org/wiki/Tag:building%3Dconstruction) points towards using this combination to specify the actual type of a building

data/presets/building/construction.json
```diff
@@ -1,5 +1,11 @@
...
+        "address",
+        "opening_date",
+        "check_date"
```
mbrzakovic, line 7: Less important, but I am curious in hearing opinions if it would makes sense to also add default building fields: `building/levels `and `height` here. I suppose the usage could be unclear, given that it could be interpreted as either planned or current. I didn't manage to find any docs on this, but personally I think it would be beneficial to use these in similar context as `check_date` (to update them during the construction) in order to track the current state and progress.

mbrzakovic (changes requested):

data/presets/building/construction.json
Pengor: Good call, I've added that one.

mbrzakovic (approved):

PR #175, opened 2021-06-13
Create railway:yard:size.json

ZEE6de:
for tag "railway=yard"

PR #176, opened 2021-06-13
Update Yard preset

ZEE6de:
Update Yard preset.

### DIFF
--- a/data/presets/railway/derail.json
+++ b/data/presets/railway/derail.json
@@ -1,5 +1,10 @@
 {
     "icon": "maki-roadblock",
+    "fields": [
+        "ref",
+        "railway:derail",
+        "railway:local_operated"
+    ],
     "geometry": [
         "vertex"
     ],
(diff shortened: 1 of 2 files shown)


natfoot:
might also want to add ref or abbreviated name.

mbrzakovic:
According to [taginfo](https://taginfo.openstreetmap.org/tags/railway=yard#overview) currently there are ~3k yards, and ~50 usages of 3 recommended railway:yard: tags.
Although osm wiki does have these tags documented in the [yard page](https://wiki.openstreetmap.org/wiki/Tag:railway%3Dyard) the tags themselves don't have appropriate pages.

Given this, I recommend we postpone this change until this usage gains some additional traction.
If this gets adopted by wider community, iD can follow with ensuring appropriate fields.
Until then, users are free to use the tag editting ui portion to specify these properties.

_Connected_: https://github.com/openstreetmap/id-tagging-schema/pull/175, https://github.com/openstreetmap/id-tagging-schema/pull/177, https://github.com/openstreetmap/id-tagging-schema/pull/178.

Moving forward, please create Pull Request in such way that it represents one self-contained logical change.
(In this case it would be more appropriate that _Connected_ and this change were part of a singular PR).

PR #178, opened 2021-06-13
Create railway:yard:purpose.json

ZEE6de:
See https://wiki.openstreetmap.org/wiki/Tag:railway=yard


mbrzakovic:
Please refer to #176

PR #179, opened 2021-06-13
Update derail.json

ZEE6de:
expand

PR #180, opened 2021-06-13
Create derail.json

ZEE6de:
create for railway=derail


tyrasd:
this needs to be included in #179

PR #181, opened 2021-06-13
Create local_operated.json

ZEE6de:


natfoot (commented):
What is being changed here?

natfoot (commented):
Looks like a good edit.

tyrasd:
this needs to be included in #179

data/fields/railway/local_operated.json
```diff
@@ -0,0 +1,10 @@
+{
+    "key": "railway:local_operated",
+    "type": "radio",
```
tyrasd, line 3: should this be a `check` field, given that the values are only `yes` and `no`?

PR #185, opened 2021-06-13
Create on_demand.json

ZEE6de:


mbrzakovic:
[crossing:on_demand](https://wiki.openstreetmap.org/wiki/Key:crossing:on_demand) doesn't look like its mature enough to be added as a field.
Of course, in iD there is still a flexibility to add this property via tag editting ui panel.
Feel free to reopen this PR if this tag gains some additional traction (currently [~2k usage](https://taginfo.openstreetmap.org/keys/crossing:on_demand#overview))

PR #187, opened 2021-06-24
Add fields in power=substation

tntchn:
Many substations have name=* and voltage=* tags but iD editor doesn't have those field.


natfoot (approved):
This looks like a good edit.

PR #193, opened 2021-07-08
Create railway:signal:position

natfoot:
I hope I am doing this right.  This would add the field of the location of the signal to the desired fields populated by the most common responses rather than a set or desired set of responses.  This will help with the addition of this data by making it front and center and will speed up my own mapping.  If this is approved I will move forward with a set of other changes in `railway:signal` to better add descriptive tags of signals.

### DIFF
--- a/data/fields/railway/signal/railway:signal:position
+++ b/data/fields/railway/signal/railway:signal:position
@@ -1,0 +1,6 @@
++ {
++     "key": "railway:signal:position",
++     "type": "combo",
++     "placeholder": "Left, Right, In Track, etc...",
++     "label": "Signal Location to Track"
++ }


data/fields/railway/signal/railway:signal:position
```diff
@@ -0,0 +1,6 @@
++ {
```
mbrzakovic, line 1: As described here: https://github.com/ideditor/schema-builder#fields, the filename should be: position.json

Additionally, I think this field should be added to the corresponding preset as a strong field. (presets/railway/signal.json)

mbrzakovic (changes requested):
Thanks for the contributions!
It would be great if you could provide fast links in the pull request initial comment which would speed-up the review process. e.g. for this pr:
https://wiki.openstreetmap.org/wiki/Tag:railway%3Dsignal 
https://taginfo.openstreetmap.org/keys/railway:signal:position#values

PR #195, opened 2021-07-10
Update access_simple.json

natfoot:
After a conversation at STOM about access keys and the need for unknown to be added as a preset in ID. 
https://wiki.openstreetmap.org/wiki/Key:access


mbrzakovic:
Connected with #196

tyrasd:
closing in favor of #196

PR #196, opened 2021-07-10
add "unknown" as a possible access value

natfoot:
After a conversation at STOM about access keys and the need for unknown to be added as a preset in ID. 
https://wiki.openstreetmap.org/wiki/Key:access

### DIFF
--- a/data/fields/access.json
+++ b/data/fields/access.json
@@ -21,6 +21,10 @@
                 "title": "Prohibited",
                 "description": "Access not allowed to the general public"
             },
+             "unknown": {
+                "title": "Unknown",
+                "description": "Access are unknown or unclear; or access could be restricted"
+            },
             "permissive": {
                 "title": "Permissive",
                 "description": "Access allowed until such time as the owner revokes the permission"


mbrzakovic:
@natfoot are you referring to [sotm](https://2021.stateofthemap.org/)*? :)
Could you please share some highlights/links on this topic, I'd prefer if some additional information was provided before moving forward with this.

Here is some of the intel that I am able to pick-up:
https://wiki.openstreetmap.org/wiki/Tag:access%3Dunknown (current status: unspecified)
Some discussion in [wiki history (subject: access=unknown ....)](https://wiki.openstreetmap.org/wiki/Talk:Key:access)
[Josm access preset](https://josm.openstreetmap.de/browser/trunk/resources/data/defaultpresets.xml#L243) (currently does not include unknown).

From these I am deriving that although the tag is being used, it still hasn't been accepted like the rest of the values in access.json

matkoniecz:
>  https://wiki.openstreetmap.org/wiki/Tag:access%3Dunknown (current status: unspecified)

I edited it to "de facto" what I consider to be actual consensus about this tag ( https://wiki.openstreetmap.org/w/index.php?title=Tag:access%3Dunknown&action=history ). Not sure why I initially missed this when I created this page.

(as usual, edits are welcome if some OSM Wiki page is wrong or missing some info)

matkoniecz:
I created https://josm.openstreetmap.de/ticket/21705 "Add access=unknown to presets"

PR #174, opened 2021-05-30
Add translating info

peternewman:

PR #200, opened 2021-07-22
Add information on translating to readme

westnordost:
This should be helpful information.


ireun:
Kind of a duplicate of #174

westnordost:
Right, true. I think though it is better to have this as a full sentence.

PR #209, opened 2021-08-02
Update abandoned.json

ZEE6de:
Change name from Abandoned to Demolished, because "abandoned" railway would be railway=disused, to prevent confusion.

### DIFF
--- a/data/presets/railway/abandoned.json
+++ b/data/presets/railway/abandoned.json
@@ -19,8 +19,12 @@
     "terms": [
         "abandoned railroad",
         "former train tracks",
-        "train right-of-way"
+        "train right-of-way",
+        "demolished railway",
+        "former railway",
+        "railbed",
+        "
     ],
     "matchScore": 0.85,
-    "name": "Abandoned Railway"
+    "name": "Demolished Railway"
 }


mbrzakovic:
I understand that the current naming conventions are not most intuitive for this situation. Here is how I am perceiving some of these informations ([disused](https://wiki.openstreetmap.org/wiki/Tag:railway%3Ddisused), [abandoned](https://wiki.openstreetmap.org/wiki/Tag:railway%3Dabandoned), [demolished](https://wiki.openstreetmap.org/wiki/Demolished_Railway)):
- If there is a track and some infrastructure, but the rail is not used, label as disused - as in: not currently used but there is a possibility to restore it with not much effort and use it in future
- If there is no track but some infrastructure (sleepers, signals etc.) label as abandoned - as in: permanently abandoned, it would take substantial effort to restore it
- If there are no tracks and very limited amount or no infrastructure, then it goes in the demolished category and by [mapping what's on the ground principal](https://wiki.openstreetmap.org/wiki/Good_practice#Map_what.27s_on_the_ground) should be strongly considered not mapping this as well.

Unfortunately, I don't think that changing the name from Abandoned to Demolished would be inline with abovementioned documentation, it could only cause some further confusion by having the different preset name and railway: tag value.

PR #210, opened 2021-08-05
Update warehouse.json

ZEE6de:
add website key

### DIFF
--- a/data/presets/building/warehouse.json
+++ b/data/presets/building/warehouse.json
@@ -2,6 +2,7 @@
     "icon": "maki-warehouse",
     "fields": [
         "{building}",
+        "website",
         "phone"
     ],
     "geometry": [


data/presets/building/warehouse.json
```diff
@@ -2,6 +2,7 @@
     "icon": "maki-warehouse",
     "fields": [
         "{building}",
+        "website",
```
mbrzakovic, line 5: [Some statistics](https://taginfo.openstreetmap.org/tags/building=warehouse#combinations) are showing that website and phone are rarely used in this scenario, which is why I would recommend placing these in [moreFields](https://github.com/ideditor/schema-builder#fieldsmorefields)

mbrzakovic (changes requested):

tyrasd:
closing. a warehouse by itself typically does not have a website. The few cases where the co-tagging occurs are typically from POIs which are mapped on the same object as the building tag (often POIs of the touristic, art, shop or similar kind).

PR #214, opened 2021-08-07
Add cardinal direction field to some route presets

### LINKED ISSUE #202: Add Direction field to route relation presets
As of openstreetmap/iD#8276, the `direction` key on a `route` relation disambiguates the relation in a list of relations. Similar to #104, the route presets should each have a Direction field. The field should accept cardinal directions such as `north` (North), which are [standard in some countries](https://wiki.openstreetmap.org/wiki/User:Minh_Nguyen/Route_directions). A combo field is probably appropriate, because other values such as `anticlockwise` and `inbound` sometimes occur on public transportation routes.

1ec5:
Added a Direction field to the Cycle Route, Road Route, and Train Route presets (and by extension other public transportation route presets). The field has only been added to route relation types that are commonly tagged with cardinal directions as of writing; I didn’t find enough usage on `hiking`, `horse`, or `piste` route relations to justify matching 9f092d884f4aa194e2b512d5d1d933e53096cab7.

Unlike the other Direction fields, this is a combo box that suggests cardinal directions. Both the `N`/`S`/`E`/`W` and `north`/`south`/`east`/`west` formats are approved and well-established overall, but the latter [predominates](https://overpass-turbo.eu/s/1a5I) [among route relations](https://overpass-turbo.eu/s/1a5J), so that’s what this PR introduces. The combo box allows user-defined values, since public transportation systems use a variety of idiosyncratic [rail directions](https://en.wikipedia.org/wiki/Rail_directions), some of which are commonly tagged in OSM.

Fixes #202.

### DIFF
--- a/data/fields/direction_cardinal.json
+++ b/data/fields/direction_cardinal.json
@@ -1,0 +1,14 @@
+{
+    "key": "direction",
+    "type": "combo",
+    "label": "Direction",
+    "strings": {
+        "options": {
+            "north": "Northbound",
+            "south": "Southbound",
+            "east": "Eastbound",
+            "west": "Westbound"
+        }
+    },
+    "autoSuggestions" : false
+}
--- a/data/fields/direction_cardinal.json
+++ b/data/fields/direction_cardinal.json
@@ -10,5 +10,8 @@
             "west": "Westbound"
         }
     },
-    "autoSuggestions" : false
+    "autoSuggestions" : false,
+    "locationSet": {
+        "include": [ "CA", "NZ", "US" ]
+    }
 }
--- a/data/presets/type/route/bicycle.json
+++ b/data/presets/type/route/bicycle.json
@@ -6,6 +6,7 @@
         "network_bicycle",
         "cycle_network",
         "network/type",
+        "direction_cardinal",
         "from",
         "to",
         "via"
--- a/data/presets/type/route/road.json
+++ b/data/presets/type/route/road.json
@@ -4,6 +4,7 @@
         "name",
         "ref_route",
         "network_road",
+        "direction_cardinal",
         "from",
         "to",
         "via"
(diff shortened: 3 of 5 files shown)


tyrasd (commented):
I don't know. Currently, only very few relations use this tag (see below). The [wiki](https://wiki.openstreetmap.org/wiki/Relation:route) only mentions usage for some specific countries: Perhaps it would make more sense to limit this to the US, Canada and New Zealand for now?

| route type | total count | count with direction | relative amount |
|---|---|---|---|
| `route=bicycle` | 87007 | 65 | 0.07% |
| `route=train` | 11038 | 42 | 0.4% |
| `route=road` | 197179 | 1413 | 0.7% |
| `route=road` (USA) | 48912 | 661 | 1.4% |

1ec5:
I limited the field to the three MUTCD countries where routes are commonly identified by cardinal direction. There’s no real-world justification for limiting it to just road routes in these countries, because bicycle and public transportation routes are commonly referred to in the same way. There’s more information in [this wiki page](https://wiki.openstreetmap.org/wiki/Route_directions), which has presented the `direction` tag as one of two approaches for many years.

tyrasd:
> because public transportation routes are commonly referred to in the same way. 

Then why not also add `route=bus`, `route=subway`, `route=tram` and `route=light_rail` as well? :thinking: 

Perhaps the non-road route types could instead have the field in the `moreFields` section?

1ec5:
So far, I’ve only added cardinal directions to this field. For public transportation routes that distinguish directions, there’s more local variation, like “inbound” versus “outbound”, so I wasn’t sure about broaching that here. By contrast, long-distance bike routes are always referred to by cardinal directions in the U.S. (not sure about Canada and New Zealand).

I agree that we could add the field to `moreFields` for the public transportation presets just in case a local community finds it suitable for their local bus or subway system. The original motivation for this PR was that iD is already using `direction` for its automatic relation labels, so I guess it would be more consistent to add it to all of them.

PR #215, opened 2021-08-07
Add some common name fields

### LINKED ISSUE #283: Show official_name if tagged already
Some objects such as https://www.openstreetmap.org/way/25506183 https://www.openstreetmap.org/relation/9465346 have regular actually used `name` and officially assigned `official_name`

Some new mappers change `name` tag to what is already in `official_name` what is a repeated annoyance.

It would happen less often if `official_name` would be shown by iD and make clear that this info is collected already

1ec5:
Added some of the most common name fields as universal fields to all presets:

* Local Name (`loc_name`)
* National Name (`nat_name`)
* Official Name (`official_name`)
* Regional Name (`reg_name`)
* Short Name (`short_name`)

For this round, I omitted some common name fields because they weren’t as straightforward:

* `alt_name` and `old_name` are likely to be localized and contain multiple values at the same time, but neither `semiCombo` nor `localized` provides all the necessary functionality.
* `sorting_name` is especially relevant in some languages but completely irrelevant in others, so it would be nice to filter the `localized` field type’s language list accordingly.

Fixes #283 and fixes openstreetmap/iD#1554.

### DIFF
--- a/data/fields/alt_name.json
+++ b/data/fields/alt_name.json
@@ -1,0 +1,16 @@
+{
+    "key": "alt_name",
+    "type": "localized",
+    "label": "Alternative Name",
+    "universal": true,
+    "terms": [
+        "aka",
+        "alias",
+        "also known as",
+        "nonstandard name",
+        "secondary name"
+    ],
+    "prerequisiteTag": {
+        "key": "name"
+    }
+}
--- a/data/fields/loc_name.json
+++ b/data/fields/loc_name.json
@@ -1,0 +1,12 @@
+{
+    "key": "loc_name",
+    "type": "localized",
+    "label": "Local Name",
+    "universal": true,
+    "terms": [
+        "colloquial name",
+        "common name",
+        "informal name",
+        "nickname"
+    ]
+}
--- a/data/fields/loc_name.json
+++ b/data/fields/loc_name.json
@@ -8,5 +8,8 @@
         "common name",
         "informal name",
         "nickname"
-    ]
+    ],
+    "prerequisiteTag": {
+        "key": "name"
+    }
 }
--- a/data/fields/nat_name.json
+++ b/data/fields/nat_name.json
@@ -1,0 +1,6 @@
+{
+    "key": "nat_name",
+    "type": "localized",
+    "label": "National Name",
+    "universal": true
+}
--- a/data/fields/nat_name.json
+++ b/data/fields/nat_name.json
@@ -2,5 +2,8 @@
     "key": "nat_name",
     "type": "localized",
     "label": "National Name",
-    "universal": true
+    "universal": true,
+    "prerequisiteTag": {
+        "key": "name"
+    }
 }
--- a/data/fields/official_name.json
+++ b/data/fields/official_name.json
@@ -1,0 +1,11 @@
+{
+    "key": "official_name",
+    "type": "localized",
+    "label": "Official Name",
+    "universal": true,
+    "terms": [
+        "formal name",
+        "full name",
+        "legal name"
+    ]
+}
--- a/data/fields/official_name.json
+++ b/data/fields/official_name.json
@@ -7,5 +7,8 @@
         "formal name",
         "full name",
         "legal name"
-    ]
+    ],
+    "prerequisiteTag": {
+        "key": "name"
+    }
 }
--- a/data/fields/reg_name.json
+++ b/data/fields/reg_name.json
@@ -1,0 +1,6 @@
+{
+    "key": "reg_name",
+    "type": "localized",
+    "label": "Regional Name",
+    "universal": true
+}
--- a/data/fields/reg_name.json
+++ b/data/fields/reg_name.json
@@ -2,5 +2,8 @@
     "key": "reg_name",
     "type": "localized",
     "label": "Regional Name",
-    "universal": true
+    "universal": true,
+    "prerequisiteTag": {
+        "key": "name"
+    }
 }
--- a/data/fields/short_name.json
+++ b/data/fields/short_name.json
@@ -7,5 +7,8 @@
         "abbreviation",
         "acronym",
         "initialism"
-    ]
+    ],
+    "prerequisiteTag": {
+        "key": "name"
+    }
 }
--- a/data/fields/short_name.json
+++ b/data/fields/short_name.json
@@ -1,0 +1,11 @@
+{
+    "key": "short_name",
+    "type": "localized",
+    "label": "Short Name",
+    "universal": true,
+    "terms": [
+        "abbreviation",
+        "acronym",
+        "initialism"
+    ]
+}
--- a/interim/source_strings.yaml
+++ b/interim/source_strings.yaml
@@ -3150,6 +3150,11 @@ en:
         label: Lit
         # 'terms: lamp,lighting'
         terms: '[translate with synonyms or related terms for ''Lit'', separated by commas]'
+      loc_name:
+        # loc_name=*
+        label: Local Name
+        # 'terms: colloquial name,common name,informal name,nickname'
+        terms: '[translate with synonyms or related terms for ''Local Name'', separated by commas]'
       location:
         # location=*
         label: Location
@@ -3499,6 +3504,10 @@ en:
         placeholder: Common name (if any)
         # 'terms: label,title'
         terms: '[translate with synonyms or related terms for ''Name'', separated by commas]'
+      nat_name:
+        # nat_name=*
+        label: National Name
+        terms: '[translate with synonyms or related terms for ''National Name'', separated by commas]'
       natural:
         # natural=*
         label: Natural
@@ -3680,6 +3689,11 @@ en:
           union: Labor Union Office
           # office=water_utility
           water_utility: Water Utility Office
+      official_name:
+        # official_name=*
+        label: Official Name
+        # 'terms: formal name,full name,legal name'
+        terms: '[translate with synonyms or related terms for ''Official Name'', separated by commas]'
       oneway:
         # oneway=*
         label: One Way
@@ -4410,6 +4424,10 @@ en:
         label: Taxiway Name
         # ref_taxiway field placeholder
         placeholder: e.g. A5
+      reg_name:
+        # reg_name=*
+        label: Regional Name
+        terms: '[translate with synonyms or related terms for ''Regional Name'', separated by commas]'
       relation:
         # type=*
         label: Type
@@ -4862,6 +4880,11 @@ en:
       shop:
         # shop=*
         label: Type
+      short_name:
+        # short_name=*
+        label: Short Name
+        # 'terms: abbreviation,acronym,initialism'
+        terms: '[translate with synonyms or related terms for ''Short Name'', separated by commas]'
       shower:
         # shower=*
         label: Showers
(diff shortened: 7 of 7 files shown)


tyrasd (commented):
Should the `name` field be added as a `prerequisiteTag` of these additional name fields?

> For this round, I omitted some common name fields because they weren’t as straightforward:
> `alt_name` and `old_name` are likely to be localized and contain multiple values at the same time, but neither `semiCombo` nor `localized` provides all the necessary functionality.

The wiki says that multiple valued names are the exception: _“In rare cases, the key is used for multiple semicolon-separated names, e.g. alt_name=name1;name2;name3, but this usage is not preferred.”_ I would say that a regular `localized` field is also good enough for `alt_name` and `old_name`.

> `sorting_name` is especially relevant in some languages but completely irrelevant in others, so it would be nice to filter the `localized` field type’s language list accordingly.

ok, let's omit that one (it's also used relatively less often than the other mentioned "names" here: 20k occurrences vs. 100k+ for the others). It might be included in the future on a per-region basis, perhaps.

1ec5:
> Should the `name` field be added as a `prerequisiteTag` of these additional name fields?

That may be too strict, since I’ve heard of mappers using combinations like `highway=*` `ref=*` `noname=yes` `official_name=*` when the official name is too obscure for renderers or routers to present to end users.

tyrasd:
OK, but in my opinion that seems to be rather an exception / edge case. Also iD currently doesn't "support" the `noname` tag anyway. The benefit of having `name` as a prerequisite tag here would be that users would be nudged towards not only mapping the specialized names, but always also map the canonical `name` of a feature. :thinking:

1ec5:
`noname` + `*_name` also struck me as an odd edge case at first, but the memorial highway name designations raised in [this discussion](https://lists.openstreetmap.org/pipermail/talk-us/2021-November/thread.html#21408) are very common in some regions. It’s a complex situation, so I was worried about the editor seeming opinionated about where to put these memorial name designations. But you have a good point about `noname` not having a field of its own. The negation properties (including `not:brand:wikidata`) are usually the domain of validator suggestions rather than fields.

I’m willing to try making it a `prerequisiteTag` and revisiting later if it turns out to be a problem. When applied to a `localized` field, should iD require for example `name:fr` in order to show `alt_name:fr`, or should it only operate on the main part of the key rather than the language subkeys?

tyrasd:
> When applied to a localized field, should iD require for example name:fr in order to show alt_name:fr, or should it only operate on the main part of the key rather than the language subkeys?

Good question. I would say that for a first iteration it's ok to have the fields operate independently (only based on the main part of the key), but perhaps this could be improved further down the line. It could be slightly tricky to balance the UI for this, though: Some places (e.g. major capitals) have lots of `name:<lng>` tags while typically much fewer (translated) alternative names. A different situation would be in [multilingual regions](https://wiki.openstreetmap.org/wiki/Multilingual_names) where it could perhaps make sense to sync up locales between the fields. :thinking:

1ec5:
> I would say that for a first iteration it's ok to have the fields operate independently (only based on the main part of the key), but perhaps this could be improved further down the line.

Would this be considered a showstopper for this PR, or can the PR land as is for now while we track this idea in a separate issue? These fields are hidden by default unless already set, so clutter wouldn’t be an issue.

1ec5:
This PR was mentioned in [this forum discussion](https://community.openstreetmap.org/t/multiple-delimited-names-in-the-name-tag/6803/86?u=minh_nguyen) as a way to help mappers avoid listing multiple values in `name`, at least in some cases.

1ec5:
Also mentioned in [this forum discussion](https://community.openstreetmap.org/t/multilingual-names-in-bulgaria/98762/5) as a way to keep `name:en` from being conflated with `int_name`.

1ec5:
@tyrasd can you please give this PR another look? Mappers keep making [controversial edits](https://www.openstreetmap.org/changeset/141303559) (politically sensitive in this case) possibly because they don’t know about `alt_name`.

tordans:
@1ec5 is there a preview on this PR? Maybe its older than the preview-generator; in this case another change to the PR should re-trigger the preview generator, I think.

 I wanted to look into the preview to understand if those new files are "moreFields" or always visible. I don't think having them visible always where a good idea. It will lead to people duplicating terms just to fill out fields…

1ec5:
> I wanted to look into the preview to understand if those new files are "moreFields" or always visible. I don't think having them visible always where a good idea. It will lead to people duplicating terms just to fill out fields…

These are “universal” fields, which behave like “moreFields” but are available regardless of the preset.

Local Name, Name, National Name, Official Name, and Regional Name all listed under Add field and not in the inspector by default

tordans:
@1ec5 I looked this over to get it merged. I read the thread the way that Martin's request was to add `name` as  `prerequisiteTag` for the fields:
- https://github.com/openstreetmap/id-tagging-schema/pull/215#issuecomment-996542562
- https://github.com/openstreetmap/id-tagging-schema/pull/215#issuecomment-995975389

I would feel comfortable to merge this without further discussion once that is applied.

1ec5:
`name` is now a prerequisite for the other five name fields. I’ve also added an Alternative Name field for `alt_name=*`, as requested in https://github.com/openstreetmap/id-tagging-schema/pull/215#pullrequestreview-831413464.

As you can see from [the preview](https://pr-215--ideditor-presets-preview.netlify.app/id/dist/#background=Santa_Clara_CA_2022&disable_features=boundaries&id=w229832392&locale=en&map=21.00/37.32500/-121.91345), a `localized` field only appears based on the base key, not any localized subkeys. So even though [this way](https://www.openstreetmap.org/way/229832392) has `name=*`, `name:es=*`, and `alt_name:es=*`, the Alternative Name field doesn’t appear by default unless you manually select it from the “Add fields” menu. At that point, iD reveals that there are alternative names after all, just not one in the feature’s primary language:

Name field without Alternative Name field Name field and blank Alternative Name field with two Multilingual Name subfields for English and Spanish

This applies to the other name fields as well. I think this behavior is suboptimal because some languages simply don’t have an alternative name to speak of while others do, but I think we can track it in openstreetmap/iD#10323. It isn’t a showstopper for landing this PR.

As of openstreetmap/iD#10181, universal fields now sort below preset-specific fields in the “Add field” menu. For some presets with many optional fields, inexperienced users will be less likely to find these name fields. It’s unfortunate but not a blocker for landing this PR.

tordans:
@1ec5 the general setup of this works as expected.

I checked the wikidata item descriptions which look good as well.

One thing I am still wondering about is, what the difference between those three names actually are https://taghistory.raifer.tech/#***/loc_name/&***/reg_name/&***/nat_name/
Wikidata just repeats the name…
And the wiki is even less helpful https://wiki.openstreetmap.org/wiki/Names#Local_names_(loc_name).

My question is, as a mapper, when I see the options in the more tags dropdown, how will I know when to pick local, regional, national?

I wonder if we should at least remove the regional name with only 40k usage. Or maybe make that an unsearchable field (is there such a thing?) so we only expose what is there, but not offer it in the dropdown.

The difference for local vs. national seems a bit easier to make. Even though I would expect the name or official_name to be the national name …

1ec5:
The `loc_*`/`reg_*`/`nat_*`/`int_*` keys are consistent with `*_ref`, and we can see the same pattern in the three-letter acronyms for `network=*` in recreational route relations and some of the most common values of `flag:type=*`. (I left out `int_name=*` but should probably include it too…) The terms are perfectly intuitive to me in American English, but maybe not in some other dialects or languages?

I don’t think these fields or the documentation should clearly spell out when “local”, “regional”, and “national” apply. There are already complaints about the global “Key:name” page being too long and difficult to digest, so I don’t think going into detail about each country’s peculiarities would help matters at all. Instead, local communities such as [Canada](https://wiki.openstreetmap.org/wiki/Canada/Tagging_guidelines#Named_major_highways) should document country-specific expectations on their respective pages.

The raw usage numbers are not particularly meaningful. These keys often clarify geopolitical disputes (some of them quite petty). One I’m familiar with is [this lake](https://www.openstreetmap.org/relation/2552396), for which the national government insists on one name, the locals insist on another, and the state government has sided with the locals. `reg_name=*` will naturally see less usage than the other two, because most countries aren’t so large and complex that three-way naming discrepancies would ever arise.

This is not to say that the four-way name classification system is perfect: in the U.S., we sometimes see disagreements at a local/county/state/national level over the name of something, not to mention cross-border disagreements like [this lake](https://www.openstreetmap.org/relation/223999) that’s been the subject of a bitter disagreement over the name between two states. But those situations don’t have clear tagging solutions, whereas these tags are clear solutions for enough situations.

> Or maybe make that an unsearchable field (is there such a thing?) so we only expose what is there, but not offer it in the dropdown.

I don’t think this option currently exists. But the stakes are already _very_ low if all we’re doing is putting some more items in this menu. If this PR had been merged in its current state three years ago, I’m pretty sure we’d look back and agree that the benefits would have outweighed any confusion from not knowing the difference between a local or regional name.

This key demonstrates the point I made in https://github.com/openstreetmap/id-tagging-schema/pull/1229#pullrequestreview-2165065099: `reg_name=*` has been used widely for almost 14 years. Nominatim supports `reg_name=*`, as do Telenav’s products. Though usage of this key isn’t increasing dramatically, do we have any reason to believe that it would generate global controversy or get rejected in a vote?

tordans (approved):

tordans:
Thanks for taking the time to get into this more @1ec5. I know this is tedious, but in feel better looking at more angles for a case like this (where I plan to merge something prev. reviewed, long open and global…).

I will merge now, based on the thinking that those fields will mainly present data that is already there – which is a big win – but not cause any "lets fill out every field just because it is there" kind of mapping (which we should avoid with our presets if possible). I think the `prerequisiteTag` was a good addition to that.

I am still unsure that mappers will really understand when to use `loc_name` vs. `reg_name`. And I still thing that the wiki should give guidance on how to use those tags. Eg. one thing I notices with spot checks is that both tags are used with the same value. Which to me falls in the "better save then sorry" and "don't know what each tag will to, cannot hurt to add both" category. Which are indications for a lack of guidance and documentation. But I agree, the tags are used all over the place, so we should make them visible (https://taginfo.openstreetmap.org/keys/reg_name#map vs https://taginfo.openstreetmap.org/keys/loc_name#map vs https://taginfo.openstreetmap.org/keys/nat_name#map). And this kind of wiki updates it not a prerequisite for us merging here, IMO. Fixing the double tagging – or even talking about if that is actually a problem – is also out of scope for this repo. 
On the topic of "Unsearchable field": Agreed, the schema builder does not talk about it and the impact is low.

---

Unfortunately we missed the 1k days mark since the PR was opened (1,072 days says ChatGPT) and are a few weeks short of the actual birthdays. ;-)

1ec5:
Thanks, I do agree that there should be documentation about `reg_name=*` _somewhere_, just not on the already overloaded “Key:name” page. Someone started a bare-bones “ES:Key:reg_name” page in Spanish; a dedicated page about the key could elaborate, though I think country-specific pages would be more sustainable in the long run. Otherwise, a global page will become a pointless comparison exercise, as it has for `admin_level=*`.

#1288 tracks adding a field for `old_name=*` and #1289 tracks adding one for `int_name=*`. I omitted both fields from this PR due to more severe disagreements about how they should be used.

PR #216, opened 2021-08-07
Add Expressway field to Trunk and Primary Road presets

1ec5:
Added an Expressway field to the Trunk Road and Primary Road presets based on the [`expressway`](https://wiki.openstreetmap.org/wiki/Key:expressway) key that is commonly used in the United States. This checkbox field allows users to indicate that a road has the physical characteristics of an expressway, regardless of the road’s prominence or connectedness to the overall road network. That said, the field has been relegated to More Fields for the Primary Road preset, because a primary road is less likely to have been tagged as an expressway.

## Background

Historically, there was a consensus among U.S.-based mappers to abuse `highway=trunk` as the tag for expressways. This led to fragmented map rendering at low zoom levels, especially in states that upgrade surface streets to expressways in piecemeal fashion. Earlier this year, U.S.-based mappers [began pushing to harmonize](https://wiki.openstreetmap.org/wiki/United_States/Highway_classification) the definition of a trunk road with global practices. At the same time, American maps conventionally give expressways special treatment, similar to freeways: openmaptiles/openmaptiles#1148 ZeLonewolf/openstreetmap-americana#22. The previously obscure `expressway` key has become an attractive way to avoid dataloss as some trunk roads get reclassified as primary roads.

`expressway` is orthogonal to the [`motorroad`](https://wiki.openstreetmap.org/wiki/Key:motorroad) key commonly used in Europe.

## Caveats

The term “expressway” being used here is traffic engineering jargon. It’s unfamiliar and unintuitive outside the U.S., frequently eliciting mild confusion from mappers overseas, so I’ve limited this preset to the U.S. for now, even though `expressway` does see some use in a few other countries.

Even in the U.S., it can be a confusing term, because it isn’t the street name type that’s used interchangeably with “highway” in everyday American English. Unfortunately, this is the best we can do. Each regional dialect of American English has familiar words for high-performance roads that conflict with other dialects; there isn’t a commonly understood term for expressways across the entire country.

### DIFF
--- a/data/presets/highway/trunk.json
+++ b/data/presets/highway/trunk.json
@@ -6,6 +6,7 @@
         "oneway",
         "maxspeed",
         "lanes",
+        "expressway",
         "surface",
         "structure",
         "access"
@@ -31,6 +32,8 @@
         "highway": "trunk"
     },
     "terms": [
+        "expressway",
+        "highway",
         "road",
         "street"
     ],
--- a/data/fields/expressway.json
+++ b/data/fields/expressway.json
@@ -1,0 +1,8 @@
+{
+    "key": "expressway",
+    "type": "check",
+    "label": "Expressway",
+    "locationSet": {
+        "include": ["US"]
+    }
+}
--- a/data/presets/highway/primary.json
+++ b/data/presets/highway/primary.json
@@ -14,6 +14,7 @@
         "charge_toll",
         "covered",
         "cycleway",
+        "expressway",
         "flood_prone",
         "incline",
         "junction_line",
--- a/interim/source_strings.yaml
+++ b/interim/source_strings.yaml
@@ -948,6 +948,10 @@ en:
       expected_rwn_route_relations:
         # expected_rwn_route_relations=*
         label: Adjacent Walking Nodes
+      expressway:
+        # expressway=*
+        label: Expressway
+        terms: '[translate with synonyms or related terms for ''Expressway'', separated by commas]'
       faces:
         # faces=*
         label: Faces
@@ -5749,7 +5753,7 @@ en:
       highway/trunk:
         # 'highway=trunk\n\nTranslate the primary name. Optionally, add equivalent synonyms on newlines in order of preference (press the Return key).'
         name: Trunk Road
-        # 'terms: road,street'
+        # 'terms: expressway,highway,road,street'
         terms: '<translate with synonyms or related terms for ''Trunk Road'', separated by commas>'
       highway/trunk_link:
         # 'highway=trunk_link\n\nTranslate the primary name. Optionally, add equivalent synonyms on newlines in order of preference (press the Return key).'


Kovoschiz:
As a non-US resident. I would prefer you require it to be signed, instead of relying on users to judge "a road has the physical characteristics of an expressway" when you agreed it "The term “expressway” being used here is traffic engineering jargon". US citizens could too be confused by what this means.

1ec5:
> I would prefer you require it to be signed, instead of relying on users to judge "a road has the physical characteristics of an expressway"

Unfortunately it isn’t that straightforward: in the U.S., there is a standard sign indicating the end of an expressway, where an expressway becomes a normal surface street and the motorist has to beware of cross traffic, but there’s no sign for the beginning of an expressway. In most states, there’s also no sign for where an expressway becomes a freeway. So an expressway that connects to a freeway on both ends would have no sign, even if it’s otherwise identical to an expressway that connects to a surface street on both ends. The wiki page documents the Expressway Ends sign but does not require it because of this discrepancy.

Incidentally, freeways are also signposted in exactly the same manner, with Freeway Ends signs but no Freeway Entrance sign, except in California and a couple other states. The term “freeway” is also traffic engineering jargon that has no consistent colloquial term nationwide, yet we still tag `highway=motorway` on freeways using a Motorway preset that’s even less familiar. (Where I’m from, “motorway” might even be misinterpreted as “motor speedway”. 🏎)

Perhaps we can tighten up the definition in the wiki page’s intro paragraph and infobox so that someone unfamiliar with the term “expressway” will know what it’s about when clicking the ℹ️ button.

MoiraPrime:
A way I like to put it is that an expressway would be a road with limited cross traffic that has divided ways and has the fastest non-motorway speed limit for its region.

Mississippi examples would be US 84, which has 2 lanes in each direction, divided with median, across the entire state, bypasses every town along its route (except Laurel), has regular turnaround points and no controlled access, and has a speed limit of 65 miles per hour (as opposed to the speed of a motorway, which is 70, or a standard 2 lane road, which is 55)

1ec5:
The best definition of an expressway that I’ve come across so far is @rwelty1889’s definition of `highway=trunk` as “[a motorway wannabe](https://lists.openstreetmap.org/pipermail/talk-us/2010-February/002834.html)”. (This was back when there was consensus among U.S. mappers about tagging expressways as `highway=trunk`.)

cliffordsnow:
I'm not sure that a checkbox, adding `wannabe` or `expressway like` or  is going to stop the mis-tagging, if anything, it's likely to add more confusion, especially to people that haven't been involved in the trunk tagging issues.

1ec5:
> I'm not sure that a checkbox, adding `wannabe` or `expressway like` or is going to stop the mis-tagging, if anything, it's likely to add more confusion, especially to people that haven't been involved in the trunk tagging issues.

The point of this field is not to reduce confusion about the current contentious definition of `highway=trunk` in the United States. That’s what [this proposed redefinition of `trunk`](https://wiki.openstreetmap.org/wiki/United_States/Highway_classification) aims to do. Rather, adding this field would introduce some stability into highway classification. Edit wars over trunk road tagging will unfortunately continue until the revised `highway=trunk` definition gains widespread acceptance. As a more objective key than `highway`, `expressway` will provide some continuity so that local communities can choose to adopt the new definition at their own pace or not at all without worrying about dataloss or ambiguity.

data/presets/highway/trunk.json
```diff
@@ -31,6 +32,8 @@
         "highway": "trunk"
     },
     "terms": [
+        "expressway",
+        "highway",
```
tyrasd, line 36: if we add `highway` here as a term, we should also add it to all other `highway` presets, shouldn't we?

tyrasd (commented):
My 5 cents (I'm not from the US): This tag seems to be rather hard to use, hard to verify and is just generally quite subjective. Even after reading the [2021 us highway classification guide](https://wiki.openstreetmap.org/wiki/United_States/2021_Highway_Classification_Guidance) I think one already has to know what an "expressway" is supposed to be in order to map it "correctly". Wikipedia also [doesn't help](https://en.wikipedia.org/wiki/Expressway).

I would tend towards only including this in iD as an optional (`moreFields`) field for `highway=trunk` because then the field is only shown if the value is already set to `yes` or `no`. Always showing the field would potentially urge less experienced users to "guess" the state of the road, which could lead to errors.

Currently, about 11% of `highway=trunk` roads are mapped with the `expressway` tag (about 8,900 miles with the `yes` value and about 750 miles with `no`). `highway=primary` has an almost negligible relative usage (580 miles `expressway=yes`, 20 miles `=no`).

image

data/presets/highway/trunk.json
1ec5: The default localization is American English. In a generic American dialect, “highway” only refers to high-speed, limited- or controlled-access roads. It has a broader meaning in some dialects, but “road” or “street” is always preferred for surface streets. Meanwhile, in California, it means `route=road` among the general public and only means `highway=*` among civil engineers.

My understanding is that “highway” means `highway=*` in British English, but iD has an en-GB localization for things like that.

1ec5:
> This tag seems to be rather hard to use, hard to verify and is just generally quite subjective.

I’m sorry the documentation on the wiki leaves you with that impression. Expressways are not subjective or unverifiable, but the standards intentionally allow highway authorities wide discretion in real life. The documentation is playing a tricky balancing act as it tries to explain the concept both to non-Americans who are unfamiliar with it and to Americans who are familiar with it but have no word for it.

Alternatively, we could eliminate much of the detail and simply define an expressway as a non-freeway that qualifies for expressway-grade signage or is given the classification beneath “freeway” and “tolled” on American paper maps (https://github.com/ZeLonewolf/openstreetmap-americana/issues/22#issue-939056581), but that would be baffling to those who help us map the U.S. from abroad.

Compounding the problem is the fact that the two most established terms, “expressway” and “limited-access highway”, can both be misunderstood to mean “motorway”. But we already have significant naming problems in American English with both “motorway” and “track road”, and both of those are presets rather than mere fields.

> Currently, about 11% of highway=trunk roads are mapped with the expressway tag

This is being heralded as a successful milestone in OSMUS Slack. 😄 Until now, there was little motivation to tag `expressway` because we had been using `highway=trunk` for this purpose. The purpose of the proposed field is to support the effort of aligning American `trunk` tagging to the established international definition. It provides transparency to American mappers who would otherwise expect whatever comes after Motorway in the list of presets to correspond to expressways, either because that’s how we’ve been tagging them for years in the U.S., or because that’s how American maps and atlases classify highways.

> highway=primary has an almost negligible relative usage

As the trunk reclassification progresses in more states, we can expect the number of primary ways to increase, though maybe not the total length because these expressway segments are by definition short and disconnected.

If you think it would be prudent to phase in Expressway field, we could introduce it in `moreFields` for now with an eye toward moving it to the default fields in the future. Perhaps that could happen once usage reaches a certain level or the highway classification proposal is voted on.

data/presets/highway/trunk.json
tyrasd: OK, that makes sense, thanks!

tyrasd:
Thanks for the clarifications. 

> It provides transparency to American mappers who would otherwise expect whatever comes after Motorway in the list of presets to correspond to expressways

:+1: This is a good argument for having it it as a regular field for `highway=trunk`s.

PR #217, opened 2021-08-07
Expand usage of Advisory Speed Limit field

1ec5:
The Advisory Speed Limit field is now shown either by default or in More Fields for the Motorway Link, Trunk Link, Primary Link, Secondary Link, and Tertiary Link presets.

In Canada and the United States, if not elsewhere, advisory speed limits are normally posted at entrance and exit ramps but regulatory speed limit signs are rarely posted in these situations. Tagging `maxspeed` on a ramp could potentially lead to a more dangerous experience for users reliant on navigation UI to remind them about the most relevant speed limit. Mappers unaware of the distinction or unaware of the separate [`maxspeed:advisory`](http://wiki.openstreetmap.org/wiki/Key:maxspeed:advisory) key have been inaccurately setting `maxspeed` to the advisory speed limit. Presenting the normal field by default will hopefully make such errors less common.

When this point was originally raised in openstreetmap/iD#4870, it generated some controversy, which this PR attempts to sidestep by limiting the changes to places where it’s known to be correct. Unfortunately, the schema lacks an affordance for saying that a field should be under `fields` in some locations but `moreFields` in other locations. As an ugly workaround, this PR duplicates the Speed Limit and Advisory Speed Limit fields everywhere they’re used. At some point, we’ll need to come up with a real solution before we can consider other variations of speed limits, such as `maxspeed:hgv`, to avoid a proliferation of similar fields.

This PR also adds the Advisory Speed Limit field to the Motorway, Trunk Road, Primary Road, Secondary Road, and Tertiary Road presets, but under More Fields. Advisory speed limits are tagged on main roads in cases where road geometry requires extra care, such as around curves or hidden driveways.


tyrasd (commented):
> Unfortunately, the schema lacks an affordance for saying that a field should be under fields in some locations but moreFields in other locations. As an ugly workaround, this PR duplicates the Speed Limit and Advisory Speed Limit fields everywhere they’re used.

Did you consider to instead duplicate the `*_link` presets themselves instead of the the `maxspeed*` fields? IMHO it would be a cleaner solution, since it is actually a property of the ramps that they are typically signposted differently in the US and Canada.

1ec5:
> Did you consider to instead duplicate the `*_link` presets themselves instead of the the `maxspeed*` fields?

Oh, good idea. It would still be messy but not as messy.

> PS: not sure if mutcd is a universally known acronym (I had to google it 😅). Perhaps one could use _ca_us as a suffix, since that is what the locationSet setting actually uses?

I think you’re right about that being a better suffix. Variations of the MUTCD are used in some other countries too, but as far as I know, ramp advisory speed limit signs aren’t used in all of them.

PR #223, opened 2021-08-21
Create shoulder.json

ZEE6de:
```
{
    "key": "shoulder",
    "type": "check",
    "label": "Shoulder",
}
```


ZEE6de:
https://wiki.openstreetmap.org/wiki/Key:shoulder

shoulder.json
```diff
@@ -0,0 +1 @@
+
```
matkoniecz, line 1: This creates an empty file, is it intentional and sufficient?

matkoniecz (changes requested):

tyrasd (changes requested):
please fix the error pointed out by @matkoniecz above and please also change the highway presets directly in this pull request.

PR #225, opened 2021-08-21
Update primary.json

ZEE6de:
add shoulder tag


tyrasd:
this should be updated directly in #223

PR #227, opened 2021-08-21
Add layer=1 to building=roof preset

karsonkevin2:
`Building=roof` is used to denote a structure which is just a roof, passage underneath this roof via foot or car is possible.
https://wiki.openstreetmap.org/wiki/Tag:building%3Droof

It is pretty typical that their might be a `highway=service` &`service=drive-through` or a `highway=footway` passing underneath the roof. With the current roof preset, this leads to conflict as the roof and highways default to layer=0. In essentially all cases, the roof should be `layer=1` or a higher layer if necessary. Defaulting to `layer=1` for roof objects represents this reality.

Currently, 43% of roofs include `layer=*`, which would have to have been manually entered.
https://taginfo.openstreetmap.org/tags/building=roof#combinations

### DIFF
--- a/data/presets/building/roof.json
+++ b/data/presets/building/roof.json
@@ -8,8 +8,11 @@
         "area"
     ],
     "tags": {
+        "building": "roof"
+    },
+    "addTags": {
         "building": "roof",
-        "layer": 1
+        "layer": "1"
     },
     "matchScore": 0.5,
     "name": "Roof"
--- a/data/presets/building/roof.json
+++ b/data/presets/building/roof.json
@@ -8,7 +8,8 @@
         "area"
     ],
     "tags": {
-        "building": "roof"
+        "building": "roof",
+        "layer": 1
     },
     "matchScore": 0.5,
     "name": "Roof"


tyrasd (changes requested):
Could you please change this to use the [`addTags`](https://github.com/ideditor/schema-builder#addtags) syntax for this? Here's an [example](https://github.com/openstreetmap/id-tagging-schema/blob/v3.1.0/data/presets/man_made/bridge.json#L23). Otherwise a roof with, say, `layer=2` would not match the preset anymore.

data/presets/building/roof.json
```diff
@@ -8,7 +8,8 @@
         "area"
     ],
     "tags": {
-        "building": "roof"
+        "building": "roof",
+        "layer": 1
     },
```
peternewman, line 13: I think this will fix @tyrasd comment:
https://github.com/openstreetmap/id-tagging-schema/pull/227#pullrequestreview-830575632

Keen to get this in too having hit the same issue:
```suggestion
        "building": "roof"
    },
    "addTags": {
        "building": "roof",
        "layer": "1"
    },
```

tyrasd (approved):

data/presets/building/roof.json
tyrasd: thanks!

PR #230, opened 2021-08-23
add preset for Digital Screen

k-yle:
This adds a preset for digital screens ([`man_made=video_wall`](https://wiki.osm.org/Tag:man_made=video_wall))

video wall

### DIFF
--- a/data/fields/min_height.json
+++ b/data/fields/min_height.json
@@ -1,0 +1,9 @@
+{
+    "key": "min_height",
+    "minValue": 0,
+    "type": "number",
+    "label": "Height of bottom (Meters)",
+    "prerequisiteTag": {
+        "key": "height"
+    }
+}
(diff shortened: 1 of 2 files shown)


tyrasd (changes requested):
Could you please reformat the JSON similar to [other presets](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/advertising/billboard.json), i.e. arrays distributed to multiple lines.

k-yle:
> Could you please reformat the JSON similar to other presets

Done

tyrasd (approved):

data/fields/min_height.json
```diff
@@ -0,0 +1,9 @@
+{
+    "key": "min_height",
+    "minValue": 0,
+    "type": "number",
+    "label": "Height of bottom (Meters)",
```
1ec5, line 5: This should be in Title Case like other field labels.

tyrasd: fixed in 4e60539d328c444bf37b94e96e189be99565b261, thanks for noticing!

PR #232, opened 2021-09-06
Adding maxheight to parking garage entrance/exit

cicku:
A lot of garage entrances/exits have height restriction. Tagging the service road towards the entrance is not very accurate, so I would like to propose this change.

### DIFF
--- a/data/presets/amenity/parking_entrance.json
+++ b/data/presets/amenity/parking_entrance.json
@@ -6,7 +6,7 @@
         "access_simple",
         "address",
         "level",
-	"maxheight"
+        "maxheight"
     ],
     "geometry": [
         "vertex"
--- a/data/presets/amenity/parking_entrance.json
+++ b/data/presets/amenity/parking_entrance.json
@@ -5,7 +5,8 @@
         "ref",
         "access_simple",
         "address",
-        "level"
+        "level",
+	"maxheight"
     ],
     "geometry": [
         "vertex"


tyrasd:
Is there consensus for mapping height restrictions on the entrances of parking facilities? I don't see it mentioned on the [wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dparking_entrance).

FWIW, about 7% of `amenity=parking_entrance` objects have also a `maxheight` tag ([source: taginfo](https://taginfo.openstreetmap.org/tags/?key=amenity&value=parking_entrance#combinations)).

andrewharvey:
I think given 7% already have the tag, that's good enough consensus. I can't see why it would be an issue. I've added it to the wiki as an optional tag combination.

tyrasd:
I agree that this looks like something the proposal simply forgot to include.[^1] But I still think that this repository is not the right place to fix a "mistake" in a tagging proposal. Also, the definition of the [`maxheight`](https://wiki.openstreetmap.org/wiki/Key:maxheight) tag in the wiki also kind of contradicts the proposed usage (highlight by me):

> The tag `maxheight=*` specifies the maximum height limit for vehicles **using the way** to which the tag is added

[^1]: Or, perhaps the intended way to tag such a height restriction would be to combine the parking entrance with a nearby placed [`barrier=height_restrictor`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dheight_restrictor) node? I don't know.

cicku:
Using the way? 

Time to update the wiki.

tyrasd:
> Time to update the wiki.

@cicku are you still interested to fix this mistake in the wiki? :blush:

PR #235, opened 2021-09-11
add some NHD:* tags to discarded.json

ZEE6de:
Please add this NHD-import stuff. https://wiki.openstreetmap.org/wiki/National_Hydrography_Dataset
NHD:ComID, NHD:ReachCode and NHD:way_id have the same purpose as Tiger:tlid, NHD:Resolution is always "high", and FCode is duplicated by FType, which uses a human-readable value (NHD:FType=StreamRiver instead of NHD:FCode=46006) https://wiki.openstreetmap.org/wiki/National_Hydrography_Dataset#Attributes-to-OSM-tags 

Short: All added tags are import cruft, are just in the way and provide no value to anyone.

### DIFF
--- a/data/discarded.json
+++ b/data/discarded.json
@@ -11,6 +11,16 @@
 
   "geobase:datasetName": true,
   "geobase:uuid": true,
+  
+  "NHD:ComID": true,
+  "nhd:com_id": true,
+  "NHD:FCode": true,
+  "nhd:fcode": true,
+  "NHD:RESOLUTION": true,
+  "NHD:Resolution": true,
+  "NHD:ReachCode": true,
+  "nhd:reach_code": true,
+  "NHD:way_id": true,
 
   "osmarender:nameDirection": true,
   "osmarender:renderName": true,
--- a/data/discarded.json
+++ b/data/discarded.json
@@ -18,8 +18,6 @@
   "nhd:fcode": true,
   "NHD:RESOLUTION": true,
   "NHD:Resolution": true,
-  "NHD:ReachCode": true,
-  "nhd:reach_code": true,
   "NHD:way_id": true,
 
   "osmarender:nameDirection": true,


tyrasd:
Can you please point me to some discussion which confirms that there is consensus about the automatic discarding of these tags? At least https://wiki.openstreetmap.org/wiki/National_Hydrography_Dataset#Past_mistakes_and_NHD_fixup would indicate that there is still "use" for (some of) the `NHD:FCode`s:

> […] add intermittent=yes to streams with FCodes 46003 and 46007 on which the tag was left out for some reason.

matkoniecz:
See https://lists.openstreetmap.org/pipermail/tagging/2020-June/053408.html https://www.openstreetmap.org/changeset/3382975 

Noone really protested about `nhd:com_id` but there was also no clear aupport.

I just asked again on the Slack.

matkoniecz:
`nhd:fcode` means "intermittent stream". On `waterway=stream` with `intermittent=yes` is a pointless duplication.

On `waterway=stream` without `intermittent=yes` - not sure is it trustworthy enough to replace with `intermittent=yes`

On everything else? No idea.

matkoniecz:
> I just asked again on the US Slack.

> https://osmus.slack.com/archives/C029HV951/p1639609335055400

There is definitely distinct lack of clear support for silent mass deletion (two people commented so far, neither supported this idea of silent mass deletion in general - it appears that @ZEE6de has not checked what relevant community prefers)

1ec5:
> FCode is duplicated by FType, which uses a human-readable value (NHD:FType=StreamRiver instead of NHD:FCode=46006) https://wiki.openstreetmap.org/wiki/National_Hydrography_Dataset#Attributes-to-OSM-tags

This table shows that a single FType can correspond to many FCodes, so if anything, the FType can be deleted in favor of the FCode.

In a sense, the FCodes are duplicative of standard OSM tags. However, the mapping from FCodes to OSM tags isn’t straightforward and there are some nuances that got papered over in the original imports. These import-specific keys make it possible to trace the original intent and retag with more confidence. For example, `nhd:fcode` was the only way to understand why there were [so many apparent bodies of saltwater](https://lists.openstreetmap.org/pipermail/talk-us/2021-November/021403.html) in New Mexico and Wisconsin. `salt` had been added based on an FCode that has since been removed from newer versions of NHD, likely due to – ahem – _upstream_ quality control issues, so the community felt much more confident in recommending that the `salt` tags be removed.

ZeLonewolf:
US mapper here.  I think in general, if an import tag is nonsense, duplicating other import tags, or always the same value, it's a good candidate for discarding.

If an import tag is "information about what an authoritative source (in this case, the US government) thinks a feature is", that is unique and distinct information that should be retained unless there's community agreement that there's a problem with that data that warrants its removal.

PR #240, opened 2021-09-26
Add shooting presets

### LINKED ISSUE #237: shooting range is missing (and the current closest match is quite unfortunate, especially for people from US)
screen01

1ec5:
Added presets for Shooting Range (`leisure=pitch` `sport=shooting`), Shooting Range Facility (`leisure=sports_centre` `sport=shooting`), and Paintball Field (`leisure=pitch` `sport=paintball`) to avoid the unfortunate string match reported in #237. All three presets use the [crosshairs](https://commons.wikimedia.org/wiki/File:Font_Awesome_5_solid_crosshairs.svg) icon from FontAwesome.

A “shooting range facility” contains one or more shooting ranges and can be indoors or outdoors. This type of feature isn’t explicitly mentioned [on the wiki](https://wiki.openstreetmap.org/wiki/Tag:sport%3Dshooting), but the general pattern of sports centers containing individual pitches is well established. These facilities are often marketed as shooting clubs or rifle clubs, so there’s some overlap with [`club=shooting`](https://taginfo.openstreetmap.org/tags/club=shooting), but I left that out of this PR for simplicity.

Although [this wiki page](https://wiki.openstreetmap.org/wiki/Key:shooting) lists [`shooting=paintball`](https://taginfo.openstreetmap.org/tags/shooting=paintball) as a type of shooting range, I created a separate preset for [`sport=paintball`](https://taginfo.openstreetmap.org/tags/sport=paintball) instead. The latter tag is used several times as often, and paintball is played quite differently than most of the shooting disciplines that would be practiced at a shooting range.

Fixes #237.


data/fields/shooting.json
```diff
@@ -1,12 +0,0 @@
-{
-    "key": "shooting",
-    "type": "semiCombo",
-    "label": "Type",
-    "strings": {
-        "options": {
-            "rifle": "Rifle",
-            "clay_pigeon": "Clay Pigeon",
-            "pistol": "Pistol"
```
1ec5: I removed the Type field based on feedback from @westnordost in OSMUS Slack. I was under the impression that the `shooting=*` key indicates a discipline (e.g., [clay pigeon shooting](https://en.wikipedia.org/wiki/Clay_pigeon_shooting)), not a specific type of target or weapon used in the discipline, but I can see how these values are confusing without additional explanation. I removed the field entirely instead of turning it into a freeform combo box, because the most common value, `range`, is considered superfluous.

For what it’s worth, my only personal reference point for this sport is [this shooting club](https://www.openstreetmap.org/way/717832152) that I mapped once. I used `shooting=*` to clarify which range was for which discipline, but the particulars aren’t that important to me. The most important thing is not allowing the School preset to come up first in a search for the terms being added here.

1ec5:
Can we get some eyes on this pull request? I’m no fan of shooting sports or guns in general, but it’s such an unfortunate coincidence that searching for a shooting range currently brings up the School Building preset. It would be unjustifiable to someone who doesn’t know how presets work.

bhousel:
seems ok to me, thanks @1ec5

PR #241, opened 2021-09-27
Update wind turbine preset

cicku:
Add rotor:diameter (Meters).

image
image

### DIFF
--- a/data/presets/power/generator/source/wind.json
+++ b/data/presets/power/generator/source/wind.json
@@ -4,15 +4,14 @@
         "ref",
         "operator",
         "generator/type",
-        "generator/output/electricity"
+        "generator/output/electricity",
+        "rotor/diameter"
     ],
     "moreFields": [
         "{power/generator}"
     ],
     "geometry": [
-        "point",
-        "vertex",
-        "area"
+        "point"
     ],
     "tags": {
         "power": "generator",


matkoniecz:
It is something not really editable by a regular mapper, right? How they would get this info?

cicku:
In USA, everything can be found at USGS:

https://eerscmap.usgs.gov/uswtdb/

matkoniecz:
But it would be not used by a typical mapper, right? Even in USA?

Is this change resulting in display of that data if tagged or also showing an empty box to fill it?

1ec5:
I suppose the rationale for including `rotor:diameter` in the main fields would be the same as including `height` in the main fields for various presets, right?

cicku:
I'm not an expert on tagging schema side, sorry.

data/presets/power/generator/source/wind.json
```diff
@@ -4,15 +4,14 @@
...
-        "generator/output/electricity"
+        "generator/output/electricity",
+        "rotor/diameter"
     ],
     "moreFields": [
         "{power/generator}"
     ],
     "geometry": [
-        "point",
-        "vertex",
```
tyrasd, line 14: I would tend to keep `vertex` as an allowed geometry type, since a wind turbine might be mapped as being directly connected to an underground power cable for example. see #270

tyrasd (changes requested):
Thank you for the suggestion. This needs a little more work, however:

* move the new field to `moreFields`
* add the JSON definition for the new field `rotor/diameter` (should be quite similar to the [`height`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/height.json) field, if you need a template)

tyrasd (approved):

tyrasd:
Sorry, I've accidentally merged this PR prematurely. #361 reverted the merge. Would you mind to open a new PR for this branch @cicku? Please note that some issues would still need to be addressed: see https://github.com/openstreetmap/id-tagging-schema/pull/241#pullrequestreview-830537649

eumiro:
> It is something not really editable by a regular mapper, right? How they would get this info?

Many wind turbines accessible by hikers/cyclists have some sort of a small plaque or information board with some technical information on the turbine that can be used by mappers.

Came here from https://github.com/Zverik/every_door because it uses `id-tagging-schema` for presets. There are a few more tags that could be used, such as `height:hub` instead of `height`, or missing `generator:model` (or `manufacturer:type` or `model`?).

tyrasd:
89aaad1 adds the `rotor:diameter` and `height:hub` fields.

@eumiro… for the model/type I'm currently not sure if it makes sense to include in iD. At first glance, this seems to be rather specific technical information which a regular mapper would have a hard time to find out, isn't it? If you like, you may open a new issue to discuss this further. :slightly_smiling_face:

PR #249, opened 2021-10-14
Add Branch field to accompany Brand field

1ec5:
Added a Branch field corresponding to the [`branch`](https://wiki.openstreetmap.org/wiki/Key:branch) key. Added the field to every preset that has a Brand field.

## Rationale

This field represents the name of an individual location affiliated with a store chain, restaurant chain, etc. Since openstreetmap/iD#8305 (I think), iD’s validator has been splitting some `name`s into a combination of `brand` and `branch`: openstreetmap/iD#8734. Adding this field to the “More fields” of each preset keeps subsequent mappers from thinking the branch name has gone missing and unknowingly adding some redundancy in `name`. This field will also complement the other name fields being added in #215.

## Caveats

In American English, this concept is only called a “branch” for certain kinds of POIs that are rarely franchised, such as banks and post offices. For other kinds of POIs, such as restaurants and supermarkets, the term “store location” is much more common. (For example, this is the term that usually precedes the branch name on a receipt.) But naming the field “Store Location” would mislead some mappers to enter a description of the store’s location, along the lines of the `is_in` key. I think we’d just have to live with the lack of a great name for this key. That said, if it’s even harder to come up with a sensible field name in other languages, we could split up the field into more variations to make translation easier.

I named the field’s definition file “branch_brand.json” because the `branch` key is also commonly used on rail features, where it means something quite different: the name of a branch line. This PR doesn’t add a field for the branch line name, because [this wiki page](https://wiki.openstreetmap.org/wiki/Key:branch#Railway_branch) is unclear as to whether that usage is still considered correct.

### DIFF
--- a/data/fields/branch_brand.json
+++ b/data/fields/branch_brand.json
@@ -1,0 +1,5 @@
+{
+    "key": "branch",
+    "type": "text",
+    "label": "Branch"
+}
--- a/data/presets/amenity/atm.json
+++ b/data/presets/amenity/atm.json
@@ -8,6 +8,7 @@
         "drive_through"
     ],
     "moreFields": [
+        "branch_brand",
         "brand",
         "covered",
         "height",
--- a/data/presets/amenity/bank.json
+++ b/data/presets/amenity/bank.json
@@ -12,6 +12,7 @@
     ],
     "moreFields": [
         "air_conditioning",
+        "branch_brand",
         "brand",
         "currency_multi",
         "email",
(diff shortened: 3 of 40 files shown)


bhousel:
> Oh, that script can define fields that the tagging schema doesn’t define as part of the unbranded preset? That’s powerful.

It can maybe, but I don't know what will happen when iD goes to localize the field.  It might throw a warning about "no English translation for the 'branch' field".

PR #251, opened 2021-10-19
Add translatable tunnel types

1ec5:
Added human-readable, translatable display names for some documented values of `tunnel=*`.


tyrasd:
Should we also add these strings to https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/tunnel.json? As far as I can see, the `tunnel_combo` field is only used in the `man_made=tunnel` preset, which has [relatively little usage](https://taginfo.openstreetmap.org/compare/man_made=tunnel/tunnel/man_made=bridge) I would say.

PR #252, opened 2021-10-20
add some popular cuisines that appear to be valid

matkoniecz:
this would allow translating this values

triggered by https://github.com/streetcomplete/StreetComplete/issues/103#issuecomment-947204418

based on https://taginfo.openstreetmap.org/keys/cuisine#values

### DIFF
--- a/data/fields/cuisine.json
+++ b/data/fields/cuisine.json
@@ -26,6 +26,7 @@
             "thai": "Thai",
             "korean": "Korean",
             "american": "American",
+            "seafood": "Seafood",
             "sushi": "Sushi",
             "ice_cream": "Ice Cream",
             "german": "German",
--- a/data/fields/cuisine.json
+++ b/data/fields/cuisine.json
@@ -32,7 +32,7 @@
             "german": "German",
             "greek": "Greek",
             "vietnamese": "Vietnamese",
-            "turkish": "Turkish"
+            "turkish": "Turkish",
             "lebanese": "Lebanese"
         }
     },
--- a/data/fields/cuisine.json
+++ b/data/fields/cuisine.json
@@ -14,7 +14,6 @@
             "japanese": "Japanese",
             "mexican": "Mexican",
             "kebab": "Kebab",
-            "kebab;pizza": "Kebab and pizza",
             "indian": "Indian",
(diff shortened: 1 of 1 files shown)


data/fields/cuisine.json
```diff
@@ -14,18 +14,26 @@
             "japanese": "Japanese",
             "mexican": "Mexican",
             "kebab": "Kebab",
+            "kebab;pizza": "Kebab and pizza",
```
1ec5, line 17: It isn’t necessary to include semicolon-delimited lists, because the `semiCombo` field allows users to mix and match an arbitrary number of these individual values to form a semicolon-delimited list in the raw tag.

Here’s what the field looks like in iD right now. None of the options reflect the strings in this file yet, probably because the latest version of the tagging schema hasn’t been pulled in yet, but at least you can get a sense for how it’ll eventually look:

semiCombo

data/fields/cuisine.json
```diff
@@ -16,16 +16,23 @@
...
             "asian": "Asian",
             "french": "French",
+            "spanish": "Spanish",
+            "portugese": "Portugese",
+            "russian": "Russian",
+            "polish": "Polish",
             "thai": "Thai",
+            "korean": "Korean",
             "american": "American",
             "sushi": "Sushi",
```
peternewman, line 29: If we're adding fish, we should definitely add the broader seafood (and probably fish_and_chips as well):
```suggestion
            "seafood": "Seafood",
            "sushi": "Sushi",
```

data/fields/cuisine.json
```diff
@@ -16,16 +16,24 @@
             "kebab": "Kebab",
             "indian": "Indian",
             "chicken": "Chicken",
+            "fish": "Fish",
             "asian": "Asian",
             "french": "French",
+            "spanish": "Spanish",
+            "portugese": "Portugese",
```
peternewman, line 23: Fix the typo Codespell found:
```suggestion
            "portuguese": "Portuguese",
```

tyrasd:
Thanks for this improvement!

> After this PR is merged/rejected/get requested changes I would be happy to further look for valid cuisine values.

Please go ahead! :smiley:

PR #261, opened 2021-11-03
Add Population Date and Population Source fields

1ec5:
Added Population Date and Population Source fields that only show up if Population is already tagged on the feature. Unlike the Source field, Population Source completes the most popular values on taginfo instead of a fixed set of known values, since there are so many possible values depending on the region.

Ideally, [`census:population`](https://wiki.openstreetmap.org/wiki/Key:census:population#Cleanup) would be deprecated in favor of a combination of `population` and `population:date`: openstreetmap/iD#6678. However, I don’t think the deprecated.json format supports splitting a tag value across two keys (which is one of the problems with `census:population` in the first place).

Fixes openstreetmap/iD#6672 except for the validation ideas in there. In my opinion, it would be worth visually grouping the fields first to set a user expectation that the fields are related: openstreetmap/iD#8104.


tyrasd:
> In my opinion, it would be worth visually grouping the fields first to set a user expectation that the fields are related: openstreetmap/iD#8104.

:+1: 

> Ideally, census:population would be deprecated in favor of a combination of population and population:date: openstreetmap/iD#6678. However, I don’t think the deprecated.json format supports splitting a tag value across two keys (which is one of the problems with census:population in the first place).

After reading the [cleanup](https://wiki.openstreetmap.org/wiki/Key:census:population#Cleanup) section of the corresponding wiki page, I think this is a bit more complex than what one could reasonably model in the context of the `deprecated.json` data. To me it looks like as if such situations would need to get "hard" coded validators for each explicit tag, wouldn't they?

1ec5:
Right, I only meant that `census:population` would ideally be deprecated, but we’re far from that point and I don’t think iD’s tag deprecation mechanism is the right tool for the job. (It’s probably a job for MapRoulette.) This PR would assist any manual efforts at cleaning up population data by making the approved tags more visible.

PR #262, opened 2021-11-04
Rename turn restriction presets to more common names

1ec5:
Renamed most of the turn restriction presets (just the names, not the tags) to match the most common names for these restrictions in English. For example, “Only Right Turn” is now “Right Turn Only”, which reads more naturally. The presets were originally named after the tag values verbatim, but the values were apparently optimized for pattern matching, not human readability.

### DIFF
--- a/data/presets/type/restriction/only_straight_on.json
+++ b/data/presets/type/restriction/only_straight_on.json
@@ -11,5 +11,5 @@
         "key": "restriction",
         "value": "only_straight_on"
     },
-    "name": "Only Straight On"
+    "name": "No Turns"
 }
--- a/data/presets/type/restriction/only_straight_on.json
+++ b/data/presets/type/restriction/only_straight_on.json
@@ -11,5 +11,5 @@
         "key": "restriction",
         "value": "only_straight_on"
     },
-    "name": "No Turns"
+    "name": "Straight Ahead Only"
 }
--- a/data/presets/type/restriction/no_straight_on.json
+++ b/data/presets/type/restriction/no_straight_on.json
@@ -11,5 +11,5 @@
         "key": "restriction",
         "value": "no_straight_on"
     },
-    "name": "No Straight On"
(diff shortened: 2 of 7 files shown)


data/presets/type/restriction/only_straight_on.json
```diff
@@ -11,5 +11,5 @@
         "key": "restriction",
         "value": "only_straight_on"
     },
-    "name": "Only Straight On"
+    "name": "No Turns"
```
1ec5, line 14: This is the most drastic change in the PR. As [this wiki page](https://wiki.openstreetmap.org/wiki/Relation:restriction#Road_signs) notes, signs in the United States literally say [No Turns](https://commons.wikimedia.org/wiki/File:MUTCD_R3-3.svg), whereas “Only Straight On” is an odd-sounding construction for a couple different reasons.

RudyTheDev: I would add that this wording makes a somewhat English-centric assumption. "No turns" does not imply "no U-turn" in languages where "turn" and "u-turn" are different words. As a similar example, "No left turn" (in right-side driving countries) does not forbid a "u-turn" by default. I feel like "No turns" sounds too much like "No left or right turn". Something like "straight only" removes this ambiguity and is easier to parse. Anyway, that's my impression from a non-US/non-native English speaker.

1ec5: This repository’s default locale is American English (`en-US`), so unfortunately there’s no other way to customize the wording for American English speakers other than to change the source string.

In principle, any language that would make a distinction between “turns” and U-turning can still use the appropriate terminology for this preset, including a literal translation of “Straight Only”. Unfortunately, I don’t think there’s currently a way for these preset files to supply comments to translators, so that context might be lost and translators might end up translating the string more literally.

In the meantime, “Straight Ahead Only” would be a reasonable compromise; thanks for the suggestion!

Issue #254, opened 2021-10-21
Make denomination field translatable


matkoniecz:
It would be useful to provide translations for non-English speakers

1ec5:
Should we split up the Denomination field into multiple fields, one per religion? That would allow us to filter out irrelevant values, which would be a significant usability improvement. On the other hand, the list of denominations for some religions is rather open-ended, and the less common denominations are the ones that benefit the most from a taginfo-based list.

matkoniecz:
> Should we split up the Denomination field into multiple fields, one per religion? That would allow us to filter out irrelevant values, which would be a significant usability improvement. 

That would be also very useful for some other data consumers. For example in https://github.com/streetcomplete/StreetComplete/issues/1737 I would be able to limit to list of Christian denomination/rites/churches without maintaining own list which `denomination` values are Christian.

> On the other hand, the list of denominations for some religions is rather open-ended, and the less common denominations are the ones that benefit the most from a taginfo-based list.

Is it possible to show Christian denominations/rites/churches and then taginfo values which exclude known Muslim/etc denominations?

And for mosques show Muslim denominations/variants and then taginfo values which exclude Christian/etc denominations?

Even just showing all denominations specific to given religion and then all taginfo values would be significant usability improvement.

1ec5:
> Is it possible to show Christian denominations/rites/churches and then taginfo values which exclude known Muslim/etc denominations?
> 
> And for mosques show Muslim denominations/variants and then taginfo values which exclude Christian/etc denominations?

Not really, because the relationships between the religions and denominations isn’t defined anywhere in the tagging system. Only Wikidata has that information in machine-readable format.

We could start to curate denominations by creating a separate denomination field for each religion. Each field would have `type` set to `combo` and `autoSuggestions` set to `false`. Users would still be able to enter unlisted values, as long as `customValues` remains `false`, but we’d have to keep an eye on the most popular `denomination` values to make sure they’re getting surfaced in the dropdown as localizable suggestions.

matkoniecz:
> Not really, because the relationships between the religions and denominations isn’t defined anywhere in the tagging system. Only Wikidata has that information in machine-readable format.

To clarify: I was proposing to start recording this info in this preset.

And OSM Wiki has already this classification for at least more popular ones

tyrasd:
> Users would still be able to enter unlisted values, as long as `customValues` remains `false`

you mean as long as `customValues` remains `true`, right? :wink:

PR #263, opened 2021-11-04
Add translatable strings for Religion field

1ec5:
Added translatable strings to the Religion field for the most common documented values of the `religion` key. This PR includes each value with at least 100 occurrences and some mention on [this wiki page](https://wiki.openstreetmap.org/wiki/Key:religion). There are many more `religion` values in use, but the less common values would likely be more difficult for translators to translate. They’ll still be visible in the dropdown, just not translated yet.

Working towards #254.

### DIFF
--- a/data/fields/religion.json
+++ b/data/fields/religion.json
@@ -8,7 +8,7 @@
             "benzhu": "Benzhu",
             "buddhist": "Buddhist",
             "caodaism": "Caodaist",
-            "chinese_folk": "Chinese Folk",
+            "chinese_folk": "Chinese Folk Religion",
             "christian": "Christian",
             "confucian": "Confucian",
             "hindu": "Hindu",
@@ -24,7 +24,7 @@
             "taoist": "Taoist",
             "tenrikyo": "Tenrikyo",
             "unitarian_universalist": "Unitarian Universalist",
-            "vietnamese_folk": "Vietnamese Folk",
+            "vietnamese_folk": "Vietnamese Folk Religion",
             "voodoo": "Voodoo"
         }
     }
--- a/data/fields/religion.json
+++ b/data/fields/religion.json
@@ -1,5 +1,31 @@
 {
     "key": "religion",
     "type": "combo",
-    "label": "Religion"
+    "label": "Religion",
+    "strings": {
+        "options": {
+            "bahai": "Bahá’í",
+            "benzhu": "Benzhu",
+            "buddhist": "Buddhist",
+            "caodaism": "Caodaist",
+            "chinese_folk": "Chinese Folk",
+            "christian": "Christian",
+            "confucian": "Confucian",
+            "hindu": "Hindu",
+            "jain": "Jain",
+            "jewish": "Jewish",
+            "multifaith": "Multifaith",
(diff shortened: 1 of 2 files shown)


data/fields/religion.json
```diff
@@ -1,5 +1,31 @@
...
     "type": "combo",
-    "label": "Religion"
+    "label": "Religion",
+    "strings": {
+        "options": {
+            "bahai": "Bahá’í",
+            "benzhu": "Benzhu",
+            "buddhist": "Buddhist",
+            "caodaism": "Caodaist",
+            "chinese_folk": "Chinese Folk",
```
1ec5, line 11: This and “Vietnamese Folk” read a bit awkward. I wonder if the strings should be nouns rather than adjectives; “Chinese Folk Religion” sounds less awkward.

matkoniecz: Chinese Folk Religion seems far better but I am not a native speaker

tyrasd (commented):
> I wonder if the strings should be nouns rather than adjectives

My first tendency was also pro nouns instead of adjectives, but I'm also not a native speaker. :speak_no_evil: 

I think what we should do is using the same scheme for this field and the [`denomination`](https://wiki.openstreetmap.org/wiki/Key:denomination) field. Does that change anything for this PR? :thinking:

data/fields/religion.json
1ec5: Maybe we can make an exception for the particular entries that don’t have a natural adjective form.

1ec5:
> I think what we should do is using the same scheme for this field and the denomination field. Does that change anything for this PR? 🤔

Denomination will be a much more difficult task, because it’s so open-ended. (My past self from 2010 should’ve campaigned hard to make it a freeform text key similar to `operator`/`operator:wikidata`…) I’m unsure which form to use for that field. For some denominations (`roman_catholic`), the adjective form is more natural, but plenty of other denominations lack an adjective form in the first place (`assemblies_of_god`, `iglesia_ni_cristo`). We can get by with a few exceptions in Religion, but Denomination will be all over the place.

matkoniecz:
@tyrasd Is some extra research/help/review required here? I am not a native speaker but may try recruiting some.

(I am interested in denomination stuff as I plan to use it for https://github.com/streetcomplete/StreetComplete/issues/1737 )

1ec5:
Upon reflection, I think we should go with the adjective forms of religions, where available. People sometimes need to coin new `religion` values. If they see that most of the menu items are -isms, they may perceive a need to include -ism in the raw tag value, even though that hasn’t been the general practice. This consideration is specific to English, which happens to be the language of raw tag values; translators can choose the form most suitable in their respective languages.

tyrasd (approved):
> Upon reflection, I think we should go with the adjective forms of religions, where available.

Sorry for the delay in the review of this PR. I'm OK with merging it as is, but I do also agree that "Chinese/Vietnamese Folk" sounds a bit strange. It's also not really the adjective form of the religion as far as I understand. I'd therefore tend to make an exception for these entries, and use the full "* Folk Religion" name of these religions instead. Or what do you think @1ec5 ?

PR #264, opened 2021-11-06
Replace obsolete `cycleway=shared` with `cycleway=shared_lane`

MingweiSamuel:
`cycleway=shared` (and `:left`/`:right` versions) should be replaced with `cycleway=shared_lane`. The former is obsolete and doesn't render on the most popular cycle maps.

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1754,5 +1754,17 @@
   {
     "old": {"wood": "mixed"},
     "replace": {"leaf_cycle": "mixed", "leaf_type": "mixed"}
+  },
+  {
+    "old": {"cycleway": "shared"},
+    "replace": {"cycleway": "shared_lane"}
+  },
+  {
+    "old": {"cycleway:left": "shared"},
+    "replace": {"cycleway:left": "shared_lane"}
+  },
+  {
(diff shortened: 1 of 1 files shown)


tyrasd:
Can you point to a source which supports this claim?

https://wiki.openstreetmap.org/wiki/Key:cycleway doesn't mention that `cycleway=shared` is obsolete or deprecated[*].

> `cycleway=shared_lane`
> Cyclists share a lane with motor vehicles, and there are markings indicating that motorists and cyclists should share this lane.
> […]
> `cycleway=shared`
> In general, this indicates that cyclists share space with other traffic on this highway.

From what I can see there is a distinction: `shared_lane` must have markings, while `shared` doesn't.

[*] except for the use in combination with `highway=cycleway` where the wiki says that:

> [`cycleway=shared`] has also formerly been used on cycleways which were mapped as separate ways tagged as `highway=cycleway` before the `segregated=*` tag was formalized. Its use with cycleways is now considered obsolete: purpose-built cycleways are assumed to be shared.

PS: usage comparison

taghistory

MingweiSamuel:
Seems I misread that * on the wiki.

PR #266, opened 2021-11-10
Catch and fix a cuisine typo in tagging

peternewman:
Off the back of #252

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -455,6 +455,10 @@
     "replace": {"diet:kosher": "*"}
   },
   {
+    "old": {"cuisine": "portugese"},
+    "replace": {"cuisine": "portuguese"}
+  },
+  {
     "old": {"cuisine": "vegan"},
     "replace": {"diet:vegan": "*"}
   },


matkoniecz:
7 cases worldwide, may be better suitable for manual fix https://taginfo.openstreetmap.org//search?q=cuisine%3Dportugese

peternewman:
Possibly, but as your typo shows there's no guarantee more won't appear in future.

1ec5:
The deprecation feature is primarily intended for intentional changes in the tagging conventions, but there’s such a long tail of possible typos. Better to merge #252 and enhance iD so that the suggested values are filtered based on fuzzy matching based on edit distance, similar to how preset search currently catches typos.

matkoniecz:
JOSM already detect automatically such typos (close matches to values in presets).

Maybe making specialized script/editor that is using https://github.com/codespell-project/codespell dictionary and/or fuzzy matching, with human reviewing typos would be useful?

1ec5:
A dictionary would be for catching typos in freeform text, but since values such as `portuguese` are already known, iD can simply rank suggestions by edit distance: openstreetmap/iD#8802.

PR #288, opened 2021-12-02
rename highway=track preset

tyrasd:
This would bring it in line with the [OSM wiki](https://wiki.openstreetmap.org/wiki/Tag:highway=track), solve at least some ambiguous situations (like https://github.com/openstreetmap/iD/issues/6954 for example) and IMHO makes the description more precise. See also the entry in the [controversial decisions](https://wiki.openstreetmap.org/wiki/ID/Controversial_Decisions#Calling_a_highway.3Dtrack_an_.22unmaintained_track_road.22) page of the OSM wiki. What do you think?

//edit: some alternative spelling versions have been proposed below:

* `Land Access Road`
* `Track Road`
* `Land Access Track Road`
* `Land Access / Track Road`
* `Off Road / Not Maintained`


bhousel:
"Minor Land Access Road" is even more ambiguous and doesn't raise enough of a signal to untrained mappers that these roads are potentially hazardous to drive and route on. 

[We really did put a lot of thought into the original name](https://github.com/openstreetmap/iD/issues/4092#issuecomment-306639206), and it bothers me a lot that this keeps coming up for discussion.

ArloJamesBarnes:
Also, it makes me wonder what a minor land is.

bhousel:
> Also, it makes me wonder what a minor land is.

It implies the existence of a "Major Land Access Road"

ZeLonewolf:
I think "minor" is unneeded.  I would prefer it simply be "Land access road".  IMO that's the most succinct description at this point given the very extensive community discussions on this topic.

I also agree that "minor" implies that there are also major land access roads which doesn't make sense to me. Land access road are minor roads by definition.

zekefarwell:
I was the one who came up with the phrase "minor land access road" for the wiki page, and while I think it is a good description, I don't think it makes a very good preset name. The complete phrases from the wiki page are even more descriptive, but much too long for a name:

> minor land access roads that are not considered part of the general-purpose road network

> Minor land access roads like farm or forest tracks

I think it is important to include the actual tag name "track" in the preset name so mappers using different editors aren't using wildly different names to refer to the same thing.  

Other road presets include the underlying tag name:
- Motorway
- Trunk Road
- Primary Road
- Secondary Road
- Tertiary Road
- Minor/Unclassified Road
- Residential Road
- Service Road

Seems to me that "Track Road" would work just fine for a preset name.  It may not  mean much to a new mapper or be easily misinterpreted, but for better or worse most of the other road preset names suffer from this same problem.

ZeLonewolf:
Missing from this thread is a link to the extensive March 2021 discussion on the tagging list, where the current wiki description was discussed:
https://lists.openstreetmap.org/pipermail/tagging/2021-March/060353.html

adamfranco:
How about "**Land Access Track Road**"?

It gets at the land-access concept that differentiates `highway=track` from `highway=service|unclassfied|residential`, includes the actual tag name, and doesn't make a maintenance distinction that might not be valid in some cases.

Dimitar5555:
`Agricultural Road` could also be an option.

zekefarwell:
> Also, it makes me wonder what a minor land is.

This is a good point.  The wording on the wiki page should probably be tweaked to "minor, land-access road" so as not to be interpreted as "minor-land, access road" (whatever that would be 😀).

ZeLonewolf:
> How about "**Land Access Track Road**"?
> 
> It gets at the land-access concept that differentiates `highway=track` from `highway=service|unclassfied|residential`, includes the actual tag name, and doesn't make a maintenance distinction that might not be valid in some cases.

If we follow the example of `highway=unclassified`'s preset, this could be:

Land Access / Track Road

pamolloy:
> "Minor Land Access Road" is even more ambiguous and doesn't raise enough of a signal to untrained mappers that these roads are potentially hazardous to drive and route on.

I don't think this is universally true. It is generally accurate in the United States, but not in Germany where I currently live. Note that `highway=track` may be paved. There are other tags that describe whether it is hazardous.

bhousel:
> I don't think this is universally true. It is generally accurate in the United States, but not in Germany where I currently live. Note that `highway=track` may be paved. There are other tags that describe whether it is hazardous.

It is not universally true and was never meant to be (like any definition of any tag).  It does not mean "paved", and never has.   The German localization of iD calls it a Forest/Farm Road or something, and that's fine..  

All of this has been hashed out _extensively_ going back years.  We're shouldn't break a thing that's working mostly around the world just fine.

tyrasd:
> The […] localization of iD calls it […] or something, and that's fine..

This argument is invalid in my opinion. I strongly think that the English locale is special in the sense that it is the default for languages which iD is not yet translated in and there are also quite a lot of people choosing to run their OS in English instead of any local language. English is also the basis from which all translations of iD are derived from. It is comparable to the status of the English pages of the tag descriptions in the OSM wiki.

Therefore, the English version of any of iD's preset strings must take particular attention to be precise, unambiguous, clear and uncontroversial. We owe this to the OSM community.

bhousel:
> Therefore, the English version of any of iD's preset strings must take particular attention to be precise, unambiguous, clear and uncontroversial. We owe this to the OSM community

I agree you owe it to the OSM community.  "Minor Land Access Road" is not any of those things.  But anyway go ahead and rename it.  I've said why I think it's wrong, and you're going to do whatever you want anyway.

tyrasd:
> you're going to do whatever you want anyway

I am going to do whatever the community is in favor for.

bhousel:
> I am going to do whatever the community is in favor for. 

You're going to do whatever 5 loud people are in favor for, but not the thing that makes the map better or the thing that makes iD easier for new users to comprehend.

tyrasd:
At the moment you are the loudest one here, Brian. We definitely heard you and know what you are in favor for.

Let's wait and listen to the all arguments.

woodpeck:
Good decision and long overdue. Currently, anyone with an English-langauge ID editor clicks on a forest track in Germany is informed that this is "unmaintained" which is not what highway=track means in Germany. Same for many other countries. I agree with the sentiment that the English locale needs to work everywhere and cannot be tailored to whatever the situation in one English-speaking country might be. I wish the community had been involved in a similar fashion before word "unmaintained" was introduced.

bhousel:
For comparison, Waze calls them "Off Road / Not Maintained", and has a pretty good description and guidance of how to use this road class in their editor.
https://wazeopedia.waze.com/wiki/USA/Road_types#Off-road_.2F_Not_maintained_Road

It does sound like the situation in Germany is kind of special.  Are your tracks plowed after a snow storm?  If so, why not just call these maintained / paved / publicly accessible roads `highway=service`?   I've never really gotten a great answer to that question.  The whole point of adding this stuff is to build a useful map.  Please consider what Waze does an an example worth emulating.

matkoniecz:
"Land access road" seems to be a fitting name for me, "Land Access Track Road" and "Land Access / Track Road" also seems fine

Thanks to @ZeLonewolf for linking https://lists.openstreetmap.org/pipermail/tagging/2021-March/060353.html

> Agricultural Road could also be an option.

Many are for accessing only forest land and while technically forest can be defined as special kind of agriculture it is unlikely to be clear. Also, see https://lists.openstreetmap.org/pipermail/tagging/2021-March/060353.html - it discusses use of this value to describe access to land that is neither forest nor fields.

> We really did put a lot of thought into the original name, and it bothers me a lot that this keeps coming up for discussion.

It is mostly caused by fact that it ignored how it is actually used in many places outside USA and described all `highway=track` as not maintained

Maybe significant thought was put into it, but it was attempt to change meaning how this tag is actually used and its meaning.

> why not just call these maintained / paved / publicly accessible roads `highway=service`?  I've never really gotten a great answer to that question. 

iD issue tracker is not a good place to change meaning of established tags or to introduce new ones without discussing it with wider community. Doing it is why "this keeps coming up for discussion"

Maybe it is silly to tag maintained paved forestry roads as `highway=track` and should be done differently but issue tracker of a  specific editor is not place to make such changes in general OSM tagging.

Maybe it would be useful to have special road type for "hazardous road" and so on, but in my opinion editors and presets should not be place where new tagging appears without discussion.

To be clear: often it is very useful to change established tagging of fix documentation. But authors of editors should consult it with others rather than instruct other to obey their ideas. Even if they think that their ideas are clearly superior. No idea whether I was successful with this, but when I encounter poor tagging scheme or misleading documentation I try to consult with others and discuss it in cases where there situation is not fully clear. 

> Are your tracks plowed after a snow storm?

In Poland it would be really unusual, and in places where it happens it was never clear to me why it was done.

But there are also other types of road maintenance. Which are often applied to road of this type (especially `tracktype=grade1` `=grade2` `=grade3`), so describing them as not maintained is a poor idea and misleading.

And even if that is poor idea - `tracktype=grade1`, `tracktype=grade2` were quite clearly established with their meaning. So I am not surprised that there is opposition to "Unmaintained Track Road" used to describe all `highway=track`

cliffordsnow:
I live in an area filled with `highway=track`, both for traditional agriculture use as well as for logging. Suggesting the use of _minor land access_ would just be confusing to most mappers that don't spend time reading the wiki. I suspect that many mappers just rely on what the editor suggest. It's difficult enough to determine if the road should be a service or a track. For example, when does the way stop being `highway=service` and become `highway=track` on a farm. 

@bhousel suggestion of using Waze term, "Off Road / Not Maintained" make much better sense. It clearly defines logging roads much better than minor land access. Trust me that `highway=track` is not a minor road when logging is involved. And they certainly are not suitable for passenger vehicles which is another problem with minor land access.

matkoniecz:
> Off Road / Not Maintained

This is quite good and not implying that all `highway=track` are not maintained. But still would be a poor label for `highway=track surface=asphalt tracktype=grade1` 

Maybe separate label for paved / `tracktype=grade1`  would be a good idea?

zekefarwell:
> For comparison, Waze calls them "Off Road / Not Maintained", and has a pretty good description and guidance of how to use this road class in their editor.
https://wazeopedia.waze.com/wiki/USA/Road_types#Off-road_.2F_Not_maintained_Road

While there is overlap between what Waze calls "Off-road / Not Maintained" and what OSM calls "highway=track", they don't entirely line up.  The distinction between highway=service and highway=track is what the road's primary purpose is.  highway=service primarily provides access to buildings or other human structures/facilities, service.  highway=track primarily just provides access to land (forest, fields, etc).  My attempt to illustrate visually: 
 
access-roads-osm

With Waze it looks like the same spectrum of roads is covered by three types:  "Off-road / Not Maintained", "Private Road", and "Parking Lot Road".  The lines between these types fall in different spots:

access-roads-waze

As the graphics show, physical conditional matters for Waze classification, but not for OSM classification.  I prefer the way Waze does it, but that's just not how is in OSM.  In my area the distinction between roads that a normal car can definitely drive on vs one that may require high clearance or 4 wheel drive is pretty important, but perhaps in other parts of the world this is less useful.  In OSM some Waze Private Roads would be service and some would be track.  Most Off-road / Not Maintained roads would be track, but a few might be service if they provide access to structures or facilities.  An attempt could be made to try and change the meaning of highway=track and highway=service to line up with the Waze classifications, but to me that seems like a major uphill battle doomed to failure.

zekefarwell:
> For example, when does the way stop being `highway=service` and become `highway=track` on a farm.

Roads leading from the public road to the farm house, barn, and other farm buildings are `highway=service`, roads leading beyond the buildings into the fields are `highway=track`.  Why is this distinction useful?  I don't know ¯\_(ツ)_/¯.  But that is my understanding of how the tags are intended to be used.

lgommans:
What I think when I hear...

> Unmaintained Track Road

Definitely not a good quality asphalt path, which `highway=track` very well can be

As for unmaintained, if the farmer maintains it himself it can be maintained just fine. She or he definitely has a vested interest in accessing major fields, can look better maintained than some official roads.

> Minor Land Access Road

`highway=service` is for access roads, and "land" is super broad. Someone who owns a building also owns the land, so any minor roads leading to that land, like a service road on the backside? This is simply the first thing I think of, I'm not purposefully trying to stretch the definition, but this sounds like it would cause a lot of mistagging for private roads. (Plus the "minor land" ambiguity, but that's resolvable with a comma.)

> Track

Anything where tracks show, typically from a tractor. I guess we all agree that this would be a bad name, but basically what we should look for is the difference between what track means and what the tag means.

> Land Access Track Road

Same "land" access problem. For me it really just triggers the idea of access to the land that you/someone own/s. Maybe that's just me though, maybe someone should ask an OSM noob, e.g. girl/boyfriend, since those are most likely to actually be the audience impacted by this change.

What I like about this option is that it includes the actual tag name, making it somewhat recognisable also for experts.

> Agricultural Road

I find this the most clear value so far suggested in the thread, but then it got 5 downvotes 0 upvotes at the time of writing. I dunno what's unclear about a road meant for agriculture because that's exactly what it is. Matkoniecz argued that "*Many [track roads] are for accessing only forest land and while technically forest can be defined as special kind of agriculture it is unlikely to be clear.*" That's how I would interpret it, so for me it is clear. But perhaps that's because I'm already in the know? Hard to say for certain.

> Land Access / Track Road

Here the *Track Road*, as a separate part, seems to clarify *Land Access* from "anything that accesses someone's land" by communicating "typically found as tracks in the ground and it's about land access".

This sounds quite clear to me, given how hard-to-define a category this road type is. It might result in pretty good tagging behavior by those that didn't read the wiki, but it isn't great because track road is not actually fitting the definition of highway=track.

> Off Road / Not Maintained

This might be a major problem because:

- Same problem as current tagging, see top.
- I was going to joke that: people might also re-tag Belgium if we change it to this, since those roads also fit the description (jk).
- But then I realized that I used to tag many roads in Africa as highway=track before learning that highway=track does not mean "some tracks in the dirt". In low-income countries, highway=primary can be Off Road and the smaller ones (all of: secondary, tertiary, unclassified, residential, service) can very well look Not Maintained. Such roads perfectly fits the name, but are very very wrong. This might be worse than the current name which at least signals there should be *tracks* rather than a proper road.

---

What probably makes the most sense is what Matkoniecz suggested (if I understood the suggestion correctly):

> Maybe separate label for paved / `tracktype=grade1` would be a good idea?

Just have two tags. Avoid the whole problem by having two presets that result in `highway=track`. If it's hard to capture all the relevant conditions in one short label, then don't use one short label.

Even the option I think would make the most sense to little Luc (when he was still a noob), *Land Access / Track Road*, does not really fit for paved roads. Or *Agricultural Road* would be clear for little Luc because it's about the function (rather than its appearance), but apparently unclear for other people. I do think that *Agricultural Road* would trigger people to open the wiki more often to see what falls within this agriculture category, so might potentially be a good option *because* it's not crystal clear. But then perhaps the same can be said for *Land Access*, that also signals it's about function while being rather vague (to me).

Perhaps we should collect some of the more promising ideas and try it with on friends and family. Show (the same set of) pictures of some roads, give some preset names that somewhat fit (residential, tertiary, track) as available options, see what they pick. Maybe automate it into a minor unmaintained website to have the results be double blind. It doesn't have to result in one good option, it might tell us which ones are actually misleading (or in which countries they are misleading) and we can work with that information to get to better solutions.

ZeLonewolf:
> > Agricultural Road
> 
> I find this the most clear value so far suggested in the thread, but then it got 5 downvotes 0 upvotes at the time of writing. I dunno what's unclear about a road meant for agriculture because that's exactly what it is.

There are many, many usages of `highway=track` that are used for access to land that isn't agricultural, for example, in recreational areas.  This is a _very_ common usage in the US.  While in Europe the agriculture/logging usage of `highway=track` is the best understood, in the US the tag is quite common as usually unpaved general land access roads, or for roads which have been abandoned and are now used by hikers for land access.  This is why "land access" came to be recognized as a general term for global usage -- it's a reasonably adequate umbrella term that covers both the European and American understanding of how `highway=track` should be used.  After all, agricultural roads are also essentially providing access to the land.

zekefarwell:
I'm going to make another pitch for simply **Track Road** as the default preset name.  I really think it's the simplest, and best option.  Clicking for more information shows the short description of what a Track Road is from the wiki and then the user can click _View documentation_ for the full wiki page.  This seems perfectly fine.

track

**Service Road** is the same way.  Many people with think of a frontage road when the see the term service road, but when they click for more info the short description shows up explain that in OSM terminology "service road" means something different.  This is also fine.  There's no need to change the preset name to **Building-access / Service Road** in an attempt to clarify.
 
service
 
Same for tertiary.
 
tertiary
 
Track Road would fit with all the other presets.  If a simple name is good enough for them, it's good enough for track.

bhousel:
Here's a road that I've tagged as `highway=track` on OSM.

track

- it is paved
- it has a name
- it has lines
- it connects 2 roads of higher classification (yes it's a shortcut that a router software would love to use)
- it does not have anything to do with farming or forestry or anything like that
- nobody lives here
- (does it "access land"? .. I guess?)
- one might call it "abandoned", but it gets regular recreational use
- this is the normal state of the road (not like the day after a storm or anything)
- it is cleared of debris occasionally.  Yes, we can take a truck on it to do chainsaw work. (yes this is "maintenance")
- it is NOT maintained to standards that would allow people to drive on it most of the time.

I am again asking everyone - please don't get so caught up in the precise definitions of what the wiki says that we forget why we are doing this.  Consider adjusting those definitions when presented with examples of roads that don't look like the tracks on the wiki or where you live.

We are here to build a useful map.  One feature of a useful map is that it doesn't get people _stuck in the woods_.

The purpose of the iD presets is to help relatively untrained users contribute useful map data.  Not to "bring it in line with the OSM wiki".  Not to use hit a precise definition of "how the tags are intended to be used".  (both phrases that appear in comments on this thread).   This is not what OSM is about.

If it sounds like I'm being a pain in the ass about tracks, it's because OSM's wiki definition of a track is preventing OSM from being a useful map for routing in a lot of places.   Waze has solved this problem pretty well, and OSM still hasn't.

zekefarwell:
I'd say `track` is the most appropriate option for that road.  If nobody lives there and the only thing people use it for is outdoor recreation, that fits under the term of land access.  I know its a vague term and not something anyone would say in normal conversation (that's why I'm arguing against using it as the preset name).  However, I do think it accurately captures all the things track roads are used for worldwide in a concise way.

andrewharvey:
+1 to "Land Access Road", dropping "minor". I'd still support localisation, but perhaps the locale code just doesn't suit our localisation needs, and instead we need region+language.

So for Germans with their UI set to English it might say "Land Access Road", but for Americans with their UI set to English it might say "Track Road".

lgommans:
> [There are many, many usages of highway=track that are used for access to land that isn't agricultural, for example, in recreational areas. This is a very common usage in the US.](https://github.com/openstreetmap/id-tagging-schema/pull/288#issuecomment-985959374)

Wait, that's a redefinition of the tag I wasn't aware of. I see this was introduced in mid-2020 but phrased as "some people use these tracks for recreation" rather than "the tag can be used for roads that are meant exclusively for recreation":

https://wiki.openstreetmap.org/w/index.php?title=Tag:highway%3Dtrack&diff=prev&oldid=1996976

This change has now been morphed into saying that using highway=track for recreational roads is correct, no longer reflecting the original function. (Note that this change only happened over the last year, OSM has a lot more history during which highway=track for such roads was a tagging mistake.)

By that logic, whenever enough people make a mapping mistake, we should redefine the tag to include that possible use. Thus we need to now say that track means any unpaved road because that reflects the typical use of the tag. (Heck, we could write on the highway=raceway page that it's fine to map motorways as such because that's how "some people" use the road.)

I'm no longer sure what we're trying to do here.

Are we making a clear name for what highway=track is *supposed* to be used for such that it isn't misused, or do we want to have an all-encompassing name to cover what mappers *have been* using it for?

zekefarwell:
> This change has now been morphed into saying that using highway=track for recreational roads is correct, no longer reflecting the original function. (Note that this change only happened over the last year, OSM has a lot more history during which highway=track for such roads was a tagging mistake.)

The wiki may have only started documenting the full meaning recently, but `highway=track` has been used extensively for roads that have nothing to do with agriculture or forestry for over a decade.  Call it a mistake if you want, but that's the reality in the database.  The phrase "a minor land-access road like a farm or forest track", is meant to capture the spirit of the previous description while clarifying that many other, similar uses fit as well.  These same kinds of roads are found in all different type of open land (deserts, mountains, grasslands,  forests etc) and are used for all sorts of purposes including 4WD/off-road recreation, agriculture, mountain biking, forestry, environmental research, hunting, and fire suppression.  The similarity between all these activities is that the are done away from human settlements and infrastructure, on areas of open land.  Mappers around the world needed a tag for the minor roads on this open land that didn't really go anywhere important except the land itself.  They intuitively chose `highway=track` as the best fit.

zekefarwell:
> I'd still support localisation, but perhaps the locale code just doesn't suit our localisation needs, and instead we need region+language.
> 
> So for Germans with their UI set to English it might say "Land Access Road", but for Americans with their UI set to English it might say "Track Road".

I like this idea.  If there could be a default English (International) localization, separate from the English (United States) localization, that would be great.  As far as I can tell this may not exist though.  I've checked the language options in Windows, Mac OS, Chrome, Firefox, etc and the options are:  English (Australia), English (Canada), English (United Kingdom), English (United States), etc.  I didn't see an English (International), or just plain English option in any of the places I checked.

andrewharvey:
> I like this idea. If there could be a default English (International) localization, separate from the English (United States) localization, that would be great. As far as I can tell this may not exist though. I've checked the language options in Windows, Mac OS, Chrome, Firefox, etc and the options are: English (Australia), English (Canada), English (United Kingdom), English (United States), etc. I didn't see an English (International), or just plain English option in any of the places I checked.

Could be something implemented in the editor, it could use geoip to detect your region for the default but let you change your locale region independently of language. Otherwise we'll keep running into this kind of issue where US mappers have a different term to other English speaking countries for the same thing.

matkoniecz:
> OSM's wiki definition of a track is preventing OSM from being a useful map for routing in a lot of places

Maybe how `highway=track` is used is problematic and unhelpful but

- there was significant effort to bring OSM Wiki definition into synch with actual use
- OSM Wiki definition is now matching actual use quite well
- trying to force meaning change with iD preset is neither a good idea, nor helpful to mappers nor solving actual problem
- pretending that problem does not exists does not help
- iD changing description of some object is not changing how this tags were used and how people not using iD are mapping
- iD trying to redefine tags in this way just results in significant confusion and is not helpful

Maybe `highway=track` should be used in a different way, maybe it should be partially or fully replaced. But what iD attempted with `highway=track` was not a good way to achieve this.

> The purpose of the iD presets is to help relatively untrained users contribute useful map data. Not to "bring it in line with the OSM wiki". Not to use hit a precise definition of "how the tags are intended to be used". (both phrases that appear in comments on this thread). This is not what OSM is about.

"bring it in line with the OSM wiki" is part of what is necessary to make iD presets an useful help, iD having diverging definition both from tags are actually used and OSM wiki and intended meaning is not helping new mappers at all.

Note that in many cases OSM wiki is gibberish and needs to be changed to that synchronization is achieved, but it should not be done in communication with wider OSM community. Especially for clearly established practices, even if that practices are unhelpful and dubious.

For example `highway=unclassified` is a really poor name. But preset maintainer suddenly, without discussion with wider OSM community, commiting change that deprecates it or describes in a new way not matching either OSM Wiki nor actual use would not be helpful.

> Here's a road that I've tagged as highway=track on OSM.

Based on the description it qualifies as `highway=track` better than to other OSM road classes. Though I am not familiar with either location or its use and how it is actually used was not really described. So I am unsure how it would be conterexample.

And yes, some `highway=track` are not maintained. But it is not their defining attribute.

simonpoole:
I find the "land access" variants language-wise somewhat grating and in reality this is not the only use.

highway=track is infrastructure usable for double tracked vehicles that isn't part of the regular public road network and mostly used for access to areas not accessible from that network. While that is often for farming and agriculture, I don't see a reason to exclude other uses just as it is IRL. 

How the " public road network" is defined is obviously country dependent, but in developed areas there is typically a process that designates infrastructure as a public road and tracks typically do not go through this process. This doesn't imply that tracks are not publicly accessible at all, just that they do not have the same legal status. I point to the differences between the UK and German speaking Europe as an example of how widely access can differ in practice. 

As to the details, for example maintained vs unmaintained, I don't see why we can't simply have 5 presets for the 5 tracktype variants which can then be named appropriately instead of trying to find a one-size fits all solution to this.

ArloJamesBarnes:
What would that look like in practice? I don't think either "Even Mix of Hard and Soft Materials Track Road" or "Grade 3 Track Road" would go over well.

simonpoole:
> What would that look like in practice? I don't think either "Even Mix of Hard and Soft Materials Track Road" or "Grade 3 Track Road" would go over well.

Well I think something like "Overgrown/unused track", "Unmaintained track", "Unpaved track" would work well for 5-3, 2 and 1 are a bit tricky, but I'm sure we can find something that fits.

bhousel:
> Could be something implemented in the editor, it could use geoip to detect your region for the default but let you change your locale region independently of language. Otherwise we'll keep running into this kind of issue where US mappers have a different term to other English speaking countries for the same thing.

Yes since 4359caf7ef and https://github.com/openstreetmap/iD/pull/8305 iD now has the flexibility for presets to have any scope we want - region, country, custom geojson, etc.  This is relatively recent work that I only finished this year.  Today if a few countries wanted a special track preset, we can support it.  

Really I don't think we need a special preset.  We can just replace the word "Unmaintained" - which again, never meant "unpaved" or "grade>1" - with another word or phrase that implies "something less than a service road".

tyrasd:
After some time has passed, I don't think that my proposal to rename the preset to "[minor] land access road" is the best solution: I agree with what has been said above that it is still unclear what exactly it means, not a common term and basically only repeats the description the presets already displays when clicking on the (i) icon.

I see two (maybe tree) concrete solutions to go on with:

* a) Just simplify the preset name to `Track Road` (see the argument made by @zekefarwell [above](https://github.com/openstreetmap/id-tagging-schema/pull/288#issuecomment-985966000)).
* b) Use `Land Access / Track Road` (which, according to @lgommans's [comment](https://github.com/openstreetmap/id-tagging-schema/pull/288#issuecomment-985957660), seems to be a good way to communicate that a track is _something less than a service road_.
* c) I was wondering if one could rewrite solution b) also as `Track / Land Access Road` to emphasize the track aspect a bit more. I would appreciate the opinion of a native speaker on this one :bow: 

> I don't see why we can't simply have 5 presets for the 5 tracktype variants which can then be named appropriately instead of trying to find a one-size fits all solution to this.

While this sounds like the optimal approach at first, we would always still need to have a preset for the 67% of `highway=track` features without a mapped `tracktype` and for new tracks to be added where the classification is unknown to the mapper.

<hr>

I would like to hopefully soon come to a conclusion on this issue. Currently, I'm tending slightly towards solution b) `Land Access / Track Road` because I think it is a good compromise solution which covers the broadest spectrum of uses of the tag and also conveys the most important aspects of the tag/preset.

ZeLonewolf:
Regarding the request for a native speaker, I'm a native speaker of en_US (though I've on occasion be described as a native speaker of en_US_RI, but that's a discussion for another day).  Here's my native speaker perspective.

There is no such thing as a "track" to describe a road in en_US.  There is no such thing as a "track road" in en_US.  The closest term I'm aware of is a "two-track road", and it's used mostly by outdoorspeople to describe any dirt road with two ruts.

In pictures:

This is a **track**, aka an athletic track:
image

This is also a **track**, aka the footprints of an animal:
image

This is a **two-track road**:
image

In short,
* **land access road** is something that exists in en_US and would be understood.
* **track road** is a made-up term that was invented from OSM tagging.

andrewharvey:
It's going to be locale specific, to me as an Australian, a track is a walking path, and trail is more commonly something you drive a 4wd vehicle on. For en_AU it's mostly just a "Fire Trail" unless it's on a farm. 

If you're talking about en_US then only Americans can answer this and they alone should be deciding what to use. So my point from StreetComplete still applies here that it may be useful to have a global english distinct from en_US.

westnordost:
@andrewharvey It seems to me that such suggestion falls flat exactly at source strings that are different words in different dialects of English. I.e. at the very strings your suggestion is trying to solve. 
For example `highway=footway + footway=sidewalk`. What is it going to be - "Sidewalk" or "Pavement"? To paraphrase in such a case, e.g. "usually paved walk for pedestrians at the side of a street" won't work either - too long and going to be very confusing for translators.

The only true international source string would be the tag key+value itself (e.g. `highway=track`) and treat everything else including en-US as translations. But this excludes translators from participating if they are not veteran mappers.
Hence, the best compromise is that the tag key+value is mentioned as a developer note and this is already done, see

image

andrewharvey:
@westnordost fair point, you've convinced me, then we must make it clear that the source language is en_US and leave it for the Americans to decide their own localisations.

Transifix has localisations for Australia, India, New Zealand and United Kingdom, so basically the rest of the world is left out. Are there users from other parts of the world who do prefer to use the app in English but need their own regional variations?

1ec5:
From a usability perspective, we’d ideally choose a preset name such that users can find it when they need it and won’t mistakenly use for something different. A circumlocution like “(minor) land access road” is precise with respect to the wiki’s current definition, so it wouldn’t be misapplied to something unrelated. However, it’s also obscure, so it risks underuse. (Memorability is also helpful for clients like the Overpass turbo wizard. I don’t foresee anyone entering `"minor land access road" in Wyoming`.)

“Track road” looks elegant, but it suffers from the opposite problem, that a layperson is liable to confuse it for something wholly different. It’s pretty unlikely to be misapplied to a running track, but a more likely mistake would be applying it to a race**track** (which is a kind of road) or the service **road**s fronting a railroad **track**. “Track road” actually means “towpath”, though that usage is limited to dictionaries at this point.

By the way, this is not the only `highway` tag with a problematic name. “Motorway”, “trunk road”, “service road”, and “slipway” are all baffling or grossly misleading to speakers of various dialects of American English. In the Upper Midwest, new mappers routinely overclassify state routes, which are known there as “trunk highways”, so misnomers aren’t a purely theoretical concern. Ultimately, in the absence of an elegant, intuitive, precise term in American English, do we care more about misapplication or underuse?

> highway=track is infrastructure usable for double tracked vehicles that isn't part of the regular public road network and mostly used for access to areas not accessible from that network. While that is often for farming and agriculture, I don't see a reason to exclude other uses just as it is IRL.

Just for fun, [this way](https://www.openstreetmap.org/way/723305483/history) is County Road 8 (CR&nbsp;8), part of the regular public road network. 🤦

[Ophir Pass Road (723305483)](https://www.openstreetmap.org/way/723305483/history)

Here’s what it’s like in real life, courtesy of [YouTube](https://youtu.be/QOBI6Vce0wo?t=1164) and [Mapillary](https://www.mapillary.com/app/?pKey=747873812554895). It’s a shame we need six different tags to express that it’s not a `track` but [might as well be](https://www.kjct8.com/2021/06/18/delivery-trucks-attempt-go-over-4x4-mountain-pass-goes-horribly-wrong/). But here we are.

> It's going to be locale specific, to me as an Australian, a track is a walking path, and trail is more commonly something you drive a 4wd vehicle on. For en_AU it's mostly just a "Fire Trail" unless it's on a farm.

It’s really tempting to call `highway=track` a “[4WD road](https://pubs.usgs.gov/gip/TopographicMapSymbols/topomapsymbols.pdf#page=4&zoom=180,-7,632)” or “[Jeep trail](https://en.wikipedia.org/wiki/Jeep_trail)”.[^jeep] One of these terms probably would’ve replaced `highway=track` if Americans had gotten a chance to design a ~~`highway`~~ `road` tagging scheme of our own. 😏

Colorado DOT 2010

> EDIT: I just saw Brian's comment at [#288 (comment)](https://github.com/openstreetmap/id-tagging-schema/pull/288#issuecomment-986919641) looks like there may be support outside of Transifex.

There are a few distinct things:

* The user interface already loads a localization specific to **the user’s locale** (language+country+script). This is granular enough to account for many dialectal differences, but [not all] [of them](https://github.com/Project-OSRM/osrm-text-instructions/issues/188#issuecomment-406827751).
* Some presets are scoped to a geography and filtered by **the location of the selected feature**. These presets are translated into all languages, including languages that aren’t widely spoken in that geography. This can be an effective mechanism for regional tagging idiosyncrasies or real-world differences in construction, like [expressways](https://github.com/openstreetmap/id-tagging-schema/pull/216) and possibly [school classification](https://github.com/openstreetmap/id-tagging-schema/pull/331).
* Presets are not scoped to **the location in which the user is located**. That would require the Geolocation API on the Web or a similar API on other platforms. We’d still need to come up with a global default for each language in case the user declines to share their location with the application.

[^jeep]: To be clear, “Jeep trail” does not imply `highway=path` `jeep=dismount`. 💪

matkoniecz:
> It’s really tempting to call highway=track a “[4WD road](https://pubs.usgs.gov/gip/TopographicMapSymbols/topomapsymbols.pdf#page=4&zoom=180,-7,632)” or “[Jeep trail](https://en.wikipedia.org/wiki/Jeep_trail)”

On the other hand, that would make many European `highway=track` highly confusing - many are passable with a regular car (legal restriction and gating is more blocking)

1ec5:
That was merely an observation, not a proposal. I’m aware there would’ve been downsides. 😉

simonpoole:
I don't think there is going to be a really satisfactory solution, naturally, as has been pointed out, we have an added layer of confusion because of the use of en-US in iD, but that's a done deed. 

"land access road" would seem to work even if a bit bumpy and descriptive, it even maps reasonably well to German "Waldweg" and "Feldweg" .

1ec5:
It isn’t just American English that lacks a natural word for `highway=track` as defined. Other languages such as Spanish also experience significant confusion about this tag. Despite the highly constrained official definition, people gravitate toward it for a variety of purposes that might be described as “pseudo-road” (also not a proposal).

westnordost:
What do native American English speakers think about ...

`Land access road / local country road` (or alternatively add the second as an `alias`)?

"Country road" is mentioned in the definition of "dirt track" in the Oxford dictionary and [in Merriam-Webster](https://www.merriam-webster.com/dictionary/country%20road) as "a usually unpaved rural road off the main highway". This sounds pretty similar to the `highway=track` defnition that seems to be somewhat leaned on the meaning of German "Feldweg  / Waldweg" definition.

ZeLonewolf:
`local country road` sounds like any local road in the country?  "The country" in this context means a rural area.  Doesn't really make a lot of sense to me and definitely doesn't describe the kind of road we'd tag `highway=track`.

westnordost:
Ok, nevermind that then.

zekefarwell:
As a native en_US speaker I'd say `Land Access / Track Road` or `Track / Land-Access Road` both sound fine.  They are not entirely satisfactory, but as others have noted, that is probably not possible with this tag.  I added a hyphen to "land-access a compound word.  Otherwise, "land access road" suggests something that is both a "land road" and an "access road" since the two adjectives operate independently.

cliffordsnow:
Native en_US speaker here. I live in an area full of highway=track logging roads and have no problem differentiate these from the typical service road. I like @1ec5 post about about 4X4 road. (I'd drop the seasonal since that only applies to areas with seasonal changes.) One of the problems with highway=track is that a typical family vehicle isn't suitable for driving on them. Yes people do get away with it but it's not something OSM should support. 

Land access road is too vague. It could be a service road or a county road.

westnordost:
FYI all, iD presets allow both `aliases` and `terms` to be added for a preset.
- **Terms** are simply terms by which the particular preset should also be found (e.g. type "fruit" and iD displays "greengrocer")
- **Aliases** are alternative names for a preset. The difference to terms will _likely_ be (see https://github.com/openstreetmap/iD/issues/6139) that 1. if you type an alias, that alias will also be displayed in the preset search result list alongside the "primary/proper name" and 2. of course it has a higher priority for matching than terms. (Hence), aliases do not need to be synonymous with the primary name, they can also describe a subtype - as long as there is not another preset that covers that

I mention this because since so much brain power was now activated into finding the "primary" en-US name for `highway=track`, I want to suggest to put the other suggestions that have been made here into `terms` or `aliases` respectively, so it can be found better.

Scanning suggestions made in this ticket, it could be...

**Name**
land-access road

**Aliases**
- track road (or maybe not, see https://github.com/openstreetmap/iD/issues/9117)
- fire trail
- two-track road
- logging road
- forest road
- forestry road
- farm road
- agricultural road
- unmaintained road

**Terms**
- 4WD
- jeep
- high clearance
- off-road
- overgrown
- unmaintained

matkoniecz:
> It isn’t just American English that lacks a natural word for highway=track as defined. 

Also in Polish! There are some terms that are just close enough to be maximally confusing and misleading. I would deprecate `highway=track` and `highway=unclassified` and `highway=path` if I could.

blacklocust3:
(tl;dr at bottom) Hey sorry new here but had an issue and was told to post here. Trying to add some maintained gravel forest roads, but only options on the iD editor were "minor road", "service road", and "unmaintained track road".

Per the Wiki and checking with other OSM editors, the road is indeed a track because its primary purpose is land access. However, the iD editor labels all tracks as "unmaintained track roads."

I think this is an issue because if I'm reading the wiki documentation for tracks correctly,

https://wiki.openstreetmap.org/wiki/Tag:highway=track

tracks are defined by their purpose (land access) and not by their quality or maintenance. Even a paved road can be a track. I think labeling the track tag as "unmaintained track" in the editor is inaccurate because it excludes maintained tracks.

Could it be changed to "track"? I saw another option was "track / land access road" and that also seems fine to me.

Tl;dr Tracks are currently labeled as "unmaintained track roads" which excludes maintained tracks. Change label to "track" or "land access road / track"?

pamolloy:
> Tl;dr Tracks are currently labeled as "unmaintained track roads" which excludes maintained tracks. Change label to "track" or "land access road / track"?

Thanks for posting @blacklocust3. This is the core problem and the conversation has perhaps gotten lost in trying to rename it. _Unmaintained_ is objectively wrong in that most of these roads are in fact maintained (dirt roads need maintenance too!) and it gives the false impression (which you can see in this thread) that track means off-roading or 4x4 roads. Even off-roading or 4x4 roads can need maintenance!

tyrasd:
Thanks to everyone who joined the conversation!

I've merged the version `Track / Land-Access Road` now, as this seems to be the solution which most people find acceptable.

To summarize the arguments which ultimately lead to this specific name:

* remove the _unmaintained_ adjective from the preset name, as it is objectively wrong in many valid use cases of the tag
* include the term "land access road", as this is the most accurate description of the intended usage of the tag
* include "track" as it is a very common "technical term" in OSM-speech (other highway presets also reflect the tag value in their preset names)

1ec5:
> I've merged the version Track / Land-Access Road now, as this seems to be the solution which most people find acceptable.

Would the maintainers of this repo be open to continuing to explore better names for this preset in a separate discussion? There seemed to be a lot of discomfort or concern about each element of this compromise. The more I think about this issue, the less certain I am that American English _doesn’t_ have the right word for this concept. But it would only be worth dragging out this discussion further if there’s any chance that a less unwieldy name would be approved.

tyrasd:
> Would the maintainers of this repo be open to continuing to explore better names for this preset in a separate discussion?

Sure! Feel free to open an issue. :relaxed:

PR #296, opened 2021-12-14
updating cuisines based on commonly used values from taginfo

andrewharvey:


andrewharvey:
I think at some point we should sort these alphabetically rather than by date added/popularity, though it can be done later.

tyrasd:
see #252

data/fields/cuisine.json
```diff
@@ -33,7 +33,21 @@
             "greek": "Greek",
             "vietnamese": "Vietnamese",
             "turkish": "Turkish",
-            "lebanese": "Lebanese"
+            "lebanese": "Lebanese",
+            "indonesian": "Indonesian",
+            "malaysian": "Malaysian",
+            "taiwanese": "Taiwanese",
+            "juice": "Fruit Juice",
```
tyrasd, line 40: `juice` could also be for vegetable juices, couldn't it?

data/fields/cuisine.json
```diff
@@ -33,7 +33,21 @@
...
+            "bubble_tea": "Bubble Tea",
+            "hot_dog": "Hot Dog",
+            "cake": "Cake",
+            "pankcake": "Pancake",
+            "pasta": "Pasta",
+            "salad": "Salad",
+            "chocolate": "Chocolate",
```
tyrasd, line 47: was there a specific reason to include this value? It seems to not be used too often and is also not included on the [`cuisine`](https://wiki.openstreetmap.org/wiki/Key:cuisine) page on the wiki.

tyrasd:
I'm wondering if we should do this a bit more methodological this time…

For example, include everything from https://wiki.openstreetmap.org/wiki/Key:cuisine which has at least a certain amount of usage (say 500 times globally). 

 > I think at some point we should sort these alphabetically rather than by date added/popularity

:+1: IMO we could do this right away.

data/fields/cuisine.json
andrewharvey: It's got 119 uses, and when I was looking through the list it's one that I could think of a number of places where their main cusine is chocolate.

andrewharvey: I've added to the wiki now

andrewharvey:
> For example, include everything from https://wiki.openstreetmap.org/wiki/Key:cuisine which has at least a certain amount of usage (say 500 times globally).

You can, for now I just added those I feel confident in.

PR #297, opened 2021-12-15
upgrade path for amenity=youth_centre

westnordost:
As promised in #83, here is a follow-up for the upgrade path of deprecated `amenity=youth_centre` -> `amenity=community_centre` + `community_centre=youth_centre`. It looks like _another_ upgrade path was already added in the meantime, but it is not quite correct or at least not consistent with the tagging in #83:

`community_centre:for` specifies actually rather a property of a community centre/youth centre, while `community_centre` specifies the type of community centre, i.e. is part of what it defines as a map feature. See PR #83

---

I don't know, hypothetical example: There could be a general community centre for bingo-playing and a community centre dedicated to hands-on creating art and music each in a residential complex for old people. Both are **for** old people, but there are different kinds of community centres. The "for" does not define the type of community centre necessarily and that is why both `community_centre` key exist and `community_centre:for` key exist.

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -185,11 +185,11 @@
   },
   {
     "old": {"amenity": "youth_center"},
-    "replace": {"amenity": "community_centre", "community_centre:for": "juvenile"}
+    "replace": {"amenity": "community_centre", "community_centre": "youth_centre"}
   },
   {
     "old": {"amenity": "youth_centre"},
-    "replace": {"amenity": "community_centre", "community_centre:for": "juvenile"}
+    "replace": {"amenity": "community_centre", "community_centre": "youth_centre"}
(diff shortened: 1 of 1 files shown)


tyrasd:
>  It looks like another upgrade path was already added in the meantime

Actually, it was already added a while ago in https://github.com/openstreetmap/iD/commit/aca699ecca70e0778472a0fd9c5365a7d5c82ff9. It would indicate that these upgrade paths were copied over from osmose. Not sure if these rules are still in there somewhere, but at first glance I can't find them over at the [osmose](https://github.com/osm-fr/osmose-backend) repository.

Anyway thanks for the PR (even though there are now almost no `amenity=youth_centre` remaining)!

PR #302, opened 2021-12-18
Update en-AU.json

tastrax:
Remove an actual live phone number and replace with template number

### DIFF
--- a/dist/translations/en-AU.json
+++ b/dist/translations/en-AU.json
@@ -9,7 +9,7 @@
                     "label": "Colour"
                 },
                 "fax": {
-                    "placeholder": "+61 2 9000 0000"
+                    "placeholder": "+61 1 2345 6789"
                 },
                 "height": {
                     "label": "Height (Metres)"
@@ -24,7 +24,7 @@
                     "label": "Length (Metres)"
                 },
                 "phone": {
-                    "placeholder": "+61 2 9000 0000"
+                    "placeholder": "+61 1 2345 6789"
                 },
                 "recycling_type": {
                     "options": {


matkoniecz:
See https://github.com/openstreetmap/id-tagging-schema#translations - are you sure that it is not editable on Transifex?

matkoniecz:
Also, see https://www.acma.gov.au/choose-your-phone-number

>  Use phone numbers for fiction

> Do not use real phone numbers in radio, books, films and television. People often call them.

> We have phone numbers you can use for the work you publish or broadcast.

It would be nice to use one of them, if currently used is not such case

tastrax:
The sample phone numbers they show are for specific geographic areas within Australia. +61 1 2345 6789 is generic for the whole country. Area code 1 doesn't not get used anywhere in the country so its safe to use. I am not a transifex user so feel free to edit in there if you are.

tastrax:
Added as suggestions in transifex @andrewharvey

andrewharvey:
> Added as suggestions in transifex @andrewharvey

I think you should be able to make the change directly, but I've gone ahead and saved those changes, hopefully it should come through automatically but I'm not across how that happens.

matkoniecz:
> hopefully it should come through automatically but I'm not across how that happens

Yes, translations are automatically pulled in from time to time. See https://github.com/openstreetmap/iD/commit/34759ae3f2c41e6fc07b83b5f96750b9b342e1bf where this happened in iD

tyrasd:
I've pulled the latest translations in bbe7b25623f406eb64a9b1f99d8ce25f4fa90483. This looks to be fine now.

PR #303, opened 2021-12-18
Update en-AU.min.json

tastrax:
Change phone number from an actual live number to a template number


matkoniecz:
See https://github.com/openstreetmap/id-tagging-schema#translations

tastrax:
Added as suggestions in transifex @andrewharvey

PR #305, opened 2021-12-18
add "railway" term to landuse=railway

### LINKED ISSUE #301: "railway landuse" should be able to find railway=landuse preset
Yes, unlikely to be used by newbies but may help people who know OSM tags and terminology

screen02

matkoniecz:
may fix #301


tyrasd:
I don't think that this will actually fix #301, but it doesn't hurt either.

PR #308, opened 2021-12-18
Create glass_bottles

### LINKED ISSUE #294: Add preset for "Altglascontainer" glass bottle container
It looks like this is a thing for Germany only(?).
But we have them everywhere, and I always pick the wrong preset first. So it would be great to have a specific preset for this.

Bildschirmfoto 2021-12-14 um 20 22 32

https://wiki.openstreetmap.org/wiki/Altglascontainer
```
amenity=recycling
recycling_type=container
recycling:glass_bottles=yes
```

Other wiki pages
* https://wiki.openstreetmap.org/wiki/Tag:amenity%3Drecycling
* https://wiki.openstreetmap.org/wiki/Key:recycling_type

Alternatively, we could improve the search keywords on https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/recycling_container.json.

Icon: I would keep the existing generic recycling icon, but we could make a composite icon with  from https://wiki.openstreetmap.org/wiki/Tag:amenity%3Drecycling#Rendering.

kjonosm:
Add preset for Glass Bottle Containers, closes #294

### DIFF
--- a/data/presets/amenity/recycling/container/glass_bottles
+++ b/data/presets/amenity/recycling/container/glass_bottles
@@ -1,0 +1,35 @@
+{
+    "icon": "maki-recycling",
+    "fields": [
+        "{amenity/recycling_container}"
+    ],
+    "moreFields": [
+        "{amenity/recycling_container}"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "bottle",
+        "bottle bank",
+        "bottle container",
+        "bottle recycling",
+        "bottle collection",
+        "glass bottle",
+        "glass bottle bank",
+        "glass bottle container",
+        "glass bottle recycling",
+        "glass bottle collection"
+    ],
+    "tags": {
+        "amenity": "recycling",
+        "recycling_type": "container",
+        "recycling:glass_bottles": "yes"
+    },
+    "reference": {
+        "key": "recycling:glass_bottles",
+        "value": "yes"
+    },
+    "name": "Glass Bottle Container"
+}
--- a/data/presets/amenity/recycling/container/glass_bottles
+++ b/data/presets/amenity/recycling/container/glass_bottles
@@ -31,5 +31,5 @@
         "key": "recycling:glass_bottles",
         "value": "yes"
     },
-    "name": "Glass Bottle Container"
+    "name": "Single Glass Bottle Container"
 }
--- a/data/presets/amenity/recycling/container/glass_bottles
+++ b/data/presets/amenity/recycling/container/glass_bottles
@@ -1,14 +1,23 @@
 {
     "icon": "maki-recycling",
     "fields": [
-        "{amenity/recycling_container}"
+        "operator",
+        "recycling_accepts"
     ],
     "moreFields": [
-        "{amenity/recycling_container}"
+        "opening_hours",
+        "opening_hours/covid19",
+        "collection_times",
+        "colour",
+        "covered",
+        "indoor",
+        "level",
+        "manufacturer",
+        "material",
+        "ref"
     ],
     "geometry": [
-        "point",
-        "area"
+        "point"
     ],
     "terms": [
         "bottle",
@@ -31,5 +40,5 @@
         "key": "recycling:glass_bottles",
         "value": "yes"
     },
-    "name": "Single Glass Bottle Container"
+    "name": "Recycling Container for Glass Bottles"
 }


matkoniecz:
What about name suggesting to use it only for aone container?

To avoid problematic tagging where there are separate objects for metal recycling container, glass bottles recycling container and paper recykling containers standing in one group?

kjonosm:
I'm no sure I understand your comment. Even if the Glass Bottle Container stands in a group of several other containers it's still a Glass Bottle Container and could therefor be tagged like that.

matkoniecz:
Container group like
external-content duckduckgo com
really should be tagged as single `amenity=recycling` object

kjonosm:
Thanks for explaining. The name for the preset is already singular "Glass Bottle Container".
Would it help to change the name to "Single Glass Bottle Container"?

matkoniecz:
Maybe Lone Glass Bottle Container? Maybe Alone Glass Bottle Container?

But I am not a native speaker.

tordans:
The established mapping practice in Berlin is, to tag the location of 1-n glass bottle container with https://wiki.openstreetmap.org/wiki/Altglascontainer. AFAIK that is the case for whole Germany. And it is the way I read the wiki example section and page https://wiki.openstreetmap.org/wiki/Tag:amenity%3Drecycling showing two container but the same tagging, not multiple sets of tags to represent each container..

We had a discussion in our local community how the accepted colors should be mapped. The consensus was, to keep one node for all colors. There is no established way to map the accepted glass bottle colors AFAIK.

data/presets/amenity/recycling/container/glass_bottles
```diff
@@ -0,0 +1,35 @@
...
+    "tags": {
+        "amenity": "recycling",
+        "recycling_type": "container",
+        "recycling:glass_bottles": "yes"
+    },
+    "reference": {
+        "key": "recycling:glass_bottles",
+        "value": "yes"
+    },
+    "name": "Single Glass Bottle Container"
```
tordans, line 34: To keep it more in line with https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/recycling_container.json#L41 and to show that it is a subgroup of that, we could name it
```suggestion
    "name": "Recycling Container for Glass Bottles"
```

data/presets/amenity/recycling/container/glass_bottles
```diff
@@ -0,0 +1,35 @@
...
+    "fields": [
+        "{amenity/recycling_container}"
+    ],
+    "moreFields": [
+        "{amenity/recycling_container}"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
```
tordans, line 12:
```suggestion
    "geometry": [
        "point"
    ],
```

The Wiki page for recycling container lists points and areas, but for this more specific preset of just glass bottle container, I suggest we start by just using points. This is more in line with the mapping practice I am aware of and with the numbers:
image

https://taginfo.openstreetmap.org/tags/recycling%3Aglass_bottles=yes

data/presets/amenity/recycling/container/glass_bottles
```diff
@@ -0,0 +1,35 @@
+{
+    "icon": "maki-recycling",
+    "fields": [
+        "{amenity/recycling_container}"
+    ],
+    "moreFields": [
+        "{amenity/recycling_container}"
+    ],
```
tordans, line 8: Interesting, I did not know there was a way to reference values for another preset.
When I was looking into it before, I wondered whether to move some of the default fields to the moreFields section. Why…
- IMO, they are not that useful for this glass bottle container preset (see notes below)
- there is not recommendation on https://wiki.openstreetmap.org/wiki/Altglascontainer
- there is no recommendation on https://wiki.openstreetmap.org/wiki/Tag:amenity%3Drecycling for glass bottle container
- on the combination tab at https://taginfo.openstreetmap.org/keys/recycling%3Aglass_bottles, I cannot see any relevant info that would argue for one for those additional tags being very visible


```suggestion
    "fields": [
        "operator",
        "recycling_accepts"
    ],
    "moreFields": [
        "opening_hours", // moved, AFAIK they nearly always are place along the road / in public paces so they don't have opening hours. Therefore moved to "moreFields".
        "opening_hours/covid19", // moved, same
        "collection_times", // moved, I know of no glass bottle container that has the collection times written on the container/nearby. Its not that relevant, since one does not plan the drop of this way. Therefore moved to "moreFields".
        "colour",
        "covered",
        "indoor",
        "level",
        "manufacturer",
        "material",
        "ref"
    ],
```

For comparison: https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/recycling_container.json#L3-L9

tordans (changes requested):
Thank for starting this!

data/presets/amenity/recycling/container/glass_bottles
kjonosm: Thanks for the hint. I updated the name accordingly.

data/presets/amenity/recycling/container/glass_bottles
kjonosm: Good catch, limited to "point"

data/presets/amenity/recycling/container/glass_bottles
kjonosm: I just copied that section from the existing Green Waste Container preset: green_waste.json
But all your suggestions make sense.

tordans (approved):
Looks good to me.

tyrasd (commented):
I'm afraid that this preset could turn out to be problematic: As far as I can see, there is currently no way to map singular recycling containers. The tag `recycling=container` is used for recycling sites with one to many containers, where each container can potentially accept different recycling goods. Here are a few examples of mixed-use recycling container sites:

* https://www.openstreetmap.org/node/1247422556
    ```
    amenity = recycling
    recycling_type = container
    recycling:clothes = yes
    recycling:glass_bottles = yes
    recycling:shoes = yes
    ````
    
* https://www.openstreetmap.org/node/2309361218/
    ```
    amenity = recycling
    recycling_type = container
    recycling:cans = yes
    recycling:glass_bottles = yes
    ```
    
* https://www.openstreetmap.org/node/560073337
    ```
    amenity = recycling
    recycling_type = container
    recycling:cans = yes
    recycling:glass_bottles = yes
    recycling:paper = yes
    recycling:plastic = yes
    location = underground
    ```

All of these non-pure recycling container sites would match this new preset, "falsely" identifying them as a _Recycling Container for Glass Bottles_. To some extent this effect is already in place with the [`electrical_items`](https://github.com/openstreetmap/id-tagging-schema/blob/v3.1.0/data/presets/amenity/recycling/container/green_waste.json) and [`green_waste`](https://github.com/openstreetmap/id-tagging-schema/blob/v3.1.0/data/presets/amenity/recycling/container/green_waste.json) presets, as can be seen in this example:

* https://www.openstreetmap.org/node/9038856195
    ```
    amenity = recycling
    recycling_type = container
    recycling:electrical_items = yes
    recycling:glass = yes
    recycling:paper = yes
    ```
    is labeled as being an _E-Waste Container_, even though it accepts also _glass_ and _paper_:
    

While not ideal, I think the electrical items and green waste presets do have a small enough overlap with other recycling types to be acceptable. The glass recycling containers on the other hand would introduce much too many false positive matches for my liking.

Currently, there is no way in iD's tagging schema to only match a preset if only a specific tag from a list of tags matches, which would be needed for this preset to work as intended.

tyrasd:
Closing until https://github.com/ideditor/schema-builder/issues/25 is fixed upstream.

PR #322, opened 2022-01-01
Fix typo in German translations

wielandb:
"Haudach-Solaranlage" -> "Hausdach-Solaranlage"


tordans:
@wielandb it looks like you did this change directly in the code, right? Here is the links + info on how to change translations: https://github.com/openstreetmap/id-tagging-schema#translations

tyrasd:
I've fixed this typo on transifex now.

PR #327, opened 2022-01-05
change default highway value for golf cart paths to "path"

### LINKED ISSUE #48: Golf Cartpath is insufficiently distinguished from service roads for cars
The Golf Cartpath preset adds a `highway=service` tag, relying on `golf=cartpath` and `golf_cart=designated` to distinguish the path from a regular motorized service road:

https://github.com/openstreetmap/id-tagging-schema/blob/0df7b20019c48c969fa9f9124724117de321f563/data/presets/golf/cartpath.json#L14-L16

openstreetmap/iD#6165 added the `highway=service` tag to the preset because major renderers weren’t on board with rendering `golf=cartpath` specifically, for example in gravitystorm/openstreetmap-carto#3734. Unfortunately, the workaround of adding `highway=service` creates ambiguity for routers, which are no more likely to special-case `golf=cartpath`. `golf_cart=designated` means the path is intended for use by golf carts, but routers can’t assume that ordinary motor vehicles are banned from these paths. (In fact, `golf_cart=designated` could apply to an ordinary `highway=residential` in a golf club community, or to any low-speed public street [in Ohio](https://www.theintelligencer.net/news/community/2016/12/golf%u2008carts-legal-on-ohio%u2008streets-beginning-sunday/).)

A quick fix would be to [add `motor_vehicle=no`](https://www.openstreetmap.org/way/805817989) to this preset. This will prevent routers from potentially sending full-size vehicles down narrow, winding, and [sometimes impassable](https://www.mlive.com/news/2020/11/amazon-driver-takes-wrong-turn-onto-michigan-golf-course-wedges-truck-in-cart-tunnel.html) paths designed for tiny golf carts. (A golf cart is a kind of motor vehicle, but `golf_cart=designated` will override `motor_vehicle=no`.) However, for all I know, some mappers could disagree with tagging access restrictions that in many cases may be implicit. After all, there’s no guarantee that a golf course would post [No Motor Vehicles](https://commons.wikimedia.org/wiki/File:MUTCD_R5-3.svg) signs all over.

In the long term, it may be worth getting the community on board with a dedicated `service=cartpath` tag or (fancifully) a `highway=cartway` tag, so that mappers can succinctly indicate the restriction and the reason for the restriction without pretending there’s a No Motor Vehicles sign. Or maybe `highway=path` is a safer replacement for `highway=service`. But that would be outside the scope of this repository, probably better for the tagging mailing list.

tyrasd:
This is to fix #48:

I went with the approach to just use `highway=path` (+ `golf_cart=designated`) because it: avoids explicitly adding an implicit access restriction and is similar to the well established tagging scheme for multi-use paths (such as foot+bike paths), as well as the corresponding `golf=path` preset.

This PR also activates the access field, so users can set see and set further access information.

### DIFF
--- a/data/presets/golf/cartpath.json
+++ b/data/presets/golf/cartpath.json
@@ -12,8 +12,8 @@
     },
     "addTags": {
         "golf": "cartpath",
-        "golf_cart": "designated",
-        "highway": "service"
+        "highway": "path",
+        "golf_cart": "designated"
     },
     "name": "Golf Cartpath"
 }
--- a/data/presets/golf/cartpath.json
+++ b/data/presets/golf/cartpath.json
@@ -15,7 +15,7 @@
     "addTags": {
         "golf": "cartpath",
         "highway": "path",
-        "golf_cart": "customers"
+        "golf_cart": "yes"
     },
     "terms": [
         "cartpath"
--- a/data/presets/golf/cartpath.json
+++ b/data/presets/golf/cartpath.json
@@ -1,6 +1,8 @@
 {
     "icon": "temaki-golf_cart",
     "fields": [
+        "name",
+        "highway_cartpath",
         "{golf/path}",
         "maxspeed"
     ],
@@ -13,7 +15,10 @@
     "addTags": {
         "golf": "cartpath",
         "highway": "path",
-        "golf_cart": "designated"
+        "golf_cart": "customers"
     },
+    "terms": [
+        "cartpath"
+    ],
     "name": "Golf Cartpath"
 }
--- a/data/fields/highway_cartpath.json
+++ b/data/fields/highway_cartpath.json
@@ -1,0 +1,12 @@
+{
+    "key": "highway",
+    "type": "typeCombo",
+    "label": "Type of Path",
+    "strings": {
+        "options": {
+            "path": "Cartpath",
+            "service": "Service Road"
+        }
+    },
+    "autoSuggestions": false
+}
--- a/data/presets/golf/path.json
+++ b/data/presets/golf/path.json
@@ -5,6 +5,7 @@
         "surface",
         "width",
         "structure",
+        "access",
         "incline"
     ],
     "geometry": [
--- a/data/presets/golf/path.json
+++ b/data/presets/golf/path.json
@@ -16,8 +16,10 @@
     },
     "addTags": {
         "golf": "path",
-        "highway": "path",
-        "foot": "designated"
+        "highway": "footway"
     },
+    "terms": [
+        "golf path"
+    ],
     "name": "Golf Walking Path"
 }
--- a/interim/source_strings.yaml
+++ b/interim/source_strings.yaml
@@ -1351,6 +1351,14 @@ en:
       highway:
         # highway=*
         label: Type
+      highway_cartpath:
+        # highway=*
+        label: Type of Path
+        options:
+          # highway=path
+          path: Cartpath
+          # highway=service
+          service: Service Road
       historic:
         # historic=*
         label: Type
@@ -5759,6 +5767,7 @@ en:
       golf/cartpath:
         # golf=cartpath | Translate the primary name. Optionally, add equivalent synonyms on newlines in order of preference (press the Return key).
         name: Golf Cartpath
+        # 'terms: cartpath'
         terms: <translate with synonyms or related terms for 'Golf Cartpath', separated by commas>
       golf/clubhouse:
         # golf=clubhouse | Translate the primary name. Optionally, add equivalent synonyms on newlines in order of preference (press the Return key).
@@ -5789,6 +5798,7 @@ en:
       golf/path:
         # golf=path | Translate the primary name. Optionally, add equivalent synonyms on newlines in order of preference (press the Return key).
         name: Golf Walking Path
+        # 'terms: golf path'
         terms: <translate with synonyms or related terms for 'Golf Walking Path', separated by commas>
       golf/rough:
         # golf=rough | Translate the primary name. Optionally, add equivalent synonyms on newlines in order of preference (press the Return key).


tyrasd:
@1ec5, @matkoniecz: what do you think about this approach?

matkoniecz:
Sadly, I am unfamiliar with cart paths so in this case I cannot offer guidance

(I learned about them via https://github.com/gravitystorm/openstreetmap-carto/issues/3734 - see especially https://github.com/gravitystorm/openstreetmap-carto/issues/3734#issuecomment-481244275 )

systemed:
I feel your pain on that one!

1ec5 (commented):
Thanks for this fix – I think it’s important to minimize the risk of large vehicles traveling over these paths. `highway=path` follows the principle of least surprise (by adopting an unsurprisingly chaotic tag).

If I recall correctly, `highway=path` was relatively commonly used for golf cart paths in the U.S. prior to openstreetmap/iD#5862. I mentioned it in passing in gravitystorm/openstreetmap-carto#3734, but I don’t recall why the preset wound up with `highway=service` instead of `highway=path` when it became clear that some `highway=*` value was needed. Based on the response to chadrockey/TGC-Designer-Tools#64, it’s likely that golf mappers will adhere to whatever this preset uses without edit-warring, as long as the preset retains `golf=cartpath`.

Note that the `golf=path` preset is very similarly tagged as `highway=path` `foot=designated`. As things stand, this will result in many non-golf-specific renderers giving both kinds of paths the same appearance. Consider using `highway=footway` for that preset.

Do we need to introduce any deprecations for the older tagging scheme being replaced here?

> My recollection was that their proposal was documented on the wiki (at least as a proposal).

I looked but haven’t been able to find a formal proposal from that time, unfortunately. There were a lot of parallel discussions at the time.

Later on, the discussions in #48 and elsewhere led to [a proposal for `highway=golf_cart_path`](https://wiki.openstreetmap.org/wiki/Proposed_features/golf_cart_path), which would make `golf=cartpath` redundant. An RFC went out for that proposal in 2020, kicking off lots of discussion and back-and-forth over the optimal primary feature tag. It would be premature to adopt the proposal in its current form, since existing renderers wouldn’t recognize `highway=golf_cart_path` at all. However, I think the mailing list and wiki discussions garnered sufficient support for the idea of some `highway` value other than `service`.

bhousel:
> However, I think the mailing list and wiki discussions garnered sufficient support for the idea of some `highway` value other than `service`.

The commit  https://github.com/openstreetmap/iD/commit/df7a2f978e7cabc3d4e99387b62251a2746033c5 that closed https://github.com/openstreetmap/iD/issues/6165 contained:
- a preset for a cartpath combined with a `highway=path` and
- a preset for a cartpath combined with a `highway=service` 

So mappers always did have this choice of which one to use.

1ec5:
This PR would modify the latter to use `highway=path`, but there’s still a semantic distinction between the two presets, so the suggestion was to also preserve that distinction in the primary feature tags, not just in the access tags.

bhousel:
> This PR would modify the latter to use `highway=path`, but there’s still a semantic distinction between the two presets, so the suggestion was to also preserve that distinction in the primary feature tags, not just in the access tags.

https://github.com/openstreetmap/iD/pull/5862#issuecomment-460917301 indicates that the purpose of these was for multiuse situations, like where the golf carts are expected to be driving through a parking lot or service area:

> For situations where both situations apply: example: a retirement community built around a golf course where the course's cart paths serve the purpose for the game as well as transportation routes for the residents, then the most appropriate tagging will be highway path/track with the golf:cartpath tag added.

(they originally were proposing `highway=track` and I steered them towards `highway=service`)

bhousel:
Ok I'm doing way more digging into this history of this than I'd like to, but the reason for this is because I _know_ we had this correct back in https://github.com/openstreetmap/iD/commit/e045c0f1781a59362b422563ac844990247ff419

and it got messed up somewhere along the way, I guess here: https://github.com/openstreetmap/iD/commit/df7a2f978e7cabc3d4e99387b62251a2746033c5

The intent of these presets was:
- a path-like preset with `golf=path` named "Golf Walking Path" 
(for where carts are not driving.  `highway=path` was added later, to appease carto)
- a path-like preset with `golf=cartpath` named "Golf Cartpath"
(`highway=path` was added later, to appease carto)
- a service-llike preset with `golf=cartpath` _and_ `highway=service` named "Golf Cart Multipurpose" 
(later "Golf Cartpath / Service Road", name has gone through several iterations for clarity but _`service` should not be dropped_ - it is literally for where carts drive on service roads)

I remember having a lot of discussion with Chad about what he wanted these for.   I even drew an icon for it.

1ec5:
> * a service-llike preset with `golf=cartpath` _and_ `highway=service` named "Golf Cart Multipurpose"
>   (later "Golf Cartpath / Service Road", name has gone through several iterations for clarity but _`service` should not be dropped_ - it is literally for where carts drive on service roads)

Oof, so this preset basically got conflated with the normal “Golf Cartpath” preset. That would explain all the problematic `highway=service` usage out there. Thanks for digging up this history.

The case where golf carts can drive on actual service roads sounds like a good fit for the Access field rather than a separate preset. Since the last time these presets were modified, [`golf_cart` was approved](https://wiki.openstreetmap.org/wiki/Proposed_features/Key:golf_cart) as an access key for any kind of road or path. A new Golf Carts row in the Access field would also give golf mappers an obvious alternative to turning _residential_ roads into golf cart paths or [drawing golf cart paths](https://osmcha.org/changesets/82899214/) [right on top of them](https://www.openstreetmap.org/changeset/90872348), as they so often do today around golf clubs. 😩 Maybe the extra could be limited to just the Residential Road and Service Road presets, since golf carts are unlikely to be street-legal for higher-classed roads anyways.

As long as we restore the golf cart path/service road combination in the form of an expanded Access field, it should be safe to revert the plain “Golf Cart Path” preset to `highway=path`, right? That would leave at least as many distinctions as you had in the presets at one point.

tyrasd:
Ok, I didn't know that these were three presets originally. I have solved the issue of not being able to easily map "multipurpose" service+cartpaths by adding a field for the highway tag for golf cartpaths where one can choose between a "cartpath" (the default) and "service road". This has the side benefit that manually entered highway values (e.g. `residential) will also be recognized.

data/presets/golf/cartpath.json
```diff
@@ -12,8 +14,11 @@
     },
     "addTags": {
         "golf": "cartpath",
-        "golf_cart": "designated",
-        "highway": "service"
+        "highway": "path",
+        "golf_cart": "customers"
```
1ec5, line 18: Sorry for splitting hairs, but I think `designated` would avoid making assumptions we can’t be sure about. It’s very common for [golf courses](https://www.openstreetmap.org/relation/49218) to be intermingled with residential neighborhoods, in which case the golf course is also open to residents or even the general public, not just members of the country club. There’s even [a town near Atlanta](https://lists.openstreetmap.org/pipermail/tagging/2020-December/057429.html) that maintains a public network of golf cart paths as a major local mode of transportation.

1ec5: All I can say is that golf courses intermingled with residential subdivisions are the norm in the U.S. My only experience with them comes from riding the school bus on a long route that included many of them.

Wouldn’t the more conservative option be `yes`? I’m sure there are country clubs that are strict enough to care that a non-member has brought their golf cart all the way there, but I’d imagine more are preoccupied with making sure golf carts stay on the paths in the first place.

tyrasd: `yes` sounds like a good compromise. changed in b41923c2ccdad057e4a5c7550471c466550044ab

PR #329, opened 2022-01-05
add Footway Area area:highway label

### LINKED ISSUE #131: Describe area:highway=footway area as "footway area", not "road area"
See for example https://www.openstreetmap.org/way/300555812

Similar would be useful for `area:highway=cycleway` `area:highway=path`

https://taginfo.openstreetmap.org/keys/area%3Ahighway#values

screen02

matkoniecz:
fixes #131

sadly, not tested because I have no idea how to do this https://github.com/openstreetmap/id-tagging-schema/issues/131#issuecomment-1005913970

If this will pass then I will create at least `area:highway=cycleway` one (`area:highway=path` is a bit tricky)

### DIFF
--- a/data/presets/area/footway.json
+++ b/data/presets/area/footway.json
@@ -1,6 +1,5 @@
 {
     "fields": [
-        "area/highway",
         "surface"
     ],
     "geometry": [
--- a/data/presets/area/footway.json
+++ b/data/presets/area/footway.json
@@ -1,0 +1,22 @@
+{
+    "fields": [
+        "area/highway",
+        "surface"
+    ],
+    "geometry": [
+        "area"
+    ],
+    "terms": [
+        "area:highway",
+        "edge of pavement",
+        "footway area",
+        "footway shape",
+        "pavement",
+        "sidewalk shape",
+        "sidewalk area"
+    ],
+    "tags": {
+        "area:highway": "footway"
+    },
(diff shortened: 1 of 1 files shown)


data/presets/area/footway.json
```diff
@@ -0,0 +1,22 @@
+{
+    "fields": [
+        "area/highway",
```
tyrasd, line 3: this field is not necessary anymore in this specific preset, since it would just duplicate the information already present in the title of the preset.

```suggestion
```

matkoniecz: I admit that I have no idea why exactly it will work but I applied it

data/presets/area/footway.json
```diff
@@ -0,0 +1,21 @@
+{
+    "fields": [
+        "surface"
+    ],
+    "geometry": [
+        "area"
+    ],
+    "terms": [
+        "area:highway",
```
1ec5, line 9: By the way, this preset will already be searchable by “footway” regardless of language, because iD will match tag values: https://github.com/openstreetmap/iD/issues/8869#issuecomment-1004306198. If matching keys is also desired, maybe that should be a more general, non-preset-specific feature on the iD side, so other consumers like Overpass turbo can decide not to match keys if it doesn’t make sense for their use case.

PR #330, opened 2022-01-05
Allow areas for charging_station

nchristensen:
Charging stations often take up a fixed number of parking spaces which can be easily mapped as a fixed area.

### DIFF
--- a/data/presets/amenity/charging_station.json
+++ b/data/presets/amenity/charging_station.json
@@ -17,7 +17,8 @@
         "manufacturer"
     ],
     "geometry": [
-        "point"
+        "point",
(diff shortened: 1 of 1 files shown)


tordans:
We should also update the German Wiki https://wiki.openstreetmap.org/wiki/DE:Tag:amenity%3Dcharging_station which only recommends points where the english Wiki https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station recommends both.

tyrasd:
Usage on ways is currently quite low (approx. 300 out of 60k), but I see no reason why these should be limited to just nodes.

PR #339, opened 2022-01-07
Parcel locker category change

ttomasz:
Implements proposal changing parcel lockers tagging which was recently approved: https://wiki.openstreetmap.org/wiki/Proposed_features/amenity%3Dparcel_locker

Essentially just moves presets that had vending_machine=parcel_pickup to amenity=parcel_locker and adds old tags to deprecated.json

Also added some fields (description, ref, payment) and some terms


data/deprecated.json
```diff
@@ -175,6 +175,22 @@
     "old": {"amenity": "vending_machine", "vending": "news_papers"},
     "replace": {"amenity": "vending_machine", "vending": "newspapers"}
   },
+  {
+    "old": {"amenity": "vending_machine", "vending": "parcel_pickup;parcel_mail_in"},
+    "replace": {"amenity": "parcel_locker", "parcel_mail_in": "yes"}
+  },
+  {
+    "old": {"amenity": "vending_machine", "vending": "parcel_mail_in"},
+    "replace": {"amenity": "parcel_locker", "parcel_mail_in": "yes"}
```
tyrasd, line 184: shouldn't this one have `"parcel_pickup": "no"` ?

tyrasd (dismissed):
Strictly speaking this is missing a preset for "drop off only" parcel locker, right? I guess those are pretty uncommon, but they would be misclassified as a _"Parcel Pickup/Dropoff Locker"_ with the proposed solution, which is not great. Previously they didn't match any of the presets.

Why not consolidate the two presets into a single "Parcel Locker" one with two new checkbox-fields for the tags `parcel_pickup` (placeholder could be "assumed to yes") and `parcel_mail_in` respectively?

tyrasd:
PS: It would also be good if there was a proper (non-proposal) wiki page before we include the preset in iD.

matkoniecz:
> Why not consolidate the two presets into a single "Parcel Locker" one with two new checkbox-fields for the tags parcel_mail_in (placeholder could be "assumed to yes") and parcel_mail_in respectively?

This way tagging it would require more clicks (instead of selecting it from the list -> select basic packstation, select both checkboxes)

data/deprecated.json
ttomasz: You're right. Fixed it.

ttomasz:
> Strictly speaking this is missing a preset for "drop off only" parcel locker, right? I guess those are pretty uncommon, but they would be misclassified as a _"Parcel Pickup/Dropoff Locker"_ with the proposed solution, which is not great. Previously they didn't match any of the presets.

You're right. This was a mostly theoretical possibility for me since I have never seen such locker but as it's allowed it's good to have a preset for it. I added it.

> Why not consolidate the two presets into a single "Parcel Locker" one with two new checkbox-fields for the tags `parcel_mail_in` (placeholder could be "assumed to yes") and `parcel_mail_in` respectively?

I guess one preset might be better than three. On the other hand during RFC some people were quite adamant about marking parcel_pickup as not required at treating it as assumed to be "yes". With simple checkboxes it could lead to people adding these fields when not necessary. I don't know what's better... I'll try to merge these into single preset in the next commit.

tyrasd:
Thanks for the update! But I still see one (or two) additional potential issue(s) here: If a mapper only knows that a parcel locker exists at a given location, but doesn't know whether is also accepts dropping off of parcels or not, they would be forced to choose one of the three presets. The correct one to choose would be _"Parcel Pickup Locker"_. The second (somewhat related) issue is that the current "Parcel Pickup Locker" preset doesn't allow to explicitly specify that the dropping off of parcels is actually not possible: since `parcel_mail_in` is not set, it could be the case that it's value was just unknown when the feature was mapped.

> With simple checkboxes it could lead to people adding these fields when not necessary.

I think this is not really the case when the description of the "default" case is well chosen. For example, the `oneway` field displays _"assumed to be no"_ for the unset case, and I think most mappers actually don't explicitly set the checkbox to `no` on regular non-oneway streets. This should work here just fine as well.

> This way tagging it would require more clicks

I think two clicks is just fine for such a feature. I think we should go for it.

ttomasz:
Didn't know that there were such options :sweat_smile: I updated the preset, let me know if this is correct or needs some fixes :)

Issue #340, opened 2022-01-07
Use embassy=yes instead of embassy=embassy


overflorian:
Hi,
In iD, I propose to stop suggesting the use of embassy=embassy
An embassy should instead be tagged embassy=yes, as detailled in the [documentation](https://wiki.openstreetmap.org/wiki/Key:embassy).

### Current situation
For an embassy the following tags are used:
* office=diplomatic
* diplomatic=embassy
* both values embassy=embassy and embassy=yes are proposed to the user

### Proposed evolution
The editor stop to suggest embassy=embassy and only propose the following tags to use, as described in the [documentation](https://wiki.openstreetmap.org/wiki/Key:embassy):
* embassy=yes - an classic embassy with an ambassador
* embassy=high_commission - high commission
* embassy=nunciature - diplomatic mission sent from the Vatican
* embassy=interests_section - a country that represents another sovereign state in a country where it lacks its own diplomatic representation
* embassy=mission - diplomatic mission to or from a multilateral organization such as the United Nations
* embassy=delegation - a delegation is headed by an ambassador
* embassy=branch_embassy - like trade office or cultural centre
* embassy=residence

tyrasd:
Do I understand you correctly that the issue is that the value `embassy` is proposed in the "Type" field of the embassy preset? See screenshot below:

<img src="https://user-images.githubusercontent.com/1927298/148597618-0f970fbe-bc56-426f-ab60-19f9cdca3eb5.png" width="300">

overflorian:
Yes

matkoniecz:
https://github.com/openstreetmap/id-tagging-schema/pull/341 is related

overflorian:
Hi, you closed this issue but I still see the "embassy" value in the form. Should I just wait for your implementation to appear?
Thanks.

tyrasd:
The changes will be published (i.e. made available for the public iD editor) in the next [release](https://github.com/openstreetmap/id-tagging-schema/releases) of this package. I'm still in the process of working through the backlog of the issues of this repository right now, but am planning to publish a release relatively soon (in one or two weeks presumably).

PR #341, opened 2022-01-08
embassy=embassy to embassy=yes (new deprecation)

matkoniecz:
embassy=embassy is an unwanted tagging fueled by iD presets showing top taginfo suggestions

In discussion on tagging mailing list noone considered embassy=embassy as a good idea

see https://lists.openstreetmap.org/pipermail/tagging/2022-January/063459.html and say https://lists.openstreetmap.org/pipermail/tagging/2022-January/063501.html

https://taginfo.openstreetmap.org/tags/embassy=embassy#chronology

excluding embassy=embassy from taginfo suggestions also would work - but AFAIK there is no way to achieve it right now

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -563,6 +563,10 @@
     "replace": {"embankment": "yes"}
   },
   {
+    "old": {"embassy": "embassy"},
+    "replace": {"embassy": "yes"}
+  },
+  {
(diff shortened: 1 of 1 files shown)


tyrasd:
Thanks! FYI: I've combined this one with #343 (implementation of #340).

PR #342, opened 2022-01-10
Add direction field to benches

fdr:
I've noticed a number of benches with directions already in OSM, but without this set, it's something of a hassle to do while using GoMap.

### DIFF
--- a/data/presets/amenity/bench.json
+++ b/data/presets/amenity/bench.json
@@ -8,6 +8,7 @@
     ],
     "moreFields": [
         "access_simple",
+        "direction",
         "height",
         "inscription",
         "level",


tyrasd (commented):
I'm wondering if this one would be better included as an optional field (in `"moreFields"`)? Currently, about 1.65% of benches have the direction tag specified (for comparison, the other fields `backrest`, `seats`, `material` all all well above 10%).

tordans:
+1 fore moreFields. AFAIK GoMap does not distinguish between fields and moreFields (just lists them all in one list), so this change would not affect the goal of getting the tag in the default-list of GoMap.

tyrasd:
There are no hard rules here. Usually, "fields" are only used for the essential properties of map features. This is because mappers (especially beginners) sometimes feel urged to fill out all fields that are displayed. Now, if the direction of a bench should count as essential or not is of course quite a subjective matter. From my personal experience (and going by the numbers mentioned above), the direction is just not very often mapped for benches, that's why I suggested "moreFields".

PS: Looking through the other presets, I can see that there are quite a few more presets where the "direction" could be made optional by using the same argument.

fdr:
Alright, well, done that way, as you suggest.

tyrasd:
thanks!

PR #343, opened 2022-01-10
specify values for "embassy" tag

### LINKED ISSUE #340: Use embassy=yes instead of embassy=embassy

tyrasd:
closes #340

I'm not completely sure with some of the labels I used here. 
If someone could perhaps proof-read these would be very much appreciated! :blush:  I was especially unsure about the following ones:

* `"residence": "Official Residence of an Ambassador"` – the [wiki](https://wiki.openstreetmap.org/wiki/Key:embassy) describes it as the _Official residence of an ambassador or other diplomatic chief of mission_.
* `"delegation": "Delegation"` – AFAICS this one seems to exclusively (?) for the diplomatic missions of the EU (see [wikipedia](https://en.wikipedia.org/wiki/Foreign_relations_of_the_European_Union)).
* `"interests_section": "Interests Section"`

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -579,6 +579,10 @@
     "replace": {"embankment": "yes"}
   },
   {
+    "old": {"embassy": "embassy"},
+    "replace": {"embassy": "yes"}
+  },
+  {
     "old": {"entrance": "emergency_exit"},
     "replace": {"entrance": "emergency"}
   },
--- a/data/fields/embassy.json
+++ b/data/fields/embassy.json
@@ -1,5 +1,18 @@
 {
     "key": "embassy",
+    "label": "Type",
     "type": "combo",
-    "label": "Type"
+    "autoSuggestions": false,
+    "strings": {
+        "options": {
+            "yes": "Embassy",
+            "residence": "Official Residence of an Ambassador",
+            "high_commission": "High Commission",
+            "mission": "Diplomatic Mission",
+            "branch_embassy": "Branch of an Embassy",
+            "nunciature": "Diplomatic Mission of the Holy See",
+            "delegation": "Delegation",
+            "interests_section": "Interests Section"
+        }
+    }
 }
--- a/interim/source_strings.yaml
+++ b/interim/source_strings.yaml
@@ -1064,6 +1064,23 @@ en:
       embassy:
         # embassy=*
         label: Type
+        options:
+          # embassy=branch_embassy
+          branch_embassy: Branch of an Embassy
+          # embassy=delegation
+          delegation: Delegation
+          # embassy=high_commission
+          high_commission: High Commission
+          # embassy=interests_section
+          interests_section: Interests Section
+          # embassy=mission
+          mission: Diplomatic Mission
+          # embassy=nunciature
+          nunciature: Diplomatic Mission of the Holy See
+          # embassy=residence
+          residence: Official Residence of an Ambassador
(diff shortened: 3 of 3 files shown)


ZeLonewolf:
The labels are fine.

However, the tag `embassy=yes` is useless cruft that should not get added to the database.  It provides _no_ information beyond `amenity=embassy`.  It should be deleted unless the mapper specifies a specific type of embassy.

tyrasd:
Thanks.

I thought `embassy=yes` was useful since it explicitly marks a diplomatic "site" as being an actual "regular" embassy; that it is not any of the other possible kinds (e.g. branch, mission, delegation, etc.):

osm wiki for [`embassy=yes`](https://wiki.openstreetmap.org/wiki/Tag%3Aembassy%3Dyes):
> Bears a sign labeling it as an embassy and is headed by an ambassador. 

A simple omission of the `embassy` tag would leave the actual kind unspecified, wouldn't it?

ZeLonewolf:
Okay, fair enough. It looks like the original proposal did intentionally make an embassy=yes, which is still weird but that's OSM for you.

Bibi56:
> Okay, fair enough. It looks like the original proposal did intentionally make an embassy=yes, which is still weird but that's OSM for you.

Exactly. `embassy=full_range_embassy` or whatsoever would have made more sense. It's the residence (or not!) and the chancery and the consular services (or not!). 

- Residence included: [Embassy of Cameroon](https://www.openstreetmap.org/node/7926524568)
- Residence excluded: [French Embassy](https://www.openstreetmap.org/relation/3964772)

- Consular services included: [French Embassy](https://www.openstreetmap.org/relation/3964772)
- Consular services excluded: [Embassy of India (Chancery)](https://www.openstreetmap.org/way/66646386)

Tag|Value
-- | --
diplomatic|embassy
embassy|yes
diplomatic:services:citizen_services | no
diplomatic:services:immigrant_visas | no
diplomatic:services:non-immigrant_visas | no

but [Embassy of India - Consular Wing](https://www.openstreetmap.org/way/66790229): 

Tag|Value
-- | --
diplomatic|consulate


To make more fun the usage changed over the years:
The former French embassy is now the [French ambassador's residence](https://www.openstreetmap.org/way/66790387).
The current [French embassy](https://www.openstreetmap.org/relation/3964772) was the French General Consulate.
At that time the French chancery was in [another place](https://www.openstreetmap.org/way/67069872)...

Issue #333, opened 2022-01-06
Introduce shared, automatic code formatting


tordans:
It would be a great help, if we had a shared code formatter config and documentation.

In https://github.com/openstreetmap/id-tagging-schema/pull/331 my Editor VS Code Reformatted all manner of things. I added a `.prettierrc` in https://github.com/openstreetmap/id-tagging-schema/pull/331/commits/23da8f4ba40540a6d2121a7d50633837e62ecf8a which made thinks better (https://prettier.io/docs/en/options.html).

- Should we start using prettier?
  - Are the settings/options good?
- Should re re-format all files once, or only once they are changed?
- Should we use a different formatter instead?
- Should we enforce the formatter via a Github Action as an error?
   - Or even apply the formatting changes via Github Actions as commit?

matkoniecz:
> Should re re-format all files once

Yes, this prevents unexpected linter changes mixed with actual PR changes.

> Should we enforce the formatter via a Github Action as an error? 

That would make contributing harder (right now format is sufficiently clear that it is possible to contribute without running code)

Maybe have it as an info warning, but not blocking merge?

> Or even apply the formatting changes via Github Actions as commit?
That seems nice if reasonable to setup and maintain.

tyrasd:
In general, I would not be opposed to add such a thing. Right out of the box, I don't particularly like the idea of of automated linting fixes to be honest. Also, I didn't experience any big issues with mixed-up indentations or other code formatting issues. Did you? If anything what sometimes feels slightly annoying is that the order of the different "sections" in the presets/fields are not always the same (e.g. sometimes the `name` is on top, sometimes on the bottom). A consistent order of these properties could make the files easier to read. Are there code linting tools which can do such a thing as well?

Since this would probably touch and change a lot of files, and it's low priority at the moment, I added this for the to do list of a next major release (working title `v4`) for now.

1ec5:
The only potential issue I can think of is unreadable diffs caused by automatic reformatter like the one @tordans has installed. Messy diffs could make mistakes harder to spot in code review. GitHub has an “ignore whitespace” option, but I don’t think it handles cases like where an array gets flattened onto one line instead of one line per item.

tordans:
> But to be honest, right out of the box, I don't particularly like the idea of of automated linting fixes. 

Those are two things. Prettier only changes formatting and makes sure your frontend code is formatted the same way across developers.

Linting would also be great, but a different tool/config.

[...]

PR #347, opened 2022-01-10
Create `.prettierrc`

tordans:
Solves https://github.com/openstreetmap/id-tagging-schema/issues/333.

Adds a config so users of https://prettier.io/ use the same formatting that does not change too much.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,9 @@
+{
+	"tabWidth": 4,
+	"useTabs": false,
+	"semi": false,
+	"printWidth": 80,
+	"proseWrap": "never",
+	"bracketSpacing": true,
+	"endOfLine": "lf"
+}


tyrasd:
Are you sure this works? As far as I can see, prettier is only meant to format javascript code and doesn't do anything with `.json` files. See https://github.com/prettier/prettier/issues/322#issuecomment-283432929. Am I overlooking something here?

What we could do pretty easily is to instead add a [`.editorconfig`](https://editorconfig.org/) file to define indentation and newline properties. Something like this should work.

```
root = true

[*]
charset = utf-8
end_of_line = lf
insert_final_newline = true
trim_trailing_whitespace = true

[**.json]
indent_style = space
indent_size = 4

[{package.json,.package-lock.json}]
indent_style = space
indent_size = 2
```

It doesn't support more advanced formatting rules, like `bracketSpacing` for example. :thinking:

PR #350, opened 2022-01-11
Add strings to castle_type.json

kjonosm:
Add some often used values as strings for translation

### DIFF
--- a/data/fields/castle_type.json
+++ b/data/fields/castle_type.json
@@ -1,5 +1,18 @@
 {
     "key": "castle_type",
     "type": "combo",
-    "label": "Type"
+    "label": "Type",
+    "strings": {
+        "options": {
+            "defensive": "Defensive castle",
+            "palace": "Palace",
+            "stately": "Stately home",
+            "manor": "Manor house",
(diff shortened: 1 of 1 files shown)


data/fields/castle_type.json
```diff
@@ -1,5 +1,18 @@
...
-    "label": "Type"
+    "label": "Type",
+    "strings": {
+        "options": {
+            "defensive": "Defensive castle",
+            "palace": "Palace",
+            "stately": "Stately home",
+            "manor": "Manor house",
+            "fortress": "Fortress",
+            "castrum": "Roman fort",
```
tyrasd, line 12: the convention is to use [title case](https://en.wikipedia.org/wiki/Title_case) for labels. There are some exceptions here and there, but I would say that we should write it like this here:

```suggestion
            "defensive": "Defensive Castle",
            "palace": "Palace",
            "stately": "Stately Home",
            "manor": "Manor House",
            "fortress": "Fortress",
            "castrum": "Roman Fort",
```

tyrasd:
thanks!

PR #355, opened 2022-01-15
Add `bin` field to `vending=excrement_bags`

FloEdelmann:
Many excrement bag dispensers have an attached trash bin, so it makes sense to include it in the preset.

### DIFF
--- a/data/presets/amenity/vending_machine/excrement_bags.json
+++ b/data/presets/amenity/vending_machine/excrement_bags.json
@@ -3,10 +3,10 @@
     "fields": [
         "bin",
         "fee",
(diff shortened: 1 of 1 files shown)


tyrasd:
Thanks! I've tweaked the preset slightly to also include the `fee` and `operator` tags (as suggested on https://wiki.openstreetmap.org/wiki/Tag:vending%3Dexcrement_bags), and added a few fields as optional (`moreFields`) ones (i.e. the ones which are shown for the "parent" preset for all `amenity=vending_machines`).

FloEdelmann:
Cool, even better!

PR #356, opened 2022-01-17
more precise notary presets for common law countries

### LINKED ISSUE #300: Change the name of `Lawyer office`
Currently `office=lawyer + lawyer=notary` and `office=lawyer` have the same name (`Notary office`) which can cause confusion if the translator doesn't see the tags. Is it possible to change the preset name of `office=lawyer` to something else?

tyrasd:
* make `lawyer=notary` searchable
* make `office=notary` only searchable in common law countries
* more precise names: “_civil law notary_” + “_notary public_” in common law countries, “_notary office_” + “_notary public (common law)_” elsewhere

see #300


data/presets/office/lawyer/notary.json
```diff
@@ -0,0 +1,34 @@
...
+    },
+    "locationSet": {
+        "exclude": [
+            "US",
+            "UK",
+            "CA",
+            "AU",
+            "IE",
+            "NZ"
+        ]
```
1ec5, line 23: I think this has it backwards: non-lawyer notaries in civil law countries but lawyer notaries in common law countries?

tyrasd:
I'm wondering if there are more countries which have this distinction of notaries apart from the list I used for now (US, Canada, UK, Ireland, Australian and New Zealand).

data/presets/office/lawyer/notary.json
1ec5: Oh, I see, this preset only exists to have a more generic name in civil law countries, where “notary” is assumed to mean “civil law notary”.

simonpoole:
Notaries and lawyers tend to be distinct even say in DACH. I suspect that required qualifications are being confused with the service offered. 

In other words I would expect office=notary for a notary in CH too and not a variant of a lawyers office.

matkoniecz:
see https://github.com/openstreetmap/id-tagging-schema/issues/300#issuecomment-3547269476

closing this PR as discussed with @tyrasd

PR #365, opened 2022-01-22
Properly set parcel_pickup to yes in deprecation fix

### LINKED ISSUE #364: Proposed fix for parcel locker tagged with deprecated scheme suggests wrong tag value
The value for parcel_pickup is incorrectly proposed as `no` although it was previously tagged as available (`vending=parcel_pickup[;...]`).

**Actual**
```diff
- amenity=vending_machine
+ amenity=parcel_locker
+ parcel_mail_in=yes
+ parcel_pickup=no
- vending=parcel_pickup;parcel_mail_in
```

**Expected**
```diff
- amenity=vending_machine
+ amenity=parcel_locker
+ parcel_mail_in=yes
+ parcel_pickup=yes
- vending=parcel_pickup;parcel_mail_in
```

iD screenshot

**Overpass turbo query to find affected elements**
```
node
  [amenity=vending_machine]
  [vending="parcel_pickup;parcel_mail_in"]
  ({{bbox}});
out;
```
This change was implemented in #339 / #358.

riQQ:
Fixes #364

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -177,7 +177,7 @@
   },
   {
     "old": {"amenity": "vending_machine", "vending": "parcel_pickup;parcel_mail_in"},
-    "replace": {"amenity": "parcel_locker", "parcel_mail_in": "yes"}
+    "replace": {"amenity": "parcel_locker", "parcel_mail_in": "yes", "parcel_pickup": "yes"}
   },
   {
     "old": {"amenity": "vending_machine", "vending": "parcel_mail_in"},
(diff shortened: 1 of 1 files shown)


tyrasd:
Hey. Thanks for this suggestion. I solved it slightly differently in 609c0178b75b2e4907098e75514031c337737e05. Basically, I only removed the upgrade path of the `parcel_mail_in` case, which was causing the issue. Your solution would have also worked and would arguably have been an even better solution, because it would have preserved the upgrade path for mail-in-only parcel lockers features. I did however go with the other approach, because adding `parcel_pickup=yes` is somewhat discouraged by the [osm wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dparcel_locker#Related_tags), see also https://github.com/openstreetmap/id-tagging-schema/pull/339#issuecomment-1007670190. Since drop-off only parcel lockers are relatively rare, I think this solution is acceptable for the moment.

Ideally, there would be a way to indicate in the `deprecated.json` that upgrading a tag should explicitly not set a particular tag (e.g. something like this: `"replace": {"amenity": "parcel_locker", "parcel_mail_in": "yes", "parcel_pickup": null}`), but that would need to be implemented as a separate enhancement of this repository.

PR #366, opened 2022-01-23
Add Golf Pin preset

arch0345:
There is already a Golf Hole and Golf Tee preset, so it makes sense to have a [Golf Pin](https://wiki.openstreetmap.org/wiki/Tag:golf%3Dpin) preset. I've added a `ref_golf_hole` field in the case that the Golf Hole isn't mapped.

### DIFF
--- a/data/presets/golf/pin.json
+++ b/data/presets/golf/pin.json
@@ -1,0 +1,13 @@
+{
+    "icon": "temaki-golf_green",
+    "fields": [
+        "ref_golf_hole"
+    ],
+    "geometry": [
+        "vertex"
+    ],
+    "tags": {
+        "golf": "pin"
+    },
+    "name": "Golf Pin"
+}


tyrasd:
Interesting. I would have assumed that the pin locations should not be mapped in OSM, because the exact position of the pins are changed quite frequently on almost all golf courses in order to prevent excessive and uneven wear of the grass on the green (see for example this [source](https://www.alresfordgolf.co.uk/ufaqs/pin-positions-changed-greens/)). An exception could be golf courses with artificial turf greens perhaps, but I think those are pretty rare… I would have thought that the last node of the `golf=hole` would be sufficient to indicate the approximate or average location of the pin of a particular golf hole.

PS: Even the [wiki](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dgolf_course) recommends against mapping them:

> [`golf=pin`] For the location of the hole. Since this does change over time (as often as daily) to keep wear on the green even, it is not recommended to map it.

Is there some aspect of this that I'm overlooking?

arch0345:
Ah I didn't check the description on the golf course wiki page. I hadn't really thought about how pins are moved around: I'm going to close this because as you said, exceptions to this are pretty rare.

PR #368, opened 2022-01-24
Add presets for pedestrian crossings with traffic signals for lines

Dimitar5555:


tyrasd:
for reference: this would complement the vertex-presets added in #192

1ec5:
Somehow the terms for the traffic calming version of this preset wound up in the non–traffic calming version:

https://github.com/openstreetmap/id-tagging-schema/blob/3fc31f0a3cff4e227fd4f9da79c502c06731cfe0/data/presets/highway/footway/signals.json#L31-L42

tyrasd:
> traffic calming version

what version do you mean, exactly? after https://github.com/openstreetmap/id-tagging-schema/commit/4644b468ac5a9c1fc3d1521792e4265feb5a21b4 there should be no more "traffic calming" versions of the crossing presets.

1ec5:
`traffic_calming=table`-related terms like “hump” should be terms of the Raised field, not the crossing presets. #508 would move the terms to the correct place. (I should’ve posted the above comment to 4644b468ac5a9c1fc3d1521792e4265feb5a21b4, not this issue.)

tyrasd:
Ah, I see. That's now fixed in 3da4577, thanks for the hint.

PR #372, opened 2022-01-28
Add more disc golf features

arch0345:
This PR adds a preset for [Disc Golf Basket](https://wiki.openstreetmap.org/wiki/Tag:disc_golf=basket), [Disc Golf Hole](https://taginfo.openstreetmap.org/tags/disc_golf=hole), and [Disc Golf Tee](https://wiki.openstreetmap.org/wiki/Tag:disc_golf=tee). I decided to use the `par` field for the Disc Golf Hole preset rather than `disc_golf:par` as recommended by the wiki as it looks like the majority (735 vs 199) use `par` instead.

### DIFF
--- a/data/presets/disc_golf/hole.json
+++ b/data/presets/disc_golf/hole.json
@@ -1,0 +1,15 @@
+{
+    "icon": "temaki-disc_golf_basket",
+    "fields": [
+        "name",
+        "ref_golf_hole",
+        "par"
+    ],
+    "geometry": [
+        "line"
+    ],
+    "tags": {
+        "disc_golf": "hole"
+    },
+    "name": "Disc Golf Hole"
+}
--- a/data/presets/disc_golf/hole.json
+++ b/data/presets/disc_golf/hole.json
@@ -2,7 +2,7 @@
     "icon": "temaki-disc_golf_basket",
     "fields": [
         "name",
-        "ref_golf_hole",
+        "ref_disc_golf_hole",
         "par"
     ],
     "geometry": [
--- a/data/fields/ref_disc_golf_hole.json
+++ b/data/fields/ref_disc_golf_hole.json
@@ -1,0 +1,6 @@
+{
+    "key": "ref",
+    "type": "text",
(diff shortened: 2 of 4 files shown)


data/presets/disc_golf/hole.json
```diff
@@ -0,0 +1,15 @@
+{
+    "icon": "temaki-disc_golf_basket",
+    "fields": [
+        "name",
+        "ref_golf_hole",
```
tyrasd, line 5: :thinking: while this seems to work fine in English (where the term _hole_ is also used in the context of disc golf courses), it doesn't work in other Languages. For example, in German, different terms are used for disc golf and regular golf (_Bahn_ vs. _Loch_).

I think it would be better to add a new field for `ref_disc_golf_hole` or alternatively to use the generic `ref` field.

tyrasd (changes requested):
Thanks for this PR. Seems good in general and quite useful. See a few minor suggestions below.

I'm a bit torn on the `par` vs. `disc_golf:par` aspect. Your reasoning seems legit (usage is currently 700 for `par` vs. 200 for `disc_golf:par`[^1][^2]), but I think this should at least be first documented on the wiki as an alternative tagging possibility.

[^1]: `par` was also used earlier than `disc_golf:par`: [](https://user-images.githubusercontent.com/1927298/151529090-20c2b4c2-4c8d-4721-aaa0-d3a243293e08.png).

[^2]: PS: about 2,000 holes don't have any par specified.

data/presets/disc_golf/hole.json
arch0345: Thanks for letting me know about this, I've added a new `ref_disc_golf_hole` like you suggested.

arch0345:
> I think this should at least be first documented on the wiki as an alternative tagging possibility.
Just added it [here](https://wiki.openstreetmap.org/w/index.php?title=Tag:disc_golf%3Dhole&diff=prev&oldid=2253532)

tyrasd (approved):

Issue #76, opened 2020-12-07
A few new presets for NSI


bhousel:
NSI is depending on this repo for presets now - see https://github.com/osmlab/name-suggestion-index/issues/4718! 

Here are some features being tracked in NSI that we didn't find iD presets for in the latest build:
- [ ] Warning - no iD source preset found for amenity/motorcycle_rental
- [ ] Warning - no iD source preset found for office/security
- [ ] Warning - no iD source preset found for shop/repair
- [ ] Warning - no iD source preset found for emergency/lifeboat_station
- [ ] Warning - no iD source preset found for route/aerialway

bhousel:
Thanks! 👍

quincylvania:
@bhousel Sure! Thanks for all your great work on NSI 🚉

PR #382, opened 2022-02-01
update lifeboat station preset

k-yle:
The "Lifeboat Station" preset was added in 072d2dd (#76), but without any fields to further describe the lifeboat station. 

This PR adds the [`seamark:rescue_station:category`](https://wiki.openstreetmap.org/wiki/Seamarks/Categories_of_Objects#Rescue_Stations_.28CATRSC.29) field to let you specify what amenities the lifeboat station has (e.g. a lifeboat, seaplane, first aid, etc.)

This PR also suggests adding [`seamark:type=rescue_station`](https://wiki.openstreetmap.org/wiki/Tag:seamark:type=rescue_station) to these features, so that they render in OpenSeaMap.

### DIFF
--- a/data/fields/seamark/rescue_station/category.json
+++ b/data/fields/seamark/rescue_station/category.json
@@ -1,0 +1,19 @@
+{
+    "key": "seamark:rescue_station:category",
+    "type": "semiCombo",
+    "label": "Amenities",
+    "strings": {
+        "options": {
+            "lifeboat": "Lifeboat (on land)",
+            "lifeboat_on_mooring": "Lifeboat (at a mooring)",
+            "seaplane": "Seaplane",
+            "aircraft": "Aircraft",
+            "hovercraft": "Hovercraft",
+            "radio": "Radio Station",
+            "first_aid": "First aid",
+            "refuge_shipwrecked": "Shipwreck Refuge",
+            "refuge_intertidal": " Intertidal Refuge",
+            "tug": "Salvage tug"
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


data/fields/seamark/rescue_station/category.json
```diff
@@ -0,0 +1,19 @@
+{
+    "key": "seamark:rescue_station:category",
+    "type": "semiCombo",
+    "label": "Amenities",
+    "strings": {
+        "options": {
+            "lifeboat": "Lifeboat (on land)",
+            "lifeboat_on_mooring": "Lifeboat (at a mooring)",
```
tyrasd, line 8: I think we should title-case also the words in parentheses?

```suggestion
            "lifeboat": "Lifeboat (on Land)",
            "lifeboat_on_mooring": "Lifeboat (at a Mooring)",
```

data/fields/seamark/rescue_station/category.json
```diff
@@ -0,0 +1,19 @@
+{
+    "key": "seamark:rescue_station:category",
+    "type": "semiCombo",
+    "label": "Amenities",
```
tyrasd, line 4: Hm… _Amenities_ sounds a bit too unspecific here. Maybe the following would be better?
```suggestion
    "label": "Life Saving Equipment",
```

tyrasd (changes requested):
Thanks! A few small remarks are below.

PS: It's a bit annoying that there is also [`amenity=lifeboat_station`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dlifeboat_station) with very similar usage than [`emergency=lifeboat_station`](https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dlifeboat_station) (many features are currently mapped using both of these tags). But I think the preset is fine since the `emergency` is arguably a more precise tag.

k-yle:
thanks, I've made both changes

tyrasd (approved):

PR #391, opened 2022-02-05
Fix data/presets/amenity/ticket_validator format

quotquot:
`data/presets/amenity/ticket_validator` is not valid JSON, with trailing commas and a dangling curly bracket. This PR fixes these issues.


tyrasd:
Thanks! FYI: The file was also missing the `.json` file ending (see 7efb634), that's why the tests didn't catch the issues.

quotquot:
Thanks to you and other for this fantastic work.

I found these errors because I have a tool that processes the schema. Besides the errors fixed by the PR, my tool also choked up on `emergency/water_reservoir`, which has the same tags as `emergency/water_tank`. In fact, the tags are different in their respective preset files, i.e. `data/presets/emergency/water_reservoir.json`:
```json
    {
        "emergency": "water_tank",
        "natural": "water",
        "water": "reservoir"
    }
```
and `data/presets/emergency/water_tank.json`;
```json
    {
        "emergency": "water_tank",
        "landuse": "reservoir"
    }
```
but they are the same in `dist/presets.json` (which my tool uses), both being the latter two-key tags.  Is `dist/presets.json` out of date or is there a weird bug somewhere that sets the wrong tags for `emergency/water_reservoir` (or am I misunderstanding something)? Thanks in advance.

PR #407, opened 2022-02-20
Add Lighting Mast preset

arch0345:
This PR adds a preset for Lighting Masts (`man_made=mast` + `tower:type=lighting`). [`tower:type=lighting`](https://wiki.openstreetmap.org/wiki/Tag:tower:type%3Dlighting) is currently 'in use' and [is being used over 40,000 times](https://taginfo.openstreetmap.org/tags/?key=tower%3Atype&value=lighting). For now, the generic mast icon is being used for this preset but ideally this should get its own icon as suggested in the `#id` channel on the OSM World Discord server:

image

Edit: it's `tower:type` not `mast:type`

### DIFF
--- a/data/presets/man_made/mast/lighting.json
+++ b/data/presets/man_made/mast/lighting.json
@@ -1,0 +1,29 @@
+{
+    "icon": "temaki-mast",
+    "fields": [
+        "{man_made/mast}",
+        "direction_point"
+    ],
+    "moreFields": [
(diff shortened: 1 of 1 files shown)


tyrasd:
https://github.com/ideditor/temaki/pull/73 is a proposal for a potential icon for this preset:

1ec5:
#424 ran `npm run build` to add localizable strings for this preset.

PR #412, opened 2022-03-03
Adding Rice

waso99:
Adding Rice


tyrasd:
Hi. Thans for this proposal, but this particular tag (`cuisine=rice`) is used too rarely ([currently 27 times](https://taginfo.openstreetmap.org/tags/cuisine=rice)) to be included in the list of predefined values. It's further also not listed on the table on the OSM wiki page for the [cuisine tag](https://wiki.openstreetmap.org/wiki/Key:cuisine).

PR #416, opened 2022-03-08
remove construction=* when changing road presets

k-yle:
Closes openstreetmap/iD#8877 and closes openstreetmap/iD#8886

When changing a road-under-construction to a normal road, the `construction=*` tag will now be removed.

### DIFF
--- a/data/presets/highway/construction.json
+++ b/data/presets/highway/construction.json
@@ -17,6 +17,11 @@
         "highway": "construction",
         "access": "no"
     },
+    "removeTags": {
+        "highway": "construction",
+        "access": "no",
+        "construction": "*"
+    },
     "terms": [
         "closure",
         "construction",


tordans:
I wonder, if this makes the situation better … or just different.
It's better in the sense, that it will make sure no non-fitting tags remain.
But it's worse in the sense, that is …
* does not account for the complexity of access-tagging
* does remove values that would ideally be reused

About the access:
I am pretty sure there are edge cases, where just removing "access=no" will result in non-ideal data. The mapper should choose the right access tag more explicitly, I think. Like for situations where access would more precisely be customers or private or permissive.

About the reuse:
The value of construction would ideally be used to choose the new preset. The UX could use the data to guide the user a lot more. https://github.com/openstreetmap/iD/issues/6168 had a few good ideas about the UI for this (but went stale for other reasons).

_Maybe we should look at it this way:_ There could be an iD feature that uses the existing data to guide the user to a better re-tagging. However, until that is build, cleaning the data like this PR does is better than keeping outdated data around(?).

k-yle:
> I wonder, if this makes the situation better … or just different.

This is definitely an improvement. Something like `highway=trunk` + `construction=primary` is not just confusing, it's wrong and will cause some routers to avoid that road. For example, [these directions](https://www.openstreetmap.org/directions?engine=fossgis_osrm_car&route=-37.72016%2C175.24382%3B-37.72334%2C175.24218#map=18/-37.72211/175.24308) take a detour because [Keston Crescent](https://osm.org/way/285634300) has `highway=residential` + `construction=residential`.


> I am pretty sure there are edge cases, where just removing "access=no" will result in non-ideal data

Just so we're on the same page: iD already removes `access=no` when changing a closed road to any other preset. I'm not proposing to change that, but it might be worth discussing.

PR #417, opened 2022-03-08
add name field to beach preset

k-yle:
It's weird how this preset has the `surface` tag at the top, instead of `name`. I've seen multiple mappers accidently add the name into the `surface` field, because it's the top one. 

It's hard to determine how many of these errors have been made, but taginfo alone suggests there are over 100[^1] that have gone unnoticed. It's also not clear whether these were caused by iD.

[^1]: https://taginfo.osm.org/search?q=surface%3Dbeach (83), https://taginfo.osm.org/search?q=surface%3Dplage (16), https://taginfo.osm.org/search?q=surface%3Dplaya (32), https://taginfo.osm.org/search?q=surface%3Dпляж (8), https://taginfo.osm.org/search?q=surface%3Dstrand (1)

### DIFF
--- a/data/presets/natural/beach.json
+++ b/data/presets/natural/beach.json
@@ -1,6 +1,7 @@
 {
     "icon": "temaki-beach",
     "fields": [
+        "name",
         "surface"
     ],
     "geometry": [


tordans:
I am wondering: From your list of examples …

>https://taginfo.osm.org/search?q=surface%3Dbeach (83), https://taginfo.osm.org/search?q=surface%3Dplage (16), https://taginfo.osm.org/search?q=surface%3Dplaya (32), [https://taginfo.osm.org/search?q=surface%3Dпляж](https://taginfo.osm.org/search?q=surface%3D%D0%BF%D0%BB%D1%8F%D0%B6) (8), https://taginfo.osm.org/search?q=surface%3Dstrand (1) [↩](https://github.com/openstreetmap/id-tagging-schema/pull/417#user-content-fnref-1-a48f331b2577eb717adae42f63370d7d)

… those don't sound to me like values that should be place in the nave field instead. `natural=beach + name=Beach` sounds like a miss-use of the name field the same way as it is the wrong value for the surface field.

---

Another approach could be, to add a translated list of surface values to the preset which suggest [the list from the wiki](https://wiki.openstreetmap.org/wiki/Tag:natural=beach#How_to_map) "sand, gravel, pebbles, stones, shells".

AFAIK that would make iD suggest this list primarily and not show the taginfo top-x-list anymore. However, AFAIK it would still allow for custom values for edge cases.

---

Just for reference: https://taginfo.openstreetmap.org/tags/natural=beach#combinations gives surface as the top combination and name right after. So I don't see any reason not to add it; but still think it will not solve the initial problem.

* 40 72526.97% | surface=* | 40 725
* 38 82525.71% | name=* | 38 825
* 31 19320.66% | surface=sand | 31 193

PR #419, opened 2022-03-10
use correct word for restrooms in default en-US translation

westnordost:
`amenity=toilets` are correctly referred to _Restrooms_ in American-English. "Toilets" seem to be considered vulgar.

I don't know if "toilets" is acceptable in British-English, but for that, a translation is maintained in Transifex.

This change was triggered by a discussion on the #language channel on OSM-Slack where I asked whether it would be weird to say _"This toilet has a baby changing table"_
The overwhelming answer was "yes" as a "toilet" is simply understood as the thing you take a shit on, not the room + facilities. But `amenity=toilets` is supposed to be the room + facilities.

### DIFF
--- a/data/presets/amenity/toilets.json
+++ b/data/presets/amenity/toilets.json
@@ -36,9 +36,9 @@
     "tags": {
         "amenity": "toilets"
     },
-    "name": "Toilets",
+    "name": "Restroom",
     "aliases": [
-        "Restroom",
+        "Toilets",
         "Bathroom",
         "Lavatory",
         "Water Closet"


ZeLonewolf:
I agree that "restroom" is the correct term in US English.

matkoniecz (approved):

zymurgic:
'toilets' are perfectly acceptable in British English, and not at all vulgar.

westnordost:
(Same in German. ) But well, that explains why the tag is `amenity=toilets` and not `amenity=restrooms` as the tags use in general British English. The source strings for the iD presets are however in American English.

For British English, a translation is maintained on transifex ( https://www.transifex.com/openstreetmap/id-editor/ ) where all strings that differ between British and American English are added in the **English (United Kingdom)** localization.

PR #423, opened 2022-03-22
Garage(s): Only keep capacity as `fields`

tordans:
Move fields from from https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/building.json to the moreFields section to only show very relevant fields by default.

When using those presets I noticed that I tend to add the capacity to the level or height fields instead of the capacity fields. Since those (and all other) are not really common on garages, we should move the to the moreFields.

---

Would it be possible to write the code like this which could result in the same but a better data structure. However, I am not sure if the moreField would just merge the fields + moreFields values if used like this.
```
{
    "icon": "fas-warehouse",
    "fields": [
        "capacity"
    ],
    "moreFields": [
        "{building}"
    ],
}
```

### DIFF
--- a/data/presets/building/garage.json
+++ b/data/presets/building/garage.json
@@ -1,7 +1,10 @@
 {
     "icon": "fas-warehouse",
     "fields": [
-        "{building/garages}",
+        "{building/garages}"
+    ],
+    "moreFields": [
+        "{building/garages}"
     ],
     "geometry": [
         "area"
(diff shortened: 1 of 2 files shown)


tyrasd:
> Would it be possible to write the code like this which could result in the same but a better data structure.

This inheritance works separately for `fields` and `moreFields`: In your example, this would only copy the `moreFields` from the `building` preset into the `garages` preset. I've tweaked your PR to use this inheritance to get rid of some of the fields in the list.

See https://github.com/ideditor/schema-builder/issues/53 for further potential enhancements in this direction.

PR #424, opened 2022-03-22
Avoid snake_case on not:name

1ec5:
Disabled automatic snake_case formatting on values of the Incorrect Names (`not:name`) field. Freeform text shouldn’t be automatically lowercased.


interim/source_strings.yaml
```diff
@@ -7217,6 +7217,11 @@ en:
         name: Television Broadcast Mast
         # 'terms: antenna,broadcast tower,communication mast,communication tower,guyed tower,television mast,television tower,transmission mast,transmission tower,tv mast,tv tower'
         terms: <translate with synonyms or related terms for 'Television Broadcast Mast', separated by commas>
+      man_made/mast/lighting:
+        # man_made=mast + tower:type=lighting | Translate the primary name. Optionally, add equivalent synonyms on newlines in order of preference (press the Return key).
+        name: Lighting Mast
```
1ec5, line 7222: These strings were excluded from #407.

PR #425, opened 2022-03-24
Update wastewater_plant.json

zymurgic:
adding a ref field, because they're often on the signage, and they match official open-data listings for correlation, so hinting to the users to fill out if known helps.
Also, they're known as wastewater treatment works

### DIFF
--- a/data/presets/man_made/wastewater_plant.json
+++ b/data/presets/man_made/wastewater_plant.json
@@ -21,6 +21,8 @@
         "sewage*",
         "water treatment plant",
         "wastewater treatment works",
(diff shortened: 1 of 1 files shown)


data/presets/man_made/wastewater_plant.json
```diff
@@ -19,6 +20,7 @@
     "terms": [
         "sewage*",
         "water treatment plant",
+        "wastewater treatment works",
```
1ec5, line 23: Yep, and there are a few other words for this in American English too, “wastewater treatment plant” and “WWTP” being the most common.

PR #427, opened 2022-03-26
Add some more values to beauty=*

Zverik:
Added some frequent non-ambigous values from [taginfo](https://taginfo.openstreetmap.org/keys/?key=beauty#values) and [wiki](https://wiki.openstreetmap.org/wiki/Key:beauty).

### DIFF
--- a/data/fields/beauty.json
+++ b/data/fields/beauty.json
@@ -7,7 +7,7 @@
             "nails": "Manicure / Pedicure",
             "tanning": "Tanning",
             "cosmetics": "Cosmetics",
-            "spa": "Hydrotherapy",
+            "spa": "Day Spa",
             "skin_care": "Skin Care",
             "waxing": "Waxing",
             "hair_removal": "Hair Removal"
--- a/data/fields/beauty.json
+++ b/data/fields/beauty.json
@@ -6,7 +6,10 @@
         "options": {
             "nails": "Nails",
             "tanning": "Tanning",
-            "cosmetics": "Cosmetics"
+            "cosmetics": "Cosmetics",
+            "hair": "Hair",
+            "spa": "Spa",
+            "skin_care": "Skin Care"
         }
     }
 }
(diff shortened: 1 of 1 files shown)


tyrasd:
see my tweaks in 7ef7f8d: what do you think?

Zverik:
Yeah, looks good. `hair_removal` is indeed described in the wiki and has enough usages.

Zverik:
Although I'd argue "Spa" is a wide-understood term that people would search for.

tyrasd:
> Although I'd argue "Spa" is a wide-understood term that people would search for.

I had always associated the term "spa" exclusively with thermal baths (see wikipedia: https://en.wikipedia.org/wiki/Spa) to be honest. Perhaps "Spa Treatment" would be a better description instead of "Hydrotherapy" (but on the other hand, [spa treatment](https://en.wikipedia.org/wiki/Spa#Treatments) can apparently be used for basically any beauty treatment)?
@1ec5 would you mind sharing your opinion as a native speaker?

data/fields/beauty.json
```diff
@@ -4,9 +4,13 @@
     "label": "Services",
     "strings": {
         "options": {
-            "nails": "Nails",
+            "nails": "Manicure / Pedicure",
             "tanning": "Tanning",
-            "cosmetics": "Cosmetics"
+            "cosmetics": "Cosmetics",
+            "spa": "Hydrotherapy",
```
1ec5, line 10: I’m not an expert on this industry, but day spas or medspas (medical spas) are very common. They’re distinct from hydrotherapy facilities, which became virtually nonexistent in the U.S. generations ago, and also completely independent of any natural spas.

[](https://commons.wikimedia.org/wiki/File:Philipe's_Day_Spa.jpg)

tyrasd: thanks, let's use _Day Spa_. I guess the corresponding row of the table on the [wiki](https://wiki.openstreetmap.org/wiki/Key:beauty) could be updated as well. :wink:

Zverik: Updated the wiki, and the list looks good to me now :)

PR #428, opened 2022-03-28
Add refs to atms, payment terminals, and vending machines

Zverik:
These things usually are set in bulk by a same company, hence they have reference numbers. I've been contacted by a person who likes to collect these.


tordans (approved):
I think the moreFields are a good place for those.

PR #441, opened 2022-04-13
Add Tetherball Court preset

arch0345:
[Tetherball](https://wiki.openstreetmap.org/wiki/Tag:sport=tetherball) is a sport often played in schoolyards. [`sport=tetherball`](https://taginfo.openstreetmap.org/tags/sport=tetherball) currently has around 1K uses.

### DIFF
--- a/data/presets/leisure/pitch/tetherball.json
+++ b/data/presets/leisure/pitch/tetherball.json
@@ -1,0 +1,24 @@
+{
+    "icon": "maki-pitch",
+    "fields": [
+        "{leisure/pitch}",
+        "access_simple"
+    ],
+    "geometry": [
+        "area",
+        "point"
+    ],
+    "tags": {
+        "leisure": "pitch",
+        "sport": "tetherball"
+    },
+    "reference": {
+        "key": "sport",
+        "value": "tetherball"
+    },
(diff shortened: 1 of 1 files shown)


data/presets/leisure/pitch/tetherball.json
```diff
@@ -0,0 +1,24 @@
+{
+    "icon": "maki-pitch",
+    "fields": [
+        "{leisure/pitch}",
+        "access_simple"
+    ],
+    "geometry": [
+        "area",
+        "point"
```
1ec5, line 9: As noted on [the wiki](https://wiki.openstreetmap.org/wiki/Tag:sport%3Dtetherball#Alternative_tagging), `playground=tetherball` is also commonly used to indicate the tetherball pole itself. That would be a separate preset, but in the meantime, this preset would be available for points as well as areas, consistent with other kinds of pitches. I think that’s OK; it seems unlikely that a mapper unfamiliar with that tag would convert it into a tetherball court inside a tetherball court.

1ec5 (approved):
Looks good to me. This preset will make it easier to map schoolyards (#134), which in the U.S. typically feature an array of tetherball courts.

PR #442, opened 2022-04-24
Add Radar Speed Sign preset

arch0345:
This PR adds a preset for [`highway=speed_display`](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dspeed_display). [Radar speed signs](https://en.wikipedia.org/wiki/Radar_speed_sign) are commonly used in urban America as traffic calming devices and can sometimes be found under speed limit signs, for example:
<a title="Downtowngal, CC BY-SA 3.0 &lt;https://creativecommons.org/licenses/by-sa/3.0&gt;, via Wikimedia Commons" href="https://commons.wikimedia.org/wiki/File:Your_speed_is.jpg">Your speed is</a>
[`highway=speed_display`](https://taginfo.openstreetmap.org/tags/?key=highway&value=speed_display#) currently has around 1.5K uses.

### DIFF
--- a/data/presets/highway/speed_display.json
+++ b/data/presets/highway/speed_display.json
@@ -1,0 +1,24 @@
+{
+    "icon": "maki-square-stroked",
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "fields": [
+        "direction_point",
+        "direction_vertex"
+    ],
+    "tags": {
+        "highway": "speed_display"
+    },
+    "terms": [
+        "driver feedback sign",
+        "dynamic speed display",
+        "speed awareness message",
+        "speed display board",
+        "speed radar sign",
+        "traffic calming sign",
+        "your speed sign"
+    ],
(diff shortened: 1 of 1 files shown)


data/presets/highway/speed_display.json
```diff
@@ -0,0 +1,24 @@
+{
+    "icon": "maki-square-stroked",
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "fields": [
+        "direction_point",
+        "direction_vertex"
```
1ec5, line 9: The wiki also suggests `maxspeed`; there’s an existing field we can use for that.

arch0345: > The wiki also suggests `maxspeed`; there’s an existing field we can use for that.

Decided to also add `maxspeed/advisory` for cases like this:
image

tyrasd:
thanks for the ping. it looks good to me :+1:

PR #444, opened 2022-04-26
Add `sidewalk` field

arch0345:
This PR adds a field for the `sidewalk` key to be used on `highway=*` presets. So far I've only added values from [the wiki](https://wiki.openstreetmap.org/wiki/Key:sidewalk#Values), which are:
* no (no sidewalk)
* both (both sides)
* left (left side only)
* right (right side only)
* separate (mapped as a separate way)

Ideally this field should also support `sidewalk:left` and `sidewalk:right` for complex scenarios as described [here](https://wiki.openstreetmap.org/wiki/Key:sidewalkSeparately mapped sidewalks on only one side).

### DIFF
--- a/data/fields/sidewalk.json
+++ b/data/fields/sidewalk.json
@@ -1,0 +1,35 @@
+{
+    "keys": [
+        "sidewalk"
+    ],
+    "reference": {
+        "key": "sidewalk"
+    },
+    "type": "sidewalk",
+    "label": "Sidewalk",
+    "placeholder": "none",
+    "strings": {
+        "options": {
+            "no": {
+                "title": "None",
+                "description": "No sidewalk"
+            },
+            "both": {
+                "title": "Both",
+                "description": "A sidewalk is on both sides of the road"
+            },
+            "left": {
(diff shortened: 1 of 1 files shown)


bhousel:
Please don't.
Ton of discussion here: https://github.com/openstreetmap/iD/issues/1762 
(some of the comments got kind of nasty 😞 ).

The fact is - consumers working with sidewalk data these days prefer for the sidewalks to be mapped as separate features that can carry their own attributes about their suitability for use.

1ec5:
There’s also [a more recent ongoing discussion](https://community.openstreetmap.org/t/is-there-consensus-on-mapping-pavements-sidewalks-separately-to-roads/1067) about whether there’s a consensus about which representation of sidewalks is better. Spoiler alert: no consensus, for some definition of consensus. But at least the discussion has been edifying so far. If you have any opinions on the matter, you’re welcome to chime in there.

bhousel:
> Spoiler alert: no consensus, for some definition of consensus.

Then why bother?

PR #445, opened 2022-04-26
Add Padel Court preset

danieldegroot2:
Adds a Padel Court preset. Not to be confused with Paddle tennis. Padel Courts have a layout similar to tennis, but without the marked margins on the side.
(from a quick look, padel has four times more uses on ways than pickleball)


danieldegroot2:
_cc: @1ec5 request for review_

PR #447, opened 2022-05-01
Update power presets following transformer tagging extension

### LINKED ISSUE #415: Tagging changes for power transformers
Dear maintainers,

Recent tagging additions has been made to power transformers tagging and some of them could be relevant for ID presets.

[transformer=main](https://wiki.openstreetmap.org/wiki/Tag:transformer%3Dmain) is now documented on wiki

It is intended to replace discouraged values :
* [transformer=traction](https://wiki.openstreetmap.org/wiki/Tag:transformer%3Dtraction)
* [transformer=auto](https://wiki.openstreetmap.org/wiki/Tag:transformer%3Dauto) to be replaced by [windings:auto=yes](https://wiki.openstreetmap.org/wiki/Key:windings:auto)
* transformer=minor_distribution

Some voltage keys should also be discouraged. Is it possible to raise a warning regarding the following please ?
* voltage=* in combination with power=transformer
* [voltage-high=*](https://wiki.openstreetmap.org/wiki/Key:voltage-high)
* [voltage-low=*](https://wiki.openstreetmap.org/wiki/Key:voltage-low)

Furthermore, [voltage:primary](https://wiki.openstreetmap.org/wiki/Key:voltage:primary), voltage:secondary, voltage:tertiary are valid with (already in presets) **and should require transformer=*** (not in validator) to be used.
https://wiki.openstreetmap.org/wiki/Tag:power%3Dtransformer#Voltage_tagging

[transformer=*](https://wiki.openstreetmap.org/wiki/Key:transformer) should only be used on nodes (it doesn't raise a warning on area currently), just like power=transformer (which accurately raises a warning when used on areas).

Let me know if any edit is useful in data items, file or elsewhere regarding those improvements.

Best regards.

flacombe:
I propose theses changes following power transformer tagging extension.
It's my first PR here, I hope I didn't break anything and accurately followed contributing rules.

I didn't manage to introduce theses requirements in presets. Can someone help me please?
* [voltage:primary](https://wiki.openstreetmap.org/wiki/Key:voltage:primary), voltage:secondary, voltage:tertiary are valid with (already in presets) **and should require transformer (not in validator)** to be used.
https://wiki.openstreetmap.org/wiki/Tag:power%3Dtransformer#Voltage_tagging
* [transformer=*](https://wiki.openstreetmap.org/wiki/Key:transformer) should only be used on nodes (it doesn't raise a warning on area currently), just like power=transformer (which accurately raises a warning when used on areas).

Closes #415

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1652,6 +1652,18 @@
     "replace": {"amenity": "townhall", "townhall:type": "village"}
   },
   {
+    "old": {"transformer": "auto"},
+    "replace": {"windings:auto": "yes"}
+  },
+  {
+    "old": {"transformer": "minor_distribution"},
+    "replace": {"transformer": "distribution"}
+  },
+  {
+    "old": {"transformer": "traction"},
+    "replace": {"transformer": "main"}
+  },
+  {
     "old": {"tunnel": "1"},
     "replace": {"tunnel": "yes"}
   },
@@ -1742,6 +1754,22 @@
   {
     "old": {"volcano": "extinct"},
     "replace": {"volcano:status": "extinct"}
+  },
+  {
+    "old": {"voltage-high": "*"},
+    "replace": {"voltage:primary":"$1"}
+  },
+  {
+    "old": {"voltage-low": "*"},
+    "replace": {"voltage:secondary":"$1"}
+  },
+  {
+    "old": {"voltage-high": "*", "transformer":"generator"},
+    "replace": {"voltage:secondary":"$1", "transformer":"generator"}
+  },
+  {
+    "old": {"voltage-low": "*", "transformer":"generator"},
+    "replace": {"voltage:primary":"$1", "transformer":"generator"}
   },
   {
     "old": {"wall_type": "noise_barrier"},
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1653,7 +1653,7 @@
   },
   {
     "old": {"transformer": "auto"},
-    "replace": {"windings:auto": "yes"}
+    "replace": {"transformer": "yes", "windings:auto": "yes"}
   },
   {
     "old": {"transformer": "minor_distribution"},
@@ -1756,20 +1756,20 @@
     "replace": {"volcano:status": "extinct"}
   },
   {
-    "old": {"voltage-high": "*"},
-    "replace": {"voltage:primary":"$1"}
-  },
-  {
-    "old": {"voltage-low": "*"},
-    "replace": {"voltage:secondary":"$1"}
-  },
-  {
     "old": {"voltage-high": "*", "transformer":"generator"},
     "replace": {"voltage:secondary":"$1", "transformer":"generator"}
   },
   {
     "old": {"voltage-low": "*", "transformer":"generator"},
     "replace": {"voltage:primary":"$1", "transformer":"generator"}
+  },
+  {
+    "old": {"voltage-high": "*"},
+    "replace": {"voltage:primary":"$1"}
+  },
+  {
+    "old": {"voltage-low": "*"},
+    "replace": {"voltage:secondary":"$1"}
   },
   {
     "old": {"wall_type": "noise_barrier"},
--- a/data/fields/transformer.json
+++ b/data/fields/transformer.json
@@ -5,11 +5,11 @@
     "strings": {
         "options": {
             "main": "Forwards power",
-            "distribution": "Feeds final consumers, outside a substation",
-            "generator": "Steps-up voltage in a power plant",
-            "converter": "Feeds a power converter",
+            "distribution": "Feeds final consumers, installed outside substations",
+            "generator": "Steps-up voltage in power plants",
+            "converter": "Feeds power converters",
             "phase_angle_regulator": "Regulates phase angle",
-            "auxiliary": "Feeds internal systems in a substations",
+            "auxiliary": "Feeds internal systems in substations",
             "yes": "Unknown role"
         }
     },
--- a/data/fields/transformer.json
+++ b/data/fields/transformer.json
@@ -4,14 +4,13 @@
     "label": "Type",
     "strings": {
         "options": {
-            "main": "Main",
-            "distribution": "Distribution",
-            "generator": "Generator",
-            "converter": "Converter",
-            "phase_angle_regulator": "Phase Angle Regulator",
-            "compensator": "Compensation feed",
-            "auxiliary": "Auxiliary",
-            "yes": "Unknown"
+            "main": "Forwards power",
+            "distribution": "Feeds final consumers, outside a substation",
+            "generator": "Steps-up voltage in a power plant",
+            "converter": "Feeds a power converter",
+            "phase_angle_regulator": "Regulates phase angle",
+            "auxiliary": "Feeds internal systems in a substations",
+            "yes": "Unknown role"
         }
     },
     "autoSuggestions": false,
--- a/data/fields/transformer.json
+++ b/data/fields/transformer.json
@@ -4,12 +4,12 @@
     "label": "Type",
     "strings": {
         "options": {
+            "main": "Main",
             "distribution": "Distribution",
             "generator": "Generator",
             "converter": "Converter",
-            "traction": "Traction",
-            "auto": "Autotransformer",
             "phase_angle_regulator": "Phase Angle Regulator",
+            "compensator": "Compensation feed",
             "auxiliary": "Auxiliary",
             "yes": "Unknown"
         }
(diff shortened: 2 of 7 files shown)


data/deprecated.json
```diff
@@ -1743,6 +1755,22 @@
...
+    "replace": {"voltage:secondary":"$1"}
+  },
+  {
+    "old": {"voltage-high": "*", "transformer":"generator"},
+    "replace": {"voltage:secondary":"$1", "transformer":"generator"}
+  },
+  {
+    "old": {"voltage-low": "*", "transformer":"generator"},
+    "replace": {"voltage:primary":"$1", "transformer":"generator"}
+  },
```
tyrasd, line 1773: I _think_ these two should be included before line 1758, because the other deprecation rule would otherwise potentially overrule these ones.

data/deprecated.json
```diff
@@ -1651,6 +1651,18 @@
     "old": {"townhall:village": "yes"},
     "replace": {"amenity": "townhall", "townhall:type": "village"}
   },
+  {
+    "old": {"transformer": "auto"},
+    "replace": {"windings:auto": "yes"}
```
tyrasd, line 1656: to me it looks like as if a `windings:auto=yes` tag alone without any value specified for `transformer=*` doesn't make much sense. I don't know if the previous `transformer=auto` tag was only intended for substation transformers which are now to be tagged as `transformer=main`, or if it could have been any kind of autotransformer (e.g. auxiliary, distribution, etc.). If it is the first option, then the `replace` line should have also the `transformer=main` tag included, but if it is the second option, it does not make sense to add this as a deprecation rule here:

A mapper would need to do research or perform a survey to fix these values. Such a task could be better coordinated through a maproulette project for example.

tyrasd (changes requested):
Hi. Thanks for this update! In general it looks good, but please find some comments inline and below:

> Some voltage keys should also be discouraged. Is it possible to raise a warning regarding the following please ?
> 
>     voltage=* in combination with power=transformer

The question is what should be done with the value of the tag if it is found? It surely should not be removed, because it can still be useful to determine the (proper) values for voltage:primary/secondary… In my opinion this is not a good fit for a validation in iD itself, but rather something for an external QA tool like osmose or maproulette. 

>     [voltage-high=*](https://wiki.openstreetmap.org/wiki/Key:voltage-high)
>     [voltage-low=*](https://wiki.openstreetmap.org/wiki/Key:voltage-low)

These are included in the deprecation rules you provided.

> Furthermore, [voltage:primary](https://wiki.openstreetmap.org/wiki/Key:voltage:primary), voltage:secondary, voltage:tertiary are valid with (already in presets) and should require transformer=* (not in validator) to be used.
> https://wiki.openstreetmap.org/wiki/Tag:power%3Dtransformer#Voltage_tagging

Similar question as above: If iD can't provide is no good "upgrade" approach, and removing the data would add more harm than leaving it as it, it would be better to implement such a check in an external QA tool. Also, iD shows users the `transformer` field to fill in: What more do you expect?

> [transformer=*](https://wiki.openstreetmap.org/wiki/Key:transformer) should only be used on nodes (it doesn't raise a warning on area currently), just like power=transformer (which accurately raises a warning when used on areas).

You mean when a `transformer` tag is present, but no `power=transformer`? I would estimate that that is quite rarely the case (because it would have had to be manually added directly in the tag editor by an "experienced" mapper)…

data/fields/windings/auto.json
```diff
@@ -0,0 +1,12 @@
+{
+    "key": "windings:auto",
+    "type": "check",
+    "label": "Auto-transformer",
```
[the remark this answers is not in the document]
flacombe: Auto-transformer sounds right according to Electropedia
https://electropedia.org/iev/iev.nsf/display?openform&ievref=421-01-11

data/deprecated.json
flacombe: It's indeed necessary to look for the appropriate transformer value (it could be main, auxiliary, converter...)
I can update this deprecation with transformer=yes, is it acceptable?

flacombe:
Hi @tyrasd and thank you for the comments.

Here are additional answers complementary with previous discussion upside.

> The question is what should be done with the value of the tag if it is found? It surely should not be removed, because it can still be useful to determine the (proper) values for voltage:primary/secondary… In my opinion this is not a good fit for a validation in iD itself, but rather something for an external QA tool like osmose or maproulette.

Given problem is we don't know what voltage=* means in combination with transformers. It can be primary, secondary, tertiary or even a list of them. That's why it should not be used.
There are two different problems here
* People can create a new osm feature with transformer in combination with voltage and they should be advised to use voltage:primary, secondary instead
* An existing transformer is found with voltage=* and it's pretty impossible to guess if it's primary, secondary... just like a consumer won't be able to get any value from that.

> Similar question as above: If iD can't provide is no good "upgrade" approach, and removing the data would add more harm than leaving it as it, it would be better to implement such a check in an external QA tool.

I'd say iD can put the greatest voltage:primary, voltage:secondary or voltage:tertiary in voltage=* and drop others, if possible.

> Also, iD shows users the transformer field to fill in: What more do you expect?

It's not a matter of fields, but a matter of combinations.
voltage:primary, voltage:secondary... requires transformer=* to be used. Using them without transformer=* should raise a warning.
highway=street_lamp + voltage:primary=* is invalid and doesn't expect transformer=* to become valid.

> You mean when a transformer tag is present, but no power=transformer? I would estimate that that is quite rarely the case (because it would have had to be manually added directly in the tag editor by an "experienced" mapper)…

That's right, we can keep that apart for this PR.
That would be useful to check what is added directly as well, I'm not only experienced mappers directly add tags.

flacombe:
I'll resolve change requests in a few hours, wait for it.

data/fields/windings/auto.json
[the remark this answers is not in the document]
tyrasd: Both spellings seem to be valid and in use, but the one without hyphen seems to be more common and is used by popular sources like wikipedia (https://en.wikipedia.org/wiki/Autotransformer).

data/deprecated.json
tyrasd: >  I can update this deprecation with transformer=yes, is it acceptable?

Yes, please!

data/fields/transformer.json
```diff
@@ -4,12 +4,12 @@
...
     "strings": {
         "options": {
+            "main": "Main",
             "distribution": "Distribution",
             "generator": "Generator",
             "converter": "Converter",
-            "traction": "Traction",
-            "auto": "Autotransformer",
             "phase_angle_regulator": "Phase Angle Regulator",
+            "compensator": "Compensation feed",
```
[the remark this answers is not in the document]
flacombe: That's indeed too early to document this value, I've removed it

data/fields/transformer.json
```diff
@@ -4,12 +4,12 @@
     "label": "Type",
     "strings": {
         "options": {
+            "main": "Main",
```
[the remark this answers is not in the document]
flacombe: I've updated labels with more verbose ones. Hope it's not too long

data/deprecated.json
flacombe: Good point, I've swapped them

flacombe:
Any news on this PR @tyrasd?

tyrasd (approved):

flacombe:
Thank you @tyrasd for review and final validation.

Be sure I will report the need to have dependency in tagging rules in iD repo.

Best regards

data/fields/transformer.json
```diff
@@ -4,14 +4,13 @@
...
-            "auto": "Autotransformer",
-            "phase_angle_regulator": "Phase Angle Regulator",
-            "auxiliary": "Auxiliary",
-            "yes": "Unknown"
+            "main": "Forwards power",
+            "distribution": "Feeds final consumers, installed outside substations",
+            "generator": "Steps-up voltage in power plants",
+            "converter": "Feeds power converters",
+            "phase_angle_regulator": "Regulates phase angle",
+            "auxiliary": "Feeds internal systems in substations",
```
1ec5, line 12: Were the previous strings incorrect? The new strings can end up being so lengthy in translation that they ideally would be displayed as tooltips, though I’m unsure if schema-builder supports localizable combo option tooltips yet. For now, perhaps each option could start with the short name, like “Distribution: Feeds final consumers, installed outside substations”.

tyrasd: They were probably not incorrect, but I thought that having an almost 1:1 copy of the raw tag value isn't a big help for mappers (otherwise this field could have been a regular combo field), that's why I asked for more descriptive labels.

:+1: for adding the tag values as part of the labels, I've tweaked it in a33d777 – does that look better?

PR #449, opened 2022-05-02
Add flashing_lights field

arch0345:
This PR adds a field for [`flashing_lights`](https://wiki.openstreetmap.org/wiki/Key:flashing_lights) which will be used for Unmarked/Marked Crossing presets. The key [`flashing_lights`](https://taginfo.openstreetmap.org/keys/?key=flashing_lights) currently has around 2.5K uses. I've added values listed on the wiki article (`no`, `yes`, `button`, `always`, `sensor`, and `button;sensor`). Crossings with flashing lights are pretty common in the US, here's an example from Alexandria, VA:
image

### DIFF
--- a/data/fields/flashing_lights.json
+++ b/data/fields/flashing_lights.json
@@ -1,0 +1,29 @@
+{
+    "key": "flashing_lights",
+    "type": "combo",
+    "label": "Flashing Lights",
+    "strings": {
+        "options": {
+            "button": {
+                "title": "Button activated",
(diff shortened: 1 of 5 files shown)


arch0345:
> Good suggestion! I would only request one small change: given that the current usage is very small (less than 0.5% of crossings use the tag), I think it would be better to add this one as an optional field ([`moreFields`](https://github.com/ideditor/schema-builder#fieldsmorefields)). What do you think?

Yeah, this makes sense

PR #452, opened 2022-05-05
Add man_made=carpet_hanger

Zverik:
I've started micromapping in Tallinn and was suprised at how many of these things are here. Literally one in every yard.

The tag is [documented](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dcarpet_hanger) and used ~600 times on the map.

### DIFF
--- a/data/presets/man_made/carpet_hanger
+++ b/data/presets/man_made/carpet_hanger
@@ -1,7 +1,13 @@
 {
     "icon": "maki-picnic-site",
+    "fields": [
+        "material",
+        "height",
+        "colour"
+    ],
     "geometry": [
         "point",
+        "line"
     ],
     "terms": [
         "rack",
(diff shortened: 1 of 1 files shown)


data/presets/man_made/carpet_hanger
```diff
@@ -0,0 +1,21 @@
+{
+    "icon": "maki-picnic-site",
+    "fields": [
+        "material",
+        "height",
+        "colour"
+    ],
+    "geometry": [
+        "point",
+        "line"
```
tyrasd, line 10: According to the wiki it's only [intended](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dcarpet_hanger) to be used on nodes. Currently, there are only 14 mapped as ways (compared to 574 as nodes). But I see the rationale and similarity to how [`amenity=bench`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbench) was later [extended](https://wiki.openstreetmap.org/w/index.php?title=Tag:amenity%3Dbench&diff=prev&oldid=705902) to linear features as well.

My request would be to update this on the wiki fist.

Zverik: Thanks for the tip, I've just updated the wiki.

PR #453, opened 2022-05-09
Additional fields for `crossing=traffic_signals` presets

arch0345:
Adds fields for:
* [`button_operated`](https://wiki.openstreetmap.org/wiki/Key:button_operated)
* [`traffic_signals:sound`](https://wiki.openstreetmap.org/wiki/Key:traffic_signals:sound)
* [`traffic_signals:vibration`](https://wiki.openstreetmap.org/wiki/Key:traffic_signals:vibration)
* [`traffic_signals:arrow`](https://wiki.openstreetmap.org/wiki/Key:traffic_signals:arrow)
* [`traffic_signals:countdown`](https://wiki.openstreetmap.org/wiki/Key:traffic_signals:countdown)
* [`traffic_signals:minimap`](https://wiki.openstreetmap.org/wiki/Key:traffic_signals:minimap)

based on keys listed on [this wiki article](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtraffic_signals#Traffic_signals_for_pedestrians). I added the latter 3 to `moreFields` since they're not as commonly used as the former.

### DIFF
--- a/data/fields/button_operated.json
+++ b/data/fields/button_operated.json
@@ -1,0 +1,5 @@
+{
+    "key": "button_operated",
+    "type": "check",
+    "label": "Button Operated"
+}
--- a/data/fields/traffic_signals/countdown.json
+++ b/data/fields/traffic_signals/countdown.json
@@ -1,0 +1,5 @@
+{
+    "key": "traffic_signals:countdown",
+    "type": "check",
+    "label": "Countdown"
+}
--- a/data/fields/traffic_signals/sound.json
+++ b/data/fields/traffic_signals/sound.json
@@ -1,0 +1,21 @@
+{
+    "key": "traffic_signals:sound",
+    "type": "combo",
+    "label": "Sound Signals",
+    "strings": {
+        "options": {
+            "yes": "Yes",
+            "no": "No",
+            "locate": {
+                "title": "Locate",
+                "description": "There is only a signal to find the pole"
+            },
+            "walk": {
+                "title": "Locate",
+                "description": "here is only the signal when walking is allowed"
+            }
+        }
+    },
+    "autoSuggestions": false,
+    "customValues": false
+}
--- a/data/fields/traffic_signals/arrow.json
+++ b/data/fields/traffic_signals/arrow.json
@@ -1,0 +1,5 @@
+{
+    "key": "traffic_signals:arrow",
+    "type": "check",
+    "label": "Tactile Arrow"
+}
--- a/data/fields/traffic_signals/minimap.json
+++ b/data/fields/traffic_signals/minimap.json
@@ -1,0 +1,5 @@
+{
+    "key": "traffic_signals:minimap",
+    "type": "check",
+    "label": "Tactile Map"
+}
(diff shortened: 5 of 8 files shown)


tyrasd:
Thanks!

data/fields/traffic_signals/sound.json
```diff
@@ -0,0 +1,21 @@
+{
+    "key": "traffic_signals:sound",
+    "type": "combo",
+    "label": "Sound Signals",
+    "strings": {
+        "options": {
+            "yes": "Yes",
+            "no": "No",
+            "locate": {
+                "title": "Locate",
```
1ec5, line 10: This matches the raw value, but it feels a bit ungrammatical in this context. [This technology has multiple names in English](http://www.apsguide.org/chapter4_pushbutton.cfm), but “Call Button Locator” would be more grammatical and self-documenting.

data/fields/traffic_signals/sound.json
```diff
@@ -0,0 +1,21 @@
...
+                "description": "There is only a signal to find the pole"
```
1ec5, line 11: This would be more grammatical and descriptive:

> There is only a signal for finding the pole with the call button

data/fields/traffic_signals/sound.json
```diff
@@ -0,0 +1,21 @@
...
+            },
+            "walk": {
+                "title": "Locate",
```
1ec5, line 14: `locate` and `walk` both have the same `title` of “Locate”. I think this would be “Walk”, but “Walk Indication” would be consistent with “Call Button Locator” above and [American English terminology](http://www.apsguide.org/chapter4_walkindication.cfm).

data/fields/traffic_signals/sound.json
```diff
@@ -0,0 +1,21 @@
...
+                "description": "here is only the signal when walking is allowed"
```
1ec5, line 15: Typo:

> There is a signal only when walking is allowed

1ec5 (commented):
This is mostly belated copyediting for clarity, but one of the strings has it backwards.

data/fields/button_operated.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "button_operated",
+    "type": "check",
+    "label": "Button Operated"
```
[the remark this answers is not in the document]
arch0345: I decided to just go with `Call Button`

data/fields/traffic_signals/countdown.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "traffic_signals:countdown",
+    "type": "check",
+    "label": "Countdown"
```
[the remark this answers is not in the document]
arch0345: >This key can also be used with highway=traffic_signals. Should we add it to the Traffic Signals preset?

I've added it to `moreFields` in #456

arch0345:
Thanks for your feedback @1ec5, I've opened up #456 to resolve these issues

PR #454, opened 2022-05-12
Add `sidewalk` field

tyrasd:
Reopened #444:

I disagree with the reasons the PR was closed: The fact that there is no consensus on a specific tagging variant, doesn't give the editors a "free pass" to choose a single one of them and ignore all others. On the contrary: this would only be the case when there is consensus that a specific tagging variant is "approved".

In cases like this one, in my opinion, map editors should at least show data that has been added by mappers using either of the tagging variants. This would allow people to find out whether sidewalk data already had been mapped on a road when adding sidewalks as separate ways, thus reducing situations where sidewalks en up being mapped in a duplicate way.

//cc @arch0345 @bhousel @1ec5

### DIFF
--- a/data/fields/sidewalk.json
+++ b/data/fields/sidewalk.json
@@ -1,0 +1,35 @@
+{
+    "keys": [
+        "sidewalk"
+    ],
+    "reference": {
+        "key": "sidewalk"
+    },
+    "type": "sidewalk",
+    "label": "Sidewalk",
+    "placeholder": "none",
+    "strings": {
+        "options": {
+            "no": {
+                "title": "None",
+                "description": "No sidewalk"
+            },
+            "both": {
+                "title": "Both",
+                "description": "A sidewalk is on both sides of the road"
+            },
+            "left": {
+                "title": "Left side only",
+                "description": "A sidewalk is on the left side of the road"
+            },
+            "right": {
+                "title": "Right side only",
+                "description": "A sidewalk is on the right side of the road"
+            },
+            "separate": {
+                "title": "Mapped separately",
+                "description": "A sidewalk is mapped separately"
+            }
+        }
+    }
+}


tordans:
@tyrasd doesn't this PR miss the part where the key is added to the preset? – I might still be missunderstanding how the whole system works, though.

And for that part, I wonder if the `moreFields` would be a good start for this key. That would expose existing data but also not push for the tag actively. 

---

Another thing is the tagging schema that has the `:both` as part of the key, mainly relevant for `sidewalk:both=separate` ([Example](https://www.openstreetmap.org/way/155595024#map=19/52.47352/13.44777)).

The reasoning is, that we have to be able to tag "sidewalk:right=yes + sidewalk:left=separate" in case only one side is mapped separately. Or, maybe more realistically "sidewalk:right=no + sidewalk:left=separate".

Unfortunately, the whole `key:side=value` topic is not specified well (either). For Berlin, we tend to add the `*:both` explicitly.

data/fields/sidewalk.json
```diff
@@ -0,0 +1,35 @@
+{
+    "keys": [
+        "sidewalk"
+    ],
+    "reference": {
+        "key": "sidewalk"
+    },
+    "type": "sidewalk",
```
1ec5, line 8: This isn’t a valid type – unless the intention was to introduce a new type to schema-builder? Perhaps the `cycleway` type used by cycleway.json was replaced by `sidewalk` without realizing the purpose of `type`. But cycleway.json implements `cycleway:left/right=*`, only setting `cycleway=*` if the left and right values agree. It would be possible to piggyback on the `cycleway` type, but the only `options` would be `no`, `yes`, and `separate`, and `sidewalk:left` and `sidewalk:right` would need to be defined as `types`.

1ec5 (commented):
This PR as written seems to require a change to schema-builder that hasn’t shown up in a PR yet.

The merits of relegating pedestrian infrastructure to an auxiliary detail of motorized transportation can be discussed at further length in [other venues](https://community.openstreetmap.org/t/is-there-consensus-on-mapping-pavements-sidewalks-separately-to-roads/1067). However, if the justification for this PR is transparency to the user about existing data – like an unsearchable preset but for a field – then we’d have to account for both the `sidewalk=left/right/both` and `sidewalk:left/right=yes` schemes. Not an easy thing to fit into a single field.

tyrasd:
> doesn't this PR miss the part where the key is added to the preset?

> [`"type": "sidewalk"`] isn’t a valid type

Correct. I've just used the PR from @arch0345 as a starting ground. It's definitely not ready as is. I did convert the PR to a draft now.

> And for that part, I wonder if the `moreFields` would be a good start for this key.

:+1: that would have been my initial idea.

>  then we’d have to account for both the `sidewalk=left/right/both` and `sidewalk:left/right=yes` schemes

The argument to be made here is that while `sidewalk=left/right/both` (~400,000 km when using a factor of 2 for the lengths of `sidewalk=both`) and `footway=sidewalk` (~400,000 km) are found in OSM data in astoundingly equal amounts, the current global usage of `sidewalk:left/right/both` (~10,000 km) is about two orders of magnitude less than that.

Another argument for including the `sidewalk` tag as a field is that this is (AFAIK) the only way to express that there does _not_ exist a sidewalk alongside a road or street.

> `sidewalk:both=separate`

I hadn't thought about that particular use case. At first glance, I would have argued that `sidewalk:both=separate` is just a synonym of `sidewalk=separate` (which could be hardcoded in iD if we decide it's important to show). But this still leaves the question of `sidewalk:left/right=separate` which are even more commonly used in practice than `sidewalk:both=separate` (but still many magnitudes less often than `sidewalk=left/right/both`). :thinking: At the moment, I don't yet see an easy solution for this particular tagging detail, to be honest.

1ec5:
Of the tags discussed here so far, `sidewalk=separate` and `sidewalk:left/right=separate` are the ones that matter most for pedestrian routing. I can’t think of any mainstream pedestrian routing profile that makes a major distinction between two streets that allow foot traffic (based on their access tags) where one has a sidewalk tag while the other does not. On the other hand, `separate` matters for ensuring that routers don’t needlessly switch between sidewalks and streets, resulting in unusable guidance instructions. Coverage of `separate` is currently far below where it needs to be to account for separately drawn sidewalk ways.

data/fields/sidewalk.json
arch0345: >This isn’t a valid type – unless the intention was to introduce a new type to schema-builder?

Yeah, my intention was to create a new `type` that's similar to the `cycleway` type [as shown in this pdf](https://github.com/openstreetmap/id-tagging-schema/files/8692005/Sidewalk.field.concept.pdf), but I'm wasn't sure how (or if) this could be implemented.

bgo-eiu:
I'm a bit stubborn about sidewalk=separate not being usable for a few reasons

*  Sidewalks as a property of the road are useful for public transport routing, since sidewalks are never included in transport route relations. It is common for bus stops not to have sidewalks, so being able to discern if a sidewalk is present, and if so, which sides it is on, makes it a lot easier to derive information about where to board and alight with respect to the way without having to download the footway network in conjunction with the route relations.

* Sidewalk=separate is impossible to survey without the corresponding OSM data in front of you. If I am writing down notes from a survey, or using a mobile editor with aerial imagery, I have no way to tell whether something has been mapped separately. It also would confuse new mappers.

* It's likely that introducing it as an option will result in people removing valid information about sidewalks. Querying for the tag in most places results in a number of uses on ways where the sidewalk is not mapped separately. A lot of these aren't even mistakes, one of the most common uses of it is for tagging sidewalks separated from the road by segregated cyclepaths.

image

(Here only the cyclepaths have been filled in.)

Other interpretations include: parking aisles for which the sidewalk is around the parking lot itself rather than the aisle, and bridges with separate parallel sidewalks.

Based on this, it seems like every use of sidewalk=separate in India is on a way for which the sidewalk is not mapped separately: https://overpass-turbo.eu/s/1iPx

Much of these sidewalk=separate tags on ways which don't have sidewalks drawn separately can be attributed to StreetComplete. As StreetComplete is intended to be used for surveying, the sidewalk quest option of "is this sidewalk separate?" doesn't get read as the meaning of it being a separate way for most people. Examples:
* https://www.openstreetmap.org/way/232496610/history
* https://www.openstreetmap.org/way/962980312/history
* https://www.openstreetmap.org/way/226981388/history
* https://www.openstreetmap.org/way/435272758/history
* https://www.openstreetmap.org/way/802404456/history

I agree that tags in use should be reflected in the editor, and that there should be a way to do this without picking a "side" but adding a sidewalk=separate option would result in more of this, which goes against the objectives of even the people advocating for sidewalk=separate.

1ec5:
I think there’s a misunderstanding. `sidewalk=separate` is merely an (unenforced) aspect of the migration path from `sidewalk=both` etc. on the roadway to `footway=sidewalk` on a separate way. Without it, there may or may not be adverse effects, depending on the vagaries of any given routing profile. Mappers aren’t very good at eyeballing edge weights, so `sidewalk=separate` just makes it clear that the intention is to follow a parallel path if present.

If this tag is present but the parallel path is absent, the router would take it to mean that any sidewalks have been drawn separately in this area, and there is none to speak of here. The router might still route along the roadway, just as it would if there were no sidewalk tags to speak of.

> Sidewalk=separate is impossible to survey without the corresponding OSM data in front of you. If I am writing down notes from a survey, or using a mobile editor with aerial imagery, I have no way to tell whether something has been mapped separately. It also would confuse new mappers.

I find it no more difficult to spot a parallel `footway=sidewalk` way in Go Map!! than in iD. Doesn’t StreetComplete show any parallel `footway=sidewalk` way on the map while asking this question?

> As StreetComplete is intended to be used for surveying, the sidewalk quest option of "is this sidewalk separate?" doesn't get read as the meaning of it being a separate way for most people.

As of streetcomplete/StreetComplete#4030, the option is called “sidewalk, but displayed separately on map”. If you’re concerned that this wording causes people to use `sidewalk=separate` where no `footway=sidewalk` way has been mapped yet, then that’s good feedback to report in StreetComplete’s issue tracker, a feature request for iD’s validator, or grist for a MapRoulette cleanup challenge. But I don’t see how that would have any bearing on a tag that [routers already](https://github.com/graphhopper/graphhopper/blob/ee0703d53ad3fc26413c15f80f13168a58c000de/core/src/main/java/com/graphhopper/routing/util/FootTagParser.java#L88) [find useful](https://github.com/valhalla/valhalla/blob/72e986c2ac6fa483f24491a87a6a7362befa9b72/lua/graph.lua#L126) or a field intended for an editor that would show the separately mapped way.

bgo-eiu:
Even if a router doesn't change its behavior because of sidewalk=separate having no separate way, it makes the rest of the sidewalk tags less effective, particularly sidewalk=no, if sidewalk=no and sidewalk=both alike are getting replaced with sidewalk=separate. The result would just be the sidewalk tag including no information. It also makes queries of route relations like I mentioned harder to do, because you don't necessarily need information about the adjacent footways in order to do that kind of analysis.

In the locations in India linked, there's no sidewalks mapped as ways yet so there's nothing to use as a reference point, or there are other types of footways that are rendered similarly, so it is easy to see how someone wouldn't be able to tell what the quest means. I don't think changing the wording or how its drawn would make it any clearer. . Even iD, if there aren't already footways mapped it would be very easy not to know that they're absent. Maybe this is unusual but when my phone battery runs out I write down this sort of thing on sticky notes, anything mappable should be describable without an app. When I use the quest in street complete, sometimes I'll do things like map the ways separately first and then answer yes/no in the quest to verify on site. (More necessary in areas with tall buildings really, since it's easy to miss where a sidewalk disappears to give way to a dumpster next to apartments if you're mapping in iD. Looking closely at how things are drawn on StreetComplete defeats the purpose of surveying anyway, because it in that common situation, it is too easy to see a separate path a block away and think that means you shouldn't check - if it's next to an apartment building or office tower, you probably should.)

If I am understanding these links correctly, if I were to add sidewalk=separate or sidewalk:left=separate where applicable on Traction Street, it would route along the north side of Traction Street's sidewalk instead of along the road. In real life though, everybody would walk in Traction Street on the roadway itself, and if they did get on the sidewalk, a good number would do it at the point shown. On a narrow street with discontinuous sidewalks like that, there often are too many obstructions to walk anywhere but the road, and you're not any less likely to get hit by a car walking to the side anyway. Adding sidewalk=separate would make the router produce a worse result rather than a better one, and the spatial inference it's making in the absence of that tag is actually a really good one in this case. If the one reason to use sidewalk=separate is to prevent the router from going in the roadway when sidewalks are present, that's not even an assumption that should be made anyway.

image

If I wanted to know what percentage of the city's alleyways have sidewalks, or if there's a correlation between an alley having a name and whether or not an alley has sidewalks, or even analyze if the difference between an alley and a residential street is easily discernable based on sidewalk attributes, sidewalk=no and sidewalk=left/right/both have clear utility for answering those questions. People replacing that where it is present solely based on whether or not they see it in the editor - not even necessarily the imagery - would be making those questions harder to answer, especially for information that can only be verified on foot. I'm fairly sure I've extended a sidewalk way too long before realizing the problem on foot later at least once before.

Making the router follow the sidewalk doesn't seem like the one good reason to remove information from the map that's usable for answering questions like that is what I'm saying. What would be even messier to deal with is if people started tagging sidewalk=no on roadways that have sidewalks which shouldn't be preferred over the roadway - that means trying to analyze what sidewalk quality or use looks like on the whole impossible altogether because then the information about the "bad" sidewalks is gone from the data. If there's some specific areas where the sidewalk=separate makes the router work particularly well, maybe there could be some kind of overlay layer for those areas so that it's clear what it's for and where it works.

matkoniecz:
>  On the contrary: this would only be the case when there is consensus that a specific tagging variant is "approved".

Even then - there could be a case where one variant is approved but competing is used at least as much or even more dominating

matkoniecz:
> (Here only the cyclepaths have been filled in.)

untrue - see https://www.openstreetmap.org/way/329492554#map=19/51.38560/-0.12814 (has `foot=yes` `segregated=yes` - so it marks both cyclepath along road and footway along road)

> I'm a bit stubborn about sidewalk=separate not being usable for a few reasons

That is not a good place to try deprecating massively used schema. Many people use it and it is useable.

> Even if a router doesn't change its behavior because of sidewalk=separate having no separate way, it makes the rest of the sidewalk tags less effective, particularly sidewalk=no, if sidewalk=no and sidewalk=both alike are getting replaced with sidewalk=separate

Correctly mapped sidewalk=no NEVER cannot be replaced with correctly mapped sidewalk=separate

> What would be even messier to deal with is if people started tagging sidewalk=no on roadways that have sidewalks which shouldn't be preferred over the roadway

that is as invalid as mapping sidewalk=both where sidewalk is not existing

> Sidewalk=separate is impossible to survey without the corresponding OSM data in front of you. If I am writing down notes from a survey, or using a mobile editor with aerial imagery, I have no way to tell whether something has been mapped separately

Mapping in OSM always requires being aware what is mapped - you cannot making edit adding POI/roads etc without checking what is mapped.

Also, typical mobile editor allows to display existing data.

---------------------------------------

Overall: there could be a valid reason to ignore widely used tag, but in this case it seems personal dislike based on bunch of misunderstandings.

bgo-eiu:
The vast majority of sidewalk=separate uses in India are sidewalk=no. If entire countries don't have a single use that matches the description, that is not a widely understood tag. Here is the problem with the situation in India I am pointing out:

* If you look in most of the places in India that have sidewalk=separate where the imagery is clear enough, you can see that it is really sidewalk=no.
* In an urban area which is too built up to discern sidewalks clearly from imagery, you have sidewalk=separate down Laxmi Road (middle) here - you have two options as an editor, assume that this means sidewalk=no based on the fact that this is what it mostly means in India, or you assume that it means sidewalk=yes based on the idea that sidewalk=separate means mapped as a separate way.

image

* If you are in iD, and you see sidewalk=separate with no separate way, you might draw the separate way without realizing that this usually means sidewalk=no in India.
* Alternatively, you might switch it to sidewalk=both or sidewalk=no depending on your interpretation since you have seen there is no separate way.

Meanwhile, the imagery still does not tell us if there is actually a sidewalk here, and unless someone surveys this independently of the OSM data - meaning surveying first, checking how it's mapped later - this question of whether or not there is a sidewalk here looks "answered." Nowhere in this situation is the mapper ever able to correct this issue in iD if they think sidewalk=separate is widely understood to be sidewalk=both. This is what I mean by removing information, it makes sidewalk=no getting mapped as sidewalk=both a possibility if we ignore the fact that this is mainly used to mean sidewalk=no in many countries.

The street level imagery at that location in Croydon shows that the cyclepath and sidewalk are segregated from eachother. Mapping them together as one way when they have different geometries is not mapping the sidewalk separately.

I am not suggesting deprecating anything, I am more just pointing out that the existing use of sidewalk=separate does not even show what people are saying it does.

Surveying requires you to not look at the map in some situations. For example, one of the areas I've surveyed that I only have handwritten notes on (not mapped yet) is a stretch of road where the sidewalk and driveways were reconstructed over the course of months. On the first day it was open to walk on again, I used my phone to take photos (not possible to have OSM open at the same time) and then used the handwritten notes to get down details that can't be discerned from imagery. (For example, temporal details if you don't have images of what it used to look like but know from memory. That way you can add start_date=* to tactile paving pads that you know are there but aren't going to be in aerial imagery for another few years.)

eginhard:
> The vast majority of sidewalk=separate uses in India are sidewalk=no. If entire countries don't have a single use that matches the description, that is not a widely understood tag. 

There are only [63 instances](https://taginfo.openstreetmap.in/tags/sidewalk=separate) of `sidewalk=separate` in India, a tiny portion of the global total and of all sidewalk tags in India, so it doesn't seem like a big issue.

bgo-eiu:
Right, I am just concerned about what the tag would look like if it were presented as an option while not being widely understood the same way. As far as I can tell, there are only a few countries where the number of uses of the tag is more than 2,000. The concern about it being self-referential is still there; putting data that describes OSM data in the map seems like it could cause these kinds of issues. Anyway, I've said my piece on that, if people don't agree it would at least be interesting to see what happens with it

1ec5:
> If I am understanding these links correctly, if I were to add sidewalk=separate or sidewalk:left=separate where applicable on Traction Street, it would route along the north side of Traction Street's sidewalk instead of along the road. In real life though, everybody would walk in Traction Street on the roadway itself, and if they did get on the sidewalk, a good number would do it at the point shown. On a narrow street with discontinuous sidewalks like that, there often are too many obstructions to walk anywhere but the road, and you're not any less likely to get hit by a car walking to the side anyway. Adding sidewalk=separate would make the router produce a worse result rather than a better one, and the spatial inference it's making in the absence of that tag is actually a really good one in this case. If the one reason to use sidewalk=separate is to prevent the router from going in the roadway when sidewalks are present, that's not even an assumption that should be made anyway.

You’re assuming that `sidewalk=separate` would be determinative, preventing anyone from using the roadway even if there’s no other way to get to the destination – this is not the case. When the OSRM developers discovered that sidewalks were being mapped as separate ways, they were understandably alarmed because suddenly two dense routing networks were being connected very strongly to one another, producing alternations between the two: Project-OSRM/osrm-backend#4790.

Below are some examples of pedestrian routers taking unlawful and unsafe shortcuts in the absence of `sidewalk=separate`. You can play around with the endpoints in this neighborhood to find countless more misleading routes. (Local mappers do want to add `sidewalk=separate` soon, so these examples may not last long.)

* [Hillview Park across Ocala Avenue](https://www.openstreetmap.org/directions?engine=graphhopper_foot&route=37.33883%2C-121.82533%3B37.33987%2C-121.82138#map=18/37.33887/-121.82344) – crosses five lanes without a crosswalk, speed limit 35&nbsp;mph (average speeds much higher), even though a slight detour to the east or west would find a suitable crosswalk (reproduces in GraphHopper, OSRM, and Valhalla)
* [Orlando Drive to Bermuda Way](https://www.openstreetmap.org/directions?engine=graphhopper_foot&route=37.33227%2C-121.83469%3B37.33112%2C-121.83207) – a classic case of finagling through a signalized intersection instead of going around it on marked crosswalks (reproduces in GraphHopper)
* [Southbound Almaden Expressway](https://www.openstreetmap.org/directions?engine=graphhopper_foot&route=37.27279%2C-121.87883%3B37.27711%2C-121.88038#map=17/37.27495/-121.87965) – ignores perfectly usable but circuitous sidewalks in favor of walking along an expressway with a 50&nbsp;mph speed limit

Sometimes the route does need to traverse the roadway due to obstructions or [discontinuities](https://www.openstreetmap.org/directions?engine=graphhopper_foot&route=39.02048%2C-84.22399%3B39.02049%2C-84.22205#map=19/39.02068/-84.22302) – that’s fine. A router should be able to cope with those situations by falling back to the roadway as a last resort. But in order to treat it as a last resort, it has to recognize that the sidewalk is associated with the roadway. Algorithmically associating sidewalks with roadways is hard: https://github.com/OpenSidewalks/OpenSidewalks-Schema/issues/2#issuecomment-1074404560.

I’ll make the same point I did in [defending pedantic sidewalk ways](https://community.openstreetmap.org/t/is-there-consensus-on-mapping-pavements-sidewalks-separately-to-roads/1067/12?u=minh_nguyen) and in [arguing against pedantic interpretations of access restrictions](https://lists.openstreetmap.org/pipermail/tagging/2022-April/064438.html): the basic job of a router is to recommend the “best” route, however “best” is defined. If you want to develop a pedestrian routing profile that recommends taking shortcuts through streets without the benefit of local knowledge and produces very confusing guidance instructions, that’s your prerogative, but don’t force that decision on more vulnerable pedestrians.

The introduction of this field would mark the first time iD presents two different ways to indicate the presence of a sidewalk, exposed in two very different places. Without a `separate` option, one mapper will likely learn a different mapping style than another, leading to more redundancy than today, but routers will miss an opportunity to cleanly distinguish the two mapping styles.

hungerburg:
> The merits of relegating pedestrian infrastructure to an auxiliary detail of motorized transportation can be discussed at further length

This is not the way, such is perceived in my area: Here a street is open for all public traffic, any means of transportation. Maybe in some parts of the world, streets are for cars only and pedestrians merely regarded obstacles. Sidewalks here are but lanes. If they exist, pedestrian have to use them, except in rare occasions, where they are to walk on the carriageway, even though sidewalks exist, e.g. when carrying bulky goods, or in winter, when a certain sidewalk is not serviced. Very rare :)

tordans:
@tyrasd I ran into this twice the last few days. Once with GoMap and just now in iD. The GoMap case would ideally be mapped as separate sidewalk but there was a bridge and crossings involved which I did not had the time to map on the go, so I wanted to add the sidewalk key on the centerline at least. And just now in iD on a highway=service where not even Mapillary images are available (which is pretty uncommon in Berlin…), so separate geometries are not feasible for me for this case as well.

Meaning, we need this preset – the same way we have a bicycle preset.

I write this to ping this again so this may get pushed over the finish line.

1ec5:
> The GoMap case would ideally be mapped as separate sidewalk but there was a bridge and crossings involved which I did not had the time to map on the go, so I wanted to add the sidewalk key on the centerline at least.

iD’s validator makes it easy to add crossings and bridges (albeit with placeholder lengths). If Go Map!! lacks these conveniences, that would be another enhancement to pursue in parallel.

> Meaning, we need this preset – the same way we have a bicycle preset.

I assume you’re referring to the `cycleway` field. Unlike the `sidewalk` field in this PR, the `cycleway` field has a `separate` option:

https://github.com/openstreetmap/id-tagging-schema/blob/993b6b093571b5cfd3b647e6153e115a01d9bc26/data/fields/cycleway.json#L47-L50

Even so, there’s a lot of inconsistency out there: `cycleway=separate` without a corresponding `highway=cycleway`, `highway=cycleway` alongside `cycleway=lane`, etc., and the slow-motion edit wars to go with it.

> The introduction of this field would mark the first time iD presents two different ways to indicate the presence of a sidewalk, exposed in two very different places. One mapper will likely learn a different mapping style than another. 

I think part of the solution would be more robust validation. Editors need to warn when the in-street and off-street mapping contradict each other. Unfortunately, it isn’t straightforward to programmatically associate a separate cycleway or footway with its street, and tagging proposals to that effect have stalled.

tordans:
Ideally, we would extend the [`directionalCombo`](https://github.com/ideditor/schema-builder?tab=readme-ov-file#combodropdown-fields) in a similar way to the [just merged ](https://github.com/openstreetmap/iD/pull/9587)"both or not both" pattern to also handle `sidewark=SIDE` pattern.

@k-yle can you share an opinion on how that might be added to the `directionalCombo` code in iD?

My understanding is, that we would ideally support those three patterns equally while using only one as the version that iD uses when nothing is there. Another question is if iD should update the tagging pattern when uses change some tagging or stickt with what is there. I assume it would update to the "recommended" version.

The variations are:
- `sidewalk=left`, `sidewalk=right`, `sidewalk=both`
- `sidewalk:left=yes|no|separate`, `sidewalk:right=yes|no|separate`, `sidewalk:both=yes|no|separate`
- `sidewalk:left=yes|no|separate`, `sidewalk:right=yes|no|separate`, `sidewalk=yes|no|separate` ([also used a lot](https://taginfo.openstreetmap.org/keys/sidewalk#values))

tordans:
Apart from this technical requirement, **we will need to pick our recommended tagging schema**. 
I am not sure how to best tackle this. Looking at historic usage and number, it would be the `sidewalk=SIDE` schema. But that does not have a solution for the `separate` case. Also, the `sidewalk:SIDE` schema is the more modern one that I consider a lot more future prone. [I recently tried to get clarity](https://community.openstreetmap.org/t/rfc-feature-proposal-make-cycleway-both-the-default-to-indicate-both-sides/116585) on the recommendation to always use `:both` for cycleways which turned out to be a lot harder than I hoped. The cleanest approach would be to follow the proposal process. Is anyone up to doing that? Or any other suggestions on how this repo should come to a consensus that will not result in a flood of "bug reports"?

tordans:
We talke about next steps during [the Community Chat](https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2025-03-05), yesterday.

* There was agreement we should add a field to make the existing data visible
* There was agreement we need to handle all three variations of tagging but can only tag one of the modern once
* Which one to use is still to be decided (`sidewalk=*` or `sidewalk:both=*`); we agreed to first look at some number — see below

---

**Which one to use, `sidewalk=*` or `sidewalk:both=*`:**

| key | numbers | link |
|---|----|---|
| `sidewalk:both=*` | 332.969 | [Taginfo](https://taginfo.openstreetmap.org/keys/sidewalk%3Aboth)|
| `sidewalk=separate\|yes\|no\|none` | 1.940.681 | [Spreadsheet](https://docs.google.com/spreadsheets/d/1utbQRkEDA2gXyjOPwXC9tOW29t_tnzI9BAtyDaaeXNI/edit?gid=1060052597#gid=1060052597) |
| `sidewalk=left\|right\|both` | 2.248.360 | same Spreadsheet |

Looking at those numbers, `sidewalk=separate|yes|none` is the clear winner over `sidewalk:both=separate|yes|none`

The arguments that `sidewalk:both=*` is more explicit in its meaning that "both sides" are explicitly tagged. During the `cycleway` vs. `cycleway:both` discussion in the forum, some argued that the "both" is implicit and given. However we had clear cases like oneway streets where this clarity was clearly lacking.

However, I don't see those edge cases being that relevant for sidewalk tags and since iD is able to handle both now, we can just pick the one with most usage.


---

My recommendation is:
- Whenever data is edited in iD, we migrate to
    `sidewalk=yes|no|separate`, `sidewalk:left=yes|no|separate`, `sidewalk:right=yes|no|separate`
- We display all three variations
    `sidewalk=both`, `sidewalk=left`, `sidewalk=right`
    `sidewalk:both=yes|no|separate`, `sidewalk:left=yes|no|separate`, `sidewalk:right=yes|no|separate`
    `sidewalk=yes|no|separate`, `sidewalk:left=yes|no|separate`, `sidewalk:right=yes|no|separate`
- We need to add addition logic to iD to handle
    `sidewalk=both`, `sidewalk=left`, `sidewalk=right`
    I created https://github.com/openstreetmap/iD/issues/10839 for that

matkoniecz:
> Looking at those numbers, sidewalk=separate|yes|none is the clear winner over sidewalk:both=separate|yes|none

the problem here is that `sidewalk=separate` is widely used also when ONE side has sidewalk

so here (for `separate`) you compare "at least one side has sidewalk" with "both sides have sidewalk"

though `sidewalk=no` has no such ambiguity

> Whenever data is edited in iD, we migrate to

any tag at all or sidewalk tags specifically?

> Looking at those numbers

I wonder how it would look if one would discount tagging not consciously chosen by mappers (skip ones tagged in SC and iD, probably) - but this would likely bring `both:` lower

tordans:
> > Looking at those numbers, sidewalk=separate|yes|none is the clear winner over sidewalk:both=separate|yes|none
> 
> the problem here is that `sidewalk=separate` is widely used also when ONE side has sidewalk
> so here (for `separate`) you compare "at least one side has sidewalk" with "both sides have sidewalk"

That was the (and my) argument for cycleway as well but I did not find any proof for that. Given the huge difference (1.607.712 more for the "nil" case), I don't see how we could spin this to use the `both` variant without additional discussion. I would prefer to get this field merged and going. We can always change the default later once someone managed to get a different agreement.

> > Whenever data is edited in iD, we migrate to
> 
> any tag at all or sidewalk tags specifically?

I don't understand the question. When a user changes things in the directional combo, we always tag the one, "chosen" variant. Mixing variants is not supported by the directionalCombo nor do I think it is a good idea.

> > Looking at those numbers
> 
> I wonder how it would look if one would discount tagging not consciously chosen by mappers (skip ones tagged in SC and iD, probably) - but this would likely bring `both:` lower

Filtering the data based on other changeset tags like editor is super complex and I don't thing it will allow us to make a different decision, given the huge difference.

matkoniecz:
> That was the (and my) argument for cycleway as well but I did not find any proof for that.

for `sidewalk=separate` I can easily find plenty of those (at least in Poland)

in fact, for roads with sidewalk on one side `sidewalk=separate` may be more popular tagging than `sidewalk:left=separate`/`sidewalk:right=separate`



> > >  Whenever data is edited in iD, we migrate to

> >  any tag at all or sidewalk tags specifically?

> I don't understand the question.

lets say that user changes surface - is iD going to change `sidewalk:both=separate` to `sidewalk=separate` ?

tordans:
@Lumikeiju The way I understand the position on those matters for this repo – documented in [Guidelines](https://github.com/openstreetmap/id-tagging-schema/blob/main/GUIDELINES.md) – the repo is not the place to make this decision. Given such a big difference in numbers, we need some indication of consensus. Lacking something like clear documentation or a accepted proposal we need to look at usage numbers and follow those.

I am very happy if you pick this up and get us something different we can rely on. I suggest looking into https://wiki.openstreetmap.org/wiki/Proposal:Make_cycleway:both_the_default_to_indicate_both_sides and the linked Forum discussion. IMO that is what is needed to get a clear basis for a different decision here. I did not bring this to a vote (yet), so maybe it will go through but the discussion showed that there are many?/loud? different opinions.

Disclaimer: There might also be something in the wiki that we can rely on; I did not re-check that.

The other question we have to ask ourselves: Is the "wrong" key really so bad that we want this preset to be postponed even longer … or can we live with what I propose to get something going finally. And then have the discussion to change the default separately. — Personally I am in favor of this approach.

tordans:
> > > > Whenever data is edited in iD, we migrate to
> > > any tag at all or sidewalk tags specifically?
> > I don't understand the question.
> lets say that user changes surface - is iD going to change `sidewalk:both=separate` to `sidewalk=separate` ?

Each field is a separate entity. No one field can change data from another field. The directionalCombo handles the three mentioned tags but only those. Please test the cycleway field in iD to see how it currently behaves … and should behave for the sidewalk use case as well.

A good test case is: Start with `cycleway:both=no` => Change one side => Change it back so both are "no". It will be `cycleway=no` now, because that is the specified both-key.

tordans:
> The use counts for `sidewalk=*` and `sidewalk:<side>=*` are roughly 4,182,195 to 1,256,155, respectively, with the greater momentum behind the sidedness-formatted one, so I don't think there is a clear consensus based on usage alone.

Can you share the derivation (is that the right word?) for those numbers? I don't think it is valid to take all the `sidewalk=*` numbers given that they hold [values from different schemas](https://taginfo.openstreetmap.org/keys/sidewalk#values).
You could add the :left/:right to the numbers I quoted ([Source](https://docs.google.com/spreadsheets/d/1utbQRkEDA2gXyjOPwXC9tOW29t_tnzI9BAtyDaaeXNI/edit?gid=1060052597#gid=1060052597)) but then you have to add them to both schema, because we cannot know which schema users where following when mapping … which would result in the same diff of 1.607.712. See [Source](https://docs.google.com/spreadsheets/d/1utbQRkEDA2gXyjOPwXC9tOW29t_tnzI9BAtyDaaeXNI/edit?gid=1060052597#gid=1060052597&range=I10:K25) for more.

> with the greater momentum behind the sidedness-formatted one

I agree with that. SC, PWG, [other groups like us in Berlin](https://wiki.openstreetmap.org/wiki/Verkehrswende-Meetup/Gehwege) that work on the topic recommend `:both`. For the cycleway thread the counter argument was, that SC does not really count because mappers don't make a tagging decision so it's one apps preference…

> Yes. I would much rather this be discussed with wider audiences on other channels, especially the Forum, rather than decided by a small handful (in this case, including me) on a platform with much lower visibility.

I would like to understand that better. I think we agree that the status quo for iD is bad. The only way mappers can add sidewalk data is by manually adding tags. — The new system would respect all three variations of the tagging. So recommendations by the PWG and eg. in Berlin are still valid. They can be used and they will be displayed correctly. Users that add left/right data with iD will use the left/right schema. Only when a user assigns the same value for both sides, iD will pick the non-both-`cycleway=*` schema. — All in all, that is a win for me. 

But maybe I am too much of a "small iterations/steps towards a bigger goal" state of mind?

tordans:
> > Can you share the derivation (is that the right word?) for those numbers?
> 
> Unsided: sum count from Taginfo ([linky](https://taginfo.openstreetmap.org/keys/sidewalk#values)) for `sidewalk=no,both,right,separate,left,none,yes` = 4,190,469 Sided: sum count from Taginfo ([linky](https://taginfo.openstreetmap.org/search?q=sidewalk%3A)) for `sidewalk:left,right,both=*` = 1,259,047

That is not the argument that we have to overcome. The argument is, that `sidewalk=yes|separate|none` is the same but to be preferred over `sidewalk:both=yes|separate|none`. IMO we have three variations and you are mixing numbers between them. Also, there is no way to cleanly split the numbers with any reasonable effort.

> > Only when a user assigns the same value for both sides, iD will pick the non-both-cycleway=* schema.
> 
> Why not, in the case when a user assigns the same value for both sides, iD will condense them into `sidewalk:both=*`?

It could, of course. There is an issue on that in https://github.com/openstreetmap/id-tagging-schema/issues/1025 with a liked PR which would change the default and the link to the proposal to make that change official (which took a ton of work) and the long discussion of people on the forum about it, some against it. In the proposal and forum I argue that IMO the `:both` is the only clear choice given a UI like iD uses (which shows explicit sides but internally tags "both").

Maybe the next step is to bring this cycleway proposal to a vote so we have something to rely on for sidewalk as well. I will not have time to look into this more for at least 2 month. And still … all this will postpone getting something in iD now.

tordans:
Here is a new way to look at the numbers:

(See https://github.com/openstreetmap/id-tagging-schema/pull/454#issuecomment-2702860236 for my last comment on numbers.)

I am looking the numbers to see that mappers choose the "`:both`" tagging variation when they want to specify both side. The biggest difference in the numbers comes from the "no" tagging. I think we can argue, that a "no" always is a "both sides no". 

So lets look at the numbers without `no|none`:

  |     |   | Diff |  
-- | -- | -- | -- | --
`sidewalk:both=yes` |   11.513 | [Taginfo](https://taginfo.openstreetmap.org/tags/sidewalk%3Aboth=yes) |    
`sidewalk=yes` |   6.798 | [Taginfo](https://taginfo.openstreetmap.org/tags/sidewalk=yes#overview) | -4.715 | in favor of "both"
`sidewalk:both=separate` |   285.361 |[Taginfo](https://taginfo.openstreetmap.org/tags/sidewalk%3Aboth=separate) |    
`sidewalk=separate` |   326.661 |[Taginfo](https://taginfo.openstreetmap.org/tags/sidewalk=separate) | 41.300 | in favor of "nil"
  |   |    | 36.585 | 5,8%
  |   |    |   |  
`sidewalk:both=no\|none` |   42.102 |[Taginfo](https://taginfo.openstreetmap.org/tags/sidewalk%3Aboth=no),[Taginfo](https://taginfo.openstreetmap.org/tags/sidewalk%3Aboth=none) |   |  
`sidewalk=no\|none` |   1.613.303 |[Taginfo](https://taginfo.openstreetmap.org/tags/sidewalk=no),[Taginfo](https://taginfo.openstreetmap.org/tags/sidewalk=none) | 1.571.201 | more in "nil"

So its really only the `separate` value that we need to look into more.
And looking at the usage graph, it shows that users are migrating to the both variation:
https://taghistory.raifer.tech/?#***/sidewalk/separate&***/sidewalk%3Aboth/separate

> taghistory


When we now take into account that the wiki page at https://wiki.openstreetmap.org/wiki/Tag:sidewalk%3Dseparate clearly recommends to use the `:both` variation to avoid ambiguity, I think we have a good case to pick this as the variation that iD uses.

**To summarize**

* The numbers are only slightly in favor (6%) of the non-both variation (excluding `no` values)
* The wiki recommends `both` (for `separate`)
* The numbers show higher grows for `both` (for `separate`)
* The UI in iD is build in a way users make a clear choice per side, which the tagging should represent without ambiguity (which mean using `:both`)

I would argue that this a good basis for picking `sidewalk:both`. What do you think?

tordans:
@Lumikeiju do you have time – or know someone – to could tackle the effort to update https://wiki.openstreetmap.org/wiki/Key:sidewalk ? This page is still very much focused on the "old" tagging variation. It would be great if it where updated in a way that reflect the two variations, shows pro/cons and makes a recommendation that is in line which current tagging practices. This is important for the tagging schema, since this page will be linked as "reference" via the info-i from iD.

tordans:
I created https://github.com/openstreetmap/id-tagging-schema/pull/1507 as a follow up PR to this one (which I will close now). Please use the test preview to see if you notice any issues.

UW-Amy-Bordenave:
@tordans Yes, I can work on updating that (and the related) pages - can't provide a time estimate at the moment, though.

PR #455, opened 2022-05-12
Set common tags for trees

Binnette:
- Added important fields for trees identification: genus, species, taxon.
- Added others common fields

### DIFF
--- a/data/fields/diameter_crown.json
+++ b/data/fields/diameter_crown.json
@@ -3,5 +3,5 @@
     "type": "combo",
     "label": "Crown diameter",
     "snake_case": false,
-    "placeholder": "5 mm, 10 cm, 15 in…"
+    "placeholder": "4 m, 9 ft, …"
 }
--- a/data/fields/diameter_crown.json
+++ b/data/fields/diameter_crown.json
@@ -1,0 +1,7 @@
+{
+    "key": "diameter_crown",
+    "type": "combo",
+    "label": "Crown diameter",
+    "snake_case": false,
+    "placeholder": "5 mm, 10 cm, 15 in…"
+}
--- a/data/fields/genus.json
+++ b/data/fields/genus.json
@@ -1,0 +1,7 @@
+{
+    "key": "genus",
+    "type": "combo",
+    "label": "Genus",
+    "snake_case": false,
+    "placeholder": "Acer, Platanus, Quercus …"
+}
--- a/data/presets/natural/tree.json
+++ b/data/presets/natural/tree.json
@@ -1,14 +1,24 @@
 {
     "icon": "maki-park",
     "fields": [
+        "genus",
+        "species",
+        "taxon",
         "leaf_type_singular",
         "leaf_cycle_singular",
-        "denotation",
-        "height",
-        "diameter"
+        "start_date"
     ],
     "moreFields": [
+        "protected",
+        "age",
+        "sex",
+        "operator",
+        "source",
+        "height",
         "circumference",
+        "diameter_crown",
+        "denotation",
+        "name",
         "species/wikidata"
     ],
     "geometry": [
--- a/data/presets/natural/tree.json
+++ b/data/presets/natural/tree.json
@@ -15,8 +15,7 @@
         "start_date",
         "diameter_crown",
         "diameter",
-        "name",
-        "source"
+        "name"
     ],
     "geometry": [
         "point",
--- a/data/presets/natural/tree.json
+++ b/data/presets/natural/tree.json
@@ -4,6 +4,7 @@
         "leaf_type_singular",
         "leaf_cycle_singular",
         "height",
+        "denotation",
         "circumference"
     ],
     "moreFields": [
@@ -12,8 +13,8 @@
         "species/wikidata",
         "taxon",
         "start_date",
-        "denotation",
         "diameter_crown",
+        "diameter",
         "name",
         "source"
     ],
--- a/data/presets/natural/tree.json
+++ b/data/presets/natural/tree.json
@@ -9,10 +9,6 @@
         "start_date"
     ],
     "moreFields": [
-        "protected",
-        "age",
-        "sex",
-        "operator",
         "source",
         "height",
         "circumference",
--- a/data/presets/natural/tree.json
+++ b/data/presets/natural/tree.json
@@ -3,8 +3,8 @@
     "fields": [
         "leaf_type_singular",
         "leaf_cycle_singular",
-        "height",
         "denotation",
+        "height",
         "circumference"
     ],
     "moreFields": [
--- a/data/presets/natural/tree.json
+++ b/data/presets/natural/tree.json
@@ -1,21 +1,21 @@
 {
     "icon": "maki-park",
     "fields": [
-        "genus",
-        "species",
-        "taxon",
         "leaf_type_singular",
         "leaf_cycle_singular",
-        "start_date"
+        "height",
+        "circumference"
     ],
     "moreFields": [
-        "source",
-        "height",
-        "circumference",
-        "diameter_crown",
+        "genus",
+        "species",
+        "species/wikidata",
+        "taxon",
+        "start_date",
         "denotation",
+        "diameter_crown",
         "name",
-        "species/wikidata"
+        "source"
     ],
     "geometry": [
         "point",
--- a/data/fields/species.json
+++ b/data/fields/species.json
@@ -1,0 +1,7 @@
+{
+    "key": "species",
+    "type": "combo",
+    "label": "Species",
+    "snake_case": false,
+    "placeholder": "Acer platanoides, Quercus robur, …"
+}
--- a/data/fields/taxon.json
+++ b/data/fields/taxon.json
@@ -1,0 +1,7 @@
+{
+    "key": "taxon",
+    "type": "combo",
+    "label": "Taxon",
+    "snake_case": false,
+    "placeholder": "Acer platanoides Columnare, Pyrus calleryana Chanticleer …"
+}


data/presets/natural/tree.json
```diff
@@ -1,14 +1,24 @@
...
-        "denotation",
-        "height",
-        "diameter"
+        "start_date"
     ],
     "moreFields": [
+        "protected",
+        "age",
+        "sex",
+        "operator",
```
tordans, line 15: 0.19% (or is 0.43% the right number?) – so super low…

https://taginfo.openstreetmap.org/tags/?key=natural&value=tree#combinations

tordans (changes requested):
A few comments. I think we could keep this a bit simpler.

data/presets/natural/tree.json
```diff
@@ -1,14 +1,24 @@
 {
     "icon": "maki-park",
     "fields": [
+        "genus",
+        "species",
+        "taxon",
```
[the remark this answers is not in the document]
Binnette: Hello @tordans thanks for your comments. You are right genus, species and taxon are all in latin and reserved for user with some tree knowledge. But still, today genus and species are the only tag on trees that allows the user to identify a tree (Is is an Oak ? Is it a Pine ? ...) So for me, it is totally """mandatory""" that we add at least "genus" and "species".

You also have to think about the fact that some OpenData containing genus, species, start_date, ... have been imported to OSM in some cities. So the user can have the info about what kind of tree it is.

data/presets/natural/tree.json
```diff
@@ -1,14 +1,24 @@
...
         "leaf_type_singular",
         "leaf_cycle_singular",
-        "denotation",
-        "height",
-        "diameter"
+        "start_date"
     ],
     "moreFields": [
+        "protected",
```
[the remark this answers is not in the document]
Binnette: I never eared about it neither but it is noted in the main tree page: https://wiki.openstreetmap.org/wiki/Tag:natural%3Dtree

And to me, this tag make sens cause some trees are "protected" in my area, like very old trees and so on.

So yes, we can remove it for now until the definition is "more clear".

data/presets/natural/tree.json
```diff
@@ -1,14 +1,24 @@
...
+        "age",
+        "sex",
```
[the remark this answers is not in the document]
Binnette: The sex tag is also in the page https://wiki.openstreetmap.org/wiki/Tag:natural%3Dtree
But there is not a lot of tree species that have a "sex". Most of the trees are unisex.

So is still make sens to have it. But we can add it later is you want.

data/presets/natural/tree.json
```diff
@@ -1,14 +1,24 @@
...
         "leaf_type_singular",
         "leaf_cycle_singular",
-        "denotation",
-        "height",
-        "diameter"
+        "start_date"
     ],
     "moreFields": [
+        "protected",
+        "age",
```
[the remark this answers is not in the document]
Binnette: Yes. I agree. We should remove the age tag 👍

data/presets/natural/tree.json
Binnette: Ok let add the operator later 👍

data/presets/natural/tree.json
```diff
@@ -1,14 +1,20 @@
...
         "leaf_type_singular",
         "leaf_cycle_singular",
-        "denotation",
-        "height",
-        "diameter"
+        "start_date"
     ],
     "moreFields": [
+        "source",
+        "height",
```
tyrasd, line 13: I would slightly tend towards keeping this one as a regular field: It's reasonably popular, and not too hard to estimate/measure. What was the rationale behind making it an optional field?

data/presets/natural/tree.json
```diff
@@ -1,14 +1,20 @@
...
         "circumference",
+        "diameter_crown",
```
tyrasd, line 15: This field is currently not yet defined, as far as I can see. You would need to create a new file in https://github.com/openstreetmap/id-tagging-schema/tree/main/data/fields for it (you can take https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/diameter.json as a template).

data/presets/natural/tree.json
```diff
@@ -1,14 +1,20 @@
...
+        "denotation",
```
tyrasd, line 16: I would slightly tend towards keeping this one as a regular field: It's quite popular, and not too hard to estimate/measure. What was the rationale behind making it an optional field?

data/presets/natural/tree.json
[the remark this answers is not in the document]
tyrasd: I have to agree with @tordans here: These fields are too special for regular mappers: It cannot be expected that non experts know the required values. If we would display these fields by default, there is a relatively high probability that mappers would enter erroneous values like "oak" into the field(s).

Eventually it would be nice to have a special field in iD which lets people select the kind of tree from a list and the respective `species`/`taxon`/`genus` are filled in automatically. But that's something for another day.

data/presets/natural/tree.json
```diff
@@ -1,14 +1,20 @@
...
     "fields": [
+        "genus",
+        "species",
+        "taxon",
         "leaf_type_singular",
         "leaf_cycle_singular",
-        "denotation",
-        "height",
-        "diameter"
+        "start_date"
```
tyrasd, line 9: I would move this to the `moreFields` section: If there is no dedicated signage at the tree, this information is almost impossible to deduce on the ground. Imports and OGD data might make it possible to map this information, but for that use cases it is sufficient to have this field only show if the information is already present.

data/presets/natural/tree.json
```diff
@@ -1,14 +1,20 @@
...
     "icon": "maki-park",
     "fields": [
+        "genus",
+        "species",
+        "taxon",
         "leaf_type_singular",
         "leaf_cycle_singular",
-        "denotation",
-        "height",
-        "diameter"
```
tyrasd, line 8: I assume that you removed `diameter` because it is kind of duplicating the `circumference` tag and `diameter` is used less often? In principle, that would be fine for me, but I could also see potential confusion if a mapper finds that the data they entered in the field in the past is suddenly not displayed anymore. Maybe it could still make sense to continue to include the `diameter` field in the `moreFields` section?!

tyrasd (changes requested):
Thanks for the update @Binnette . Please find some further improvement suggestions from my side below.

data/presets/natural/tree.json
Binnette: Thank @tyrasd 👍  I created the file. Please review it 😃

data/presets/natural/tree.json
Binnette: Yes It was exactly my goal 😅 So I restore it 👍 Thanks @tyrasd

Binnette (commented):
Seems good to me now 👍

data/presets/natural/tree.json
```diff
@@ -1,14 +1,20 @@
...
+        "start_date"
     ],
     "moreFields": [
+        "source",
```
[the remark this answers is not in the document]
Binnette: Ok I remove it then. Thanks @tyrasd

data/presets/natural/tree.json
Binnette: Thanks @tyrasd I restored it 👍

data/presets/natural/tree.json
Binnette: Done 👍 Thanks @tyrasd

tyrasd (changes requested):
Thanks for the amendments. I overlooked earlier that the fields for `genus`, `species` and `taxon` were also missing. It's probably best to use the `combo` field type for these.

Binnette:
Thanks @tyrasd. I just created the missing files. Please review them 👍

tyrasd (approved):

Binnette:
Great! Thank @tyrasd! Have a nice day 🚀

data/fields/diameter_crown.json
```diff
@@ -0,0 +1,7 @@
+{
+    "key": "diameter_crown",
+    "type": "combo",
+    "label": "Crown diameter",
```
1ec5, line 4: Fields should be in Title Case.

data/fields/diameter_crown.json
```diff
@@ -0,0 +1,7 @@
...
+    "snake_case": false,
+    "placeholder": "4 m, 9 ft, …"
```
1ec5, line 6: According to [the unit documentation](https://wiki.openstreetmap.org/wiki/Map_features/Units), the `ft` symbol is discouraged in favor of an apostrophe (`'`), for better or worse. There’s no usage of `'` or `"` so far but also hardly any usage of `ft` either. Standardizing on a single symbol for feet would facilitate the eventual introduction of a measurement field type: ideditor/schema-builder#15.

data/presets/natural/tree.json
```diff
@@ -5,11 +5,17 @@
         "leaf_cycle_singular",
         "denotation",
         "height",
-        "diameter"
+        "circumference"
```
1ec5, line 8: Trees are conventionally measured by [diameter at breast height](https://en.wikipedia.org/wiki/Diameter_at_breast_height). However, [`circumference`](https://wiki.openstreetmap.org/wiki/Key:circumference) is the more prominently documented way to encode that information. Perhaps the idea is that it’s easier to get the circumference on a field survey using a measuring tape without having to employ trigonometry. Should we clarify the name of this field and add “diameter at breast height” and “dbh” as search terms for it? If so, we’ll probably need to split off a separate field, to avoid affecting other presets that coincidentally contain a Circumference field.

https://github.com/openstreetmap/id-tagging-schema/blob/be3954f8cc16b3150f9eed668a04b1aa9038a76f/data/fields/circumference.json#L4

tyrasd: currently only the tree preset uses this field, so we're fine. :wink: 

updated in 2c6c044

data/fields/genus.json
```diff
@@ -0,0 +1,7 @@
+{
+    "key": "genus",
+    "type": "combo",
+    "label": "Genus",
```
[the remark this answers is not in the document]
tyrasd: > Will this field ever be used for animal species

Not in the short or medium term I would say.

PR #458, opened 2022-05-15
Add area to parcel_locker

HandyHat:
As described on the wiki page (https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dparcel_locker), parcel lockers can be mapped as areas as well as nodes


tyrasd:
Thanks. It seems like we simply overlooked this in #339.

PR #459, opened 2022-05-16
Use more specific icon for Tee Box/MTB Route

arch0345:
This changes both the icon used for the Tee Box preset from `maki-golf` to [`fas-golf-ball`](https://fontawesome.com/v5/icons/golf-ball?s=solid) and the icon used for the Mountain Biking Route preset from `maki-bicycle` to [`fas-biking-mountain`](https://fontawesome.com/v5/icons/biking-mountain?s=solid)

### DIFF
--- a/data/presets/golf/tee.json
+++ b/data/presets/golf/tee.json
@@ -1,5 +1,5 @@
 {
-    "icon": "maki-golf",
+    "icon": "fas-golf-ball",
     "fields": [
         "name",
         "surface"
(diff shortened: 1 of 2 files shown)


tyrasd:
> `fas-biking-mountain`

This icon seems to be only available in the premium (non-free) _pro_ version of font-awesome, so it's not something we can use unfortunately.

> `fas-golf-ball`

Hmm… Here's how this one would look like:

image &emsp; image

I'm not sure this is so really better than the generic golf icon we use now. :thinking:

PR #462, opened 2022-05-18
use emoji and unicode for mtb:scale:imba values

andrewharvey:


data/fields/mtb/scale/imba.json
```diff
@@ -5,11 +5,11 @@
...
     "strings": {
         "options": {
-            "0": "Easiest (white circle)",
-            "1": "Easy (green circle)",
-            "2": "Medium (blue square)",
-            "3": "Difficult (black diamond)",
-            "4": "Extremely Difficult (double black diamond)"
+            "0": "⚪ Easiest (white circle)",
+            "1": "🟢 Easy (green circle)",
+            "2": "🟦 Medium (blue square)",
```
1ec5, line 10: FYI, some older operating systems may not be able to display these emoji, which were introduced in Unicode 12.0 (released in 2019).

1ec5 (commented):
Cool! This has me thinking schema-builder should have a way to associate individual options with icons: ideditor/schema-builder#56.

data/fields/mtb/scale/imba.json
andrewharvey: I think that's okay because the information is still present in the text, happy to remove the emoji again if ideditor/schema-builder#56 happens.

andrewharvey:
> Cool! This has me thinking schema-builder should have a way to associate individual options with icons: https://github.com/ideditor/schema-builder/issues/56.

Yeah I'm with you on that, so this is just a workaround until that is supported.

PR #463, opened 2022-05-20
add preset for highway=service mapped as areas

### LINKED ISSUE #431: `highway=service` areas neither provided nor accepted
Carto renders `area=yes; highway=service`, but iD suggests `area:highway=service` for the "Road Area" preset and complains about the former not being a line.

tyrasd:
Adds a preset for `highway=service` objects mapped as areas. Closes #431.

Since the only concrete use case listed on the [wiki](https://wiki.openstreetmap.org/w/index.php?title=Tag:highway%3Dservice&oldid=2321782#Linear_ways_and_areas) is for _widenings […] in an industrial estate_ I used the name _Maneuvering Area_ for this. But maybe it is too specific? My concern with _Service Road Area_ is that it could mislead many mappers to use this tag when [`area:highway`](https://wiki.openstreetmap.org/wiki/Key:area:highway) should actually be used (and the fact that it is too similar to the name of the _Road Area_ preset for these map features).

### DIFF
--- a/data/presets/highway/service_area.json
+++ b/data/presets/highway/service_area.json
@@ -33,6 +33,7 @@
         "service road",
         "service yard",
         "shunting area",
+        "staging lot",
         "yard"
     ],
     "searchable": false,
--- a/data/presets/highway/service_area.json
+++ b/data/presets/highway/service_area.json
@@ -1,0 +1,36 @@
+{
+    "icon": "iD-highway-service",
+    "fields": [
+        "name",
+        "service",
+        "maxspeed",
+        "surface",
+        "structure",
+        "access"
+    ],
+    "moreFields": [
+        "covered_no",
+        "flood_prone",
+        "lit",
+        "maxheight",
+        "maxspeed/advisory",
+        "maxweight_bridge",
+        "smoothness",
+        "trolley_wire"
+    ],
+    "geometry": [
+        "area"
+    ],
+    "tags": {
+        "highway": "service"
+    },
+    "terms": [
+        "maneuvering surface",
+        "manoeuvring area",
+        "manoeuvring surface",
+        "service road",
+        "shunting area"
+    ],
+    "matchScore": 0.8,
+    "name": "Maneuvering Area"
+}
--- a/data/presets/highway/service_area.json
+++ b/data/presets/highway/service_area.json
@@ -25,12 +25,16 @@
         "highway": "service"
     },
     "terms": [
+        "maneuvering area",
         "maneuvering surface",
         "manoeuvring area",
         "manoeuvring surface",
+        "paved lot",
         "service road",
-        "shunting area"
+        "service yard",
+        "shunting area",
+        "yard"
     ],
     "matchScore": 0.8,
-    "name": "Maneuvering Area"
+    "name": "Paved Service Area"
 }
--- a/data/presets/highway/service_area.json
+++ b/data/presets/highway/service_area.json
@@ -35,6 +35,6 @@
         "shunting area",
         "yard"
     ],
-    "matchScore": 0.8,
+    "searchable": false,
     "name": "Paved Service Area"
 }


matkoniecz:
> Maneuvering Area

this is actually quite good name and have no better ideas

> Service Road Area is that it could mislead many mappers to use this tag when [area:highway](https://wiki.openstreetmap.org/wiki/Key:area:highway) should actually be used

that is definitely serious problem, and I have seen this tag misused in this way, so often that it makes me dubious about adding support for it at all

1ec5:
> Since the only concrete use case listed on the [wiki](https://wiki.openstreetmap.org/w/index.php?title=Tag:highway%3Dservice&oldid=2321782#Linear_ways_and_areas) is for _widenings […] in an industrial estate_ I used the name _Maneuvering Area_ for this. But maybe it is too specific? My concern with _Service Road Area_ is that it could mislead many mappers to use this tag when [`area:highway`](https://wiki.openstreetmap.org/wiki/Key:area:highway) should actually be used (and the fact that it is too similar to the name of the _Road Area_ preset for these map features).

“Maneuvering area” could be misinterpreted as [runways and taxiways](https://en.wikipedia.org/wiki/Maneuvering_area). The icon may not imply anything aeronautical, but that’s no help to someone who enters “"maneuvering area"” into the Overpass turbo wizard because they’re familiar with that term. Someone attempting to map a turning circle that isn’t quite circular as an area might reach for this preset instead of Road Area.

To come up with a less ambiguous term, it would be helpful to get a more specific example of what’s considered correct usage of this tagging combination. All the examples I can think of would arguably be better represented as [`area:highway=service`](https://www.openstreetmap.org/way/35013329) or [`amenity=parking`](https://www.openstreetmap.org/way/233473075) areas with `service` ways [running through them](https://www.openstreetmap.org/way/41112573). The only term that comes to mind is “lot” (as in “parking lot” or “backlot”).

matkoniecz:
maneuvering areas in industrial sites qualify...

Maneuvering Road Area?

@ZeLonewolf

ZeLonewolf:
What the heck is a maneuvering area?

1ec5:
Yeah, the problem is that you can’t just stick arbitrary words together in any dialect of English and expect them to be intuitive, even if they would make logical sense, because the language is so rife with unpredictable compound words.

> The only term that comes to mind is “lot” (as in “parking lot” or “backlot”).

“Yard” or “service yard” can ambiguously refer to either the whole industrial site (think railyard) or just the paved area usually connected to loading docks. I don’t think there’s anything more natural than that or “lot”. If we really want a preset for this, it would end up being called something unwieldy like “Vehicle Free-for-All at Industrial Site” or something that makes assumptions like “Paved Lot”. But I still don’t see a compelling case for a preset independent of the existing Road Area preset.

tyrasd:
> What the heck is a maneuvering area?

Sorry, I just got this one as the result from a translation service. :sweat_smile: Perhaps it was giving a result which should be only used in an [airport context](https://en.wikipedia.org/wiki/Maneuvering_area).

Let's try to find a better name then.

> To come up with a less ambiguous term, it would be helpful to get a more specific example of what’s considered correct usage of this tagging combination.

:+1: 

As far as I understood one typical use case for this is for paved areas in industrial areas where trucks halt to load/unload goods (examples: [1](https://www.openstreetmap.org/way/464464273#map=17/42.21668/-83.38050), [2](https://www.openstreetmap.org/way/259663751#map=17/50.19166/8.64894)). Another example would be for the paved area in front of fire stations, bus depots, etc. ([3](https://www.openstreetmap.org/way/30214661)). Typically, in such situations there exists no clear linear linear way/path that the vehicles take. Mapping such areas as `amenity=parking` or `area:highway` doesn't really fit. I've also seen it in use for complex driveways/backyards geometries ([4](https://www.openstreetmap.org/way/52006903), [5](https://www.openstreetmap.org/way/28892118#map=18/49.41514/8.69512)), fuel stations ([6](https://www.openstreetmap.org/way/102949666#map=18/47.25838/11.37277)) or turning circles ([7](https://www.openstreetmap.org/way/464186536)), but at least some of these uses can be considered questionable since they could potentially also be mapped using `area:highway`.

The [wiki](https://wiki.openstreetmap.org/w/index.php?title=Key:area:highway&oldid=2270789#Differentiation_area:highway_vs._area=yes_on_a_highway) also says the following:

> > `[area](https://wiki.openstreetmap.org/wiki/Key:area)=[yes](https://wiki.openstreetmap.org/wiki/Tag:area%3Dyes)` on a [`highway`](https://wiki.openstreetmap.org/wiki/Key:highway)=* describes a routable highway area on which the dedicated traffic can route omnidirectionally, i.e. from any point along the edge to any other. This is typical for […] service areas where vehicles drive without prescribed lanes. The rationale is that routing engines should consider those omnidirectional objects in addition to nodes+edges in classical graph theory; thus no linear highway lines would be required within.

FWIW, I initially liked that the term _maneuvering_ in _Maneuvering Area_ kind of inherently conveys the omnidirectional driving on such paved areas.

> If we really want a preset for this

We could also choose to not support this tag, sure. But if we don't, we must not forget to fix the issue that iD currently shows a false validation warning (_Service Road should be a line, not an area_) which would be automatically fixed by introducing this preset.

> or something that makes assumptions like “Paved Lot”

I wouldn't mind this one. One possibility would be to use this name, but make the preset non-searchable to reduce the issue that users might apply it in potentially wrong situations?

ZeLonewolf:
Something like "paved service area" would probably get the idea across (assuming it's paved), but I think fundamentally we don't really have a word for this.

matkoniecz:
> make the preset non-searchable

I think that it is a good idea

1ec5:
“[Staging lot](https://osmus.slack.com/archives/C2VJAJCS0/p1661041394449079?thread_ts=1660878180.683979&cid=C2VJAJCS0)” came up in OSMUS Slack as a term for at least some of these facilities.

Issue #404, opened 2022-02-19
Add translation for `fuel:*` values


Dimitar5555:
Most common values are listed here: https://wiki.openstreetmap.org/wiki/Key:fuel

kjonosm:
@Dimitar5555
Can this issue be closed? See #621

Dimitar5555:
Yep, it can be closed.

Issue #408, opened 2022-02-21
Should iD be tagging `crossing=uncontrolled` instead of `crossing=marked`?


joepie91:
Currently, the 'marked crossing' preset will tag the crossing with `crossing=marked`. The [wiki page for `crossing`](https://wiki.openstreetmap.org/wiki/Key:crossing) claims that this is a duplicate of `crossing=uncontrolled`, and that the latter is 'approved', and indeed in terms of usage statistics it [appears to have seniority](http://taghistory.raifer.tech/#***/crossing/uncontrolled&***/crossing/marked). The [`highway:crossing` wiki page](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dcrossing) claims that both are semantically identical.

On the other hand, [the wiki page for `crossing=uncontrolled`](https://wiki.openstreetmap.org/wiki/Tag:crossing%3Duncontrolled) claims that `marked` vs. `unmarked` "seems to be more specific", though it fails to provide a clear rationale for that claim. It *implies* that `uncontrolled` is used for both marked and unmarked crossings currently, but this is not corroborated by other wiki pages on the topic, as far as I can tell.

Should iD be adopting `crossing=uncontrolled` as the canonical tag instead, that being the 'senior' tag which was already in wide use prior to iD's preset? Or is there a specific reason to continue using `crossing=marked` instead?

Dimitar5555:
`crossing=uncontrolled` is more useful for `railway=crossing`. Also `crossing=marked` and `crossing=unmarked` provide more information compared to `crossing=uncontrolled`. In my opinion the use of `crossing=uncontrolled` on `highway=crossing` should be discouraged.

joepie91:
> crossing=uncontrolled is more useful for railway=crossing.

That doesn't sound right to me - it is *primarily* used for road crossings in practice, with railway usage being [much less common](https://taginfo.openstreetmap.org/tags/?key=crossing&value=uncontrolled#combinations), and was indeed [approved for this use through a vote](https://wiki.openstreetmap.org/w/index.php?oldid=132309). What specifically makes it "more useful for railway crossings"?

> Also crossing=marked and crossing=unmarked provide more information compared to crossing=uncontrolled.

[...]

PR #467, opened 2022-05-24
Created translatable strings for fuel types, crossings, entrances and substations

Miroff:
1. Created list of translatable values for fuel types based on [Wiki](https://wiki.openstreetmap.org/wiki/Key:fuel)
2. Created limited set of values for RU based on discussion in local community. This set contains only fuel types used in the Russia. 

If this PR was merged, I'm going to create similar preset for Belarus.


Miroff:
3. Created list of translatable values for crossing types based on [Crossings wiki page](https://wiki.openstreetmap.org/wiki/Key:crossing)
4. Created list of translatable values for entrance types based on [Entrance wiki page](https://wiki.openstreetmap.org/wiki/Key:entrance)
5. Created list of translatable values for substation types based on [Substation wiki page](https://wiki.openstreetmap.org/wiki/Key:substation)

1ec5:
This would fix #404.

data/fields/crossing.json
```diff
@@ -1,5 +1,13 @@
 {
     "key": "crossing",
     "type": "combo",
-    "label": "Type"
+    "label": "Type",
+    "strings": {
+        "options": {
+          "traffic_signals": "Traffic signals and marking",
+          "uncontrolled": "Marking only",
```
1ec5, line 8: FYI, crosswalk classification is being discussed in #408.

data/fields/crossing.json
```diff
@@ -1,5 +1,13 @@
...
+          "unmarked": "No Traffic Signals, no marking",
```
1ec5, line 9: These values are using a mix of Title Case and Sentence case. I agree that title case can be annoying for long values, but the fields should at least be internally consistent.

data/fields/entrance.json
```diff
@@ -1,5 +1,19 @@
 {
     "key": "entrance",
     "type": "typeCombo",
-    "label": "Type"
+    "label": "Type",
+    "strings": {
+        "options": {
+          "yes": "Unspecified",
```
1ec5, line 7: Some possibly relevant naming considerations for entrance types in streetcomplete/StreetComplete#3949.

data/fields/fuel/fuel_multi.json
```diff
@@ -0,0 +1,66 @@
...
+    "label": "Fuel Types",
+    "strings": {
+        "options": {
+          "diesel": "Diesel",
+          "gtl_diesel": "GTL Diesel",
+          "hgv_diesel": "HGV Diesel",
+          "biodiesel": "Bio Diesel",
+          "diesel:class2": "Cold Weather Diesel",
+          "taxfree_diesel": "Untaxed Diesel",
+          "octane_80": "Octane 80",
```
1ec5, line 13: A bunch of these values need to be tweaked for American English. For starters, it would be “80 Octane”, not “Octane 80”. I think `octane_` is merely being used as a namespace.

data/fields/fuel/fuel_multi.json
```diff
@@ -0,0 +1,66 @@
...
+          "1_25": "1:25 mix",
+          "1_50": "1:50 mix",
+          "ethanol": "Ethanol",
+          "ethanol_free": "Ethanol-free",
+          "methanol": "Methanol",
+          "svo": "SVO",
+          "e5": "5% ethanol, 95% gasoline mix",
+          "e10": "10% ethanol, 90% gasoline mix",
+          "e20": "20% ethanol, 80% gasoline mix",
+          "e85": "85% ethanol, 15% gasoline mix",
```
1ec5, line 32: This is helpful as help text, but in American English it’s simply known as E85.

data/fields/substation.json
```diff
@@ -1,5 +1,18 @@
 {
     "key": "substation",
     "type": "typeCombo",
-    "label": "Type"
+    "label": "Type",
+    "strings": {
+        "options": {
+          "transmission": "HV transmission",
+          "distribution": "HV/MV distribution",
+          "minor_distribution": "LV distribution",
```
1ec5, line 9: It took me a little while to figure out what these nonstandard initialisms stand for. Can we spell them out?

1ec5 (commented):
Just a few things that jumped out at me at a glance, but this would benefit from a more detailed review.

data/fields/fuel/fuel_multi.json
1ec5: Also it probably makes sense to add strings for all the values between 80 and 100, if not beyond, because the specific octane numbers sold vary by country, region, or city, or brand.

tyrasd:
I've tweaked this quite a bit and finally merged as 95c206b (sorry for the extra commits in your main branch @Miroff). I excluded the crossings field, as that is now handled in #590. I also changed the way the localized fuel field works: instead of duplicating the preset for every region, I made the field region-specific and let it inherit the strings from the global field. That avoids that translators have to translate the same string more than once, and should scale pretty nicely if you want to add more countries. Further, I removed lesser used fuel values (cut-off approx ~1000 usages with exceptions for the H2 types) and reworded the substation value strings slightly.

data/fields/fuel/fuel_multi.json
```diff
@@ -0,0 +1,51 @@
+{
+    "key": "fuel:",
+    "type": "multiCombo",
+    "label": "Fuel Types",
+    "strings": {
+        "options": {
+            "diesel": "Diesel",
+            "GTL_diesel": "GTL Diesel",
+            "HGV_diesel": "HGV Diesel",
+            "biodiesel": "Bio Diesel",
```
1ec5, line 10: This is spelled “Biodiesel”, as one word.

data/fields/fuel/fuel_multi.json
```diff
@@ -0,0 +1,51 @@
+{
+    "key": "fuel:",
+    "type": "multiCombo",
+    "label": "Fuel Types",
+    "strings": {
+        "options": {
+            "diesel": "Diesel",
+            "GTL_diesel": "GTL Diesel",
+            "HGV_diesel": "HGV Diesel",
```
1ec5, line 9: “HGV” is a British acronym completely unknown in American English, where this type would be called “Truck Diesel”. (It’s an odd value. Apparently it refers to the kind of nozzle / what vehicle class can use the pump, not the type of fuel supplied.)

data/fields/substation.json
```diff
@@ -1,5 +1,19 @@
 {
     "key": "substation",
     "type": "typeCombo",
-    "label": "Type"
+    "label": "Type",
+    "strings": {
+        "options": {
+            "transmission": "High Voltage Transmission",
```
1ec5, line 7: Nit: “High-Voltage”, with a hyphen.

1ec5 (commented):
A couple minor things I noticed belatedly:

PR #469, opened 2022-05-25
Semantic correction to natural=stone and natural=rock

bgo-eiu:
This is hopefully not too controversial of a proposition, but this pull request is to correct the "Stone" and "Rock" presets to be semantically accurate.

Context for why I think this is necessary:
* Stone and rock do not mean the same thing. Stone is a general term which can be used to describe a variety of materials, including non-rock and man-made ones, while rock is a term which specifically describes naturally occurring aggregates of mineral material. This is already reflected else where in the tagging schema; "paving stone" is not seen a similar concept to rock, and "bare rock" would not be considered an acceptable landcover type to map over a paving stone surface. Many rocks can be stones, but stones include a lot more than just rocks.

* The recommendation from the wiki to use natural=stone to tag "loose" material and natural=rock for "attached" rocks is a historical one from before introduction of the "geological" set of tags. [Key:geological](https://wiki.openstreetmap.org/wiki/Key:geological) was adopted and documented years ago at this point, and geological=outcrop essentially outmoded any reason to use a semantically incorrect key for stones / rocks. Outcrop means attached rock, so anything that is attached to the ground can be tagged with geological=outcrop, and there is no reason anymore to tag actual loose rocks that are best described as such with the more generic term stone. It is likely the outcrop tag is not that well known so I added an explanation about outcrops to the natural=rock page and intend to link it in more places there.
  
* Currently, the preset names for both stone and rock have "Boulder" in the name after implying they are detached or attached. Boulders are definitionally detached and there is no such thing as an attached boulder anyway, so this is contradictory. [Boulder from a geologic perspective is a loose rock larger than 25.6 cm in diameter](https://en.wikipedia.org/wiki/Boulder), so I just left boulder in the rock json file since boulders are rocks.

I know sometimes there's some fuss made about iD "forcing" certain tag uses when changes like this are made, but my guess is that this is considerably less contentious than crosswalks.

### DIFF
--- a/data/presets/natural/rock.json
+++ b/data/presets/natural/rock.json
@@ -12,8 +12,7 @@
     },
     "terms": [
         "boulder",
-        "stone",
         "rock"
     ],
-    "name": "Attached Rock / Boulder"
+    "name": "Rock"
 }
--- a/data/presets/natural/stone.json
+++ b/data/presets/natural/stone.json
@@ -11,9 +11,7 @@
         "natural": "stone"
     },
     "terms": [
-        "boulder",
-        "stone",
-        "rock"
+        "stone"
     ],
-    "name": "Unattached Stone / Boulder"
+    "name": "Stone"
 }


tordans:
I am not sure this change is a good idea – at least not the way it is change right now. The goal of the "terms" and the "name" is to guide me as a user to pick the right preset. By removing the terms that match both, you assume that I know the right term from the start, which I most likely don't.

IMO, we should keep the terms with an overlap and try to find a name that match the your definition more closely. This way, I will be guided to pick the right preset from the list.

(I did not look into the semantics of this; I was only looking at the UX implications.)

bgo-eiu:
That makes sense. Stone and rock could be terms for both and then people could pick which makes sense to them.

I think simpler names tend to work better - right now I get confused when I map a rock and it calls it an "attached rock / boulder" because a lot of the time its not attached or a boulder. And there's already a tag for attached rock, geological=outcrop. For stone vs. rock, it's kind of like commercial vs. retail building. Any rock could be called a stone, but stones can also be a lot of different things besides rocks.

Maybe it would be more helpful to keep "attached rock" as a separate preset that adds natural=rock + geological=outcrop? That way people who are searching can see that option too.

tyrasd:
> IMO, we should keep the terms with an overlap

+1

<hr>

> I think simpler names tend to work better

That sounds reasonable to me. I also think that the adjective "_attached_" could be confusing to mappers who are not aware that is meant to mean "_attached to the underlying bedrock_".

> map a rock […] a lot of the time its not attached or a boulder

I'm only still slightly confused here, because the wiki still differentiates between [stone](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dstone) and [rock](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dstone) by whether they are _attached to the underlying bedrock_ or not. When exactly is a stone a rock if it is not attached?

bhousel:
For context, people should also look at @westnordost's original pull request if you want to know about why there are 2 presets and why they are named what they are.  He did a great job on this:
https://github.com/openstreetmap/iD/pull/6311

westnordost:
With Aliases, we could now use the (scientifically) correct `name` for `natural=stone` -  _["Glacial erratic"](https://en.wikipedia.org/wiki/Glacial_erratic)_ - and put "Unattached Boulder" (etc.) into `aliases`. Or the other way round.

But in any case, I'd conclude that "Stone" for `natural=stone` would be wrong, according to the research I did linked by @bhousel , thanks for that!

bgo-eiu:
The German translation context is interesting, I wasn't aware of that. When I think of fels, I think of this rock type, which doesn't imply anything about whether it is attached to the bedrock https://en.wikipedia.org/wiki/Hornfels (I don't think any derivatives of findling ever made it into English.) I think "attached" and "unattached" are really the most confusing parts here and a wording that doesn't rely on that would be ideal - possibly support for geological=outcrop if that would help too.

Looking at taginfo/Overpass it's also clear that the natural=stone tag has a better understood meaning in Germany, and is much more concentrated in Europe generally than natural=rock.

Glacial erratic seems well suited based on all the pictures I'm looking at of existing mapped natural=stone features, but my only concern would be it might be too specific. Glacial erratics are generally quite big, and I don't know to what extent people who are mapping natural=stone would consider something closer in size to a fire hydrant or smaller to be the same thing. If you know anything about this being used for smaller objects that would be helpful. It would definitely be an improvement to label it that, because I don't think many English speakers are aware of that meaning - even if they don't use it at the map as often, it looks like a lot of these natural=stone tags in Germany also have image links and names and that sort of thing, so a preset name like that would help English speakers to tell its being understood more specifically.

bgo-eiu:
> I'm only still slightly confused here, because the wiki still differentiates between [stone](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dstone) and [rock](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dstone) by whether they are attached to the underlying bedrock or not. When exactly is a stone a rock if it is not attached?

In English at least, something is a rock if it is: a naturally occurring solid, consistent of mineral material, and not a product of a biologic process. The mineral material part is maybe the harder to understand part, but it just means whether or not it is made up of crystalline (structured) material as opposed to an un-structured material. Glass would be one of the most common examples of natural non-crystal material.

Stones don't have that specific of a definition. Examples of stones which are not rocks would be:
* Types of stones defined by their man-made use: gemstones, milestones, paving stones, and so on. That kind of covers most of the things which are commonly called stones which could be defined based on any material (like concrete)
* Naturally occurring glass materials
* Other amorphous materials like opal
* Biologically-formed stones (like kidney stones)
Those last two items are usually too small to map though if there's a huge one somewhere I'd love to be proven wrong.

Loose rocks in Baltimore, MD
This is something that I've mapped as natural=rock. They're the same types of rock that you might see exposed from the ground in the area, but they were manually removed at some point to decorate this neighborhood sign. The fact that it's the same material as you would find exposed from the earth is why I'm inclined to describe them as rocks, because they can still be described with all the same properties.

bgo-eiu:
OK, reading a translation of the wikipedia article for "findling" (https://de.wikipedia.org/wiki/Findling) which is even connected to the glacial erratic article on English wiki, it does seem like it's always understood to be a larger object and even mentions that diameter is used for criteria for findlings which have some kind of protected status. I would definitely support changing the English preset name for natural=stone to "glacial erratic" so that people are aware of how specific this meaning is, and it would help mappers be on the same page about what they are talking about.

I could close this and create a separate pull request based on that. Would other people think this is a good idea, and would adding a "attached to bedrock" checkmark to the "rock" preset to add geological=outcrop be OK as well? I like these ideas more than what I originally suggested.

westnordost:
I do, at least. Since you researched a bit more here, you could also improve the list of `terms` and `aliases` if you find some that work.

I cannot find anything regarding using `geological=outcrop` in combination with "rock" on the wiki. Reading it, I do not think this tag is intended for that usage. Also, adding this tag always would be an unnecessary duplication in the best case and and leading to confusion or misunderstandings in the worst case.

bgo-eiu:
Sorry, I don't mean adding geological=outcrop for everything with natural=rock. Just adding an option for it, like a checkbox, "[Y/N] Attached to earth" So it's just offering a more specific option, if someone isn't sure then they have the option to leave it alone. It's kind of location dependent, like if a place is mostly covered in vegetation an actual outcrop where the rock is attached to the earth would be more rare/notable compared to rocks which aren't attached to anything.

The geological key tags are intended to be descriptive keys that can be paired with natural key tags where applicable. So you could do natural=rock + geological=outcrop if you were mapping a rock outcrop as a node, but you can also use things like natural=cliff + geological=outcrop for outcrops that form cliffs for example (like https://www.openstreetmap.org/way/948586833), or natural=bare_rock + geologic=outcrop for a larger area that can be described that way (https://www.openstreetmap.org/way/948586833).

I added some images to illustrate the outcrop page on the wiki page better, I can add a "how to use" or similar section to clarify this as well.

For terms and aliases, I will keep stone, rock, and boulder for both based on the points made already and try to make a list of others (I'll take a look more out how people are using these tags.)

matkoniecz:
"Attached to earth" seems weird, as all nonfloating boulders will be (vast majority of them and floating blocks of pumice are unlikely to be mappable). 

Have you intended "Attached to bedrock"? Or "Exposed bedrock"? Or "Outcrop of berdrock"?

bgo-eiu:
Maybe something is getting lost in translation in describing it that way - attached to bedrock, exposed bedrock, and outcrop of bedrock would all be accurate as well and could work if they are more easily understood.

A rock doesn't have to be very big to be to heavy to move normally, situations like the rocks picture I posted above are more common than outcrops a lot of the time. Those are too small to be described just as boulders, and boulders are often rounded off to some degree rather than broken off like above. Looking at an Overpass Turbo query, people have been using natural=rock to map things like this for example - rocks used as trail markers which aren't very big but are mappable:

Trail marker rocks in Maine

That would be the distinction, I think this is different from the German concept of findling as well.

bgo-eiu:
OK, I am closing this and have submitted #471 based on the discussion here. Any feedback on the new update would be appreciated.

PR #470, opened 2022-05-26
move aliases of `highway=track` from terms to aliases field

westnordost:
I would also have renamed the preset to `Land-Access Road` and moved `Track Road` to the aliases because with aliases, there is no need anymore to have any names with a `/` but since you [just](https://github.com/openstreetmap/iD/commit/d486ab197d624cff8b1b9f1a4a2958848cb19c62) implemented interpreting the `aliases` field in iD, I guess you are aware of this possibility and had your reasons to keep it that way.

See https://github.com/openstreetmap/id-tagging-schema/pull/288#issuecomment-1134043557


data/presets/highway/track.json
```diff
@@ -30,29 +30,34 @@
...
         "quad",
-        "ranch road",
         "rut",
-        "track",
-        "track road",
-        "two track",
         "unmaintained",
-        "woods road"
+        "unmaintained road",
+        "dirt road"
```
tyrasd, line 46: I'd like to keep terms sorted alphabetically. :nerd_face:

westnordost: I didn't realize they were before. So, should I change this?

tyrasd: yes, please

data/presets/highway/track.json
```diff
@@ -30,29 +30,34 @@
...
-        "unmaintained",
-        "woods road"
+        "unmaintained"
     ],
-    "name": "Track / Land-Access Road"
+    "name": "Track / Land-Access Road",
+    "aliases": [
+        "Agricultural Road",
+        "Carriage Road",
+        "Dirt Road",
```
zekefarwell, line 50: I don't think "Dirt Road" should be included as an alias.  In American English this term is used to mean any unpaved road regardless of purpose or importance.  Many of these roads are properly tagged with a different `highway` class and `surface=dirt/compacted/unpaved`.

zekefarwell: I'd say the likelyhood that an unmaintained road would be `track` is much higher than the likelyhood that a dirt road would be track.  I live in a state where more of the public road network is dirt than paved ([Vermont road stats](https://vtrans.vermont.gov/planning/maps/stats)).  These are roads people live on that are part of the general purpose road network, so here it is just as likely that a given dirt road should be tagged `highway=residential/unclassified` + `surface=unpaved` and `highway=track` should not be assumed just because a road is dirt.

I'm not really sure what the difference is between "aliases" and "terms", but I'd suggest not including the term "dirt road" in either to avoid confusion.

tyrasd:
> I guess whether it should be an alias depends how exactly "alias" is defined and how it is (going) to be shown in the UI.

> I'm not really sure what the difference is between "aliases" and "terms"

Well, the problem here is that this property was never documented (see https://github.com/ideditor/schema-builder/issues/28), nobody _really_ knows.

The most conservative approach would be to use `aliases` only for proper synonyms (here: _land access road_ and _track road_). Whether it makes sense to also include names which describe a strict subset of features represented by the preset can be discussed IMO (here for example: _forest track_, _agricultural road_, etc.). I would abstain from including terms as aliases which can cover features not represented by the preset (here: _dirt road_, _unmaintained road_).

westnordost:
Yeah... I will answer in https://github.com/ideditor/schema-builder/issues/28

data/presets/highway/track.json
```diff
@@ -30,29 +30,34 @@
...
+        "Land-Access Road",
+        "Farm Road",
+        "Fire Road",
+        "Forest Road",
+        "Forestry Road",
+        "Logging Road",
+        "Ranch Road",
```
1ec5, line 57: Speaking of these industry-specific road types, “[Haul Road](https://en.wikipedia.org/wiki/Haul_road)”, “Haulage Road”, and “Mining Road” refer to a kind of road that I’m pretty sure would necessarily be tagged `highway=track`.

[](https://commons.wikimedia.org/wiki/File:Wagon_on_a_dusty_haul_road_-_geograph.org.uk_-_2586796.jpg)

[](https://commons.wikimedia.org/wiki/File:Tilgate_Forest,_Haul_track_churned_up_during_harvesting_operations_-_geograph.org.uk_-_1763337.jpg)

data/presets/highway/track.json
```diff
@@ -30,29 +30,34 @@
...
+        "unmaintained"
     ],
-    "name": "Track / Land-Access Road"
+    "name": "Track / Land-Access Road",
+    "aliases": [
+        "Agricultural Road",
+        "Carriage Road",
+        "Dirt Road",
+        "Land-Access Road",
+        "Farm Road",
```
1ec5, line 52: Glad we aren’t debating [the finer points](https://www.houstonpublicmedia.org/articles/news/2018/05/04/283588/whats-the-difference-between-ranch-to-market-and-farm-to-market-roads/) of American English anymore!

[Farm Road 1](https://commons.wikimedia.org/wiki/File:FM1-SignAlongRoadside.jpg)

data/presets/highway/track.json
```diff
@@ -30,29 +30,34 @@
...
+        "Fire Road",
+        "Forest Road",
+        "Forestry Road",
+        "Logging Road",
+        "Ranch Road",
+        "Track Road",
+        "Two-Track Road",
+        "Unmaintained Road",
+        "Woods Road"
+    ]
```
1ec5, line 62: “Vehicular Trail” would be a good alias, if not a good candidate for the preset’s name. I had previously overlooked this term in all the fuss about maintenance. Sometimes it’s a synonym of “4×4 road“, but it doesn’t emphasize access restrictions or surface quality.

data/presets/highway/track.json
1ec5: iD ranks terms by their position in the list, so sorting them alphabetically promotes some rarer terms over more common terms. This might be less of a problem now that some terms are aliases, but it’s hard to test.

westnordost:
Are you sure? Is this documented somewhere (+documented for translators)?

On May 28, 2022 8:12:03 PM GMT+02:00, "Minh Nguyễn" ***@***.***> wrote:
***@***.*** commented on this pull request.
>
>
>
>
>
>
>
>>          "unmaintained",
>
>-        "woods road"
>
>+        "unmaintained road",
>
>+        "dirt road"
>
>
>
>iD ranks terms by their position in the list, so sorting them alphabetically promotes some rarer terms over more common terms. This might be less of a problem now that some terms are aliases, but it’s hard to test.
>
>
>
>-- >
>Reply to this email directly or view it on GitHub:
>
>https://github.com/openstreetmap/id-tagging-schema/pull/470#discussion_r884163417
>
>You are receiving this because you authored the thread.
>
>
>
>Message ID: ***@***.***>
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

data/presets/highway/track.json
westnordost: Are you sure? Is this documented somewhere (+documented for translators)?

westnordost:
(FYI I consider this blocked until https://github.com/ideditor/schema-builder/issues/28 came to a conclusion)

data/presets/highway/track.json
1ec5: Oh, either I’m mistaken or it’s no longer in place. It looks like nowadays iD only considers the index of a match [against the preset name](https://github.com/openstreetmap/iD/blob/0c69018f96f4f5e5303c215e03693117f4a4dc1d/modules/presets/collection.js#L93) but only filters by terms.

tyrasd:
> (FYI I consider this blocked until [ideditor/schema-builder#28](https://github.com/ideditor/schema-builder/issues/28) came to a conclusion)

see https://github.com/ideditor/schema-builder/issues/28#issuecomment-1141327355 – please only keep `Track Road` and `Land Access Road` in the list of `aliases` and move the rest back to the `terms`.

westnordost:
Applied suggestions and https://github.com/openstreetmap/id-tagging-schema/pull/470/#issuecomment-1141350491 , ready for merge

tyrasd (approved):

PR #471, opened 2022-05-27
Updated `natural=rock` and `natural=stone` presets

bgo-eiu:
These updates mostly reflect the discussion in #469 , with the objective of arriving at presets for natural=rock and natural=stone that are more clear in meaning and still relate to how they are used.

Included in this version:

* "rock," "stone," and "boulder" as terms for both since enough people are using these interchangeably that it would help to see both options. Added some miscellaneous related terms for both that I found from searching ("cairns" appears to be a term that can be used to describe rocks used as trail markers for example).

* "Glacial Erratic / Boulder" as the name of the natural=stone preset because as @westnordost clarified, this is being used more specifically than the name "stone" or "boulder" suggests alone. This would help make it more clear how people are using this tag.

* Just "Rock" for natural=rock. This is the more common tag and people have been using it for a variety of types of rocks so the simple name seems OK as long as the terms are there.

* Added "Exposed bedrock" as a checkbox field for natural=rock. Per @matkoniecz asking about this, I think that this has a much more clear meaning for adding geological=outcrop to natural=rock. Attached is too ambiguous and was confusing in the first place.

* Added dimension field options (height by default; diameter, width, circumference in additional) since these are often used in combination and it would help to know what type of object people are actually mapping.

* Added a "Lava field" checkbox field in the additional options for natural=bare_rock which adds geological=volcanic_lava_field. This tag combination is used a lot in Iceland for example. I noticed bare_rock didn't have very much in the presets so that could  be a helpful detail to have in there.

### DIFF
--- a/data/fields/geological_outcrop.json
+++ b/data/fields/geological_outcrop.json
@@ -1,7 +1,7 @@
 {
     "key": "geological",
     "type": "defaultCheck",
-    "label": "Exposed bedrock",
+    "label": "Exposed Bedrock",
     "strings": {
         "options": {
             "undefined": "No",
--- a/data/fields/geological_outcrop.json
+++ b/data/fields/geological_outcrop.json
@@ -8,4 +8,4 @@
             "outcrop": "Yes"
         }
     }
-  
+}
--- a/data/fields/geological_outcrop.json
+++ b/data/fields/geological_outcrop.json
@@ -1,0 +1,11 @@
+{
+    "key": "geological",
+    "type": "defaultCheck",
+    "label": "Exposed bedrock",
+    "strings": {
+        "options": {
+            "undefined": "No",
+            "outcrop": "Yes"
+        }
+    }
+  
--- a/data/fields/geological_volcanic_lava_field.json
+++ b/data/fields/geological_volcanic_lava_field.json
@@ -1,0 +1,11 @@
+{
+    "key": "geological",
+    "type": "defaultCheck",
+    "label": "Lava field",
+    "strings": {
+        "options": {
+            "undefined": "No",
+            "volcanic_lava_field": "Yes"
+        }
+    }
+   
--- a/data/fields/geological_volcanic_lava_field.json
+++ b/data/fields/geological_volcanic_lava_field.json
@@ -1,7 +1,7 @@
 {
     "key": "geological",
     "type": "defaultCheck",
-    "label": "Lava field",
+    "label": "Lava Field",
     "strings": {
         "options": {
             "undefined": "No",
--- a/data/fields/geological_volcanic_lava_field.json
+++ b/data/fields/geological_volcanic_lava_field.json
@@ -8,4 +8,4 @@
             "volcanic_lava_field": "Yes"
         }
     }
-   
+}
--- a/data/presets/natural/bare_rock.json
+++ b/data/presets/natural/bare_rock.json
@@ -1,5 +1,8 @@
 {
     "icon": "temaki-boulder3",
+    "moreFields": [
+        "geological_volcanic_lava_field"
+    ],
     "geometry": [
         "area"
     ],
--- a/data/presets/natural/rock.json
+++ b/data/presets/natural/rock.json
@@ -1,7 +1,14 @@
 {
     "icon": "temaki-boulder2",
     "fields": [
-        "name"
+        "name",
+        "height",
+        "geological_outcrop"
+    ],
+    "moreFields": [
+        "diameter",
+        "circumference",
+        "start_date"
     ],
     "geometry": [
         "point",
@@ -13,7 +20,10 @@
     "terms": [
         "boulder",
         "stone",
-        "rock"
+        "rock",
+        "attached rock",
+        "cairn",
+        "bedrock"
     ],
-    "name": "Attached Rock / Boulder"
+    "name": "Rock"
 }
--- a/data/presets/natural/rock.json
+++ b/data/presets/natural/rock.json
@@ -8,7 +8,6 @@
     "moreFields": [
         "width",
         "diameter",
-        "circumference",
         "start_date"
     ],
     "geometry": [
@@ -26,7 +25,9 @@
         "cairn",
         "bedrock",
         "outcrop",
-        "exposed rock"
+        "exposed rock",
+        "decorative boulder",
+        "landscape boulder"
     ],
     "name": "Rock"
 }
--- a/data/presets/natural/rock.json
+++ b/data/presets/natural/rock.json
@@ -20,7 +20,6 @@
     "terms": [
         "boulder",
         "stone",
-        "rock",
         "attached rock",
         "cairn",
         "bedrock",
--- a/data/presets/natural/rock.json
+++ b/data/presets/natural/rock.json
@@ -6,6 +6,7 @@
         "geological_outcrop"
     ],
     "moreFields": [
+        "width",
         "diameter",
         "circumference",
         "start_date"
@@ -23,7 +24,9 @@
         "rock",
         "attached rock",
         "cairn",
-        "bedrock"
+        "bedrock",
+        "outcrop",
+        "exposed rock"
     ],
     "name": "Rock"
 }
--- a/data/presets/natural/stone.json
+++ b/data/presets/natural/stone.json
@@ -16,11 +16,9 @@
         "natural": "stone"
     },
     "terms": [
-        "boulder",
         "stone",
         "rock",
         "findling",
-        "glacial erratic",
         "erratic"
     ],
     "aliases": [
--- a/data/presets/natural/stone.json
+++ b/data/presets/natural/stone.json
@@ -1,7 +1,13 @@
 {
     "icon": "temaki-boulder1",
     "fields": [
-        "name"
+        "name",
+        "height"
+    ],
+    "moreFields": [
+        "width",
+        "diameter",
+        "circumference"
     ],
     "geometry": [
         "point",
@@ -13,7 +19,10 @@
     "terms": [
         "boulder",
         "stone",
-        "rock"
+        "rock",
+        "findling",
+        "glacial erratic",
+        "erratic"
     ],
-    "name": "Unattached Stone / Boulder"
+    "name": "Glacial Erratic / Boulder"
 }
--- a/data/presets/natural/stone.json
+++ b/data/presets/natural/stone.json
@@ -6,8 +6,7 @@
     ],
     "moreFields": [
         "width",
-        "diameter",
-        "circumference"
+        "diameter"
     ],
     "geometry": [
         "point",
@@ -24,5 +23,8 @@
         "glacial erratic",
         "erratic"
     ],
-    "name": "Glacial Erratic / Boulder"
+    "aliases": [
+        "Boulder"
+    ],
+    "name": "Glacial Erratic"
 }


data/fields/geological_outcrop.json
```diff
@@ -0,0 +1,11 @@
+{
+    "key": "geological",
+    "type": "defaultCheck",
+    "label": "Exposed bedrock",
```
1ec5, line 4: Field names should be in Title Case.

data/presets/natural/stone.json
```diff
@@ -13,7 +19,10 @@
...
         "boulder",
         "stone",
-        "rock"
+        "rock",
+        "findling",
+        "glacial erratic",
+        "erratic"
     ],
-    "name": "Unattached Stone / Boulder"
+    "name": "Glacial Erratic / Boulder"
```
1ec5, line 27: It’s been a long time since my geology appreciation class in school. “Glacial erratic” is a term I’ve never encountered before, despite hailing from in a region where terms like “terminal moraine” are well known. If there’s no colloquial term for this specific concept, then “glacial erratic” is a good name for the preset. But seeing “boulder” next to it might make me inclined to choose this obscurely named preset over “Rock” when tagging a decorative boulder. Maybe “Boulder” should be an alias but not part of the name? (“Decorative boulders”, marketed as such, are so common at garden stores that they could even be a term for that preset.)

data/presets/natural/rock.json
```diff
@@ -1,7 +1,15 @@
...
     "fields": [
-        "name"
+        "name",
+        "height",
+        "geological_outcrop"
+    ],
+    "moreFields": [
+        "width",
+        "diameter",
+        "circumference",
```
1ec5, line 11: This field is currently called “Circumference (at Breast Height)” because the `natural=tree` tagging scheme derives `circumference` from a diameter at breast height (DBH) measurement. (Go figure.) We’ll need to split out a different field for circumferences of other things, like rocks, for which “breast height” would be baffling.

/ref #455

bgo-eiu: Didn't realize it was that specific - I think there are already petrified trees on the map as natural=rock so it's not totally out of the question but diameter will do.

data/presets/natural/stone.json
bgo-eiu: I was kind of on the fence since boulder is a more commonly known term. The German concept of "findling" that the tag is used for does seem to be a near exact match to "glacial erratic" though so boulder could just be an alias. Good point about decorative boulders / landscape boulders, I was thinking a lot of what people would describe as boulders count as glacial erratics anyway but those would not.

Thank you for the feedback, I will update 👍

bgo-eiu:
Updated to reflect feedback above

data/presets/natural/stone.json
```diff
@@ -13,7 +18,13 @@
     "terms": [
         "boulder",
         "stone",
-        "rock"
+        "rock",
+        "findling",
+        "glacial erratic",
```
tyrasd, line 23: the preset's name shouldn't be repeated in the list of `terms`

bgo-eiu: I'm a little unclear on how aliases work - so if boulder is an alias, it would still operate the same way as if it were in terms without duplicating it?

bgo-eiu:
Updated to avoid duplication in terms/aliases

data/presets/natural/stone.json
tyrasd: Yes, exactly. Basically the set of `name + aliases + terms` defines all the possible "words" a preset can be found when searching for it. Whether an individual "word" is contained in the `name` or `aliases` or `terms` only changes how the preset is displayed in the UI (`name`) and influences the order of search results (presets that only match a word from the list of `terms` are sorted below matching `name` or `aliases`).

data/fields/geological_outcrop.json
```diff
@@ -0,0 +1,11 @@
...
+    "key": "geological",
+    "type": "defaultCheck",
+    "label": "Exposed Bedrock",
+    "strings": {
+        "options": {
+            "undefined": "No",
+            "outcrop": "Yes"
+        }
+    }
+
```
tyrasd, line 11: missing 
```suggestion
}
```

data/fields/geological_volcanic_lava_field.json
```diff
@@ -0,0 +1,11 @@
...
+    "key": "geological",
+    "type": "defaultCheck",
+    "label": "Lava Field",
+    "strings": {
+        "options": {
+            "undefined": "No",
+            "volcanic_lava_field": "Yes"
+        }
+    }
+
```
tyrasd, line 11: missing
```suggestion
}
```

data/presets/natural/rock.json
```diff
@@ -13,7 +20,13 @@
...
+        "attached rock",
+        "cairn",
+        "bedrock",
+        "outcrop",
+        "exposed rock",
+        "decorative boulder",
+        "landscape boulder"
     ],
-    "name": "Attached Rock / Boulder"
+    "name": "Rock"
```
westnordost, line 31: > Just "Rock" for natural=rock. This is the more common tag and people have been using it for a variety of types of rocks so the simple name seems OK as long as the terms are there.

Well this would lead to `natural=rock` being tagged (a lot) on free-standing non-attached rocks which seems to be wrong according to the previous discussion/research.

If I understand your point correctly, you are saying that this is what is happening anyway already a lot, so renaming the preset name to just "Rock" is merely following the de-facto tagging situation and **not** redefining the meaning of a tag. The latter, we absolutely do not want for obvious reasons that I hope do not need explaining.

What I am missing is proof for the claim that this is indeed the de-facto tagging situation and if you have proof, it should first be documented in the wiki. How much percent of `natural=rock` is not really a rock attached to bedrock and how much percent is on the other hand tagged correctly according to the wiki definition? Without this, we cannot simply shrug this distinction away in an iD preset PR because "it has never been tagged correctly anyway". After all, the very first two sentences of the wiki article mention this distinction:

> [natural](https://wiki.openstreetmap.org/wiki/Key:natural)=rock describes a notable rock feature or small group of rocks, attached to the underlying bedrock, mostly used on a single node element.
> 
> On the contrary, single boulders **not attached** to the bedrock are better tagged as [natural](https://wiki.openstreetmap.org/wiki/Key:natural)=[stone](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dstone).

... and iD (presets), too, has been denoting the meaning of `natural=rock` correctly according to the wiki definition since the preset had been added.

westnordost (changes requested):
See https://github.com/openstreetmap/id-tagging-schema/pull/471/files#r884296415

As it stands, this would change the definition of a well-documented tag.

westnordost:
Also note that the wiki definition of `natural=stone` is 

>A large (whatever that might mean) natural boulder or rock fragment that is not attached to the underlying terrain.

So, this includes more than a glacial erratic - any boulder or rock that is not attached to the underlying terrain is a `natural=stone`. This is why I proposed that you use the `aliases` field.

data/presets/natural/rock.json
bgo-eiu: Good point - I have a strong suspicion the de facto use is for any rock due to how the word is understood in English, but I will do some more formal queries on this for documentation first.

bgo-eiu:
I am going to look into:

* How is natural=stone used outside of Europe, where mappers likely did not understand the idea of a findling? It does seem like it follows "glacial erratic" quite closely within Europe, but if there is a noticeable pattern in other uses that could complicate things. In any case, it does seem worth updating the wiki in some way to explain the findling concept and link to your research on this to clarify things.
* How is natural=rock used generally/de facto?

And document the findings for this.

westnordost:
Great! Though, in regards to this, just a disclaimer that I am not the maintainer of the iD presets, @tyrasd is. 

In particular, I am not sure what happens if your finding is that indeed, many tagged `natural=rock`s are actually stones or the other way round: Whether the guideline for contributing to this repo would allow to disregard the (original, documented) definition of a tagging and replace it with a more general one. Maybe there is a precedent for that, I do not know.

tyrasd:
Great that you're going to look into this in more detail @bgo-eiu !

> In particular, I am not sure what happens if your finding is that indeed, many tagged `natural=rocks` are actually stones or the other way round: Whether the guideline for contributing to this repo would allow to disregard the (original, documented) definition of a tagging and replace it with a more general one.

In general I do abstain from modelling presets in a way that contradicts or otherwise mismatches the documentation on the wiki. If you find that the documentation on the wiki does in fact not match real-world usage of the tags, then we first need to find out why that is: If it is because of editor presets which are not conveying the intended meaning well enough, these presets should probably be improved/clarified. If it is because the tagging concept is not well enough defined or too complex for mappers, then the tagging needs to be redefined, optimally through a tagging proposal. It could also be the case that the wiki just contained some misleading information, then it would suffice to update the wiki.

data/presets/natural/rock.json
```diff
@@ -1,7 +1,14 @@
 {
     "icon": "temaki-boulder2",
     "fields": [
-        "name"
+        "name",
+        "height",
+        "geological_outcrop"
```
tyrasd, line 6: Something that irritates me a bit is that the proposed added field for `geological=outcrop` to the `natural=stone` preset currently only has very few usages (only 30 map features have this tag combination). The [wiki](https://wiki.openstreetmap.org/wiki/Tag%3Ageological%3Doutcrop) [also](https://wiki.openstreetmap.org/wiki/Tag%3Anatural=stone) doesn't mention this particular usage.

data/presets/natural/bare_rock.json
```diff
@@ -1,5 +1,8 @@
 {
     "icon": "temaki-boulder3",
+    "moreFields": [
+        "geological_volcanic_lava_field"
```
tyrasd, line 4: For the `geological=volcanic_lava_field` tag the combined usage with `natural=bare_rock` seems to be quite popular. Perhaps it could make sense to have this combination as a dedicated preset, so that it can be more easily found when mapping such features?

data/presets/natural/rock.json
matkoniecz: @bgo-eiu ping?

I have not looked much into PR in general, but I plan to close it as abandoned given no reaction to comment over 3 years. Just to confirm: is it abandoned?

data/fields/geological_outcrop.json
```diff
@@ -1,14 +1,11 @@
...
+    "key": "geological",
     "type": "defaultCheck",
-    "label": "Barrier",
-    "geometry": [
-        "vertex"
-    ],
+    "label": "Exposed Bedrock",
     "strings": {
         "options": {
             "undefined": "No",
```
matkoniecz, line 7: `geological=undefined` has literally zero uses, it should not be pushed by iD

https://taginfo.openstreetmap.org/keys/geological#values

matkoniecz (changes requested):

data/fields/geological_outcrop.json
matkoniecz: (have not checked other parts yet)

matkoniecz: oh no, that is a special value

ops, sorry - see https://github.com/ideditor/schema-builder?tab=readme-ov-file#strings

matkoniecz:
I will close this one as abandoned. There is sadly no linked issue so I will try to recover usable parts here, skipping adding support for basically unused tags.

I have general feeling that this PR tried to do multiple things at once and it should have avoided doing this. If anyone sees useful parts I missed, please create an issue/PR - ideally for self-contained change.

matkoniecz:
And I got confused almost immediately on https://wiki.openstreetmap.org/wiki/Talk:Tag:natural%3Drock#%22Consider_adding_geological=outcrop_to_natural=stone/natural=rock_where_it_is_attached_to_underlying_bedrock%22

PR #479, opened 2022-06-01
Add `lane_markings` field

arch0345:
This PR adds a [`lane_markings`](https://wiki.openstreetmap.org/wiki/Key:lane_markings) field in `moreFields` for Living Street, Residential/Unclassified, Service, and Tertiary Road presets.

### DIFF
--- a/data/fields/lane_markings.json
+++ b/data/fields/lane_markings.json
@@ -3,7 +3,7 @@
     "type": "check",
     "label": "Lane Markings",
     "terms": [
-        "centerline", 
+        "centerline",
         "striped",
         "unstriped"
     ]
--- a/data/fields/lane_markings.json
+++ b/data/fields/lane_markings.json
@@ -1,5 +1,5 @@
 {
-    "key": "lane:markings",
+    "key": "lane_markings",
     "type": "check",
     "label": "Lane Markings"
 }
--- a/data/fields/lane_markings.json
+++ b/data/fields/lane_markings.json
@@ -1,5 +1,10 @@
 {
     "key": "lane_markings",
     "type": "check",
-    "label": "Lane Markings"
+    "label": "Lane Markings",
+    "terms": [
+        "centerline", 
+        "striped",
+        "unstriped"
+    ]
 }
--- a/data/fields/lane_markings.json
+++ b/data/fields/lane_markings.json
@@ -1,0 +1,5 @@
+{
+    "key": "lane:markings",
+    "type": "check",
+    "label": "Lane Markings"
+}
--- a/data/presets/highway/living_street.json
+++ b/data/presets/highway/living_street.json
@@ -14,6 +14,7 @@
         "cycleway",
         "flood_prone",
         "junction_line",
+        "lane_markings",
         "lit",
         "maxheight",
         "maxweight_bridge",
--- a/data/presets/highway/living_street.json
+++ b/data/presets/highway/living_street.json
@@ -14,7 +14,6 @@
         "cycleway",
         "flood_prone",
         "junction_line",
-        "lane_markings",
         "lit",
         "maxheight",
         "maxweight_bridge",
--- a/data/presets/highway/primary.json
+++ b/data/presets/highway/primary.json
@@ -18,6 +18,7 @@
         "flood_prone",
         "incline",
         "junction_line",
+        "lane_markings",
         "lit",
         "maxheight",
         "maxspeed/advisory",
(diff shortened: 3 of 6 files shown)


data/fields/lane_markings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "lane:markings",
```
1ec5, line 2: This key appears to be used [only once in OSM](https://taginfo.openstreetmap.org/keys/lane:markings). [`lane_markings`](https://wiki.openstreetmap.org/wiki/Key:lane_markings) is heavily used though:

```suggestion
    "key": "lane_markings",
```

data/fields/lane_markings.json
```diff
@@ -0,0 +1,5 @@
...
+    "type": "check",
+    "label": "Lane Markings"
```
1ec5, line 4: I think the more common term in the U.S. is “Lane Dividers”, and the official term is “lane line”, which should be a `term`. But “Lane Markings” is reasonably intuitive.

The wiki page makes it sound like this key can also indicate whether the road has a painted centerline. If so, “centerline”, “striped”, and “unstriped” should all be `terms`.

data/fields/lane_markings.json
arch0345: Thanks for catching this!

tyrasd:
> Living Street, Residential/Unclassified, Service, and Tertiary Road presets.

Can you elaborate why exactly you chose these presets for this field?

<hr> 

For context, here's an analysis showing the current usage of this tag in combination with different `highway` values:

`highway`= | total count | count with `lane_markings` | percentage
--|--|--|--
`motorway` | 1,100,000 | 1,200 | 0.11 %
`trunk` | 1,500,000 | 3,800 | 0.25 %
`primary` | 3,200,000 | 17,400 | 0.54 %
`secondary` | 4,600,000 | 42,700 | 0.93 %
`tertiary` | 7,200,000 | 82,100 | 1.14 %
`unclassified` | 15,000,000 | 96,000 | 0.64 %
`residential` | 62,000,000 | 119,000 | 0.19 %
`living_street` | 1,600,000 | 1,700 | 0.11 %
`pedestrian` | 700,000 | 480 | 0.07 %
`service` | 48,000,000 | 7,000 |  0.01 %
`*_link` | 2,200,000 | 14,000 | 0.6 %
other<br>(`track`, `cycleway`, `construction`, …) |  | 1,000 | < 0.01 %

<hr>

Especially the usage on the `highway=service` preset seems to be pretty rare. To me it seems that it would make sense to limit it to the range between `unclassified` and `primary` (i.e. everything with > 0.5% usage above) or alternatively, to simply include this field for all `highway` presets.

bgo-eiu:
For what its worth, part of why I haven't tagged `lane_markings` on many service roads is because I only recently learned of the key.

While lane markings are less common on service roads compared to other types of roads, that is a reason why it is useful to be to tag them where they do exist. Yesterday, I mapped a 4 lane service way contained entirely within a parking lot which had land markings, and forgot to add `line_markings`. It's common enough that I encounter it from time to time, but uncommon enough that it would be quite useful to be able to identify which service roads are like this.

tyrasd:
For reference: currently, only 177 ways are tagged with `highway=service` + `lane_markings=yes` (the remaining 6.9k have `lane_markings=no`). I agree that the interesting cases would be the ones with the `yes` value. But I think these numbers are simply too low for the field to be also included in the Service Road preset. If the numbers change in the future, we can reconsider.

bgo-eiu:
Alright, I didn't realize they were that low. I'll try to remember to use it more

arch0345:
> How did you come up with the list of presets for this new field?

I initially picked these presets since in my experience these lower classifications commonly have instances with both lane markings and without while higher classifications always have lane markings.

Looking at the [current usage](https://overpass-turbo.eu/s/1jk4) of `highway=primary` + `lane_markings=no` however, it looks like there are a considerable amount of roads of these higher classifications that don't have road markings.

> I think these numbers are simply too low for the field to be also included in the Service Road preset. If the numbers change in the future, we can reconsider.

Alright, I'll add it to just primary-unclassified as you suggested for now.

tordans:
Just as a +1, this sounds good to me. 

I talked to @supaplexosm to hear his thoughts because he looked at this kind of data a lot for his detailed rendering (and mapping) of Neukölln, Berlin, Germany. For this area residential streets usually only have lane markings at the beginning of the street – if any. So it's good to have this tagging as "moreTags". It will show the details, but not mislead less experienced mappers to miss-tag the whole street. One example:
- https://strassenraumkarte.osm-berlin.org/?map=micromap#20/52.47928/13.44532
- https://www.openstreetmap.org/way/980711002#map=19/52.47929/13.44521

PR #481, opened 2022-06-01
make the multipolygon relation type searchable

k-yle:
When you create a new relation, it's very hard to find the option for Multipolygon. 

- The first issue is you need to scroll down a lot to find the most common option.
- The second issue is Multipolygon is the only relation preset that can't be found by searching. This is annoying if you're using keyboard shortcuts to navigate around. 

This PR fixes both issues. Multipolygon is now treated like any other relation preset. 

Current behaviour:

<table>
<tr>
	<td>
	<td>
</table>


k-yle:
`unsearchable: true` was added 9 years ago in https://github.com/openstreetmap/iD/commit/4bdd49b5, so that it doesn't appear alongside area-presets like building. `unsearchable` isn't needed if we make this preset only apply to relations. I've tested this in an old version of iD and it works well

PR #485, opened 2022-06-01
Add preset and field for sport=table_football

Binnette:
Hi, here is a proposition for https://wiki.openstreetmap.org/wiki/Tag:sport=table_soccer

But DO NOT MERGE IT yet. Cause I created it as a preset of leisure=pitch BUT the wiki do not say it should be "combined" with leisure=pitch. It seems this tag sport=soccer can be "alone" on a node without any other tag.

Can you please tell me how we should add it in the id-tagging-schema?

### DIFF
--- a/data/fields/sport_pub.json
+++ b/data/fields/sport_pub.json
@@ -1,0 +1,13 @@
+{
+    "key": "sport",
+    "type": "semiCombo",
+    "label": "Sports",
+    "strings": {
+        "options": {
+            "table_soccer": "Table football",
+            "billiards": "Billiards",
+            "darts": "Darts"
+        }
+    },
+    "autoSuggestions": false
+}
--- a/data/fields/sport_pub.json
+++ b/data/fields/sport_pub.json
@@ -4,7 +4,7 @@
     "label": "Sports",
     "strings": {
         "options": {
-            "table_soccer": "Table football",
+            "table_soccer": "Table Football",
             "billiards": "Billiards",
             "darts": "Darts"
         }
--- a/data/presets/amenity/bar.json
+++ b/data/presets/amenity/bar.json
@@ -25,6 +25,7 @@
         "phone",
         "ref/vatin",
         "smoking",
+        "sport_pub",
         "website",
         "wheelchair"
     ],
--- a/data/presets/amenity/pub.json
+++ b/data/presets/amenity/pub.json
@@ -27,6 +27,7 @@
         "phone",
         "ref/vatin",
         "real_fire-GB-IE",
+        "sport_pub",
         "website",
         "wheelchair"
     ],
--- a/data/presets/amenity/restaurant.json
+++ b/data/presets/amenity/restaurant.json
@@ -34,6 +34,7 @@
         "ref/vatin",
         "reservation",
         "smoking",
+        "sport_pub",
         "stars",
         "takeaway",
         "wheelchair",
--- a/data/presets/leisure/pitch/table_soccer.json
+++ b/data/presets/leisure/pitch/table_soccer.json
@@ -17,9 +17,10 @@
         "value": "table_soccer"
     },
     "terms": [
-        "table football",
+        "foosball",
+        "kicker",
         "table soccer",
-        "foosball"
+        "tecball"
     ],
-    "name": "Table football"
+    "name": "Table Football Table"
 }
--- a/data/presets/leisure/pitch/table_soccer.json
+++ b/data/presets/leisure/pitch/table_soccer.json
@@ -1,0 +1,25 @@
+{
+    "fields": [
+        "name",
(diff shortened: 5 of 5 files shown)


Binnette:
Hi @westnordost thanks for your comments. Note that my PR addess this issue DoFabien/OsmGo#108:

> I found quite a few table soccer tables in public parks (foosball, "babyfoot" in French), but the lack of a preset tag on Osm Go! makes it harder to tag them. I think the table tennis preset could be used, with "Lit" and "Access".

In this issue, @maroneze report table soccer directly in parks as a POI (so out of amenities, pub, etc). So @maroneze will more than probably map then directly as a node on the map.

Please let me know your point of view @westnordost 👍

maroneze:
> I have not seen in my life an outdoor table soccer table, I don't think this is a thing. But if there was such a thing, that would be the tag to use.

I had never seen it either (and I thought it unwise due to rain), but apparently that's a thing in some French towns (I've seen at least 3 in nearby towns). I didn't think of taking a photo (I will try to do it to submit it to the OSM wiki eventually), but if you google "babyfoot table parc" in Google Images ('babyfoot' is the French term for table soccer), you can see a few examples.

tordans:
JFYI, we have them in our hood.

Mapped by @supaplexosm
```
leisure	pitch
sport	table_soccer
```

* [Mapilary](https://www.mapillary.com/app/?pKey=566012754606324&focus=photo&lat=52.470427399972&lng=13.4369098&z=17&x=0.6493107158483779&y=0.5601067415642688&zoom=0)
* [Table 1](https://www.openstreetmap.org/node/5859458696), [Table 2](https://www.openstreetmap.org/node/5859458697)

Mapped by me
```
playground	tabletop_football
```
* [Mapillary](https://www.mapillary.com/app/?pKey=2997633503847353&focus=photo&lat=52.467912668729&lng=13.468986446241&z=17)
* [Table](https://www.openstreetmap.org/node/5745608440#map=16/52.4671/13.4736)

Hufkratzer:
> I have not seen in my life an outdoor table soccer table, I don't think this is a thing. 

Table soccer tables can of course also be mapped if they are indoors, like other kinds of pitches too. "indoor=yes" may be added in these cases, example https://www.openstreetmap.org/node/9551800192

tyrasd:
Currently, the tag is used surprisingly rarely: of the ~300 features which have the tag, around half to two thirds seem to be amenity features like pubs (which would benefit from an added `field`) and the rest seems to be individually mapped table soccer tables (which would benefit from a new `preset`).

These raw usage numbers wouldn't currently warrant the inclusion as a preset/field, but I would also assume that table soccer tables are quite significantly more common in real life compared what is currently mapped. Therefore, I would not object adding this tag as a preset and/or field.

For a field, it would probably be needed to use a `semiCombo` field type for the `sport` key and allow related "pub" sports like `billiards`, `darts`, etc. (you can use https://github.com/openstreetmap/id-tagging-schema/blob/v3.3.0/data/fields/sport_ice.json as a template).

Binnette:
Hi @tyrasd, please let me know how I should amend my PR 👍

Binnette:
TY @tyrasd, i just pushed the requested changes. Can you please review them 👍

tyrasd:
Looks good to me, thanks! :+1:  I only changed the preset's name to `Table Football Table`. The double _table_ sounds a bit funny, but I guess it is acceptable. Is that ok for you?

Binnette:
Hi @tyrasd I'm not a native english speaker, so I usually trust wikipedia to translate from french.

> Table football, also called foosball and table soccer, is a table-top game...

Source: https://en.wikipedia.org/wiki/Table_football

arch0345:
I think that this should be named `Foosball Table` instead of `Table Football Table` since this game is commonly called foosball in American English. `Table Football` should instead be used for the English (United Kingdom) locale.

For example, Amazon in the US refers to this game as 'foosball' while Amazon.co.uk calls this game 'table football':

Screenshot 2022-06-15 220908
Screenshot 2022-06-15 221221

Binnette:
So we just revert to "Table football" and we can merge?

tyrasd:
Merged with "Foosball Table" as the preset's name [and field value](https://github.com/openstreetmap/id-tagging-schema/commits/24d6ffc).

PS: I just noticed that there is no icon for the preset, but it's also not easy to find something that fits. There isn't really anything on maki, temaki and not even [fontawesome](https://fontawesome.com/search?q=foosball&s=solid%2Cbrands) when searching in fontawesome's collection). Svgrepo (see https://github.com/ideditor/schema-builder/issues/60) would [have one](https://www.svgrepo.com/svg/6456/table-football-cenit-view), but it's too detailed for the size we would display it at.

PR #490, opened 2022-06-04
Made target and country case sensitive

wvanderp:
the wiki page for [country tag](https://wiki.openstreetmap.org/wiki/Key:country) and  [target tag](https://wiki.openstreetmap.org/wiki/Key:target) stats that we should use ISO 3166-1 alpha-2 this is currently dificult in ID editor.

so I made the tags case sensitive and non snake case.

any feedback is appreciated.

### DIFF
--- a/data/fields/country.json
+++ b/data/fields/country.json
@@ -1,5 +1,7 @@
 {
     "key": "country",
     "type": "combo",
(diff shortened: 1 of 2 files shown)


tyrasd:
thanks!

PR #491, opened 2022-06-05
Remove `fields` from Main Entrance

arch0345:
Closes https://github.com/openstreetmap/iD/issues/9148
Removes `fields` from Main Entrance preset to match with `fields` in the generic Entrance preset

### DIFF
--- a/data/presets/entrance/main.json
+++ b/data/presets/entrance/main.json
@@ -4,7 +4,9 @@
         "ref",
         "door",
         "access_simple",
-        "level"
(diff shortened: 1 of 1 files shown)


arch0345:
Good point @kjonosm. I removed `fields` so that it'll use the fields listed in the generic Entrance preset:
https://github.com/openstreetmap/id-tagging-schema/blob/0e2138f85737666a58f6da313dc0f565573f7225/data/presets/entrance.json#L3-L11

PR #499, opened 2022-06-16
Add preset amenity=driver_training

SteveLz:
Wiki page: [amenity=driver_training](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddriver_training)


data/presets/amenity/driver_training.json
```diff
@@ -0,0 +1,29 @@
...
+    "icon": "maki-car",
+    "fields": [
+        "name",
+        "operator",
+        "operator/type",
+        "address",
+        "opening_hours"
+    ],
+    "moreFields": [
+        "license_classes",
```
tyrasd, line 11: a field for [this tag](https://wiki.openstreetmap.org/wiki/Key:license_classes)  doesn't exist yet. you would need to create one in `data/fields/` for it.

tyrasd (changes requested):
see inline comments

data/presets/amenity/driver_training.json
```diff
@@ -0,0 +1,32 @@
...
+        "point",
+        "area"
+    ],
+    "tags": {
+        "amenity": "driver_training"
+    },
+    "name": "Driver Training Grounds",
+    "aliases": [
+        "Driving Safety Center"
+        "Driving School"
```
tyrasd, line 30: hm, I would not add "Driving School" as an alias here, as this would collide with the preset for `amenity=driving_school` (note that [`aliases`](https://github.com/ideditor/schema-builder#aliases) are meant for synonyms).

maybe we could include "diving school ground" as a search [term](https://github.com/ideditor/schema-builder#terms)/keyword:

```suggestion
    ],
    "terms": [
        "Driving School"
```

data/fields/license_classes.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "license_classes",
+    "type": "combo",
```
tyrasd, line 3: I think the type `semiCombo` would be a better fit for this field, since typically more than one license class.

```suggestion
    "type": "semiCombo",
```

tyrasd:
Thanks for the adjustments!

Only one question: I only found one `amenity=driver_training` object on OSM which also had a `license_classes` tag mapped. I saw that the [wiki page](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddriver_training) mentions its usage, but it clearly doesn't match the practical use. I would like to drop the field from the this preset until usage numbers are reasonable. But let's add this field to the [`amenity=driving_school`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddriving_school) preset, shall we?

SteveLz:
The `license_classes` are rarely tagged because it is not easy to get the exact value, and even people who have studied in it can't necessarily give a complete value to the driving school. But in East Asia, driving schools basically exist in the form of grounds, and the tag for driving schools is also applicable to `driver_training`, so I think it is necessary to keep it. Of course, I also agree to add this value to the [`amenity=driving_school`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddriving_school).

SteveLz:
The main teaching tasks of this kind of driving school are completed in the training ground. The offices and rooms for theoretical lessons are one or more buildings located next to the training ground within the driving school walls. So if `amenity=driving_school` does not apply to this kind of enclosure, maybe it can only be represented by `amenity=driver_training`.

SteveLz:
> For the "actual" driving school's facilities (i.e. offices, rooms for theoretical lessons, etc.), one would additionally also map a point or area with `amenity=driving_school`. And there the `license_classes` information can be mapped.

Indoor facilities and outdoor venues used for teaching usually belong to the same driving school, and according to the "one feature, one OSM element" principle, they should not be separated.

> I saw that the [wiki page](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddriver_training) mentions its usage, but it clearly doesn't match the practical use.

I've added a dozen `license_classes=*`, but there aren't even as many existing tags in East Asia as I've tagged.
So just a small number doesn't mean it shouldn't be used here, otherwise `license_classes=*` shouldn't be used in driving schools in East Asia.

I know it can be a little difficult trying to understand something that doesn't exist locally, you can take a look at [an example driving school](https://www.openstreetmap.org/way/407675704) that appears on the wiki page and you may find that just using buildings to represent driving schools is not appropriate.

SteveLz:
Considering that this form of driving school is not universal, it may not be appropriate to show this tag to every user. Is it possible to show it only to users in specific countries?

Dimitar5555:
> Indoor facilities and outdoor venues used for teaching usually belong to the same driving school, and according to the "one feature, one OSM element" principle, they should not be separated.

It depends on the country. I know at least 3 outdoor venues in my city which are open for everyone. Only one is used by a single company but it's free to use by anyone if you don't interrupt them (it's located in the parking lot of a big box store).

tyrasd:
> according to the "one feature, one OSM element" principle, they should not be separated.

I cannot follow this argument. From what I can see, this particular driving school consists of different features: The buildings, the driving practice area, the parking lots, and so on. The part of the site which is dedicated to the driver training could easily be mapped as a separate feature inside the perimeter of the `driving_school`. Similar to how a playground of a school is typically mapped as a separate area inside of the school area (see [schematic](https://wiki.openstreetmap.org/w/images/8/86/Amenity_school_usage_example_single_school_-_Proposal.svg)). This doesn't contradict the one feature, one OSM element principle.

SteveLz:
> > Indoor facilities and outdoor venues used for teaching usually belong to the same driving school, and according to the "one feature, one OSM element" principle, they should not be separated.
> 
> It depends on the country. I know at least 3 outdoor venues in my city which are open for everyone. Only one is used by a single company but it's free to use by anyone if you don't interrupt them (it's located in the parking lot of a big box store).

Yes, they may not be common outside of Asia, so I put it in "moreFields" instead of "fields". This will not affect most users, and users who need it can easily add it.

SteveLz:
> > according to the "one feature, one OSM element" principle, they should not be separated.
> 
> I cannot follow this argument. From what I can see, this particular driving school consists of different features: The buildings, the driving practice area, the parking lots, and so on. The part of the site which is dedicated to the driver training could easily be mapped as a separate feature inside the perimeter of the `driving_school`. Similar to how a playground of a school is typically mapped as a separate area inside of the school area (see [schematic](https://wiki.openstreetmap.org/w/images/8/86/Amenity_school_usage_example_single_school_-_Proposal.svg)). This doesn't contradict the one feature, one OSM element principle.

Yes, but if it works, I think this tag will not be widely used so soon. Because `driving_school` is considered an indoor facility and doesn't include an outdoor facility (see [wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddriver_training) and [talk](https://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Ddriver_training) page), and [carto](https://github.com/gravitystorm/openstreetmap-carto/issues/4101) does not provide rendering of the areas because of this.
If you think a `driving_school` can exist as a ground, can you post your opinion on the talk page? I think this will help us find a right direction.

tyrasd:
Thanks for the additional links. :+1: I'm still a bit surprised that `amenity=driving_school` should be limited indoor facilities of a driving school, but OK :shrug: In my humble opinion it would make sense to map driving schools just like regular schools: as the whole area of the school including the indoor facilities (buildings) and surrounding associated outdoor facilities, maybe the fact that in some countries driving schools have such associated `driver_training` areas was just overlooked, because they don't exist in Europe, North America, and other places?

tyrasd (approved):

data/fields/license_classes.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "license_classes",
+    "type": "semiCombo",
+    "label": "Driving License Classes"
```
1ec5, line 4: “Driving license” is understandable but not quite correct in any dialect of English. The source string should be changed to “driver’s license”, per American English. (The British English term would be “driving licence”; the Canadian, Australian, and New Zealand English localizations would have their own translations as well.)

1ec5: Apart from that, “driver’s license classes” is confusing in the context of a preset for driver training. At first, when I tried to translate this string, I thought it meant “classes for obtaining a driver’s license”, but this key actually means “classes of driver’s license”. On the Driver Training Grounds preset, a user might end up leaving this field blank because no classes are taught there. I’d suggest changing this string to “License Classifications” or “Classes of Driver’s License” to avoid confusion.

1ec5 (commented):
> I'm still a bit surprised that `amenity=driving_school` should be limited indoor facilities of a driving school, but OK 🤷 In my humble opinion it would make sense to map driving schools just like regular schools: as the whole area of the school including the indoor facilities (buildings) and surrounding associated outdoor facilities, maybe the fact that in some countries driving schools have such associated `driver_training` areas was just overlooked, because they don't exist in Europe, North America, and other places?

Just another data point in case it helps:

My driving “school” in Ohio consisted of two parts: classroom instruction in a [high school](https://www.openstreetmap.org/way/23702377) classroom rented out for the weekend, plus on-road sessions with a driving instructor that started and ended at [an office in a little strip mall](https://www.openstreetmap.org/node/5110361445). I didn’t spend any time at the office; the instructor had me take them on personal errands around town. Some high schools partner with third parties to offer driver training as an amenity, but there’s no dedicated training “area”  to speak of.

My driving test consisted of three parts: first a written exam at [this exam station](https://www.openstreetmap.org/node/2930154102), followed by a maneuverability test on [this normally empty parking lot](https://www.openstreetmap.org/way/31875612), followed by an on-road test going [around this little downtown](https://www.openstreetmap.org/#map=17/39.27062/-84.41198). You could call the second part a driver training area, since there’s always someone there either taking their test or practicing for one, but it’s just part of a parking lot. I’d be more inclined to model it as a member in a site relation, since the complex is shared with other facilities including a police station.

At one point, my insurance company offered a discount for taking a voluntary course on defensive driving. So I returned to the aforementioned high school, where an instructor had rented out [an empty parking lot](https://www.openstreetmap.org/way/182322801) for the weekend.

I have seen dedicated driver training facilities, especially for motorcycles in California, where that’s a popular option, but I rarely if ever see any offices or buildings associated with them. Physically, it’s just a paved lot with specialized markings on it.

data/fields/license_classes.json
tyrasd: changed the spelling in 7f0b1e4

PR #501, opened 2022-06-17
add more fields to the wilderness_hut preset

k-yle:
adds the fields `kitchen`, `mattress`, `drinking_water`, `capacity`, and `access` to the `tourism=wilderness_hut` preset. 

I also moved `address` into `moreFields` since only 0.9% of wilderness_hut's have an address. This makes sense since - by definition - wilderness huts are located in the _wilderness_, far away from streets so they tend not to have addresses.

### DIFF
--- a/data/presets/tourism/wilderness_hut.json
+++ b/data/presets/tourism/wilderness_hut.json
@@ -11,16 +11,16 @@
         "fireplace"
     ],
     "moreFields": [
-        "kitchen",
-        "mattress",
-        "drinking_water",
+        "address",
         "capacity",
+        "drinking_water",
         "gnis/feature_id-US",
         "internet_access",
         "internet_access/fee",
         "internet_access/ssid",
+        "kitchen",
+        "mattress",
         "reservation",
-        "address",
         "wheelchair"
     ],
     "geometry": [
--- a/data/presets/tourism/wilderness_hut.json
+++ b/data/presets/tourism/wilderness_hut.json
@@ -3,19 +3,24 @@
     "fields": [
         "name",
         "operator",
-        "address",
         "building_area",
+        "access_simple",
         "fee",
         "payment_multi_fee",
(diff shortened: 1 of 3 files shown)


data/fields/vhf.json
```diff
@@ -1,6 +1,6 @@
 {
     "key": "vhf",
-    "type": "text",
```
tyrasd, line 3: this change is unrelated to the wilderness hut preset, isn't it?

I'm wondering if this was made a text field on purpose because of the "special" usage as double digit numbers [wiki](https://wiki.openstreetmap.org/wiki/Key%3Avhf):

> Some people prefer for the channel numbers below 10 a leading zero (0), for instance "09".

tyrasd: (ping?)

k-yle: sorry, that change wasn't meant to be part of this PR. I tried to remove that change from a different branch and obviously messed something up...

data/presets/tourism/wilderness_hut.json
```diff
@@ -3,18 +3,23 @@
...
+        "access_simple",
         "fee",
         "payment_multi_fee",
         "charge_fee",
         "fireplace"
     ],
     "moreFields": [
+        "address",
+        "capacity",
+        "drinking_water",
```
tyrasd, line 16: this field is currently labeled _Drinkable_. It could be slightly confusing to see that come up in the context of a hut. It's probably necessary to create a dedicated version of the `drinking_water` field with a more generic label (e.g. `Drinking Water Available`).

tyrasd: //edit: the `tourism=camp_site` preset could also benefit from such a field

tyrasd: I've now implemented this in 2e3825c

PR #502, opened 2022-06-17
allow the type of fuel in a tank to be specified

k-yle:
For a storage tank, if you select `content=fuel` a new field will appear allowing you to specify the type of fuel in the tank.

### DIFF
--- a/data/fields/fuel_multi_2.json
+++ b/data/fields/fuel_multi_2.json
@@ -1,0 +1,9 @@
+{
+    "key": "fuel:",
+    "type": "multiCombo",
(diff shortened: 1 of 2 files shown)


tyrasd:
Ehrm… is this really a good tagging? This would allow tagging of multiple different fuel types for a single tank. :thinking: 

This tagging doesn't seem to be used very often in practice: If I see this correctly, out of the ~13000 storage tanks with `content=fuel`, only ~40 have a `fuel:*` subtag (and ~70 have a `fuel` tag).

PR #506, opened 2022-06-18
Add more barrier presets

### LINKED ISSUE #505: Support barrier=jersey_barrier
https://wiki.openstreetmap.org/wiki/Tag:barrier=jersey%20barrier?uselang=en

https://www.openstreetmap.org/edit?editor=id&node=9828269071#map=19/42.70637/-71.15377

screen

arch0345:
This PR adds presets for the following barriers listed on the wiki:
- [`barrier=bump_gate`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dbump_gate)
- [`barrier=bus_trap`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dbus_trap)
- [`barrier=cable_barrier`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dcable_barrier)
- [`barrier=full-height_turnstile`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dfull-height_turnstile)
- [`barrier=hampshire_gate`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dhampshire_gate)
- [`barrier=jersey_barrier`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Djersey_barrier)
- [`barrier=motorcycle_barrier`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dmotorcycle_barrier)
- [`barrier=planter`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dplanter)
- [`barrier=rope`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Drope)
- [`barrier=wicket_gate`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dwicket_gate)

I used the default barrier icon for most of these, but ideally those should get their own icons. For example, `barrier=jersey_barrier` could have an icon similar to this image depicting the general outline:
<a title="Public domain, via Wikimedia Commons" href="https://commons.wikimedia.org/wiki/File:Jersey_barrier_2.png">Jersey barrier 2</a>
The preset for `full-height_turnstile` could have an icon similar to the one used in OSM carto:
<a title="Public domain, via OpenStreetMap Wiki" href="https://wiki.openstreetmap.org/w/images/thumb/b/bb/Full-height_turnstile-14.svg">Jersey barrier 2</a>

Since the wiki article for [`barrier=bus_trap`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dbus_trap) recommends also using [`maxwidth`](https://wiki.openstreetmap.org/wiki/Key:maxwidth), I created a field for it using the same field type used for `maxheight` since their values are similar.

I also added [`barrier=ditch`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dditch), [`barrier=handrail`](https://wiki.openstreetmap.org/wiki/Tag%3Abarrier%3Dhandrail), and `barrier=log` as unsearchable presets. For [`barrier=ditch`](https://taginfo.openstreetmap.org/tags/barrier=ditch), [`waterway=ditch`](https://taginfo.openstreetmap.org/tags/waterway=ditch) has a lot more uses and having making both of these searchable might be confusing for some users. The wiki article for [`barrier=handrail`](https://wiki.openstreetmap.org/wiki/Tag%3Abarrier%3Dhandrail) mentions that there's an "Unclear difference from fence". Although `barrier=log` is an approved tag, I made it unsearchable to prevent mappers from adding fallen trees that don't obstruct a passage.

Closes #505


danieldegroot2:
`barrier=planter` should not be confused with `man_made=planter`;
> See also [man_made](https://wiki.openstreetmap.org/wiki/Key:man_made)=[planter](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dplanter) One object can be tagged both with [man_made](https://wiki.openstreetmap.org/wiki/Key:man_made)=[planter](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dplanter) and [barrier](https://wiki.openstreetmap.org/wiki/Key:barrier)=planter, planters that are not barriers should be tagged with just [man_made](https://wiki.openstreetmap.org/wiki/Key:man_made)=[planter](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dplanter).

data/presets/barrier/hampshire_gate.json
```diff
@@ -0,0 +1,13 @@
...
+    "fields": [
+        "access"
+    ],
+    "geometry": [
+        "vertex"
+    ],
+    "tags": {
+        "barrier": "hampshire_gate"
+    },
+    "name": "Hampshire Gate"
```
tyrasd, line 12: Should we add `New Zealand Gate` and `Wire Gate` as aliases or terms (from https://en.wikipedia.org/wiki/Hampshire_gate)?

data/presets/barrier/motorcycle_barrier.json
```diff
@@ -0,0 +1,13 @@
+{
+    "icon": "maki-roadblock",
+    "fields": [
+        "access"
+    ],
```
tyrasd, line 5: should we add `wheelchair` as suggested by the wiki?

```suggestion
    ],
    "moreFields": [
        "wheelchair"
    ],
```

data/presets/barrier/rope.json
```diff
@@ -0,0 +1,17 @@
+{
+    "icon": "temaki-rope_fence",
+    "fields": [
+        "access"
+    ],
+    "moreFields": [
+        "wheelchair"
```
tyrasd, line 7: I wouldn't oppose this optional field, but interestingly, currently there are zero features mapped with `barrier=rope` + `wheelchair=*`. :shrug:

data/presets/barrier/wicket_gate.json
```diff
@@ -0,0 +1,25 @@
...
+        "wheelchair",
+        "opening_hours",
+        "height",
+        "material"
+    ],
+    "moreFields": [
+        "colour",
+        "manufacturer",
+        "operator",
+        "ref"
```
tyrasd, line 19: All of these fields (except `access`), currently only have zero or one usage worldwide (out of 2k features matching the preset). :thinking:

data/presets/barrier/rope.json
arch0345: I just added it as it's listed as a useful combination [on the wiki](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Drope). I wouldn't be opposed to removing it though.

data/presets/barrier/wicket_gate.json
arch0345: I just copy-pasted the fields from the Gate preset since I thought that the same fields could apply to wicket gates. I've narrowed down the fields to just the ones listed [on the wiki](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dwicket_gate) (`access`, `entrance`, and `opening_hours` (however it looks like `opening_hours` currently [only has 5 uses](https://overpass-turbo.eu/s/1jvZ).

arch0345:
> See also [man_made](https://wiki.openstreetmap.org/wiki/Key:man_made)=[planter](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dplanter) One object can be tagged both with [man_made](https://wiki.openstreetmap.org/wiki/Key:man_made)=[planter](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dplanter) and [barrier](https://wiki.openstreetmap.org/wiki/Key:barrier)=planter, planters that are not barriers should be tagged with just [man_made](https://wiki.openstreetmap.org/wiki/Key:man_made)=[planter](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dplanter).

Because of this, I decided to change the preset to use `man_made=planter` and added a field (`planter_barrier`) to add `barrier=planter`, similar to how the `crossing_raised` field adds `traffic_calming=table` to the crossing presets.

tyrasd:
> Because of this, I decided to change the preset to use `man_made=planter` and added a field (`planter_barrier`) to add `barrier=planter`, similar to how the `crossing_raised` field adds `traffic_calming=table` to the crossing presets.

That sounds like a good approach :+1: 

But could you please rename the field to `data/fields/barrier_planter.json` (the first part of the field's file name should be the tag name it operates on)?

data/presets/barrier/rope.json
tyrasd: ok, then let's remove it for now

data/presets/barrier/wicket_gate.json
tyrasd: I'd say if there are only so few features with this combination, it's probably better to drop the field for now.

tyrasd:
> planter preset

I just noticed that, unfortunately, this approach doesn't allow us to show the `access` field for planter objects when they the `barrier=planter` tag is set. In 87e795f I've now added a non-searchable preset called `Planter (Barrier)` (for vertex nodes only) which affords this. Do you think that solution is acceptable?

arch0345:
Good thinking, looks good to me

data/presets/barrier/full-height_turnstile.json
```diff
@@ -0,0 +1,17 @@
...
+        "vertex"
+    ],
+    "terms": [
+        "baffle gate",
+        "turnstyle"
+    ],
+    "tags": {
+        "barrier": "full-height_turnstile"
+    },
+    "name": "Full-height Turnstile"
```
1ec5, line 16: Should be capitalized as “Full-Height Turnstile”.

data/presets/barrier/wicket_gate.json
```diff
@@ -0,0 +1,17 @@
...
+        "line"
+    ],
+    "fields": [
+        "access"
+    ],
+    "moreFields": [
+        "entrance"
+    ],
+    "tags": {
+        "barrier": "wicket_gate"
```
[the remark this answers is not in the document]
tyrasd: As far as I understand, the term refers only to a _secondary_ _pedestrian_ gate near (or in[^1]) a larger gate.

[^1]: I would say that in OSM the `barrier=wicket_gate` probably only applies to the situation where the wicket is actually a separate entrance (as in this [example](https://de.wikipedia.org/wiki/Schlupfpforte#/media/Datei:Burg_Friedestrom_Suedtor_02.jpg)), and not for wicket doors which are built directly into a large gate ([example](https://de.wikipedia.org/wiki/Schlupfpforte#/media/Datei:Alsfeld-Neurathhaus-Tor_der_Einfahrt.jpg)). The latter should probably be modeled as an attribute of the main gate (e.g. `barrier=gate` + `wicket=yes`), I assume.

RudyTheDev:
> Although `barrier=log` is an approved tag, I made it unsearchable to prevent mappers from adding fallen trees that don't obstruct a passage.

Perhaps it is better to rename it to something obvious like "Barrier (Fallen Tree)" rather than hide it from the search? As someone who surveys a lot of forests, this is a common item to add on paths.

PR #508, opened 2022-06-22
Clean up crossing preset terms

1ec5:
Moved terms related to `traffic_calming=table` from the “Unmarked Crossing”, “Marked Crosswalk”, and “Crossing With Pedestrian Signals” presets to the “Raised” field. These terms only made sense for the raised crossing presets, which got merged with the unraised crossing presets in 4644b468ac5a9c1fc3d1521792e4265feb5a21b4 along with introducing the “Raised” field.

/ref https://github.com/openstreetmap/id-tagging-schema/pull/368#issuecomment-1161544736


tyrasd:
merged as 18ad9ea1dfe269d580fde5d8c14fabd666d2c944

PR #509, opened 2022-06-22
[regional] add support for moped ramps in the NL

### LINKED ISSUE #484: Cycleways with `bicycle=no` but also `mofa=yes|designated` or `moped=yes|designated` should not raise a warning
There is check in `data/deprecated.json` for the combination of `bicycle=no` and `highway=bicycle`:

```json
{
    "old": {"highway": "cycleway", "bicycle": "no"},
    "replace": {"highway": "path", "bicycle": "no"}
},
```

This makes sense, and would usually point to a mapping error. However, in the Netherlands this raises this warning for a common feature on Dutch roads.

We have three variants of cycleways in our legal code. Two of them are relevant here: the `G11` which is mandatory for `bicycle` and `mofa`, but forbidden for `moped`, and the `G12a`, which is mandatory for all three of these road user categories. When a cycleway switches from `G11` to `G12a` or vice versa, mopeds use small links to move to or from the main road:

Screenshot from 2022-06-01 17-01-31

We've settled on the following tagging scheme for these links:

```
highway          cycleway
bicycle          no
foot             no
mofa             no
moped            designated
oneway           yes
traffic_sign     NL:D104 (or) NL:D103
```

[This is the link shown on that photograph on the map.](https://www.openstreetmap.org/way/731968881)

These are still `highway=cycleway` in the Dutch context, because mopeds are one of the designated road user classes for cycleways in the Netherlands (provided it is signed with a `G12a` sign).

Is there a way to exclude these from the deprecation warning? That is, keep the check, but don't suggest it if either `mofa` or `moped` is set to a permissive value?

tyrasd:
closes #484

* adds NL-specific preset for "moped links" (highway=cycleway + bicycle=no + moped=designated)
* replace deprecation rule for `highway=cycleway` + `bicycle=no` with special "replacement" preset(s) (which excludes the NL)


tyrasd:
this works, but it's slightly questionable if it is works to keep around two extra presets (plus all related translations) just to make the deprecation rule for  `highway=cycleway` + `bicycle=no` work outside of the NL.

For reference: Currently, there exist only ~160 features (amounting to 53km) outside of the NL which have this tag combination.

image

jdhoek:
What are the alternatives?

I would guess:

* Keep the deprecation, accept lots of false positives in the Netherlands
* Remove deprecation globally until more fine-grained rules are possible

jdhoek:
Issue #484 may of course exist in other countries as well, but I don't know if this is the case.

tyrasd:
I think the most reasonable alternative would be to 

> Remove deprecation globally until more fine-grained rules are possible

One could even make the argument that having the validator rule which transforms such features into a `highway=path` object is even counterproductive: Just upgrading the tags potentially hides inconsistencies/errors in the data. For example, https://www.openstreetmap.org/way/217901656 might actually better be tagged as `highway=footway` + `footway=sidewalk` (and need an update of the `width` value) when looking at a nearby [mapillary picture](https://www.mapillary.com/app/?pKey=806722490281026&focus=photo). An iD user could have simply _"upgraded"_ the tags to `highway=path` + `bicycle=no` (and left the `width` at 2.5m), leaving the map in a sub-optimal state.

PR #516, opened 2022-06-27
Add 2 vending machines presets

Kazing:
Hello,

I would like to add two new presets for vending machines I regularly see & map.
They have with same use level as other presets already in place: 
https://taginfo.openstreetmap.org/keys/vending#values

* vending=bicycle_tube (1180 uses)
* vending=pizza (317 uses)


data/presets/amenity/vending_machine/bicycle_tube.json
```diff
@@ -0,0 +1,21 @@
...
+    ],
+    "tags": {
+        "amenity": "vending_machine",
+        "vending": "bicycle_tube"
+    },
+    "reference": {
+        "key": "vending",
+        "value": "bicycle_tube"
+    },
+    "name": "Bicycle inner tube Vending Machine"
```
tyrasd, line 20: Please use title case for the preset name

```suggestion
    "name": "Bicycle Inner Tube Vending Machine"
```

PR #520, opened 2022-07-08
Add Truck Speed Limit field

arch0345:
This PR adds a field for [`maxspeed:hgv`](https://wiki.openstreetmap.org/wiki/Key:maxspeed:hgv), which I've added to the Motorway, Trunk Road, and Primary Road presets.

### DIFF
--- a/data/fields/maxspeed/hgv.json
+++ b/data/fields/maxspeed/hgv.json
@@ -1,0 +1,6 @@
+{
+    "key": "maxspeed:hgv",
+    "type": "roadspeed",
+    "label": "Truck Speed Limit",
+    "placeholder": "40, 50, 60..."
+}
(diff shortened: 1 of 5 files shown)


tyrasd:
would you like to add this field also to the _US+Canada_ regional version of the `primary_link` preset? [`primary_link-US-CA.json`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/highway/primary_link-US-CA.json)

arch0345:
> would you like to add this field also to the _US+Canada_ regional version of the `primary_link` preset? [`primary_link-US-CA.json`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/highway/primary_link-US-CA.json)

Most primary links in the US/Canada only have an advisory speed limit. Currently there are only 2 primary links in the US/Canada with `maxspeed:hgv`: https://overpass-turbo.eu/s/1k3g

image

arch0345:
Whoops, just realized I added it to the Primary Link preset instead of the Primary Road preset.

PR #521, opened 2022-07-09
Add Bike and Ride field

arch0345:
Adds a field for [`bike_ride`](https://wiki.openstreetmap.org/wiki/Key%3Abike_ride), which I added to the Bicycle Parking preset. The wiki says that this tag is to be used to indicate that a bicycle parking area is near a public transport stop, similar to how the `park_ride` key is used for parking lots. Currently, this [`bike_ride`](https://taginfo.openstreetmap.org/keys/?key=bike_ride)  is used around 4K times.

Here's an example of a Bike & Ride shed in Portland:
<a></a>

### DIFF
--- a/data/fields/bike_ride.json
+++ b/data/fields/bike_ride.json
@@ -1,0 +1,5 @@
+{
+    "key": "bike_ride",
+    "type": "check",
+    "label": "Bike and Ride"
+}
--- a/data/presets/amenity/bicycle_parking.json
+++ b/data/presets/amenity/bicycle_parking.json
@@ -12,6 +12,7 @@
         "charge_fee"
     ],
     "moreFields": [
+        "bike_ride",
         "colour",
         "indoor",
         "level",


tordans:
I wonder, what this tag will tell us. I am not aware of any usage by data consumers. And I have a hard time to come up with a relevant usage. „Nearby public transport“ is very vague. And can be derived from the location.  For park-n-ride the tag is more specific because those parking spaces are signed (right?). But for bikes, any open access parking nearby is basically bike-ride.

tyrasd:
The tag seems to be particularly well used in Germany (see [taginfo](https://taginfo.openstreetmap.org/keys/?key=bike_ride#map)). Presumably this term (and thus the tag) was popularized by the "Bike + Ride" initiative by DB (https://bikeandride.bahnhof.de/bikeandride)?

tyrasd (approved):

tyrasd:
> I wonder, what this tag will tell us.

I would have assumed that this tag could be useful to mark bike parkings which are especially signed and/or equipped to host a typical multimodal commuter which wants to park a bike safely for long-ish periods: Such a bike-and-ride parking should have sufficiently many parking spots, be in a safe location, perhaps protected from the elements, etc. 

But the description on the wiki does ineed lack a bit of detail in this regard… :thinking:

PR #522, opened 2022-07-09
Add grave preset

### LINKED ISSUE #156: support cemetery=grave
I would be happy to make PR but from what I see this repository seems inactive right now and some of my PRs are waiting

arch0345:
Adds a preset for [`cemetery=grave`](https://wiki.openstreetmap.org/wiki/Tag%3Acemetery%3Dgrave)

Closes #156

### DIFF
--- a/data/presets/cemetery/grave.json
+++ b/data/presets/cemetery/grave.json
@@ -15,13 +15,6 @@
         "cemetery": "grave"
     },
     "terms": [
-        "burial ground",
-        "cemetary",
-        "cemetery",
-        "churchyard",
-        "columbarium",
-        "grave yard",
-        "graveyard",
         "mausoleum",
         "tomb"
     ],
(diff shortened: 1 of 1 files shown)


data/presets/cemetery/grave.json
```diff
@@ -0,0 +1,29 @@
...
+    "terms": [
+        "burial ground",
+        "cemetary",
+        "cemetery",
+        "churchyard",
+        "columbarium",
+        "grave yard",
+        "graveyard",
+        "mausoleum",
+        "tomb"
```
tyrasd, line 26: I wouldn't just copy the terms of the Graveyard preset here.

Would it be better to only keep the terms which match an individual grave (and remove those terms from the `amenity=grave_yard` preset)?

```suggestion
        "mausoleum",
        "tomb"
```

PR #528, opened 2022-07-12
add two popular building values

matkoniecz:
https://taginfo.openstreetmap.org/tags/building=outbuilding#overview

https://taginfo.openstreetmap.org/tags/building=allotment_house#overview

building=outbuilding is popular, but this is fueled by imports from PRG in Poland, not an organic growth (at least as of 2022-07), so this unspecific value is unsearchable

@tyrasd Would you be interested in listing of popular unsupported values for say `shop` `amenity` `barrier` `leisure` `natural` (with adjusted thresholds, lower than 100k used for buildings)?

I believe that JOSM has similar script that is listing popular tags (keys?) and they maintain list of popular tags which are deliberately not supported.

### DIFF
--- a/data/presets/building/_outbuilding.json
+++ b/data/presets/building/_outbuilding.json
@@ -1,8 +1,5 @@
 {
     "icon": "maki-building",
-    "fields": [
-        "{building}"
-    ],
     "geometry": [
         "area"
     ],
--- a/data/presets/building/_outbuilding.json
+++ b/data/presets/building/_outbuilding.json
@@ -1,0 +1,15 @@
+{
+    "icon": "maki-building",
+    "fields": [
+        "{building}"
+    ],
+    "geometry": [
+        "area"
+    ],
+    "tags": {
+        "building": "outbuilding"
+    },
+    "matchScore": 0.5,
+    "name": "Outbuilding",
+    "searchable": false
+}
--- a/data/presets/building/allotment_house.json
+++ b/data/presets/building/allotment_house.json
@@ -1,0 +1,14 @@
+{
+    "icon": "maki-building",
+    "fields": [
+        "{building}"
+    ],
+    "geometry": [
+        "area"
+    ],
+    "tags": {
+        "building": "allotment_house"
+    },
+    "matchScore": 0.5,
+    "name": "Allotment house"
+}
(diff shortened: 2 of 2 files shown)


data/presets/building/allotment_house.json
```diff
@@ -0,0 +1,14 @@
...
+        "{building}"
+    ],
+    "geometry": [
+        "area"
+    ],
+    "tags": {
+        "building": "allotment_house"
+    },
+    "matchScore": 0.5,
+    "name": "Allotment house"
```
tyrasd, line 13:
```suggestion
    "name": "Allotment House"
```

data/presets/building/_outbuilding.json
```diff
@@ -0,0 +1,15 @@
+{
+    "icon": "maki-building",
+    "fields": [
+        "{building}"
+    ],
```
tyrasd, line 5: This is not necessary, as the fields will be inherited from the parent preset by default.

```suggestion
```

data/presets/building/allotment_house.json
```diff
@@ -0,0 +1,14 @@
+{
+    "icon": "maki-building",
+    "fields": [
+        "{building}"
+    ],
```
tyrasd, line 5:
```suggestion
```

tyrasd (changes requested):
some minor tweaks (see below)

data/presets/building/allotment_house.json
tyrasd: Do you remember which one did you use as a template? I checked a few of the existing building presets and didn't spot this issue.

matkoniecz: `allotment_house.json` and `outbuilding.json` - several others have this field and extra ones - it is likely necessary to repeat `{building}` when one adds extra fields, right?

(checked with `rg --context 3 "\{building\}"` using ripgrep)

tyrasd: correct: when you want to add extra fields to a sub-preset the `"fields": ["{parent}", "extra"]` syntax is the way to go. But to just get the same fields in the sub-preset, it is not necessary to explicitly include the parent.

tyrasd:
//cc @1ec5 do you think these names are acceptable in en-US?

tyrasd (approved):

PR #540, opened 2022-07-17
Update cycleway.json

SafetyIng:
Separate value added to avoid tagging problems. https://wiki.openstreetmap.org/wiki/Tag:cycleway%3Dseparate

### DIFF
--- a/data/fields/cycleway.json
+++ b/data/fields/cycleway.json
@@ -43,6 +43,10 @@
             "opposite": {
                 "title": "Contraflow bike lane",
                 "description": "A bike lane that travels in both directions on a one-way street"
+            },
+            "separate": {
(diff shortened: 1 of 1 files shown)


data/fields/cycleway.json
```diff
@@ -43,6 +43,10 @@
             "opposite": {
                 "title": "Contraflow bike lane",
                 "description": "A bike lane that travels in both directions on a one-way street"
+            },
+            "separate": {
+                "title": "Separate Way for cycleway",
```
tordans, line 48: How about something like „mapped separately“ to underline the fact that is has to be present / visible / mapped at this time.

SafetyIng: Thanks, fits better and I have adjusted it.

PR #550, opened 2022-07-27
Fix wrong label

Dimitar5555:


tyrasd:
since not all requests for adding a tag are about presets, but can also be about fields or field values, I chose the generic `enhancement` label on purpose.

PR #551, opened 2022-07-28
fix broken Lifeboat Station preset

k-yle:
I didn't realize `addTags` needs to repeat everything from `tags`


tyrasd (approved):

tyrasd:
oh, that has slipped through in #382. sorry for not noticing it either.

PR #554, opened 2022-08-01
Disable TagInfo-suggestions of parking:orientation

### LINKED ISSUE #553: `parking:orientation` field is showing additional undocumented rare values
I've noticed that ID is suggesting two additional undocumented and nearly unused values for `parking:orientation`:

2022-07-30-205459_2560x1440_scrot

The value `orthogonal` (85 uses) seems like an undocumented alternative for `perpendicular` and `marking` is used just 16 times (no idea what that means).

Why are these showing up in the list? Is this automated based on TagInfo data?

Is it possible to leave them out?

jdhoek:
The `parking:orientation` tag has three common and documented values. Currently no other values are in common use. With `autoSuggestions` turned on, two rare values (`orthogonal`, 80 uses, ostensibly a synonym of `perpendicular` and `marking` with only 16 uses) are included in the combo-box, making these seem like common and valid options.

Fixes #553.

### DIFF
--- a/data/fields/parking/orientation.json
+++ b/data/fields/parking/orientation.json
@@ -8,5 +8,6 @@
             "diagonal": "Diagonal in Relation to the Street",
             "perpendicular": "Meets the Street at a Straight Angle"
         }
-    }
+    },
+    "autoSuggestions": false
 }


tyrasd:
thanks!

PR #557, opened 2022-08-04
remove landuse=industrial default

Hiausirg:
Most hydroplants are mapped as buildings (due to their small size) and adding landuse=industrial would be wrong. Examples:
https://www.openstreetmap.org/edit?way=1083724073#map=19/34.52749/-82.37500
https://www.openstreetmap.org/edit?way=1060286456#map=19/34.79361/-83.54071
https://www.openstreetmap.org/edit?way=1060286451#map=19/34.75530/-83.50110
https://www.openstreetmap.org/edit?way=1076839170#map=19/34.16930/-81.90250
https://www.openstreetmap.org/edit?way=608270485#map=19/34.26120/-81.33098
https://www.openstreetmap.org/edit?way=186840913#map=19/34.05343/-81.21683
etc....

### DIFF
--- a/data/presets/power/plant/source/hydro.json
+++ b/data/presets/power/plant/source/hydro.json
@@ -15,7 +15,6 @@
     },
     "addTags": {
         "power": "plant",
-        "landuse": "industrial",
         "plant:source": "hydro",
         "plant:output:electricity": "*"
     },


tyrasd (approved):
:+1: from me, the `landuse=industrial` tag is not mentioned on the [`plant:source=hydro` wiki page](https://wiki.openstreetmap.org/w/index.php?title=Tag:plant:source%3Dhydro&oldid=2079564).

For completeness, here is a current usage graph (dark blue = objects with `plant:source=hydro` and `landuse=industrial`, light blue = remaining object tagged as `plant:source=hydro`)

PR #559, opened 2022-08-05
Gas Station preset: move `building` field to `fields`

arch0345:
[According to Taginfo](https://taginfo.openstreetmap.org/tags/amenity=fuel#combinations), a significant amount of features tagged with `amenity=fuel` also have a `building=*` tag (~23%).

### DIFF
--- a/data/presets/amenity/fuel.json
+++ b/data/presets/amenity/fuel.json
@@ -7,7 +7,7 @@
         "address",
         "fuel_multi",
         "self_service",
-        "building"
+        "building_area"
     ],
     "moreFields": [
         "branch_brand",
--- a/data/presets/amenity/fuel.json
+++ b/data/presets/amenity/fuel.json
@@ -6,12 +6,11 @@
         "operator",
         "address",
         "fuel_multi",
-        "self_service"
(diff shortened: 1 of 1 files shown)


data/presets/amenity/fuel.json
```diff
@@ -6,12 +6,11 @@
         "operator",
         "address",
         "fuel_multi",
-        "self_service"
+        "self_service",
+        "building"
```
tyrasd, line 10: we should use the `building_area` field here, as quite a lot of fuel stations are mapped as points, where the building tag does not apply.

```suggestion
        "building_area"
```

tyrasd (changes requested):

tyrasd (approved):

tyrasd:
Sorry, I overlooked that the `building_area` field sets the `building` tag to `yes` by default for new features. For fuel stations it should remain empty by default, however. I think a new field is needed for this particular case.

tyrasd:
FYI: In 65f46b5, I've split the `building_area` field into one which has the default `yes` value (called `building_area_yes`) and one which doesn't have that default value.

PR #560, opened 2022-08-05
Use "ski" instead of "Ski" in activity field

Hufkratzer:
Seems to be a typo. It seems to cause that an existing `ski=yes` tag isn't correctly detected by the activity, example:
https://www.openstreetmap.org/node/2648315141
ski


tyrasd:
thanks!

PR #561, opened 2022-08-05
Cemetery Section: use `name` instead of `sector:name`

arch0345:
As mentioned in https://github.com/openstreetmap/iD/issues/9232#issuecomment-1206783948, the vast majority of cemetery sectors are tagged with `name` rather than `sector:name`.

### DIFF
--- a/data/fields/sector/name.json
+++ b/data/fields/sector/name.json
@@ -1,6 +1,0 @@
-{
-    "key": "sector:name",
-    "type": "text",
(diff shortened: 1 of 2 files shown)


Dimitar5555:
Is it possible to add rendering for the `ref` tag in this PR so that it's easier to spot mistakes?

1ec5:
That was fixed in openstreetmap/iD#9061.

PR #562, opened 2022-08-05
Add `turning_circle` field

arch0345:
This adds a field for the key [`turning_circle`](https://wiki.openstreetmap.org/wiki/Key%3Aturning_circle), which is used to differentiate types of turning circles. [`turning_circle`](https://taginfo.openstreetmap.org/keys/turning_circle) is currently used around 3.3K times

### DIFF
--- a/data/fields/turning_circle.json
+++ b/data/fields/turning_circle.json
@@ -1,0 +1,5 @@
+{
+    "key": "turning_circle",
+    "type": "combo",
+    "label": "Type"
+}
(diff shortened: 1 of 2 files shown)


data/fields/turning_circle.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "turning_circle",
+    "type": "combo",
+    "label": "Type"
```
1ec5, line 4: “Shape” would dissuade users from polluting the `turning_circle=*` key with values representing orthogonal aspects of a turning circle.

PR #569, opened 2022-08-11
Suggest a proper tag for water runways

jleedev:


data/deprecated.json
```diff
@@ -7,6 +7,10 @@
     "old": {"aeroway": "aerobridge"},
     "replace": {"aeroway": "jet_bridge", "highway": "corridor"}
   },
+  {
+    "old": {"aeroway": "runway", "surface": "water"},
+    "replace": {"seamark:type": "seaplane_landing_area"}
```
1ec5, line 12: Looks like we don’t have a preset for this tag yet. Specifying that preset as being applicable to an area would be a good way to ensure that the user doesn’t end up with this tag on a line if the `aeroway=runway` was tagged on a line.

jleedev: OpenSeaMap is happy to render this on a line, although the T shapes pointing to one side look unintended.

image

Are you saying that if this is on an unclosed way we should simply not upgrade it automatically?

1ec5: Oh, I was under the impression that it’s invalid on an area. I could be wrong about that.

tyrasd:
A little bit of explanation would have been nice for this PR. Was there a discussion behind this tag replacement proposal?

From a "raw usage numbers" point of view neither tagging schema is clearly favored over the other (75 vs. 110 features).

FWIW, there do also exist some (rare) water runways inside "land aerodrome"s, which at first glance seem to fit to the description of a `aeroway=runway`. Here are some examples: https://www.openstreetmap.org/way/658226025, https://www.openstreetmap.org/way/387939245 or https://www.openstreetmap.org/way/137990458.

data/deprecated.json
1ec5: (Edited my previous post, which inverted some logic confusingly.)

1ec5:
> FWIW, there do also exist some (rare) water runways inside "land aerodrome"s, which at first glance seem to fit to the description of a `aeroway=runway`. Here are some examples: https://www.openstreetmap.org/way/658226025, https://www.openstreetmap.org/way/387939245 or https://www.openstreetmap.org/way/137990458.

These examples are very narrow seaplane landing areas. Most are [more rotund](https://www.openstreetmap.org/way/651185271). The only functional difference is that there’s less lateral maneuvering space on these ones. But even these more linear landing areas are physically just sets of buoys rather than a runway. (A landing area without buoys is just a pond.)

> From a "raw usage numbers" point of view neither tagging schema is clearly favored over the other (75 vs. 110 features).

The [long-documented](https://wiki.openstreetmap.org/wiki/Seamarks/Seamark_Objects) premise of [`seamark:type=seaplane_landing_area`](https://wiki.openstreetmap.org/wiki/Tag:seamark:type%3Dseaplane_landing_area) is that any place where seaplanes take off and land should be tagged as seamarks rather than runways, and the `aeroway=runway` documentation corroborates it. I was unaware of any controversy over this tag specifically, just some occasionally mentioned discomfort with how the overall seamark tagging scheme seems to copy a nautical chart very literally.

Many laypeople think of seaplane landing areas as runways. In the past, there’s been a lot of back-and-forth between mappers who are more familiar with seaplanes and those who aren’t in places where seaplane bases are common. The raw usage numbers don’t allow us to conclude that neither is favored over the other, because it’s very likely that most uses of `aeroway=runway` `surface=water` were made without any awareness about what differentiates a seaplane landing area from a conventional runway. Meanwhile, renderer developers consider `aeroway=runway` `surface=water` to be troll-tagging.

To me, the most important thing would be to have a dedicated preset for seaplane landing areas, however they’re tagged. This is the only way to ensure that they’re mapped as areas rather than lines, to avoid losing important information about both dimensions.

tyrasd:
> Many laypeople think of seaplane landing areas as runways.

I guess this is pretty widespread, as wikipedia generally lists seaplane landing areas as "runways" and even the FAA does so (example [1](https://nfdc.faa.gov/nfdcApps/services/ajv5/airportDisplay.jsp?airportId=DWH), [2](https://nfdc.faa.gov/nfdcApps/services/ajv5/airportDisplay.jsp?airportId=W55)). :shrug: 

> The long-documented premise of `seamark:type=seaplane_landing_area` is that any place where seaplanes take off and land should be tagged as seamarks **rather than** runways, and the aeroway=runway documentation corroborates it.

I cannot follow where you got the impression from that this is a question of tagging it as either a seamark or a runway. Considering that the `seamark` tag namespace is defined to be specifically for objects (and their attributes) which are relevant for nautical navigation, and the fact that seaplanes are by their hybrid nature are affecting and affected by both the nautical and aviation _realm_: Wouldn't a logical way to map the landing areas for seaplanes be to include tags from both tagging schemas: `seamark:type=seaplane_landing_area` in combination with `aeroway=runway` + `surface=water` and further tags like `ref`, etc.? Also note that `[seamark:type=seaplane_landing_area](https://wiki.openstreetmap.org/wiki/Tag:seamark:type%3Dseaplane_landing_area)` also does not "forbid" it from being used together with an `aeroway` tag.

> To me, the most important thing would be to have a dedicated preset for seaplane landing areas,

Definitely agree. :+1:

Btw, I've also seen the tag `aeroway=water_runway` ([example](https://www.openstreetmap.org/way/835919617)) used for seaplane landing areas.

matkoniecz:
edited https://wiki.openstreetmap.org/wiki/Tag:seamark:type%3Dseaplane_landing_area

PR #570, opened 2022-08-15
Add 'package locker' alias for Parcel Locker preset

CorruptComputer:
In the US it is common to call these 'Package Locker' rather than 'Parcel Locker', so having this term show up in search would be beneficial in my opinion.

### DIFF
--- a/data/presets/amenity/parcel_locker.json
+++ b/data/presets/amenity/parcel_locker.json
@@ -1,7 +1,4 @@
 {
-    "aliases": [
-        "Package Locker"
-    ],
     "icon": "temaki-vending_lockers",
     "fields": [
         "operator",
@@ -26,6 +23,7 @@
         "locker",
         "mail",
         "package",
+        "package locker",
         "packstation",
         "parcel",
         "pickup"
--- a/data/presets/amenity/parcel_locker.json
+++ b/data/presets/amenity/parcel_locker.json
@@ -24,7 +24,8 @@
         "mail",
         "packstation",
         "parcel",
-        "pickup"
+        "pickup",
+        "package locker"
     ],
     "tags": {
         "amenity": "parcel_locker"
--- a/data/presets/amenity/parcel_locker.json
+++ b/data/presets/amenity/parcel_locker.json
@@ -1,6 +1,6 @@
 {
     "aliases": [
-        "package locker"
+        "Package Locker"
     ],
     "icon": "temaki-vending_lockers",
     "fields": [
--- a/data/presets/amenity/parcel_locker.json
+++ b/data/presets/amenity/parcel_locker.json
@@ -1,4 +1,7 @@
 {
+    "aliases": [
+        "package locker"
+    ],
     "icon": "temaki-vending_lockers",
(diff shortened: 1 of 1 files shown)


data/presets/amenity/parcel_locker.json
```diff
@@ -24,7 +24,8 @@
         "mail",
         "packstation",
         "parcel",
-        "pickup"
+        "pickup",
+        "package locker"
```
1ec5, line 28: This is the array of `terms`, but you can add it to a separate array of `aliases` so that the user will see Package Locker as a result when they search for “package”.

Also, this repository’s contents more or less represent the American English localization, so if “parcel locker” is less common than “package locker” in this dialect, then “Package Locker” should be the preset name, relegating “Parcel Locker” to an alias.

CorruptComputer: Ah, yeah that sounds good I can move it to there. 

I can't speak for the whole US, but for the parts that I have lived in (Pittsburgh, Houston, and Atlanta) 'Package' is a lot more common for folks to say than 'Parcel'. If you think it would be better to switch the name to Package Locker I can do that and add Parcel Locker as the alias instead.

CorruptComputer: Fixed your first point, and left the second one open for discussion in case any one else wants to chime in on it.

1ec5: In general, I would agree about “package” being a more familiar term than “parcel”. I just wasn’t sure if that translated to lockers as well, but it sounds like it does. For what it’s worth, the USPS uses “[parcel locker](https://faq.usps.com/s/article/What-is-a-Cluster-Box)” to refer to a part of a cluster mailbox that doesn’t meet the definition of an `amenity=parcel_locker`. But they also tend to use the term “parcel” where most people would say “package”.

data/presets/amenity/parcel_locker.json
```diff
@@ -1,4 +1,7 @@
 {
+    "aliases": [
+        "package locker"
```
1ec5, line 3: Unlike terms, aliases are Title Cased just like the preset name.

```suggestion
        "Package Locker"
```

CorruptComputer: Ah thanks for the heads up, should be fixed now.

CorruptComputer:
Hey @1ec5 just wanted to check if a decision had been made on if the Alias and Name should be swapped?

1ec5:
Sorry, this slipped my attention – looks like most respondents didn’t really have a word for these other than to call them by the brand name, e.g., Amazon Locker. Otherwise, “parcel locker” got more support than “package locker”. I can’t explain why it would differ from how Americans normally call parcels packages, but it’s a relatively novel feature type to begin with. Given these responses, I guess we should keep it as an alias for now and revisit in the future if usage changes.

I’m not a maintainer of this repository, so you’ll need a review from someone who is before the changes can be merged.

tyrasd:
I've merged it as a _term_ for now, as suggested by @1ec5

PR #575, opened 2022-09-03
add gender field on hairdresser

nlehuby:
it will be very useful to distinguish barber shops from regular hairdressers

### DIFF
--- a/data/presets/shop/hairdresser.json
+++ b/data/presets/shop/hairdresser.json
@@ -9,6 +9,12 @@
     ],
     "fields": [
         "name",
+        "operator",
+        "address",
+        "building_area_yes",
+        "opening_hours",
+        "payment_multi",
+        "phone",
         "gender"
     ],
     "tags": {
--- a/data/presets/shop/hairdresser.json
+++ b/data/presets/shop/hairdresser.json
@@ -7,6 +7,10 @@
     "terms": [
         "beard"
     ],
+    "fields": [
+        "name",
+        "gender"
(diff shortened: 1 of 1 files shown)


data/presets/shop/hairdresser.json
```diff
@@ -7,6 +7,10 @@
     "terms": [
         "beard"
     ],
+    "fields": [
+        "name",
+        "gender"
+    ],
```
tyrasd, line 13: there's some superfluous whitespace here
```suggestion
    ],
```

data/presets/shop/hairdresser.json
```diff
@@ -7,6 +7,10 @@
     "terms": [
         "beard"
     ],
+    "fields": [
+        "name",
```
tyrasd, line 11: Please also include the regular fields of the `shop` preset:

```suggestion
        "name",
        "operator",
        "address",
        "building_area_yes",
        "opening_hours",
        "payment_multi",
        "phone",
```

tyrasd (changes requested):
Some minor tweaks are needed (see below).

I'm not super happy with reusing the `gender` preset for this, as the description "Gender" could be maybe a bit confusing (at least in some translations). But as I'm also not sure what a better description could be (_Haircut Style_ perhaps? or _Customers Served_??) right now, so I guess it's fine for the moment.

tyrasd (approved):

nlehuby:
Thanks :+1:

PR #576, opened 2022-09-03
Add French SIRET number tag on shops

nlehuby:
The `ref:FR:SIRET` is a number from the national statistics bureau of France which allows the identification of any French business.
It can be found on shops, offices and some amenities.

### DIFF
--- a/data/fields/ref/FR/siret.json
+++ b/data/fields/ref/FR/siret.json
@@ -1,0 +1,16 @@
+{
+    "key": "ref:FR:SIRET",
+    "type": "identifier",
+    "label": "SIRET Number",
+    "urlFormat": "https://annuaire-entreprises.data.gouv.fr/etablissement/{value}",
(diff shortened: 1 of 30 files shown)


tyrasd (changes requested):
Can you please rename the field to `siret-FR.json`, as it is convention for regional fields/presets to include the applicable location set(s) in the file name.

Also, I noticed that only `tourism=hotel` is included in this PR, but not other touristic businesses (e.g. `tourism=guest_house`, etc.). Was this on purpose?

nlehuby:
> Also, I noticed that only `tourism=hotel` is included in this PR, but not other touristic businesses (e.g. `tourism=guest_house`, etc.). Was this on purpose?

Yes, the registration as a company (which implies having a SIRET number) for tourist accommodations depends on several legal parameters and I'm not sure how they translate as OSM tags at this point and which exact kind of `tourism=*` could have one. 
In any case, usage is quite low at the moment on `tourism` features. I tried to cover here the items most likely to have this attribute.

tyrasd (approved):

tyrasd:
Thanks!

PR #584, opened 2022-09-19
Add a field for the key "model"

Lukas458:
Hi,
the key "model" ( https://wiki.openstreetmap.org/wiki/Key%3Amodel ) describes the model designation of a manufactured object. Formerly, it has been used for power=generator mainly, but now its usage has increased and is spread around over various objects which can have a model-designation / model-name. For example public telephones, elevators, street cabinets, substations and so on. **It is used over 32000 times.** So I think it's worth adding it. The kind of field should be free-text generally, but I chose the field type "combo" so that people can orientate on which values are already in use. For this, there's also the case-sensivity important.

### DIFF
--- a/data/fields/model.json
+++ b/data/fields/model.json
@@ -1,0 +1,6 @@
+{
+    "key": "model",
+    "type": "combo",
+    "label": "Model",
+    "caseSensitive": true
+}


tyrasd (changes requested):
This PR is missing the inclusion of the new field in the respective presets. Can you add it add it at least to the `street_cabinet` preset and any other presets where you think it should be included, please?

It' s unfortunate that for find turbines, the `manufacturer:type` tag is used more frequently than the `model` tag. But there is a [proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/Manufacturer_and_Model) on the way to deprecate that tag in favor of `model`, so I think it is reasonable to only support `model` in iD for now.

Lukas458:
> This PR is missing the inclusion of the new field in the respective presets. Can you add it add it as an optional field (moreFields) at least to the street_cabinet preset and any other presets where you think it should be included, please?

Oh sorry, I wrongly thought that the new-field.json needed to come first and this work needs to come after then. Of course I can add it to some presets, I will do that then.

> It' s unfortunate that for find turbines, the manufacturer:type tag is used more frequently than the model tag.

Oh yes, I see, another case where for some single main-tag a special under-tag had been created/used. I share your opinion that `model` is in fact more useful in that it can be used independently for a breadth/abundance of objects while being independent from any other tags.

tyrasd (approved):

tyrasd:
merged, thanks!

Issue #507, opened 2022-06-21
Allow tagging signalized but unmarked pedestrian crossing


1ec5:
For completeness, there should be a preset or field that allows the user to indicate that a pedestrian crosswalk is signalized but unmarked.

## Rationale

As of #192 #368, this repository has a set of presets to express each of these pedestrian crossing configurations:

* Crossing with no road markings and no traffic signals (`crossing=unmarked`)
* Crossing with road markings but no traffic signals (`crossing=marked`)
* Crossing with road markings and traffic signals (`crossing=traffic_signals`)

However, there’s no affordance for tagging crossings with traffic signals but no road markings.

This configuration is standard and commonplace in some regions: https://github.com/openstreetmap/id-tagging-schema/issues/408#issuecomment-1061225441. Where it occurs, it is important to distinguish from a crossing with traffic signals and road markings, because the lack of road markings creates a higher risk profile for pedestrians, especially where there are unprotected left turns. One can imagine a pedestrian router applying the same penalty as for marked, signalized crossings, while a car router issues the same alert as for unmarked, unsignalized crossings.

## Tagging

The [most common](https://overpass-turbo.eu/s/1jvg) tag combination for this configuration is `crossing=unmarked` `crossing:signals=yes`, with 78 occurrences globally. This is likely an undercount, because the [`crossing:signals`](https://wiki.openstreetmap.org/wiki/Key:crossing:signals) key description page was created only a few weeks ago despite being [proposed](https://wiki.openstreetmap.org/wiki/Proposed_features/crossing:signals) several years ago. It’s quite likely that a lot of unmarked, signalized crossings are being tagged as just `crossing=unmarked` or just `crossing=traffic_signals`, losing valuable information.

Even if we assume that this configuration is limited to the United States (which is [false](https://www.openstreetmap.org/way/1062779200)), it is still common enough to merit support in this repository. `crossing=unmarked` `crossing:signals=yes` represents 1% of [all `crossing:signals` usage in the U.S.](http://taginfo.geofabrik.de/north-america/us/keys/crossing:signals) At this rate, over 2,000 of the `crossing=traffic_signals` that have been tagged so far in the U.S. could conceivably be candidates for `crossing:signals=no`. Of course, that can only be a rough estimate, because so much data has already been lost through the adoption of `crossing=traffic_signals`.

## Design

[...]

Issue #589, opened 2022-09-22
Add support for crossing:markings=*


kaneap:
### OSM Tag(s)

crossing:markings=*

### How would you like this tag to see supported?

As a Field

### Label

Crosswalk markings

### Terms

Crossing markings

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Key:crossing:markings

### Status of the Tag

Approved

### Usage of the tag

1,472

### Further Information

This tag was recently approved.

Pehaps display the implied values in gray (`crossing=marked`,`crossing=traffic_signals` (and maybe `crossing=uncontrolled`) implies `crossing:markings=yes`)

Also see #408 for the discussion on whether iD should be tagging `crossing=uncontrolled` instead of `crossing=marked`, and #507 for the tagging of signalized but unmarked crossings (which would be resolved by adding this field).

Thanks!

1ec5:
> #507 for the tagging of signalized but unmarked crossings (which would be resolved by adding this field).

Yes and no, a `crossing:markings` field would make it _possible_ to indicate a signalized but unmarked crossing, but perhaps not in the most natural manner. The issue specifically requested a field to indicate signalization, which I think would continue to be useful regardless. The current plan is to more formally propose the already common `crossing:signals` key to further deconflict the `crossings` key. As with the `crossing:markings` proposal, there’s already some talk about being able to elaborate beyond a Boolean value. The Crossing With Pedestrian Signals preset would probably need to be renamed or removed at that point, because a crossing controlled by the main traffic signals, without a dedicated pedestrian signal face, could not be described as having “pedestrian signals”, even if `crossing=traffic_signals` remains literally accurate.

Issue #568, opened 2022-08-10
Support traffic signals on cycleway crossings


BubbaJuice:
### OSM Tag(s)

cycleway=crossing and crossing=traffic_signals

### How would you like this tag to see supported?

Other (please indicate below)

### Label

Cycle Crossing with Pedestrian Signals

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:crossing%3Dtraffic_signals; https://wiki.openstreetmap.org/wiki/Tag:cycleway%3Dcrossing

### Status of the Tag

Other (please indicate below)

### Usage of the tag

6,603

### Regional Tag?

global

### Further Information

The existing "Cycle Crossing" preset should be improved upon with new presets for cycle crossings with pedestrian signals. These would support the same tags as the "Crossing with Pedestrian Signals" preset. `cycleway=crossing` is de facto while `crossing=traffic_signals` is approved.

`cycleway=crossing` + `crossing=traffic_signals` taginfo:
![image](https://user-images.githubusercontent.com/68205167/184038246-e37ea892-79a9-4a1f-8b74-7189bb6b1903.png)

1ec5:
Related: https://github.com/openstreetmap/id-tagging-schema/issues/507#issuecomment-1179832250.

tyrasd:
I'd like to wait on this until the voting period of the proposal for `crossing:markings` tag is over (this weekend):
https://wiki.openstreetmap.org/wiki/Proposed_features/crossing:markings, as all crossing presets would likely need to be adjusted for that anyway.

tyrasd:
The [proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/crossing:markings) for the `crossing:markings` tag was recently approved.

PR #590, opened 2022-09-23
refactor crossing presets to use approved "crossing:markings" tag: add field for the new tag, change "Marked Crosswalk" presets to use "crossing=uncontrolled" tag and add preset for "Cycle Crossing With Traffic Signals"

tyrasd:
The tag proposal `crossing:markings` was recently [approved](https://wiki.openstreetmap.org/wiki/Key:crossing:markings). This adds a field for this new tag to the corresponding crossing presets. (#589)

This also changes the tag used by the "marked crossing" preset from `crossing=marked` to OSM's accepted `crossing=uncontrolled`.

preset | before this PR | changed in this PR
--  | -- | --
Crossing With Pedestrian Signals (vertex) | `highway=crossing` + `crossing=traffic_signals` | added field for `crossing:markings` |
Crossing With Pedestrian Signals (line) | `highway=footway` + `footway=crossing` + `crossing=traffic_signals` | added field for `crossing:markings` |
Cycle Crossing With Traffic Signals | N/A (#568) | `highway=cycleway` + `cycleway=crossing` + `crossing=traffic_signals` <br>+ field for `crossing:markings` |
Marked Crosswalk (vertex) | `highway=crossing` + `crossing=marked` | `highway=crossing` + **`crossing=uncontrolled`** (#408)  <br>+ field for `crossing:markings` |
Marked Crosswalk (line) | `highway=footway` + `footway=crossing` + `crossing=marked` | `highway=footway` + `footway=crossing` + **`crossing=uncontrolled`** (#408) <br>+ field for `crossing:markings` |
Marked Cycle Crossing | `highway=cycleway` + `cycleway=crossing` + `crossing=marked` | `highway=cycleway` + `cycleway=crossing` + **`crossing=uncontrolled`** (#408) <br>+ field for `crossing:markings` |
Unmarked Crossing (vertex) | `highway=crossing` + `crossing=unmarked` | (unchanged) |
Unmarked Crossing (line) | `highway=footway` + `footway=crossing` + `crossing=unmarked` | (unchanged) |
Unmarked Cycle Crossing | `highway=cycleway` + `cycleway=crossing` + `crossing=unmarked` | (unchanged) |
Cycle & Foot Crossing | `highway=cycleway` + `cycleway=crossing` + field for `crossing` | added field for `crossing:markings`

It is now possible to tag an unmarked crossing with traffic lights by using the _Crossing With Pedestrian Signals_ preset and using the `no` value in the new field for `crossing:markings`. (#507)

//edit: this branch also adds a preset for `highway=footway` + `footway=traffic_island` (see #633).


data/presets/highway/footway/crossing/uncontrolled.json
```diff
@@ -17,18 +17,18 @@
...
     "terms": [
         "marked foot path crossing",
-        "marked crossing",
+        "marked crosswalk",
         "marked pedestrian crosswalk",
         "zebra crossing",
         "crosswalk"
     ],
-    "name": "Marked Crosswalk"
+    "name": "Marked Crossing"
```
1ec5, line 33: It would still be “crosswalk” in American English.

data/fields/crossing.json
```diff
@@ -1,5 +1,17 @@
...
+    "options": [
+        "traffic_signals",
+        "uncontrolled",
+        "unmarked"
+    ],
+    "strings": {
+        "options": {
+            "traffic_signals": "Crossing With Traffic Signals",
+            "uncontrolled": "Only Road Markings",
+            "unmarked": "No Road Markings or Traffic Lights"
```
1ec5, line 14: These three strings are inconsistent with each other. Both “traffic signals” and “traffic lights” are correct in American English, but we should stick to one. “Crossing” is redundant in this field.

1ec5: Should `no` get a string too? (It would only make sense on a node.)

data/fields/crossing/markings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "crossing:markings",
+    "type": "combo",
+    "label": "Crossing Markings"
```
1ec5, line 4: Some strings would be helpful for this field. In particular, the `zebra` value was chosen as a concession to mappers used to the terminology in Europe (and some other countries), but unlike its official definition in some European countries, here it only refers to a marking pattern, without implying other aspects of the crossing.

data/presets/highway/crossing/uncontrolled.json
```diff
@@ -17,6 +17,5 @@
         "key": "crossing",
         "value": "uncontrolled"
     },
-    "name": "Marked Crosswalk",
-    "searchable": false
+    "name": "Marked Crosswalk"
```
1ec5, line 20: `uncontrolled` isn’t just “marked”, it’s “marked without traffic signals”.

data/presets/highway/crossing/unmarked.json
```diff
@@ -1,7 +1,6 @@
 {
     "icon": "temaki-pedestrian",
     "fields": [
-        "crossing",
```
1ec5, line 4: This makes it more difficult to clarify that an unmarked crosswalk is in fact signalized, so I disagree that this PR fixes #507.

1ec5 (commented):
Though it may rest on approved tags, this PR demonstrates how messy the state of approved tagging remains. I hope there will be continued openness to evolving these presets as further proposals reorient the crosswalk tags away from a one-size-fits-all model and closer to what we have for `railway=crossing`.

data/fields/crossing.json
tyrasd: I've fixed the wording.

> Maybe `crossing=no` should be a separate preset?

I haven't included https://wiki.openstreetmap.org/wiki/Tag:crossing%3Dno in this field, as it is intended to specify the type of a crossing, where the `no` option doesn't make sense.
`crossing=no` would indeed  be useful as a standalone preset and as dedicated field to be added to presets like the `highway=traffic_signals`.

data/fields/crossing/markings.json
tyrasd: Yes, strings would be definitely useful. I haven't included them in this PR already, because I didn't feel very confident choosing labels that are . Maybe you could help me out with these? This was one of my drafts, but several entries don't quite feel right to me (e.g. surface, zebra*, ladder:paired, …).

```
            "yes": "Unspecified Markings",
            "surface": "Distinct Surface",
            "zebra": "Zebra Stripes",
            "lines": "Solid Lines on Either Side",
            "ladder": "Ladder",
            "lines:paired": "Pairs of Lines on Either Side",
            "dashes": "Dashed Lines on Either Side",
            "dots": "Dotted Lines on Either Side",
            "zebra:double": "Zebra Stripes, Split in the Middle",
            "zebra:paired": "Zebra Stripes With Paired Bars",
            "zebra:bicolour": "Zebra Stripes With Alternating Colors",
            "ladder:skewed": "Skewed Ladder",
            "ladder:paired": "Ladder With Paired Bars",
            "no": "No Markings"
```

I think that eventually, it would probably be a better user interface if these options were presented using a pictorial form, like in this graphic on wikipedia: 
[](https://commons.wikimedia.org/wiki/File:Crosswalk_styles_(en).svg)

data/presets/highway/crossing/uncontrolled.json
tyrasd: Right, but we also need to keep a balance between keeping the preset names short and including every detail of a preset in its name. I thought that the presence of the dedicated preset for crossings with traffic signals makes it clear enough that the "marked crosswalk" preset is not meant for situations where there is a traffic light. There's also the :information_source:  icon which shows this additional detail for the preset. Last but not least, the name _"Marked Crosswalk"_ worked well in the past for the tag [`crossing=marked`](https://wiki.openstreetmap.org/w/index.php?title=Tag:crossing%3Dmarked&oldid=2400462) which also implied the absence of traffic signals.

Maybe you have a better name for this preset (I'd rather avoid naming it _Uncontrolled Crosswalk_ :wink: )?

data/presets/highway/crossing/unmarked.json
tyrasd: Hmm…

The common design pattern for the presets in this repo is to not repeat the field for "main tag" of a preset, because the intended mechanism to switch between presets is to, well, use the preset switcher UI to choose the better fitting preset. But perhaps we could make an exception in this case, in order to better accommodate the necessities for situations like #507. I'm open to that.

I guess an alternative would be to get rid of the "specialized" crossing presets altogether and only work with fields. One downside of that would be that we'd loose the dedicated icons which currently distinguish the different kinds of crossings on the map view.

data/presets/highway/crossing/uncontrolled.json
1ec5: I think “Marked Crosswalk” is passable as long as there’s a field that can transform it into `crossing=traffic_signals`. It looks like you’ve left the Type field in place for this preset, which helps somewhat. The field isn’t as usable to non-English speakers, but providing localizable option strings would be fraught: several of the most common values mean the same thing – or don’t mean the same thing at all, depending on who you ask – and each has its ardent defenders.

data/fields/crossing/markings.json
1ec5: Similar to `tracktype`, we can probably get away with longer, more descriptive strings to avoid confusion. I don’t think even the traffic engineering profession has a set term for every one of these values, let alone any dialect of colloquial English. In case it helps, here are the technical terms for these patterns in the U.S., paired with my best guess at what non-U.S. markings would be called, but I think using them without a more descriptive tooltip would risk mistranslation and misuse by lay mappers:

Value | Description
----|----
`yes` | Marked Somehow
`surface` | Surface Treatment Only
`zebra` | Longitudinal Bars
`lines` | Transverse Lines
`ladder` | Ladder With Longitudinal Bars
`lines:paired` | Double Transverse Lines
`dashes` | Dashed Transverse Lines
`dots` | Dotted Transverse Lines
`zebra:double` | Triple-Four (!?!?)
`zebra:paired` | Paired Longitudinal Bars
`zebra:bicolour` | Longitudinal Bars With Alternating Colors
`ladder:skewed` | Ladder With Diagonal Bars
`ladder:paired` | Ladder With Paired Longitudinal Bars
`no` | Unmarked

zekefarwell:
> Unfortunately, the proposal for [crossing:signals](https://wiki.openstreetmap.org/wiki/Proposed_features/crossing:signals) seems to be quite inactive at the moment.

I'm not sure why this matters.  Although "approved", [crossing:markings](https://taginfo.openstreetmap.org/keys/crossing%253Amarkings) is just months old with very little usage (~2000) so far.  [crossing:signals](https://taginfo.openstreetmap.org/keys/crossing%3Asignals) has been in use for several years and has four times the usage (~8000).  In the grand scheme of things, the usage of both tags is absolutely tiny compared to [crossing](https://taginfo.openstreetmap.org/keys/crossing) with over 6 million uses.  

https://taghistory.raifer.tech/#***/crossing:markings/&***/crossing:signals/

If you're going to start populating `crossing:markings` it would make sense to start populating `crossing:signals` as well in order fully solve the problem.  On the other hand if you think it's too early to start populating `crossing:signals`, then it's also too early to start populating `crossing:markings` and neither should be implemented yet.  Personally, I'd say go ahead with both.

mxxcon:
`crossing:markings` proposal was approved only 10 days ago so few people are aware of it and as of right now I don't think any editor has presets for it, so it's natural for its usage to be low for now.

zekefarwell:
> This also changes the tag used by the "marked crossing" preset from crossing=marked to OSM's accepted crossing=uncontrolled.

Changing from `crossing=marked` to `crossing=uncontrolled` seems out of scope, given the title of the PR doesn't mention this.  I suggest either limiting the PR to only the `crossing:markings` tag, or changing the title of the PR to: **Crossing presets: replace "crossing=marked" with "crossing=uncontrolled" and add new "crossing:markings" tag**.

tyrasd:
> Changing from crossing=marked to crossing=uncontrolled seems out of scope, given the title of the PR doesn't mention this

Sorry for missing this in the title. My intention was to keep the title short, but I guess it is only fair to include every important detail in the title for larger changes such as this one. I've changed the title now.

My intention behind switching out `crossing=marked` for `crossing=uncontrolled` was to bring the used `crossing` tags in line with the [_approved_](https://wiki.openstreetmap.org/wiki/Key:crossing#Approved) values of the `crossing` tag. I'd be open to omit this tag change if arguments can be made why this would be a bad choice.

tyrasd:
> the proposal process […] is not required for a new tag to become established.

I don't disagree, but the proposal process does also have benefits: On the one hand, it makes sure that the tag in question is well defined (on the wiki) and that its relationship and interplay with other tags is documented. On the other hand, while it might not guarantee that a tag is complete consensus behind a tag, it at least makes it likely that problems/disagreements/dissensus/conflicts associated with the tag are surfaced.

In this particular case both of these aspects are relevant, as several aspects of the `crossing:signals` proposal are currently not clear in my opinion:
- It is not clear what consequences the new `crossing:markings` tag has to the (proposed) usage of the `crossing:signals` tag.
- Is the `crossing:signals` tag meant to completely replace the `crossing` tag (potentially in conjunction with the new `crossing:markings` tag), or just the `crossing=traffic_signals` value? 
- Whether the community is fine with deprecate the `crossing` tag by the combination of `crossing:markings` + `crossing:signals`.

> 8000 uses of crossing:signals and its clear upward trajectory

It's not that easy, unfortunately. As you said above, the 8000 uses are almost negligible in the "grand scheme of things". There are hundreds of times more crossings added each day which don't use the `crossing:signals` tag, but instead rely on the "established" `crossing=traffic_signals` tag… Raw usage numbers are not a sufficient condition for or against adding a tag, I'm afraid.

> feels like wasted effort when many in the community consider approved proposals essentially meaningless.

I feel you. The issue from my perspective is that I really want to avoid to just incorporate tags on a "gut feeling" basis, as this approach had led to many unnecessary controversies in the past. And I also really need tags to be properly documented somewhere[^1] before they can become useful as a preset/field/etc.. These two of my requirements are currently somewhat well covered by the wiki tag approval process, that's why I rely on it so strictly on the wiki tag approval process. If you have an idea for a better process, please let me know (but maybe open a new issue, such that it doesn't get too off-topic here).

[^1]: currently, the wiki is the only good place for this, and a good amount of iD's functionalities rely on it, e.g. the taginfo integration.

1ec5:
A lot of the criticism of the wiki’s proposal process focuses on its limited participation, which has always been a struggle. However, I think what’s more relevant to crossing tagging is that the proposal process can only capture a snapshot in time.

In 2008, maybe it was OK that something as unimportant and pedantic as pedestrian crossing classification could be coined by British mappers based on local distinctions and [approved](https://wiki.openstreetmap.org/w/index.php?oldid=132309) (with reservations) by ten mappers, all but one from Europe. Maybe it was OK that the `crossing=uncontrolled` tag [originally came from non-English-speaking countries](https://wiki.openstreetmap.org/w/index.php?oldid=137822#Current_usage) where the word’s real meaning was [mistaken](https://wiki.openstreetmap.org/wiki/Tag:crossing%3Duncontrolled#Terminology) for almost its opposite.

But times change, the project changed, and so did our collective understanding of the importance and complexity of crossings. A similar mistranslation hopefully wouldn’t pass muster today, and hopefully people have learned their lesson to vote no when they have those reservations. The rest of the wiki evolved to document the controversy around this tag, gaps in the tagging scheme, and fitful attempts at addressing its shortcomings. However, the approved proposal remains fixed, confidently presenting the tagging scheme as one that works well, even though we know it doesn’t.

I appreciate that we have a formal process for requesting comments on a tagging idea, and that it can even be brought to a vote, incentivizing mappers to come to a decision instead of trailing off like most tagging list threads. But it’s an imperfect process: besides not being representative of OSM’s internal and external stakeholders, it requires too much effort. On the surface, it’s just a matter of writing a wiki page and posting a few mailing list posts. But as we saw with the original tagging list discussion about `crossing=marked`/`unmarked`, there are a lot of special interests to appease. As a result, the process is optimized for the most determined, politically savvy community members, not necessarily the best ideas.

You’re right to insist on broader discussion and decisionmaking, because these tagging issues affect more than just iD. At the same time, the iD maintainer is a position of leadership within the community. It would mean a lot if, beyond expressing a desire for tagging improvements here, you could lend your support to reworking and promoting a `crossing:signals` proposal. Meanwhile, I think it would be healthy if we stick to debating the merits of one tag or another, without putting undue emphasis on a tag status that would not have been achieved in today’s environment. After all, there remains [plenty of grousing](https://wiki.openstreetmap.org/wiki/Special:Diff/2221804#iD) about dumping `crossing=zebra` in favor of `crossing=marked`, but `crossing=zebra` wasn’t approved either.

tyrasd:
> I think that eventually, it would probably be a better user interface if these options were presented using a pictorial form

I've implemented [this suggestion](https://github.com/openstreetmap/id-tagging-schema/pull/590#discussion_r979915579) in iD now:

image

tyrasd:
I've merged this branch now, to not let it sit around any longer and get stale. I think it does improve the tagging schema (e.g. by adding a way to specify crossing markings, bring cycle crossing up to par with foot crossings, and removing some inconsistencies between presets).

I'd still love to see someone from the community to pick-up and finish the `crossing:signals` to fully simplify these presets. If nobody steps up, I might try to give it a shot when I find some spare free time.

1ec5:
There are a couple ongoing proposals to refactor crossing tagging, and from the sound of the discussion, they’re still fluid, so you have an opportunity to contribute to the potential subkey-ification of crossing tagging.

https://github.com/openstreetmap/id-tagging-schema/pull/590

1ec5:
> I've merged this branch now, to not let it sit around any longer and get stale. I think it does improve the tagging schema (e.g. by adding a way to specify crossing markings, bring cycle crossing up to par with foot crossings, and removing some inconsistencies between presets).

Discussion on one of these changes is continuing in #658.

> I'd still love to see someone from the community to pick-up and finish the `crossing:signals` to fully simplify these presets. If nobody steps up, I might try to give it a shot when I find some spare free time.

[A new proposal for `crossing:signal`](https://wiki.openstreetmap.org/wiki/Proposed_features/Crossing_signalization) is in the works. I invite those interested in this topic to contribute their expertise and perspective to this draft ahead of a general RfC.

sekerob:
Curiously, while sitting with the tablet and ID I edit mode open in vivaldi in the week of Feb 25, 2026 touched upon a zebra crossing with the crossing=zebra tag and ID insisting in upload screen the tagging was deprecated in favour of crossing_ref=zebra. Of course in the heart of Italy middle finger in splint started throbbing again. I chose to change it to crossing=uncontrolled and added crossing:markings=zebra which I must have done a zillion times, in the heart of Italy. Sure enough, ID awarded deeply despised demirit for ignoring the suggested crossing_ref tag. So lesson forgotten and refreshed, use any other editor.
(why ID offered that tag in the heart of Italy, don't know, Got tablet set to use British English for the spellchecking is the only left field answer I could think of)

PR #597, opened 2022-09-28
add some new surface values

matkoniecz:


matkoniecz:
I wanted to add `surface=grass_paver` but I was unsure is it better to describe it as "Grass Paver" or "Grass Pavers".

data/fields/surface.json
```diff
@@ -12,11 +12,16 @@
...
             "paving_stones": "Paving Stones",
+            "sett": "Sett",
             "dirt": "Dirt",
             "grass": "Grass",
             "sand": "Sand",
             "wood": "Wood",
+            "metal": "Metal",
             "artificial_turf": "Artificial Turf",
+            "unhewn_cobblestone": "Cobblestone (uncut, rounded stones)",
+            "mud": "Mud (almost always present)",
```
tyrasd, line 23: the _almost always present_ irritated me at first, as I thought this meant something like the tag `surface=mud` would be almost always present (which of course doesn't make any sense). Maybe something like _Mud (permanent)_ would be a slightly better wording?

matkoniecz: The problem with "permanent" is that suggest that if place has heavy mud for 363 days a year then it is not eligible.

Or maybe I interpret it to strongly? (NOT a native speaker)

Pinging @ZeLonewolf for language advise (let me know if you ever prefer to stop being pinged for that)

ZeLonewolf: I'm not actually sure what you're trying to express with the mud preset?

tyrasd: essentially what is stated in wiki.openstreetmap.org/wiki/Tag:surface%3Dmud – that the tag value should be used for objects that are muddy _all the time_ or at least _most of the time_. The description should make it clear that `surface=mud` should not be applied to surfaces that are only muddy during and after rain. Examples where it should be used would be wetlands like swamps or tidal areas.

ZeLonewolf: I would call that "persistently muddy" but you could come up with quite a few other phrases I think that express the same thing.

ZeLonewolf: You could say something like "Mud (consistently wet)"

PR #598, opened 2022-09-29
upgrade to schema-builder v5

tyrasd:
see https://github.com/ideditor/schema-builder/blob/main/CHANGELOG.md#510

### DIFF
--- a/data/presets/highway/primary_link-US-CA.json
+++ b/data/presets/highway/primary_link-US-CA.json
@@ -1,5 +1,5 @@
 {
-    "name": "{primary_link}",
+    "name": "{highway/primary_link}",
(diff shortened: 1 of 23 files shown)


data/presets/highway/primary_link-US-CA.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "Primary Link",
+    "name": "{primary_link}",
```
Zverik, line 2: Shouldn't it be `"{highway/primary_link}"`?

tyrasd: of course. thanks for double-checking!

PR #603, opened 2022-10-02
Add Fire Station Building preset

arch0345:
Adds a preset for [`building=fire_station`](https://wiki.openstreetmap.org/wiki/Tag:building%3Dfire_station), which is [currently used around 13K times](https://taginfo.openstreetmap.org/tags/building%3Dfire_station). This should ideally get its own icon, but for now I'm using [`fas-house-fire`](https://fontawesome.com/icons/house-fire).

### DIFF
--- a/data/presets/building/fire_station.json
+++ b/data/presets/building/fire_station.json
@@ -1,0 +1,11 @@
+{
+    "icon": "fas-house-fire",
+    "geometry": [
+        "area"
+    ],
+    "tags": {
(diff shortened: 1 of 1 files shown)


tastrax:
I think the existing ID preset for Fire Station will need to be edited if this goes ahead. It is also listed as 'Fire Station' which should probably be changed to "Fire Station Grounds" - amenity=fire_station

PR #608, opened 2022-10-05
add descriptions of bicycle parkings

matkoniecz:
based on https://github.com/streetcomplete/StreetComplete/blob/5ef8e52715326d0269d92e44644f4d5755548870/app/src/main/res/values/strings.xml#L668 https://wiki.openstreetmap.org/wiki/Key%3Abicycle_parking https://taginfo.openstreetmap.org/keys/bicycle_parking#values

### DIFF
--- a/data/fields/bicycle_parking.json
+++ b/data/fields/bicycle_parking.json
@@ -6,10 +6,10 @@
         "options": {
             "stands": "Stand (supports bicycle frame)",
             "wall_loops": "Wheelbender (supports wheel only)",
-            "shed": "Closed shed",
-            "lockers": "Individual lockers",
-            "building": "Closed in a building (not just shed)",
-            "handlebar_holder": "Handlebar holder"
+            "shed": "Closed Shed",
+            "lockers": "Individual Lockers",
+            "building": "In a Building",
+            "handlebar_holder": "Handlebar Holder"
         }
     }
 }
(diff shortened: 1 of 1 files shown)


data/fields/bicycle_parking.json
```diff
@@ -1,5 +1,15 @@
 {
     "key": "bicycle_parking",
     "type": "combo",
-    "label": "Type"
+    "label": "Type",
+    "strings": {
+        "options": {
+            "stands": "Stand (supports bicycle frame)",
+            "wall_loops": "Wheelbender (supports wheel only)",
```
tyrasd, line 8: hmm, _wheelbender_ feels a little too colloquial… an opinion from a native speaker would be appreciated on this one. //cc @1ec5 @zekefarwell @ZeLonewolf

matkoniecz: AFAIK it is both colloquial and state-of-the-art term among people who care about bicycle parking types: cyclists.

Not sure is there even more professional term for that, and 

(the same situation is in Poland, "wyrwikółka" is colloquial and at the same time used by cyclists)

And wall loops refers to some specific subtype of them but nowadays is used for not only ones like this:

screen

zekefarwell: I was not familiar with the term wheelbender, but it makes sense and is readily found with a Google search. I also don't know of a different term for this rack type.  However, it sounds like it refers to any rack that only supports the wheel, not just wall loops.
https://bikeswelcome.org/2016/12/04/wheel-benders-whats-all-the-fuss/

PR #610, opened 2022-10-08
Add field air_purifier

meNeither:
Adds field `air_purifier` to indicate the presence of an air purifier (Corsi–Rosenthal Box or the like).

### DIFF
--- a/data/fields/air_purifier.json
+++ b/data/fields/air_purifier.json
@@ -1,0 +1,10 @@
+{
+    "key": "air_purifier",
+    "type": "check",
+    "label": "Air Purifier",
+    "terms": [
+        "air cleaner",
+        "airfilter",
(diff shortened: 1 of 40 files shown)


tyrasd:
Hi. Thanks for this suggestion. However, as far as I can see, the tag `air_purifier` is neither [documented](https://wiki.openstreetmap.org/wiki/Key%3Aair_purifier) in the OSM wiki nor [in use](https://taginfo.openstreetmap.org/keys/air_purifier) in the OSM database. Usually, new tags are discussed and documented before they are introduced into tagging presets like the `id-tagging-schema` repository here. Please refer to the [proposal process](https://wiki.openstreetmap.org/wiki/Proposal_process) for one possible way to introduce a tag.

Issue #25, opened 2020-08-15
Add Cafeteria preset


TheAdventurer64:
The tag `fast_food=cafeteria` is used for a canteen, refectory or a cafeteria, a type of food service location in which there is little or no waiting staff table service, usually found within an institution such as an office building, factory, hospital, or school. 
Source: https://wiki.openstreetmap.org/wiki/Tag:fast_food%3Dcafeteria

Tag with `amenity=fast_food`and `fast_food=cafeteria`.

qugebert:
I think this can be closed as completed.

qugebert:
@tyrasd I think this can be closed as completed in #613

PR #613, opened 2022-10-11
Add preset Cafeteria

Morion-Self:
Add preset for Cafeteria.

See also #25

### DIFF
--- a/data/presets/amenity/fast_food/cafeteria.json
+++ b/data/presets/amenity/fast_food/cafeteria.json
@@ -1,0 +1,18 @@
+{
+    "icon": "maki-restaurant",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "eat",
+        "breakfast",
+        "dining",
+        "dinner"
+    ],
+    "tags": {
+        "amenity": "fast_food",
+        "fast_food": "cafeteria"
+    },
+    "name": "Cafeteria"
+}
--- a/data/presets/amenity/fast_food/cafeteria.json
+++ b/data/presets/amenity/fast_food/cafeteria.json
@@ -5,10 +5,10 @@
         "area"
     ],
     "terms": [
-        "eat",
-        "breakfast",
-        "dining",
-        "dinner"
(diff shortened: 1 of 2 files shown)


data/presets/amenity/fast_food/cafeteria.json
```diff
@@ -0,0 +1,18 @@
...
+    "icon": "maki-restaurant",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "eat",
+        "breakfast",
+        "dining",
+        "dinner"
```
tyrasd, line 11: I would replace these generic terms with some that are specific to the preset.

```suggestion
        "canteen",
        "dining hall",
        "lunchroom",
        "refectory"
```

data/presets/amenity/fast_food/cafeteria.json
```diff
@@ -0,0 +1,18 @@
...
+    ],
+    "tags": {
+        "amenity": "fast_food",
+        "fast_food": "cafeteria"
+    },
```
tyrasd, line 16: I'd include the `access` field as some cafeteria's might be not be open to the general public.

```suggestion
    },
    "fields": [
        "{amenity/fast_food}",
        "access_simple"
    ],
```
what do you think?

tyrasd (commented):
It's a bit unfortunate that there is also a second tag for these ([amenity=canteen`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcanteen)). This tag has significantly more uses, though, so I'd be fine to go with this one.

See some comments below.

data/presets/amenity/fast_food/cafeteria.json
Morion-Self: Yes, I agree

data/presets/amenity/fast_food/cafeteria.json
Morion-Self: Yes, I agree

tyrasd (approved):

PR #620, opened 2022-10-20
generator:method=solar -> generator:source=solar

Marc-marc-marc:


tyrasd:
:+1: The osm wiki lists this as a [possible tagging mistake](https://wiki.openstreetmap.org/w/index.php?title=Tag:generator:source%3Dsolar&oldid=2423177#Possible_Tagging_Mistakes). Here are some current stats for context:

* `generator:method=solar` 8k uses ([taginfo](https://taginfo.openstreetmap.org/tags/generator%3Amethod=solar))
* `generator:source=solar` 1.4M uses ([taginfo](https://taginfo.openstreetmap.org/tags/generator%3Asource=solar))

PR #629, opened 2022-11-01
give more info about situation

matkoniecz:
if we are giving background then it is worth mentioning that it is not only project documenting tagging and not even the only preset

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ This Tagging Schema fills that need, but with a number of caveats:
 - Our primary aim is to serve the needs of iD mappers (but other tools are welcome to use this too)
 - We support tags based on practicality, usage, and community approval
 - Sometimes there are reasons we can't support a tag even if it's used or approved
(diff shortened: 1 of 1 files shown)


tordans:
Is this bullet really a "caveats" for this project in the same ways the others are. I see it more as  a separate line (without bullets) below the list. As a reference in se sense "FYI you might also want to look at …".

I think the "also" can be dropped in the sentence.

And adding links to those other projects would be great, so one can actually head over there and lern more about it. (If there is no proper URL to link to, maybe its not the right fit for this list…)

matkoniecz:
> Is this bullet really a "caveats" for this project in the same ways the others are. 

Kind of?

Note "editing tools need to know how tags are used in order to facilitate mapping" - and this makes clear that there are also other major editing tools that do not use it all.

tyrasd:
I've put this into the "related projects" section of the readme.

PR #630, opened 2022-11-03
Add Tree Stump Preset and update ATM icon

### LINKED ISSUE #588: Add a preset for `natural=tree_stump`
### OSM Tag(s)

natural=tree_stump

As a Preset

### Label

Tree Stump

### Aliases

_No response_

### Terms

cut tree, removed tree

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:natural=tree_stump

### Status of the Tag

De Facto

### Usage of the tag

4 153

### Replaces other Tag?

_No response_

### Regional Tag?

_No response_

### Further Information

_No response_

### LINKED ISSUE #626: Use diffrent icons for `amenity=atm` and `amenity=bank`
Currently both presets use the same icon which makes them indistinguishable unless you click on both of them. It would be nice if they had different icons like it's done in OSM-carto.

image
image

arch0345:
* Adds a preset for [`natural=tree_stump`](https://wiki.openstreetmap.org/wiki/Tag:natural=tree_stump) (closes #588)
* Use a more specific icon for the ATM preset (closes #626)

### DIFF
--- a/data/fields/end_date.json
+++ b/data/fields/end_date.json
@@ -1,9 +1,0 @@
-{
-    "key": "end_date",
-    "type": "text",
-    "label": "End Date",
-    "placeholder": "YYYY-MM-DD",
-    "terms": [
-        "removed"
-    ]
-}
--- a/data/fields/end_date.json
+++ b/data/fields/end_date.json
@@ -1,0 +1,9 @@
+{
+    "key": "end_date",
+    "type": "text",
+    "label": "End Date",
+    "placeholder": "YYYY-MM-DD",
+    "terms": [
+        "removed"
+    ]
+}
--- a/data/presets/amenity/atm.json
+++ b/data/presets/amenity/atm.json
@@ -1,5 +1,5 @@
 {
-    "icon": "maki-bank",
+    "icon": "temaki-atm2",
     "fields": [
         "operator",
         "network",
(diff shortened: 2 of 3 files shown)


tordans:
I think 'end_date' is a great addition for the tree_stump preset to record the date (year) when the tree was cut down. 

Also, we could keep all the tree field as more_fields for tree_stump. At least in Berlin we mostly keep the ref and Denotation tags on the object since a new tree will follow in its place 1-2 years later.

1ec5:
> I think 'end_date' is a great addition for the tree_stump preset to record the date (year) when the tree was cut down.

[`end_date`](https://wiki.openstreetmap.org/wiki/Key:end_date) is currently documented as the end date of the feature itself, rather than the end date of the thing that was there (`tree:end_date`?). If there are any exceptions to this semantic, they should be noted on the key description page for awareness.

arch0345:
> If there are any exceptions to this semantic, they should be noted on the key description page for awareness.

Yeah, I think it's best to leave out the End Date field until this specific use is documented/has wider usage.

So far there are only 28 tree stumps that have an `end_date` tag:
https://overpass-turbo.eu/s/1nt6

tyrasd (approved):

PR #640, opened 2022-11-15
Add Pipeline Substation preset

arch0345:
Adds a preset for [`pipeline=substation`](https://wiki.openstreetmap.org/wiki/Tag:pipeline%3Dsubstation). I noticed that the preset for `pipeline=valve` was under `man_made/pipeline` so I decided to also add this preset here. Would it make more since for these to get moved to its own directory as they don't have `man_made=pipeline`?


tyrasd:
> Would it make more since for these to get moved to its own directory

Yes, please. All presets in a directory should be sub-presets of the corresponding parent preset (here: `man_made/pipeline`), meaning that they should have the parent's tags (here: `man_made=pipeline`) plus a number of additional tags.

I'd move the Pipeline Valve and Pipeline Substation presets into a new top-level `pipeline` directory. Optimally alongside a generic (unsearchable, low `matchScore`) preset for [other `pipeline=*`](https://taginfo.openstreetmap.org/keys/pipeline#values) objects.

arch0345:
> I'd move the Pipeline Valve and Pipeline Substation presets into a new top-level `pipeline` directory. Optimally alongside a generic (unsearchable, low `matchScore`) preset for [other `pipeline=*`](https://taginfo.openstreetmap.org/keys/pipeline#values) objects.

I noticed other unsearchable generic presets such as the one for `power=*` don't have a specified `matchScore` so I left it out

tyrasd:
Right. Some of these fallback presets do have it (e.g. the one for `man_made=*`), but it's actually redundant since the `"searchable": false` should already be enough to keep it from matching where we don't want it to match.

data/presets/pipeline/substation.json
```diff
@@ -0,0 +1,22 @@
+{
+    "icon": "temaki-powered_pump",
+    "fields": [
+        "name",
+        "substation",
```
tyrasd, line 5: This field has strings for options which are specific for electrical substations. We should probably add a dedicated `substation/pipeline` field for this with the [corresponding values](https://wiki.openstreetmap.org/wiki/Key:substation#Pipeline_dedicated_substations).

tyrasd (changes requested):
Thanks. Looks great. I only noticed one issue with the `substation` field, see below.

data/presets/pipeline/substation.json
arch0345: Thanks for catching this!

data/presets/_pipeline.json
```diff
@@ -0,0 +1,18 @@
...
+        "operator"
+    ],
+    "geometry": [
+        "point",
+        "vertex",
+        "line",
+        "area"
+    ],
+    "tags": {
+        "natural": "*"
```
tyrasd, line 14: I assume this should have been the `pipeline` tag here, right? I've totally overlooked it when reviewing the PR. :see_no_evil:

tyrasd: This was spotted in https://www.transifex.com/openstreetmap/id-editor/translate/#pl/$/442611093 by @gdabski

It's fixed in 4f70a708b3a96c02104777f1cfecef895c73970e

PR #644, opened 2022-11-17
Adding winter diesel to multi fields

tohaklim:
At first I thought I needed to edit the fields file too, but assume not?


data/fields/fuel/fuel_multi.json
```diff
@@ -8,6 +8,7 @@
             "GTL_diesel": "GTL Diesel",
             "HGV_diesel": "Truck Diesel",
             "biodiesel": "Biodiesel",
+            "diesel:class2": "Cold-weather Diesel",
```
1ec5, line 11: I don’t think this should be added to the global `fuel:*` field, because “class 2” only means winter diesel or arctic diesel in some countries. The key `diesel:class2` would be very confusing in other countries, for example the U.S., where “Diesel #2” is ordinary diesel, but there is winter blend diesel and winterized diesel, which would have to be tagged differently because it doesn’t meet the EU’s EN590 standard as described on the wiki page.

As we add fuel types specific to other countries, we should split out new field JSON files for those countries. That will allow us to tie the string more closely to the fuel’s name in each country, so that users don’t have to be well-versed in the international oil trade to tag what they see on the gas station’s signage.

tohaklim: I admit I am not familiar with all of the winterised diesel standards worldwide, however this seems like a tag applicability rather than a schema issue? Using the same argument, one could think that biodiesel and "HGV diesel" may mean different, equally confusing, things to mappers in different countries. This will not be resolved until a single standard is implemented - which is unrealistic. From memory, there was similar confusion about octane tagging and how this is assessed in the US.
The wiki definition seems quite broad, and can be applied to various winterised diesels, however the tag only needs to be specific enough for a mapper seeing "Winter diesel" to be able to tag that. "class2" is a more generic term, looking at tag usage worldwide, than may seem.

1ec5: The difference with `biodiesel` and `hgv_diesel` is that we aren’t taking a more general concept and applying a key that says something highly specific to a particular country or region’s regulations. The analogy with octane numbers supports my point: U.S. mappers are tagging the signposted octane numbers as is, without converting them to European octane numbers (which would be impossible).

[The wiki table](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2429851#How_to_map) currently defines `fuel:diesel:class2` as:

> [Diesel for colder weather climates](https://en.wikipedia.org/wiki/Winter_diesel_fuel#Arctic_diesel) (with a threshold filtration temperature of −32 °C), on the EU EN590 scale. Can be branded as "winter diesel", "Arctic diesel", "diesel −32 °C", etc., depending on the location.

It goes on to describe a `fuel:diesel:class0` for certain winter diesels. You may consider the definition to be broad, but it’s tied to an EU standard. In regions that make different distinctions, no one would be able to map these keys back to a real-world fuel type for end users.

`fuel:diesel:class2` is only being used in Europe, with the exception of [one gas station in the U.S.](https://www.openstreetmap.org/way/824731205) where the mapper probably didn’t fully comprehend the definition on the wiki.

To be clear, I see no problem with adding `fuel:diesel:class2` to fuel_multi-RU.json, but adding it to fuel_multi.json would make it more difficult for us to promote common-sense tagging outside Europe. I think you should be able to add a `strings` object to fuel_multi-RU.json that contains just the string for `diesel:class2`, and schema-builder should continue to cross-reference fuel_multi.json for all the other strings.

tohaklim: See your point (although I think "Truck diesel" for HGV diesel is even worse of a generalisation. To me, the wiki reads like it would apply to any diesel with a filtration temp of -32.
I've changed the commit to only touch the RU multi box (assume it is specific to Russia, they use the same standard)

tohaklim: Actually, thinking about it, how about adding this with a more specific label eg "EU winter diesel", given there is no EU JSON that I see (Russia seems to be the only one)

1ec5: > To me, the wiki reads like it would apply to any diesel with a filtration temp of -32.

That’s part of the conundrum. The U.S. and Canada [define](https://en.wikipedia.org/wiki/Winter_diesel_fuel#United_States) winter blend diesel in terms of relative temperature, not on an absolute temperature scale

>  I've changed the commit to only touch the RU multi box (assume it is specific to Russia, they use the same standard)

You’ll want to also add the `strings` object to this file as I mentioned above. Otherwise, there would be no opportunity to translate `class2` to Russian in Transifex.

> Actually, thinking about it, how about adding this with a more specific label eg "EU winter diesel", given there is no EU JSON that I see (Russia seems to be the only one)

We could create a field for the whole EU just like there is for Russia. The `locationSet` property is evaluated by [country-coder](https://github.com/ideditor/country-coder/), which supports `EU` as a location.

tohaklim: Getting out of my depth here - is copying the `strings` section from global Fuel Multi enough, with class2 added?

Given how I am not familiar with the schema to be able to tell, what other dependencies there are for creating a location-specific multibox

data/fields/fuel/fuel_multi-RU.json
```diff
@@ -2,6 +2,37 @@
...
+            "octane_100": "Gasoline (100 Octane)",
+            "e10": "E10 Gasoline",
+            "e85": "E85 Gasoline",
+            "lpg": "Liquefied Natural Gas (LPG)",
+            "cng": "Compressed Natural Gas (CNG)",
+            "propane": "Propane",
+            "LH2": "Liquid Hydrogen",
+            "h70": "Gaseous Hydrogen (700 bar)",
+            "adblue": "AdBlue / AUS 32",
+            "adblue:canister": "AdBlue / AUS 32 (sold in canisters)"
```
1ec5, line 33: @tyrasd, do you know if it’s necessary to copy the whole object here? Does `stringsCrossReference` merge the cross-referenced object with the object defined here? Copying all the strings here would create some extra work for translators – not too much right now, but it could become more pronounced as we add more country-specific versions of this field.

tyrasd:
This value does not seem to be [approved](https://wiki.openstreetmap.org/w/index.php?oldid=1146539), nor does it have [much traction](https://taginfo.openstreetmap.org/keys/fuel%3Adiesel%3Aclass2#chronology)[^1] in usage.

[^1]: While looking at the existing data with this tag, I noticed that most of the mapped features with this tags are originating from an [import](https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_merge_fuel_IT.py) in Italy, where I sincerely have serious doubts about its data quality. :unamused: For example, from personal experience, I'm certain that [this fuel station](https://www.openstreetmap.org/node/609238256) (and all other fuel stations in the wider region) does not specifically sell any kind of "winter" diesel in addition to "regular" diesel (as the tags would imply). It's likely that these fuel stations (seasonally) have diesel on offer which does meet a certain low-temperature "class". But the fuel is not labeled or advertised as such. It's also unclear of which specific class this "winter diesel" in Italy actually would be. I found some [spec](https://oilproducts.eni.com/media/pdf/download/e-C1331_ENU.pdf) [sheets](https://www.tamoil.it/-/media/Project/Common/PDF-Schede-prodotto/GASOLIO-ARTICO.pdf) for diesel products from some well known Italian refineries, which match the [fuel type strings](https://github.com/osm-fr/osmose-backend/commit/cde6b8fad7dc9ea34bf4c74e96887248a6c30c88#diff-e676ce08da7e2710e4ab65e30dd55dd42171d9dfa71a5cba5b934b8acf1db2feR155-R156) used in the import, but these types of diesel by far do not meet EN 590's class 2 requirements: they have a of CFPP value of `-18°C` to `-21°C`, wile EN590 class 2 would require a CFPP of `-32°C` and cloud point of `-6°C` where a class 2 diesel would need to b `-22°C`. :shrug: 

I'd recommend to first discuss on some community channel (like the tagging mailing list) if this should be mapped, and if yes, how such variants of diesel fuel should be tagged.

data/fields/fuel/fuel_multi-RU.json
tyrasd: Well, `stringsCrossReference` indicates that the client should look up the whole `strings` property of the referenced field. Currently it's undefined what should happen when one defines both `stringsCrossReference` and `strings` in a field (FWIW, iD would currently throw away the locally defined strings).
The currently intended way for this to be used is that the global field defines all strings, and the regional variants of the field then only restrict the allowed options by defining their own `options` property.[^1]

[^1] In the case where (some) options do not fit for the global case, the strings can still be defined there, but the global field would also need to define its own `options` list.

1ec5:
Thanks for the detailed investigation. Folks in OSMUS Slack [confirmed](https://osmus.slack.com/archives/C2VJAJCS0/p1668723084953369?thread_ts=1668710567.662689&cid=C2VJAJCS0) that gas stations don’t offer a choice between winter and summer diesel blends; the standards require them to offer all diesel up to that standard based on the time of year. I [proposed on the wiki](https://wiki.openstreetmap.org/wiki/Template_talk:Fuel_types#Winter_diesel) that we remove any mention of these subkeys.

PR #648, opened 2022-11-20
This add amenity=stripclub

EvanCarroll:
There is already an issue opened on FAS for a better icon, this still looks appropriate for the time being.

https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dstripclub

Issue for better icon:

https://github.com/FortAwesome/Font-Awesome/issues/8254

### DIFF
--- a/data/presets/amenity/stripclub.json
+++ b/data/presets/amenity/stripclub.json
@@ -1,0 +1,40 @@
+{
+    "icon": "fas-place-of-worship",
+    "fields": [
+        "name",
+        "operator",
+        "address",
+        "building_area_yes",
+        "opening_hours",
+        "min_age",
+        "smoking"
+    ],
+    "moreFields": [
+        "air_conditioning",
+        "email",
+        "fax",
+        "gnis/feature_id-US",
+        "level",
+        "payment_multi",
+        "opening_hours/covid19",
+        "phone",
+        "ref/FR/siret-FR",
+        "wheelchair",
+        "fee"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "amenity": "stripclub"
+    },
+    "terms": [
+        "lap dance",
+        "adult*",
+        "gentlemen*",
+        "cabaret",
+        "strip*"
+    ],
+    "name": "Strip club"
+}
--- a/data/presets/amenity/stripclub.json
+++ b/data/presets/amenity/stripclub.json
@@ -1,5 +1,5 @@
 {
-    "icon": "fas-place-of-worship",
+    "icon": "roentgen-pole_dancer",
     "fields": [
         "name",
         "operator",


data/presets/amenity/stripclub.json
```diff
@@ -0,0 +1,40 @@
+{
+    "icon": "fas-place-of-worship",
```
tyrasd, line 2: I don't think [this icon](https://fontawesome.com/icons/place-of-worship?s=solid&f=classic) is a good match for this tag.

data/presets/amenity/stripclub.json
```diff
@@ -0,0 +1,40 @@
...
+    ],
+    "tags": {
+        "amenity": "stripclub"
+    },
+    "terms": [
+        "lap dance",
+        "adult*",
+        "gentlemen*",
+        "cabaret",
+        "strip*"
```
tyrasd, line 37: what's the purpose of the asterisks `*` here?

tyrasd (changes requested):

data/presets/amenity/stripclub.json
EvanCarroll: There are lots of things that begin with gentlemen or adult, and many of them are strip clubs? I assumed from nightclub.json that the asterisks were required. https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/nightclub.json#L33

data/presets/amenity/stripclub.json
EvanCarroll: Looks like a strip club to me. Big door, no windows, big grand entrance. The icon itself is a pretty close match.

stripclub

What is your reservation? Is there a better suggestion?

1ec5: We’re trying not to reuse preset icons so much, especially for quite unrelated features: #41. Sounds like you’re aware of the need for something more specific: https://github.com/FortAwesome/Font-Awesome/issues/8254#issuecomment-1321023839. The icon can come from [a few sources besides FontAwesome](https://github.com/ideditor/schema-builder/#icon), including a custom-drawn icon if you prefer to draw something yourself.

bhousel: I'd just use `temaki-money_hand` until something better comes along.

temaki-money_hand

EvanCarroll: `temaki-money_hand` doesn't look at all like a strip club. Normally, afaik, you don't hand strippers money either. You either (a) make it rain, or (b) affix the money to the stripper. Or so I've read.

But more important `temaki-money_hand` would also be reused.

* https://github.com/openstreetmap/id-tagging-schema/search?q=temaki-money_hand

So the question remains, what icon should I use? I have 0 artistic skill.

bhousel: > temaki-money_hand doesn't look at all like a strip club. 

And your strip club above looks like a museum.

maki-museum

1ec5: Make an icon that looks like [XXX]. Simple.

1ec5: Far too detailed; doesn’t scale well to the _very_ small sizes that these icons can appear in. These icons will be shown to everyone, not just those actively seeking out strip clubs, in the event that someone encounters one that’s already mapped. They’ll need to have some general idea of what it means independently of the preset name.

Do you have an objection to my suggestion of a simple text icon? If you’re unable to create an image that consists of the string “XXX”, then someone else can create it for you.

EvanCarroll: Does this one work? https://commons.wikimedia.org/wiki/File:Pornproject_logo.svg

1ec5: Something along those lines, yes, but it would need to be in black and white, similar to the other icons. (Alternatively, there’s already [an “X” icon for railroad crossings](https://ideditor.github.io/temaki/docs/#icon-temaki-x_oblique) that could be modified for this purpose.) [Temaki](https://github.com/ideditor/temaki/) would be the place to submit a new icon.

data/presets/amenity/stripclub.json
tyrasd: I see.  :thinking: However, I cannot find any reference in the code where anything special is done with the `*` character in preset terms. I would assume that in the early prototype versions of iD, this was supposed to indicate that iD should match all search input starting with the respective term (e.g. that searching for `discoteque` would find the `amenity=nightclub` preset which was `disco*` as a term). However, as far as I can see, this was never implemented.  @bhousel does that sound right, or is some other reason for the `*` in the terms of some presets?

data/presets/amenity/stripclub.json
1ec5: Yes, I think either your suggestion (with the color modification I mentioned) or @tyrasd’s suggestion could work. I just didn’t want us to overthink it. You’re right that many of the icons aren’t unique yet – it’s something we’re working towards, but new presets should at least avoid reusing an icon that’s completely unrelated. We’ve already have several instances where FontAwesome dramatically changes the look of an icon while keeping it semantically consistent with the file name.

EvanCarroll: I can also open the request in Temaki for a new icon. Not being a hardass on the art stuff either. I just suck with the open source software on it. I spent 2 hours trying to figure out how to draw a box, asked a question on it, and 12 years later its since helped 150k people: [How do I draw a box in GIMP?](https://superuser.com/questions/167873/how-do-i-draw-a-box-in-gimp). I still have to google for how to draw a box in GIMP. GIMP is terrible, and I don't have a VM with photoshop. =(

1ec5: Ah, I feel you. I find vector drawing in Inkscape to be much more intuitive, and as it turns out, projects like Temaki and FontAwesome require vector graphics in SVG format.

EvanCarroll: I've actually only ever heard good things about Inkscape, I'll check it out. It's open source. As long as it's not gimp.

data/presets/amenity/stripclub.json
bhousel: > However, as far as I can see, this was never implemented. @bhousel does that sound right, or is some other reason for the * in the terms of some presets?

I'm not sure , using `*` as a wildcard might have been supported a very long time ago, but I don't think the `*`'s do anything now.

Issue #529, opened 2022-07-12
popular unsupported values of major keys


matkoniecz:
Based on taginfo - so some may be fixed already (I noticed and removed `barrier=log`). Some have open PRs already. Some would be wrong to include, some would require checking.

Generated with https://github.com/matkoniecz/taginfo_api_wrapper_in_python#listing-popular-tags-unsupported-by-id

It is usable for review. But using it requires some thinking. For example I like `shop = estate_agent` and is in my opinion preferable to `office = estate_agent` (as office=estate_agent and other office=* is used both for places inaccessible to general population such as internal offices - and for ones accessible to general population, but changing id presents requires some investigation what other mappers think).

Note that some highly used values, are from imports of a low quality.

If that list is useful - then more systematic reviewing like JOSM is doing may be useful.

Other potential comparisons: list tags supported by JOSM or Vespucci and not supported by iD.

[building = outbuilding 300k](https://taginfo.openstreetmap.org/tags/building=outbuilding)
[building = allotment_house 115k](https://taginfo.openstreetmap.org/tags/building=allotment_house)
[shop = grocery 4k](https://taginfo.openstreetmap.org/tags/shop=grocery)
[shop = estate_agent 4k](https://taginfo.openstreetmap.org/tags/shop=estate_agent)
[shop = convenience;gas 2k](https://taginfo.openstreetmap.org/tags/shop=convenience;gas)
[shop = food 2k](https://taginfo.openstreetmap.org/tags/shop=food)
[shop = ice_cream 2k](https://taginfo.openstreetmap.org/tags/shop=ice_cream)
[shop = glaziery 1k](https://taginfo.openstreetmap.org/tags/shop=glaziery)
[shop = printing 1k](https://taginfo.openstreetmap.org/tags/shop=printing)
[natural = land 68k](https://taginfo.openstreetmap.org/tags/natural=land)
[natural = crevasse 36k](https://taginfo.openstreetmap.org/tags/natural=crevasse)
[natural = sinkhole 26k](https://taginfo.openstreetmap.org/tags/natural=sinkhole)
[natural = hill 24k](https://taginfo.openstreetmap.org/tags/natural=hill)
[natural = shrubbery 9k](https://taginfo.openstreetmap.org/tags/natural=shrubbery)
[natural = earth_bank 9k](https://taginfo.openstreetmap.org/tags/natural=earth_bank)
[natural = landform 7k](https://taginfo.openstreetmap.org/tags/natural=landform)
[...]

PR #649, opened 2022-11-21
Add gantry presets

arch0345:
This adds a preset for two of the tags listed in #529: `man_made=gantry`, and `power=portal`.

Although the [wiki recommends `destination` as a useful combination](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dgantry), less than 10 features tagged with `man_made=gantry` currently have this tag.

I noticed that [`gantry:type`](https://taginfo.openstreetmap.org/keys/gantry%3Atype) and [`gantry`](https://taginfo.openstreetmap.org/keys/gantry#overview) might be something we could add to `fields`/`moreFields`, but these keys aren't currently documented.

I've set the icon to `temaki-toll_gantry` for now, but I think it would be better for this preset to have its own icon to better distinguish them.

Edit: this PR also adds presets for `crane:type=gantry_crane`, `crane:type=portal_crane`, and `power=portal`

### DIFF
--- a/data/presets/man_made/crane/gantry_crane.json
+++ b/data/presets/man_made/crane/gantry_crane.json
@@ -1,0 +1,21 @@
+{
+    "icon": "temaki-crane",
+    "geometry": [
+        "point",
+        "vertex",
+        "area"
+    ],
+    "tags": {
+        "man_made": "crane",
+        "gantry:type": "gantry_crane"
+    },
+    "reference": {
+        "key": "gantry:type",
+        "value": "gantry_crane"
+    },
+    "terms": [
+        "overhead crane",
+        "bridge crane"
+    ],
+    "name": "Gantry Crane"
+}
--- a/data/presets/man_made/crane/gantry_crane.json
+++ b/data/presets/man_made/crane/gantry_crane.json
@@ -7,10 +7,10 @@
     ],
     "tags": {
         "man_made": "crane",
-        "gantry:type": "gantry_crane"
+        "crane:type": "gantry_crane"
     },
     "reference": {
-        "key": "gantry:type",
+        "key": "crane:type",
         "value": "gantry_crane"
     },
     "terms": [
--- a/data/presets/man_made/crane/portal_crane.json
+++ b/data/presets/man_made/crane/portal_crane.json
@@ -2,14 +2,15 @@
     "icon": "temaki-crane",
     "geometry": [
         "point",
+        "vertex",
         "area"
     ],
     "tags": {
         "man_made": "crane",
-        "gantry:type": "portal_crane"
+        "crane:type": "portal_crane"
     },
     "reference": {
-        "key": "gantry:type",
+        "key": "crane:type",
         "value": "portal_crane"
     },
     "terms": [
--- a/data/presets/man_made/crane/portal_crane.json
+++ b/data/presets/man_made/crane/portal_crane.json
@@ -1,0 +1,19 @@
+{
+    "icon": "temaki-crane",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "man_made": "crane",
+        "gantry:type": "portal_crane"
+    },
+    "reference": {
+        "key": "gantry:type",
+        "value": "portal_crane"
+    },
+    "terms": [
+        "level-luffing crane"
+    ],
+    "name": "Portal Crane"
+}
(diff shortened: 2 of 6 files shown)


data/presets/man_made/gantry.json
```diff
@@ -0,0 +1,26 @@
...
+    ],
+    "tags": {
+        "man_made": "gantry"
+    },
+    "terms": [
+        "gantry",
+        "highway gantry",
+        "sign holder",
+        "road sign holder",
+        "sign structure"
```
1ec5, line 23: “Mast arm” refers to just the horizontal part, but this is what people often think of these structures as. “Sign assembly” is also relevant.

data/presets/man_made/gantry.json
```diff
@@ -0,0 +1,26 @@
...
+    ],
+    "name": "Road Sign Structure"
```
1ec5, line 25: Do we need a more generic name, such as “gantry”? The [original proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/Gantry) suggests using this tag for railroad gantries as well as road sign gantries, and it may not be obvious that a “road sign structure” can hold traffic lights, an electronic toll collection device, or a traffic counter.

arch0345: I changed the name to `Gantry` and added presets for `power=portal` (gantry tower) and `crane:type=gantry_crane`

data/presets/man_made/crane/gantry_crane.json
```diff
@@ -0,0 +1,21 @@
+{
+    "icon": "temaki-crane",
+    "geometry": [
+        "point",
+        "vertex",
+        "area"
+    ],
+    "tags": {
+        "man_made": "crane",
+        "gantry:type": "gantry_crane"
```
tyrasd, line 10:
```suggestion
        "crane:type": "gantry_crane"
```

data/presets/man_made/crane/gantry_crane.json
```diff
@@ -0,0 +1,21 @@
...
+    },
+    "reference": {
+        "key": "gantry:type",
```
tyrasd, line 13:
```suggestion
        "key": "crane:type",
```

data/presets/man_made/crane/portal_crane.json
```diff
@@ -0,0 +1,19 @@
+{
+    "icon": "temaki-crane",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "man_made": "crane",
+        "gantry:type": "portal_crane"
```
tyrasd, line 9:
```suggestion
        "crane:type": "portal_crane"
```

data/presets/man_made/crane/portal_crane.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+    "reference": {
+        "key": "gantry:type",
```
tyrasd, line 12:
```suggestion
        "key": "crane:type",
```

data/presets/man_made/crane/portal_crane.json
```diff
@@ -0,0 +1,19 @@
+{
+    "icon": "temaki-crane",
+    "geometry": [
+        "point",
+        "area"
```
tyrasd, line 5: Is there a particular reason why this crane shouldn't be allowed on way nodes?

```suggestion
        "point",
        "vertex",
        "area"
```

data/presets/power/portal.json
```diff
@@ -0,0 +1,29 @@
+{
+    "icon": "temaki-power",
+    "fields": [
+        "ref",
+        "operator",
+        "structure",
+        "design",
```
tyrasd, line 7: I'd move the `design` field to `moreTags`. The values for this tag require quite some domain knowledge, and are not something a typical mapper is expected to enter.

data/presets/power/portal.json
```diff
@@ -0,0 +1,29 @@
+{
+    "icon": "temaki-power",
+    "fields": [
+        "ref",
+        "operator",
+        "structure",
```
tyrasd, line 6: I don't think this is the right field for this. :wink:  The `structure` is the field for specifying bridge/tunnel/etc. on highways and similar features.  Here, only the tag actually has the `structure` key and the allowed values are `lattice`, `solid`, and `tubular`. We'd need a new field for this.

tyrasd (changes requested):
I think you mixed up the tag keys `gantry:type` and `crane:type` in this PR. See suggested changes below. See also some additional questions below.

data/presets/power/portal.json
arch0345: Thanks for catching this! I added a `structure_power` field for this

tyrasd (approved):

PR #652, opened 2022-11-25
Update maxstay options

Zaczero:
The unit should not be abbreviated - from [wiki documentation](https://wiki.openstreetmap.org/wiki/Key:maxstay?uselang=en#Time_unit).

Ref.
https://github.com/openstreetmap/iD/issues/9387


tyrasd:
Thanks!

PR #666, opened 2022-12-04
add preset for Variable Message Sign

k-yle:
[they're quite common](https://taginfo.osm.org/tags/traffic_sign=variable_message), and we already have presets for "Radar Speed Display" and "Digital Screen" so it's helpful to have this preset which fills the gap between the other two.


tyrasd:
thx! :smiling_imp:

PR #668, opened 2022-12-04
Add oneway field to cycleway uncontrolled crossing

kjonosm:


Dimitar5555:
It looks like this PR is a duplicate of #667.

PR #673, opened 2022-12-08
parking/orientation: update and deprecate key

tordans:
This PR is for ~next week when [the voting](https://wiki.openstreetmap.org/wiki/Proposed_features/street_parking_revision#Voting) was officially concluded.

---

Following the successful street parking revision proposal https://wiki.openstreetmap.org/wiki/Proposed_features/street_parking_revision#Summary:_What_is_proposed,_changed_and_deprecated? the tag for orientation on separately mapped parking spaces is now just `orientation`.

> Regarding separately mapped parking areas, (…)
> [parking:orientation](https://wiki.openstreetmap.org/wiki/Key:parking:orientation)=* will be deprecated and replaced by [orientation](https://wiki.openstreetmap.org/wiki/Key:orientation)=*(…)

---

I added duplicated the deprecation rule for `orthogonal` to handle both the old and new key because I was not sure about the order (cardinality) of the rules / if the key deprecation would then in turn still trigger the value deprecation. Having all three should not be an issue I guess.

---

FYI @SupaplexOSM

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1174,6 +1174,14 @@
     "replace": {"parking": "lane"}
   },
   {
+    "old": {"parking:orientation": "*"},
+    "replace": {"orientation": "$1"}
+  },
+  {
+    "old": {"orientation": "orthogonal"},
+    "replace": {"orientation": "perpendicular"}
+  },
+  {
     "old": {"parking:orientation": "orthogonal"},
     "replace": {"parking:orientation": "perpendicular"}
   },
--- a/data/fields/parking/orientation.json
+++ b/data/fields/parking/orientation.json
@@ -1,5 +1,5 @@
 {
-    "key": "parking:orientation",
+    "key": "orientation",
     "type": "combo",
     "label": "Orientation",
     "strings": {


tyrasd:
> Decide how to handle the folder.

I'd prefer if `<key>:<subkey>` fields actually end up as `key/subkey.json`. Yes, it would make sense to also include other related tags in the same directory, but these should then get a `<key>_<discriminator>.json` file name. Hmm… maybe here something like `orientation_parking.json` could work?!

tordans:
@tyrasd I move the field to `fields/orientation/orientation_parking.json`. I could move it to `fields/orientation_parking.json`, if you prefer.

I also updated the PR description, since the proposal is accepted and all docs are updated. The data migration is already going strong.

kaneap:
Oops, I was only searching in issues 😅

Zaczero:
This tag is now officially deprecated but iD keep on using it. What's the blocker here?

tordans:
@tyrasd I think this can be merged.

We where asked about this recently https://github.com/osmberlin/parkraum.osm-verkehrswende.org/issues/30#issuecomment-1473718318

tyrasd (approved):

PR #674, opened 2022-12-08
Add field parking:$side and parking:$side:orientation

tordans:
Solves https://github.com/openstreetmap/id-tagging-schema/issues/675

---

### Update 2022-12-29

I created https://github.com/openstreetmap/id-tagging-schema/pull/722 (and https://github.com/openstreetmap/id-tagging-schema/pull/723) as minimal versions of this PR. More about that over there.

I consider this PR a collection of TODOs and know how. Once the issues are solved, we should create separate PRs.

### Update 2022-12-28

**This part is a must have:** 

- [ ] The schema builder needs to be released https://github.com/ideditor/schema-builder/pull/83
  - [ ] Then: This PR needs an update, including a cleanup depending on the state of the issues below

**This part is a nice to have and IMO can be follow up tickets:**

- The `:both` case needs to be solved, see https://github.com/openstreetmap/iD/issues/9212#issuecomment-1366655208 for some more info about it
   - [ ] For this PR, supporting `:left`, `:right`, `:both` would be sufficient. 
   - [ ] For the linked issue there is also the `:<null>` and `:<null>+oneway=yes` case, which makes it quite a bit more complex, so maybe we can split this in two issues?
   - [ ] The `cycleway` field will need a review once this is solved https://github.com/openstreetmap/id-tagging-schema/pull/674/commits/ede72b76aaa8b3b1d63730300e4e852b166c0007.
- [ ] The `prerequisiteTag` needs to be figured out
   - See comment below for more.
   - See also https://github.com/openstreetmap/id-tagging-schema/pull/674/commits/c4e218a63689a3d648e4fef3c84cf015df8d81ba
- [ ] We should to add "descriptions" to the fields, see https://github.com/openstreetmap/id-tagging-schema/pull/674/commits/d9fc26a4e713492162e81d5063d6a5ca6456b6a2
   - @SupaplexOSM can you help with this part?
   - The `description` shows as a tooltip when hovering the dropdown option 
   Bildschirm­foto 2022-12-28 um 14 13 15

### DIFF
--- a/data/fields/parking/parking_side_orientation.json
+++ b/data/fields/parking/parking_side_orientation.json
@@ -5,11 +5,26 @@
         "parking:righ:orientationt"
     ],
     "type": "directionalCombo",
+    "label": "Parking orientation",
     "strings": {
+        "types": {
+            "parking:left:orientation": "Left side",
+            "parking:right:orientation": "Right side",
+            "parking:both:orientation": "Both sides"
+        },
         "options": {
-            "parallel": "Parallel to the Street",
-            "diagonal": "Diagonal in Relation to the Street",
-            "perpendicular": "Meets the Street at a Straight Angle"
+            "parallel": {
+                "title": "Parallel to the Street",
+                "description": "Parking "
+            },
+            "diagonal": {
+                "title": "Diagonal in Relation to the Street",
+                "description": ""
+            },
+            "perpendicular": {
+                "title": "Meets the Street at a Straight Angle",
+                "description": ""
+            }
         }
     },
     "autoSuggestions": false,
--- a/data/fields/parking/parking_side_orientation.json
+++ b/data/fields/parking/parking_side_orientation.json
@@ -27,6 +27,13 @@
             }
         }
     },
+    "prerequisiteTag": {
+        "keys": [
+            "parking:both",
+            "parking:left",
+            "parking:right"
+        ]
+    },
     "autoSuggestions": false,
     "customValues": false
 }
--- a/data/fields/parking/parking_side_orientation.json
+++ b/data/fields/parking/parking_side_orientation.json
@@ -1,0 +1,14 @@
+{
+    "key": "parking:$side:orientation",
+    "type": "combo",
+    "label": "Type",
+    "strings": {
+        "options": {
+            "parallel": "Parallel to the Street",
+            "diagonal": "Diagonal in Relation to the Street",
+            "perpendicular": "Meets the Street at a Straight Angle"
+        }
+    },
+    "autoSuggestions": false,
+    "customValues": false
+}
--- a/data/fields/parking/parking_side_orientation.json
+++ b/data/fields/parking/parking_side_orientation.json
@@ -1,6 +1,9 @@
 {
-    "key": "parking:$side:orientation",
-    "label": "Type",
+    "keys": [
+        "parking:both:orientation",
+        "parking:left:orientation",
+        "parking:righ:orientationt"
+    ],
     "type": "directionalCombo",
     "strings": {
         "options": {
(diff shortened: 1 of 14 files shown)


data/fields/parking/parking_side_orientation.json
```diff
@@ -0,0 +1,14 @@
...
+    "type": "combo",
+    "label": "Type",
+    "strings": {
+        "options": {
+            "parallel": "Parallel to the Street",
+            "diagonal": "Diagonal in Relation to the Street",
+            "perpendicular": "Meets the Street at a Straight Angle"
+        }
+    },
+    "autoSuggestions": false,
```
tordans, line 12: I noticed in iD, that when I add `population`, that also the `population:source` and `check_date:population` is added to the fields automatically. I think that feature would be a good fit for this key as well. So I guess we need to add …

```suggestion
    "prerequisiteTag": {
        "key": "parking:$side"
    },
    "autoSuggestions": false,
```

tyrasd:
> How to handle `$side` within a present field?

In principle, this would be a good fit for the `cycleway` field type (see the [_Special_](https://github.com/ideditor/schema-builder#special) section in the schema -builder docs). 

image

<details>
  <summary>
  The field json would look like this:</summary>
  
  ```json
{
    "keys": [
        "parking:both",
        "parking:left",
        "parking:right"
    ],
    "type": "cycleway",
    "label": "Street Parking Type",
    "strings": {
        "types": {
            "parking:left": "Left side",
            "parking:right": "Right side"
        },
        "options": {
            "lane": "Roadside Lane",
            "street_side": "Street-Side",
            "on_kerb": "On Kerb",
            "half_on_kerb": "Half On Kerb",
            "shoulder": "Shoulder",
            "no": "No",
            "separate": "Parking mapped separately",
            "yes": "Yes (unspecified)"
        }
    },
    "autoSuggestions": false,
    "customValues": false
}
  ```
</details>

However, the [implementation](https://github.com/openstreetmap/iD/blob/v2.23.2/modules/ui/fields/cycleway.js#L39) is pretty much hardcoded for the literal `cycleway:*` key on iD's side. It wouldn't be too hard to generalize this into a reusable (`"type": "directional"`) field type.

For this type of field, it would make sense if the `prerequisiteTag` feature would also look at the corresponding `$side` subtags: e.g., if `parking:left` is `no` or empty and `parking:right` is set, only the `parking:right:orientation` input field should be enabled. I'm however not jet sure how to best model it in the schema. Probably it would make the most sense to be able to reference another `field` as a prerequisite in such cases:

```json
"prerequsiteTag": {
    "field": "parking/side"
}
```

tordans:
@tyrasd thanks for the input in your comment above and the update on iD. I updated [this PR-description](https://github.com/openstreetmap/id-tagging-schema/pull/674#issue-1484890414) with a list of things that need to happen next from my point of view.

Would you be up to merging this with "must have" solved? And handle the other issues as follow ups?

---

On the topic of `prerequisiteTag`:

This needs an update to the [schema-builder](https://github.com/ideditor/schema-builder). But how to specify best, what we want? IMO, we want…

> The field should appear if …
> - any of the keys `parking:left`,`parking:right`,`parking:both` is present
> - AND the values are _not_ `separate`, `no`

We need the "_not_" part, since we don't users to provide an `:orientation` when so side was given.

This has two issues:

- We need to check combinations of key+value, right? Only if all given keys have a value of `separate`, `no`, the field should be _hidden_. Otherwise …

- If `parking:right=no|separate`, then `parking:right:orientation` should not be visible. Which would be another feature of the `directionalCombo`. Either the field should be deactivated; or a validation should warn users once they added data that this is (possible/doable but) wrong.

   Maybe we could work around this, by only showing `:orientation` `directionalCombo` fields if `:right`, `:left` is given, and a regular dropdown if `:both` is given.

---

Let me know what you think.

tordans:
I will close this now in favor of https://github.com/openstreetmap/id-tagging-schema/pull/744.

It will be a while until the `prerequisiteTag` is extended enough to add more fancyness to this feature. And likely https://github.com/openstreetmap/iD/issues/9476 will be a first step by manually handling those cases in iD.

PR #679, opened 2022-12-10
Remove "line" geometry from lift gate

Discostu36:
Lift gates should only be nodes ([wiki](https://wiki.openstreetmap.org/wiki/Tag%3Abarrier%3Dlift_gate)). This error is probably the cause of a validation issue, where ID suggests converting an orphaned lift_gate node into a way instead of suggesting it to be added as a node to a way.


tyrasd (approved):

tyrasd:
Thanks!

Issue #677, opened 2022-12-09
update form from site_type to archaeological_site


b-unicycling:
With the deprecation of `site_type` for archaeological sites, it would be helpful, if the iD editor would reflect that and the [new key](https://wiki.openstreetmap.org/wiki/Proposed_features/Key:archaeological_site) would be used.
I don't know if that happens automatically, but when looking at a dataset, the `archaeological_site` should also be displayed as a link like `site_type` is now.

Thanks!

(1st GitHub post ever, I hope I did everything right.)

PR #682, opened 2022-12-10
Update archaeological_site.json

kjonosm:
Replace field "site_type" with "archaeological_site" See #677

### DIFF
--- a/data/presets/historic/archaeological_site.json
+++ b/data/presets/historic/archaeological_site.json
@@ -2,7 +2,7 @@
     "icon": "temaki-ruins",
     "fields": [
         "name",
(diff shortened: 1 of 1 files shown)


tordans:
I was wondering if this is also a case for a deprecation and thus automatic change of the tag by iD. Otherwise after this change, existing data will not be visible anymore. And if no deprecation is desired, maybe we need a non-searchable field (is there such a thing) for the old data?

tordans:
Update: I now see https://github.com/openstreetmap/id-tagging-schema/pull/680/files

BTW, it's a bit confusing to have separate PRs for one issue.

PR #695, opened 2022-12-19
Update craft.json

kjonosm:
Fix typo


tyrasd:
Thanks for catching this!

PS: I got confused by the preset for the "Insulator" craft being called [`craft/insulator.json`](https://github.com/openstreetmap/id-tagging-schema/blob/v5.1.1/data/presets/craft/insulator.json), when it should actually be named `craft/insulation.json` to match the tag value. I'll leave it as is though, in order to not unnecessarily invoke re-translations on the preset.

PR #699, opened 2022-12-21
Add translatable strings to traffic_calming type field

kjonosm:


data/fields/traffic_calming.json
```diff
@@ -7,5 +7,22 @@
...
         "speed",
         "slow"
-    ]
+    ],
+    "strings": {
+        "options": {
+            "bump": "Speed Bump",
+            "chicane": "Traffic Chicane",
+            "choker": "Traffic Choker",
+            "choked_island": "Choked Traffic Island",
```
tyrasd, line 16: Usage for this value is currently pretty low (73 occurrences as of today).

At first glance, the wording of this value could also be slightly confusing as (almost) every traffic island is inherently also a choke point, as the road lanes need to make space for the width of the traffic island. :thinking:

I've looked through a sample of places where this tag is currently used, and only found one place which I would classify as a valid use of this tag value ([node 7183898498](https://www.openstreetmap.org/node/7183898498), which is also the example used to illustrate this on the [OSM wiki](https://wiki.openstreetmap.org/wiki/Key:traffic_calming#By_causing_lateral_deviation)). All other examples I found were either misclassified regular choke points or regular traffic islands: [1](https://www.openstreetmap.org/node/347258622), [2](https://www.openstreetmap.org/way/925418954), [3](https://www.openstreetmap.org/node/7198597729), [4](https://www.openstreetmap.org/node/5012530242), [5](https://www.openstreetmap.org/node/9961619902), [6](https://www.openstreetmap.org/node/6145859060), [7](https://www.openstreetmap.org/node/9160032665), [8](https://www.openstreetmap.org/node/303870857), [9](https://www.openstreetmap.org/node/577193004).

I'd say it's probably better to omit this value for now.

PR #704, opened 2022-12-21
Update utility.json

kesterlester:
A significant distinguisher for many UK gas utility markers is colour (red vs yellow for national versus local) and material (concrete vs plastic).  I have had to add these fields to markers on probably more than on thousand occasions over the last few months while trying to edit the UK in iD, so I thought it was about time I pushed this change so that I have to type less ....

### DIFF
--- a/data/presets/marker/utility.json
+++ b/data/presets/marker/utility.json
@@ -1,6 +1,8 @@
 {
     "icon": "temaki-silo",
     "fields": [
+        "material",
+        "colour",
         "ref",
         "operator",
         "marker",


tordans:
I get the `material` 👍 .

But I wonder if we do the data a favor by promoting the `colour` tag. The way you describe it is, that the color references a type. But the type should not be resented by the `colour` field, IMO, but rather some special tag. If the type always has this color, we do not need to map it explicitly. And the other way around, if the color specifies the type, we should not allow "all colors" (`#ff0000` instead of `red`), but only some (AKA a pre defined list of values).

kesterlester:
> I get the `material` 👍 .
> 
> But I wonder if we do the data a favor by promoting the `colour` tag. The way you describe it is, that the color references a type. But the type should not be resented by the `colour` field, IMO, but rather some special tag. If the type always has this color, we do not need to map it explicitly. And the other way around, if the color specifies the type, we should not allow "all colors" (`#ff0000` instead of `red`), but only some (AKA a pre defined list of values).
> 
> Unfortunately the [Wiki for marker](https://wiki.openstreetmap.org/wiki/Key:marker) is vague about the color key and the issue above is not raised there …

It is, but you have to drill down to specific types of marker.  E.g. for use of  colour and material see https://wiki.openstreetmap.org/wiki/Tag%3Amarker%3Dpost 

I am not going to argue lots about my proposal, though, as no amount of arguing on my part will ever make up for the fact that it's all  https://en.wikipedia.org/wiki/Folksonomy . If you don't like my proposal, reject it. All I can say is that it would make my life a lot easier, and that would help me map faster and more reliably, so I'd be grateful if it was accepted.

tyrasd:
`colour` and `material` are both pretty popular, and since there is no (approved) alternative for tagging the subtype of utility markers, this is what we have to work with. So, in general a :+1:  from me.

I've merged this with some slight tweaks as 78f6fe54aa5457cbeeb7bf9e81e39c14863623ca:
* the `colour` field was already present for the Marker preset (and thus also in the Utility Marker preset). And
* the `material` field is also popular with other Marker types, so I've added the field to the generic preset alongside 
* for consistency, I've reordered the fields: `ref` and `operator` should be at the top, and attributes like `colour` and `material` come immediately after those and the _Type_ field(s).

kesterlester:
Thank you both ( @tordans and @tyrasd ) for your inputs and help.  It is much appreciated.

PR #707, opened 2022-12-23
Add translatable strings to archaeological_site type field

kjonosm:


tyrasd:
Some of the values have currently pretty few uses (e.g. [`archaeological_site=baths`](https://wiki.openstreetmap.org/wiki/Tag:archaeological_site%3Dbaths) with currently 44 occurrences in OSM.

Usually, I tend to not include translatable strings for "niche" objects which have (roughly) fewer than 1000 occurrences. On the other hand, archeological features are typically relatively significant/important local features, meaning that they could warrant a little bit of extra effort. :wink:

data/fields/archaeological_site.json
```diff
@@ -1,5 +1,22 @@
 {
     "key": "archaeological_site",
     "type": "combo",
-    "label": "Type"
+    "label": "Type",
+    "strings": {
+        "options": {
+            "baths": "Baths",
+            "city": "Historic City",
+            "crannog": "Crannóg",
```
tyrasd, line 9: I think this is spelled just _Crannog_ in English, isn't it? See e.g. https://en.wikipedia.org/wiki/Crannog

kjonosm:
@tyrasd Thanks for the review. I updated the pr accordingly

tyrasd (approved):

PR #709, opened 2022-12-23
Add translatable strings to barrier type field

kjonosm:
for values with usage > 10,000

### DIFF
--- a/data/fields/barrier.json
+++ b/data/fields/barrier.json
@@ -8,7 +8,7 @@
              "gate": "Gate",
              "wall": "Wall",
              "hedge": "Hedge",
-             "kerb": "Kerb",
+             "kerb": "Curb",
              "bollard": "Bollard",
              "lift_gate": "Boom Barrier",
              "retaining_wall": "Retaining Wall",
--- a/data/fields/barrier.json
+++ b/data/fields/barrier.json
@@ -26,7 +26,6 @@
              "chain": "Chain",
              "jersey_barrier": "Jersey Barrier",
              "height_restrictor": "Height Restrictor",
-             "sliding_gate": "Sliding Gate",
              "turnstile": "Turnstile"
          }
      }
--- a/data/fields/barrier.json
+++ b/data/fields/barrier.json
@@ -2,7 +2,7 @@
     "key": "barrier",
     "type": "typeCombo",
     "label": "Type",
-     "strings": {
+    "strings": {
          "options": {
              "fence": "Fence",
              "gate": "Gate",
--- a/data/fields/barrier.json
+++ b/data/fields/barrier.json
@@ -3,30 +3,30 @@
     "type": "typeCombo",
     "label": "Type",
     "strings": {
-         "options": {
-             "fence": "Fence",
-             "gate": "Gate",
-             "wall": "Wall",
-             "hedge": "Hedge",
-             "kerb": "Curb",
-             "bollard": "Bollard",
-             "lift_gate": "Boom Gate",
-             "retaining_wall": "Retaining Wall",
-             "block": "Large Block",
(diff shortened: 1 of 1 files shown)


data/fields/barrier.json
```diff
@@ -1,5 +1,33 @@
...
+    "label": "Type",
+     "strings": {
+         "options": {
+             "fence": "Fence",
+             "gate": "Gate",
+             "wall": "Wall",
+             "hedge": "Hedge",
+             "kerb": "Kerb",
+             "bollard": "Bollard",
+             "lift_gate": "Lift Gate",
```
matkoniecz, line 13: Note that "lift gate", term used in OSM tagging is not used in actual English. 

It seems to be referred to as a boom gate / boom barrier.

https://en.wikipedia.org/wiki/Boom_barrier

data/fields/barrier.json
```diff
@@ -1,5 +1,33 @@
...
+             "swing_gate": "Swing Gate",
+             "cattle_grid": "Cattle Grid",
+             "ditch": "Ditch",
+             "toll_booth": "Toll Booth",
+             "city_wall": "City Wall",
+             "kissing_gate": "Kissing Gate",
+             "chain": "Chain",
+             "jersey_barrier": "Jersey Barrier",
+             "height_restrictor": "Height Restrictor",
+             "sliding_gate": "Sliding Gate",
```
matkoniecz, line 29: Note that barrier=sliding_gate is used regionally by paid mappers. I would not include it as a supported value.

matkoniecz (changes requested):

data/fields/barrier.json
kjonosm: Fixed

data/fields/barrier.json
kjonosm: I will leave it out for now, though I don't really have a problem with it as the value is well documented and in widespread use...

matkoniecz: There is a high usage count but usage is not actually widespread.

See also https://taginfo.openstreetmap.org/tags/barrier=sliding_gate#chronology

screen01

kjonosm: Distribution doesn't look too bad though, see https://taginfo.openstreetmap.org/tags/barrier=sliding_gate#map

data/fields/barrier.json
```diff
@@ -1,5 +1,32 @@
...
     "type": "typeCombo",
-    "label": "Type"
+    "label": "Type",
+     "strings": {
+         "options": {
+             "fence": "Fence",
+             "gate": "Gate",
+             "wall": "Wall",
+             "hedge": "Hedge",
+             "kerb": "Kerb",
```
tyrasd, line 11:
```suggestion
             "kerb": "Curb",
```

_Kerb_ is British English, if I'm not mistaken.

data/fields/barrier.json
```diff
@@ -1,5 +1,32 @@
...
+             "bollard": "Bollard",
+             "lift_gate": "Boom Barrier",
```
tyrasd, line 13:
```suggestion
             "lift_gate": "Boom Gate",
```

to keep it similar to other types of gates.

tyrasd (changes requested):
see below

data/fields/barrier.json
```diff
@@ -1,5 +1,32 @@
 {
     "key": "barrier",
     "type": "typeCombo",
-    "label": "Type"
+    "label": "Type",
+     "strings": {
```
tyrasd, line 5: PS: the indentation does have one too many space character here and the lines below.

```suggestion
    "strings": {
```

kjonosm:
@tyrasd thanks for the review. I updated the pr accordingly.

PR #711, opened 2022-12-24
Add field for `lift_gate:type`

arch0345:
Adds a field for [`lift_gate:type`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dlift_gate#Types), which is [currently used around 9K times](https://taginfo.openstreetmap.org/keys/lift_gate:type).

### DIFF
--- a/data/fields/lift_gate/type.json
+++ b/data/fields/lift_gate/type.json
@@ -1,0 +1,15 @@
+{
+    "key": "lift_gate:type",
+    "type": "combo",
+    "label": "Type",
+    "strings": {
+        "options": {
+            "single": "Single",
+            "double": "Double"
+        }
+    },
+    "reference": {
+        "key": "barrier",
+        "value": "lift_gate"
+    }
+}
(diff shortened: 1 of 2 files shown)


tyrasd (commented):
In general :+1:  for the new field. But I'm asking myself how intuitive the labels Single/Double are… Without having to look at the example photos on the wiki page, one could for example mistakenly think that "double" is for gates which are placed before and after an obstacle like a railway line.

Perhaps we could use `Single Bar` and `Two Opposing Bars` to make it clear that the type is distinguished by the number (and orientation) of bars? Or we use `Bars` instead of `Type` as the field's label?

tyrasd:
PS: While we're at it: Should the `barrier=lift_gate` preset be labeled "Boom Gate" instead of "Lift Gate"? See https://github.com/openstreetmap/id-tagging-schema/pull/709#discussion_r1056308378 or https://en.wikipedia.org/wiki/Boom_barrier

tyrasd (approved):

PR #713, opened 2022-12-24
Add translatable strings to utility type field

kjonosm:
for common values with usage > 1,000


tyrasd (approved):

tyrasd:
ps: I found the list of "intended" values for the `utility` tag only in the archived proposal page: https://wiki.openstreetmap.org/w/index.php?oldid=2261391#Tagging

PR #722, opened 2022-12-29
fields/parking: minimal centerline parking and orientation for case left_right, both

tordans:
Allow tagging `parking:$side` and `parking:$side:orientation` on centerlines (highways).
This PR is the minimal version of https://github.com/openstreetmap/id-tagging-schema/pull/674.

The goal is to have something small but useful mergable now.

I left out – or worked around – all parts that need more work (outside of the id-tagging-schema).

- [x] field-options only have a title; we can add the description later.
- [x] there are no `prerequisiteTag` for the orientation. This means, the fields will only show up if (a) the user explicitly added them or (b) they are already filled with data. — This is good enough and a lot better than what we have right now ("hidden" data). — It will be possible to add the `prerequisiteTag` later, to have the orientation field show up automatically once the schema is extended.
- [x] The issue of "both" vs "left/right" is solved in this PR by duplicating the fields. "left/right" use the `directionalCombo` where "both" use a regular `combo`. This does expose the data well. However, the UX for switching between "left/right" <> "both" is bad. Users need to modify the ("raw") tags directly. — Still, this is better than not showing the data at all. — Once the `directionalCombo` can handle this switch, those fields can be merged. Ping https://github.com/openstreetmap/iD/issues/9212#issuecomment-1366655208

### DIFF
--- a/data/fields/parking/both/orientation.json
+++ b/data/fields/parking/both/orientation.json
@@ -1,0 +1,17 @@
+{
+    "key": "parking:both:orientation",
+    "reference": {
+        "key": "parking:orientation"
+    },
+    "type": "combo",
+    "label": "Parking orientation (both sides)",
+    "strings": {
+        "options": {
+            "parallel": "Parallel to the Street",
+            "diagonal": "Diagonal in Relation to the Street (~45°)",
+            "perpendicular": "Meets the Street at a Straight Angle (~90°)"
+        }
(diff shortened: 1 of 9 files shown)


tordans:
This approach was wrong. The new PR is https://github.com/openstreetmap/id-tagging-schema/pull/744.

PR #723, opened 2022-12-29
fields/cycleway: Split case left_right and both

tordans:
This splits the case $side:
- "left/right" as directionalCombo
- "both" as regular combo

This way we work around the limitation of directionalCombo which cannot handle all the both-cases well, yet.

This way the left/right is exposed to the users. However, there is no convenient way to switch between those types, other than changing the raw tags. Ping https://github.com/openstreetmap/iD/issues/9212#issuecomment-1366655208

---

This PR follows the same structure as https://github.com/openstreetmap/id-tagging-schema/pull/722.

### DIFF
--- a/data/fields/cycleway/both.json
+++ b/data/fields/cycleway/both.json
@@ -1,54 +1,48 @@
 {
     "keys": [
         "cycleway",
-        "cycleway:left",
-        "cycleway:right"
+        "cycleway:both"
     ],
     "reference": {
         "key": "cycleway"
     },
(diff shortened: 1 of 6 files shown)


tordans:
I learned today that this PR is the wrong approach. The `directionalCombo` will automatically manage splitting (for display) an merging (for the tags) the "both"-case. However, right now only "cycleway" is a possible "both" case. I created https://github.com/openstreetmap/id-tagging-schema/issues/743 as a follow up.

Issue #658, opened 2022-12-01
Add crossing:markings=yes to "Marked Crossing" preset, drop "crossing=uncontrolled"


akadouri:
The "Marked Crossing" preset was recently updated in https://github.com/openstreetmap/id-tagging-schema/pull/590 to remove `crossing=marked` in favor of `crossing=uncontrolled` to reflect newly adopted tagging. While this helps move the tagging forward, this isn't a straightforward logical replacement to how "Marked Crossing" was used.

When selecting "marked crossing" I would assume the preset would include a tag reflecting that language, it may make sense for the preset to include `crossing:markings=yes`. 

I also was not using "Marked Crossing" to indicate a crossing was uncontrolled. I don't think it makes sense to start adding `crossing=uncontrolled` without adding language to the preset to indicate that like "Controlled Marked Crossing" or "Uncontrolled Marked Crossing".

akadouri:
The "Unmarked Crossing" preset also tags `crossing=unmarked`, which is differing behavior. 

I would expect it would also tag `crossing=uncontrolled` if "Marked Crossing" did.

zekefarwell:
When I looked over #590 before the merge I did not notice that while add a field for `crossing:markings` was added to the Marked Crossing preset, it was left blank by default.  This seems quite bad and I wish I had noticed before.  When user chooses a preset called Marked Crossing, at least `crossing:markings=yes` should be added by default so it is clear that there are markings of some kind.  I know that originally `crossing=uncontrolled` was supposed to mean that there are markings, but it was a poorly chosen name resulting in inconsistent usage around the world.   Adding `crossing:markings=yes` may see redundant to some, but it would clarify the situation.

tyrasd:
My reasoning for (initially) not adding the `crossing:markings=yes` tag was that: a) it is not adding much information because `crossing=uncontrolled` already implies that there exists some kind of road marking and b) it would disincentivize users to tag the "proper" value (e.g. `ladder`, `zebra`, etc.) for the field if it was already filled in.

I guess we could reconsider this. Maybe my concerns could be addressed by adding an icon for the `yes` value which indicates that the value should be replaced with a more specific one… maybe something like a ":grey_question:" could work‽

> drop "crossing=uncontrolled"

For this I'd like to wait how discussions on recent [proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/Highway_crossing_cleanup)(s) evolve in the community.

akadouri:
[...]

PR #728, opened 2023-01-06
Change "Marked Crossing" preset to add `crossing:markings=yes` by default

Woazboat:
Added `crossing:markings=yes` to the tags that are added automatically by the "Marked Crossing" preset.
See https://github.com/openstreetmap/id-tagging-schema/issues/658 for related discussion.

### DIFF
--- a/data/presets/highway/crossing/uncontrolled.json
+++ b/data/presets/highway/crossing/uncontrolled.json
@@ -14,6 +14,9 @@
         "highway": "crossing",
         "crossing": "uncontrolled"
     },
+    "addTags": {
+        "crossing:markings": "yes"
+    },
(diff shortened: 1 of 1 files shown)


tyrasd:
Hi. Thanks for this proposal. Unfortunately, it has one significant issue: By adding this as an `addTags` tag, iD's validator would suggest to add the `crossing:markings=yes` tag to all existing crossings with `crossing=uncontrolled`. As others have [pointed out](https://github.com/openstreetmap/id-tagging-schema/issues/658#issuecomment-1362231029), it is not always safe to do this though, as one cannot assume that these objects always have road markings.

Pull request #732 should solve this in a proper way.

PR #731, opened 2023-01-11
add most popular alternative medicine practitioners

westnordost:
i.e. all have above 100 usages. Osteopathy has 1300+ down to... homeopathy has 100+ or something.

I used the https://fontawesome.com/icons/spa?s=solid&f=classic icon (I think this is supposed to be some leaves?) for all of alternative medicine (except ayurveda, which got the hinduism icon) because it should not have the same icon as hospital, IMO.

### DIFF
--- a/data/presets/healthcare/alternative.json
+++ b/data/presets/healthcare/alternative.json
@@ -1,26 +1,18 @@
 {
-    "icon": "maki-hospital",
+    "icon": "fas-spa",
     "geometry": [
         "point",
         "area"
     ],
     "terms": [
-        "acupuncture",
         "anthroposophical",
         "applied kinesiology",
         "aromatherapy",
-        "ayurveda",
         "herbalism",
-        "homeopathy",
         "hydrotherapy",
         "hypnosis",
-        "naturopathy",
-        "osteopathy",
         "reflexology",
-        "reiki",
-        "shiatsu",
         "traditional",
-        "tuina",
         "unani"
     ],
     "tags": {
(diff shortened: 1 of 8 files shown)


bryceco:
https://fontawesome.com/icons/leaf-heart?s=regular&f=classic would also be a reasonable icon.

westnordost:
Not all alternative medicine practices are related to "natural medicine" / herbs though, so displaying something that can clearly be identified as a leaf could be misleading. 

https://fontawesome.com/icons/spa?s=solid&f=classic is more ambivalent. TBH I don't exactly know what it ought to resemble, leaves was just a guess. Maybe it is a lotus flower. In any case, it seems to be a popular logo for alternative medicine.

For Traditional Chinese Medicine (and acupuncture, because it is actually part of it), another logo that may work well is the logo commonly used for the range of religious practices known as [Chinese folk religion](https://en.wikipedia.org/wiki/Chinese_folk_religion) since the former is heavily related if not part of these religious practices:
<a href="https://commons.wikimedia.org/wiki/File:%E7%A6%84_l%C3%B9_or_%E5%AD%90_zi_symbol---red.svg">
禄 lù or 子 zi symbol</a>

bryceco:
I understand it’s pretty hard to find something that fits everything. Looking at logos online there’s usually leaves, with the implication that the medicine is “natural” rather than pharmaceutical. I liked the leaf+heart because it adds something to relate it back to health, but I I’m happy with your original choice as well.

tyrasd:
https://fontawesome.com/icons/hand-holding-medical?s=solid&f=classic could have also potentially worked as an icon here, but I don't mind the proposed one.

> [Chinese folk religion](https://en.wikipedia.org/wiki/Chinese_folk_religion) […] 禄 lù or 子 zi symbol […] would need to be added to temaki?

yes, it would be best if this icon could be added to temaki

tyrasd (approved):

PR #735, opened 2023-01-14
Add translatable strings to material field

kjonosm:
Values with usage > 1000


tyrasd:
I removed some undocumented values (such as `gold`, `masonry`, `composite`, `earth`), see my additional commits above.

PR #746, opened 2023-01-21
Add translatable strings for fence_type field

### LINKED ISSUE #619: Add translatable fields for `fence_type`
### OSM Tag(s)

fence_type=*

As an Option of a Field

### Label

Kind of fence

### Aliases

_No response_

### Terms

_No response_

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/DE:Key:fence_type

### Status of the Tag

In Use

### Usage of the tag

500k+

### Replaces other Tag?

_No response_

### Regional Tag?

_No response_

### Further Information

I am having a hard time remebering that "wire" is the proper value for this kind of fence: 

It is top 5 in the taginfo list.

I would love for this to have a translation for the most common values.

* [`fence_type.json`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/fence_type.json)
* Taginfo for `fence_type` https://taginfo.openstreetmap.org/keys/fence_type
* Docs about how the options work: https://github.com/ideditor/schema-builder#options
* Wiki page: https://wiki.openstreetmap.org/wiki/Key:fence_type

---

FirstGoodIssue

kjonosm:
For common values, closes #619

### DIFF
--- a/data/fields/fence_type.json
+++ b/data/fields/fence_type.json
@@ -4,19 +4,58 @@
     "label": "Type",
     "strings": {
         "options": {
-            "barbed_wire": "Barbed Wire",
-            "bars": "Vertical Bars",
-            "chain_link": "Chain-Link",
-            "electric": "Electric",
-            "pailing": "Pailing",
-            "panel": "Panels",
-            "pole": "Poles",
-            "post_and_rail": "Post-And-Rail",
-            "railing": "Railing",
-            "roundpole": "Roundpole",
-            "slatted": "Slatted Boards",
-            "split_rail": "Spit-Rail",
-            "wire": "Wire"
+            "barbed_wire": {
+                "title": "Barbed Wire",
+                "description": "Fencing wire constructed with sharp edges or points arranged at intervals along the strands."
+            },
+            "bars": {
+                "title": "Vertical Bars",
+                "description": "A fence made of (usually metal) bars."
+            },
+            "chain_link": {
+                "title": "Chain-Link",
+                "description": "A woven fence usually made from galvanized or polyethylene-coated steel wire."
+            },
+            "electric": {
+                "title": "Electric",
+                "description": "A fence that uses electric shocks to prevent animals or people from crossing a boundary."
+            },
+            "pailing": {
+                "title": "Pailing",
+                "description": "A variety of fences, usually consisting of either vertical posts tied together with wire, or vertical boards nailed to horizontal braces."
+            },
+            "panel": {
+                "title": "Panels",
+                "description": "A fence made of prefabricated panels that are inserted into grooves in concrete (or wooden) fence posts."
+            },
+            "pole": {
+                "title": "Poles",
+                "description": "A fence mainly made only out of poles."
+            },
+            "post_and_rail": {
+                "title": "Post-And-Rail",
+                "description": "A variety of fences usually made of horizontal rails on top of vertical, or sometimes slanted, posts."
+            },
+            "railing": {
+                "title": "Railing",
+                "description": "A railing for example at the edge of a bridge to prevent persons or vehicles from falling down. Often made of metal or wood."
+            },
+            "roundpole": {
+                "title": "Roundpole",
+                "description": "A wooden fence typical of Finland, Sweden, Norway and Estonia. It is normally made from debarked unsplit youngish trees."
+            },
+            "slatted": {
+                "title": "Slatted Boards",
+                "description": "A fence made of wooden or concrete upright posts, with usually three horizontal wooden rails, to which vertical wooden boards are fastened."
+            },
+            "split_rail": {
+                "title": "Spit-Rail",
+                "description": "A fence constructed out of timber logs, usually split lengthwise into rails."
+            },
+            "wire": {
+                "title": "Wire",
+                "description": "A variety of fences mainly made of wire; e.g. welded vertical wires, horizontal wires strung through or between posts, wire mesh."
+            }
         }
     }
 }
--- a/data/fields/fence_type.json
+++ b/data/fields/fence_type.json
@@ -6,11 +6,11 @@
         "options": {
             "barbed_wire": "Barbed Wire",
             "bars": "Vertical Bars",
-            "chain_link": "Chain-link",
+            "chain_link": "Chain-Link",
             "electric": "Electric",
             "pole": "Poles",
             "railing": "Railing",
-            "split_rail": "Spit-rail",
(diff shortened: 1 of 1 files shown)


tyrasd (commented):
Thanks for this suggestion. 

I must admit that I'm a little bit on the fence regarding the omission of the value `wood`. While it's reasonable to not include it on the basis of being a "generic"/quasi-unspecified type, it's IMHO no so good when the respective specific types are not included. For wooden fences, this PR doesn't include many of the alternative options like: `paling`, `panel`, `roundpole`, or `slatted`. On the other hand, these values are currently relatively speaking quite rarely used, so it seems that mappers often prefer to just map a fence as a _wood fence_, not caring about the more precise design of the fence. And as far as I can see, there exist some (not uncommon) types of wood fences ([example](https://wiki.openstreetmap.org/wiki/File:Wooden_Fence_-_geograph.org.uk_-_1499882.jpg)) which don't yet have a documented `fence_type`. :thinking: Btw, the same mostly applies also to the value `metal`, which as far as I can see is (also) meant to be used for situations where a metal fence isn't of one of the other fence types (e.g. an ornamental fence which is neither `chain_link`, `wire`, `bars` or `railing`). 

On a different note, I'm not sure whether describing the values `pole`,`wire` as _Poles_, _Wire_ is all too helpful for mappers… Maybe we should be a bit more descriptive here? //cc @tordans: do you think the label `Wire` would have helped you to remember which type of fence is meant?

kjonosm:
As iD already shows the `material` field in the `barrier=fence` preset. Should iD implement deprecation rules for `fence_type=wood` in favour of `material=wood` and for `fence_type=metal` in favour of `material=metal`? This would reduce duplication and clear the way for a more detailed `fence_type=*` value.

tyrasd:
> Should iD implement deprecation rules for `fence_type=wood`

That's probably a question for the tagging mailing list…

I'd be OK with either solution: a) omitting `wood/metal` in favour of a deprecation rule, or b) including them as broad fence types with an appropriate label (e.g. something like _Wooden (unspecified/other)_?). In any case, I think we should provide the field with translatable strings for the more specific subtypes of (wooden) fences.

data/fields/fence_type.json
```diff
@@ -1,5 +1,22 @@
...
+            "electric": "Electric",
+            "pailing": "Pailing",
+            "panel": "Panels",
+            "pole": "Poles",
+            "post_and_rail": "Post-And-Rail",
+            "railing": "Railing",
+            "roundpole": "Roundpole",
+            "slatted": "Slatted Boards",
+            "split_rail": "Spit-Rail",
+            "wire": "Wire"
```
tordans, line 19: When I was looking into it, I was wondering how to deal with all the alternative terms. I think it would be better to list them in the translation than to not expose them at all. Otherwise the translations is kind of a one-to-one version of the value.

```suggestion
            "barbed_wire": "Barbed Wire",
            "bars": "Vertical or metal Bars / Pool Fence",
            "chain_link": "Chain-link / Wire mesh / Chain-wire",
            "electric": "Electric",
            "pailing": "Pailing",
            "panel": "Panels",
            "pole": "Poles",
            "post_and_rail": "Post-And-Rail",
            "railing": "Railing",
            "roundpole": "Roundpole",
            "slatted": "Slatted Boards",
            "split_rail": "Spit-Rail / Log Fence",
            "wire": "Wire"
            "concrete": "Concrete blocks / Special geometry concrete",
            "metal": "Generic metal fence",
            "wood": "Generic wooden fence",        
```

tordans (commented):
Apart from suggesting to add alternative terms to the translations…

… I wonder how to deal with the most common values in Taginfo that are not part of the wiki list:

* "split_rail", taginfo 33.9k; google image search shows image like `post_and_rail`; deprecate in favor of that value?
* "metal_bars", taginfo 4,9k; maybe this one should be documented officially?
* "hedge", taginfo 2.9k; deprecate to barrier=hedge?
* "wire_mesh", taginfo 2.7k; deprecate to `chain_link`, which is also a +1 to add the term in this translation
* "wall", taginfo 2.8k; deprecate for barrier=wall?
* "stone", taginfo 2.8k; add "Generic stone" translation?
* "chain", taginfo 8k, Wiki: use chain_link – add deprecation to iD?

We can move this topic to another issue, of course, and merge this first.

tyrasd (changes requested):
My remaining concern here is that not all of the values here are self-explaining. For example, a mapper might now know that is meant by the fence type _Pole_. One possible approach to make this less ambiguous is to use something similar as in the [camp site](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/camp_site.json) field: this allows adding a description what the value means which is displayed as the hover text of the respective value. For example:

```
        "options": {
            …,
            "pole": {
                "title": "Poles",
                "description": "A fence mainly made of simple poles."
            },
            …

kjonosm:
> "Slatted Boards" vs "Vertical Bars" vs "Poles" vs "Post-And-Rail" is just utter mystery to me.

I am a irritated... Adding those values happened because of @tyrasd's recommendation:

> In any case, I think we should provide the field with translatable strings for the more specific subtypes of (wooden) fences.

I am happy to remove these values again.

matkoniecz:
And I definitely understand irritation of one maintainer commenting that something should be done and implying it would be sufficient, while another maintainer considers it as insufficient.

But sorry, current description fail to make this fence types clearly distinguishable from each other and in current form I would be against merging this PR.

Also, I have no idea how to describe this terms to make them distinguishable as I have problem even with photos next to them to understand the difference.

In the current state I plan to (sadly) close this PR.

matkoniecz:
@tyrasd @k-yle @ZeLonewolf @1ec5 @tordans are these description clear to you and would allow to distinguish between fences? Comments are welcome, especially if you think that this PR is helpful as is and you have different review than me.

ZeLonewolf:
What is a pole fence?  Can you paste a picture?

matkoniecz:
@ZeLonewolf per https://wiki.openstreetmap.org/wiki/Key:fence_type it is https://wiki.openstreetmap.org/wiki/File:Zaun_Freilichtmuseum_Salzburg_02.JPG

ZeLonewolf:
That's a farmer's fence. I don't know if there's another name for it but I've never heard of a pole fence

matkoniecz:
ping @kjonosm do you see way to get this values described/labelled so they are actually distinguishable and identifiable?

PR #747, opened 2023-01-21
Add translatable strings to sport field

kjonosm:
For values with usage > 5000


tyrasd (approved):
Thanks!

PS: I wouldn't mind to (eventually) further extend this list. It seems like some pretty popular sports didn't make the 5,000 cutoff. For example `chess`, `ice_hockey`, `archery`, `disc_golf`, `squash`, `miniature_golf`, `table_soccer`, etc.

PR #748, opened 2023-01-22
Update pedestrian_area.json

tylershuster:
Area tag needed on this item

### DIFF
--- a/data/presets/highway/pedestrian_area.json
+++ b/data/presets/highway/pedestrian_area.json
@@ -12,7 +12,8 @@
         "area"
     ],
     "tags": {
-        "highway": "pedestrian"
+        "highway": "pedestrian",
+        "area": "yes"
     },
     "terms": [
         "center",
--- a/interim/source_strings.yaml
+++ b/interim/source_strings.yaml
@@ -7642,7 +7642,7 @@ en:
         # 'terms: bootleg trail,cow path,desire line,desire path,desireline,desirepath,elephant path,game trail,goat track,herd path,pig trail,shortcut,social trail,use trail'
         terms: <translate with synonyms or related terms for 'Informal Path', separated by commas>
       highway/pedestrian_area:
-        # highway=pedestrian | Translate the primary name. Optionally, add equivalent synonyms on newlines in order of preference (press the Return key).
+        # highway=pedestrian + area=yes | Translate the primary name. Optionally, add equivalent synonyms on newlines in order of preference (press the Return key).
         name: Pedestrian Area
         # 'terms: center,centre,plaza,quad,square,walkway'
         terms: <translate with synonyms or related terms for 'Pedestrian Area', separated by commas>


tordans:
For reference from the wiki https://wiki.openstreetmap.org/wiki/Tag:highway%3Dpedestrian

> An [area](https://wiki.openstreetmap.org/wiki/Area) (closed way or multipolygon) tagged with [highway](https://wiki.openstreetmap.org/wiki/Key:highway)=pedestrian + [area](https://wiki.openstreetmap.org/wiki/Key:area)=[yes](https://wiki.openstreetmap.org/wiki/Tag:area%3Dyes) is the most common way to map the pedestrian area of a square or plaza, where pedestrians can travel freely in all directions.

Example

> highway=pedestrian
area=yes
name=Weyher Marktplatz
place=square

tylershuster:
I read the instructions this time and updated this PR to include the built version as well

tyrasd:
iD does already add the `area=yes` tag to `highway=pedestrian` areas when mapped as a closed way[^1]. Adding the `area=yes` tag to the preset would be problematic, as pedestrian areas mapped as a multipolygon relations (`type=multipolygon` + `highway=pedestrian`) would not be recognized as pedestrian area features by iD anymore. I think it is unnecessary and frankly even a bit silly to add `area=yes` to a multipolygon relation and [most](https://taginfo.openstreetmap.org/tags/highway=pedestrian?filter=relations#combinations) pedestrian areas mapped as relations do in fact not have `area=yes` ([example](https://www.openstreetmap.org/relation/7932512)).

[^1]: This is because it knows that the tag can be used on both linear and polygonal features, so it must add `area=yes` to distinguish the two cases (at least until OSM gets support for a dedicated _area_ data type).

tylershuster:
iD might but other software that uses this schema doesn't, like gomap. Seems like this is a special case that needs to wait for dedicated area types. I only PRed it because I put in a pedestrian area in gomap and it showed up as a ring of paths instead of an area. I'll try and take it up with that software.

bryceco:
> 1. This is because it knows that the tag can be used on both linear and polygonal features, so it must add `area=yes` to distinguish the two cases. 

Are you saying that iD always adds an area=yes to closed ways when the feature contains both way and area geometries? Or is this a special case for highway=pedestrian?

tyrasd:
yes, `area=yes` is generally added to all features created from presets which can be used as both lines and areas: A few exceptions exist: e.g. `waterway=dam` or `public_transport=platform` . See https://github.com/openstreetmap/iD/issues/8985#issuecomment-1041510639 for some additional background. Did you not expect this to be the case?

tylershuster:
Looks like this is the relevant list of exceptions? https://github.com/openstreetmap/iD/blob/04a07092d24a35dd6275b423efa3e733b6563d27/modules/osm/tags.js#L19-L39

tyrasd:
yes, that's the list.

bryceco:
Yeah, I didn't realize that adding the area=yes tag was an implicit action that iD performed. I've added the equivalent code to my app so things should be good now.

tordans:
Are there other such implementation details that are invisible ATM? We could add an "implimemtation_documentation" key to the Schema that is just for the developers. Once we do the "stand alone website" story we can show this as well.

tyrasd:
> Are there other such implementation details that are invisible ATM? 

:thinking: 

I might be missing something, but iD implements special routines for handling the following cases:

* building of `type=multipolygon` relations
* `natural=coastline`
* implicit oneways (e.g. waterways, `man_made=goods_conveyor`, etc.)
* handling of lifecycle prefixes (e.g. `construction:building=yes` should be an area)

Issue #58, opened 2020-11-17
More categories for elements


lectrician1:
**For example,** 
Types of schools:
![image](https://user-images.githubusercontent.com/22477298/99413296-3fd74500-28bb-11eb-8c54-ec3b0cbbc638.png)
Exercise equipment (it seems repetitive to have the same beginning label for each of these):
![image](https://user-images.githubusercontent.com/22477298/99413468-6f864d00-28bb-11eb-8d18-4dfa808b4757.png)
Others that could possibly be created: Airport features (terminals, runways, jetways), Street features (sidewalks, Highway features (category within category), street area, lamp posts), Playground Features, etc.

**Now, I know a complaint might be,** "well by then we'd just be making a category for every Key that has multiple Values". 

I would reason _why don't_ we have a category for every Key? 
The search system should also be set up so that individual features only precede a category result when the specific value is searched for ("school" vs "music school"). 

**Benefits:**
This could help users find features quickly without having to know their full names. For example:
- You don't remember what a "tee-off" is called, but you know it's part of a golf course (Golf Features category already implemented but the situation could apply to other categories)
- You're mapping an airport and placing different types of airport features (categories could also come up in your "recently used" in the feature panel if you use more than one element from one category). 

It also means the user won't have to scroll so much to find a specific feature (see the school example).

tyrasd:
another examples of potential categories:

* advertising: (for stuff like #268)

PR #753, opened 2023-01-22
Create advertising category

kjonosm:
Add advertising category. Also see #58

### DIFF
--- a/data/preset_categories/advertising.json
+++ b/data/preset_categories/advertising.json
@@ -1,0 +1,11 @@
+{
+    "icon": "temaki-billboard",
+    "name": "Advertising Features",
+    "members": [
+        "advertising/billboard",
+        "advertising/board",
+        "advertising/column",
+        "advertising/poster_box",
+        "advertising/totem"
+    ]
+}


matkoniecz:
for some reason it has no effect when brand-new feature is created

I also found no info from where this point values are coming


screen-2026-08-10-06-14-06
screen-2026-08-10-06-13-55

matkoniecz:
needs to be added also to https://github.com/openstreetmap/id-tagging-schema/blob/main/data/preset_defaults.json#L26 if we want that

matkoniecz:
screen-2026-08-11-12-37-09

k-yle (commented):
1. worth considering if this should be added to [data/preset_defaults.json](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/preset_defaults.json). currently it will only appear in search results, since it's not listed in `preset_defaults`.
2. for english, this does not help much, because all _Advertising_ presets start with the word _Advertising_. But in other languages that might be different, so it could benefit other languages

matkoniecz:
@kjonosm are you interested in finishing it by adding to https://github.com/openstreetmap/id-tagging-schema/blob/main/data/preset_defaults.json ?

(I completely understand if not, it sadly took years to get review... In such case I will take over PR and will finish it)

matkoniecz:
replaced by https://github.com/openstreetmap/id-tagging-schema/pull/2809

matkoniecz:
It got rescued in https://github.com/openstreetmap/id-tagging-schema/pull/2809 and is now part of presets

PR #754, opened 2023-01-22
Create board:title.json

kjonosm:
Key used with information=board to describe the headline or title of the board. Also see #64

### DIFF
--- a/data/fields/board:title.json
+++ b/data/fields/board:title.json
@@ -1,0 +1,5 @@
+{
+    "key": "board:title",
+    "type": "text",
+    "label": "Board Title"
+}


tordans:
I don't thing that the wiki is clear about not using name. https://wiki.openstreetmap.org/wiki/Tag:information%3Dboard the side bar  shows name. The description is vague. 

If we do this, we should move name to more fields and or deprecate it. The latter would be cleaner but IMO require a more stable consensus.

1ec5:
There wasn’t a consensus on whether `name` or `board:title` is best the last time it came up on the tagging list in the context of map boards:

https://lists.openstreetmap.org/pipermail/tagging/2021-October/thread.html#62778
https://lists.openstreetmap.org/pipermail/tagging/2021-November/thread.html#62798

That said, there are less common situations where a board has both a title and a name, so it may make sense to expose `board:title` as an optional field for features that need it.

PR #765, opened 2023-01-28
Update discarded.json

kmpoppe:
Add [`closed:note`](https://wiki.openstreetmap.org/wiki/Key:closed:note) to discardable tags.

While this tag was never intended to be added to objects (like other discardables), preventing it being added in the first place manually would prevent cluttering objects.

### DIFF
--- a/data/discarded.json
+++ b/data/discarded.json
@@ -1,4 +1,5 @@
 {
+  "closed:note": true,
   "created_by": true,
   "converted_by": true,
 


tyrasd:
As far as I know, this tag was never added to "regular" map features. Usage is and always was limited to changeset exclusively.

> preventing it being added in the first place manually

We don't want to maintain a list of all potentially mistyped / misused tags in the `discarded.json`, as this would get out of hand very quickly. Instead, it should be a list of [discardable tags](https://wiki.openstreetmap.org/wiki/Discardable_tags) which are still in (somewhat) widespread use and should actively removed by editors to improve data quality.

PS: I further think it is incorrect to classify these changeset tags as `discardable` tags in the OSM wiki.

Issue #168, opened 2021-04-29
Encourage the use of secondary wikidata tags


kolgza:
https://wiki.openstreetmap.org/wiki/Key:wikidata#Secondary_Wikidata_links

kolgza:
Such tags should include:
- `operator:wikidata=*`
- `network:wikidata=*`
- `brand:wikidata=*`
- `architect:wikidata=*`
- `artist:wikidata=*`
- `subject:wikidata=*`
- `name:etymology:wikidata=*`
- `flag:wikidata=*`
- `burried:wikidata=*`

tyrasd:
#533 is for `subject:wikidata`

riQQ:
@tyrasd you mismatched the list entries and the links to the corresponding PRs / issues.

tyrasd:
oh, it's fixed now. thanks for the hint :blush:

matkoniecz:
`architect:wikidata` and `artist:wikidata` are quite dubious, as are typically unverifiable by survey and only importable from third-party sources.

1ec5:
> `architect:wikidata` and `artist:wikidata` are quite dubious, as are typically unverifiable by survey and only importable from third-party sources.

It depends. Most of the public art in my city is signed by the artist or has a small plaque with the artist’s name on it.

matkoniecz:
Yes, some of them have. But vast majority - do not.

1ec5:
I think that’s an unreasonable standard for including a field in a preset, especially if it ends up being hidden by default. The photos I posted above are by no means exceptional. If a public sculpture doesn’t have a visible indication of its creator, it’s fine to avoid selecting the field from the dropdown menu.

In any case, the concern you’re raising is broader than this issue, since the Artwork preset already has an Artist field and the Building preset already has an Architect field. As long as those fields exist, their associated Wikidata fields shouldn’t be subject to additional scrutiny about verifiability.

matkoniecz:
> if it ends up being hidden by default

In such case it would be fine, I guess

GreenAloe:
[model:wikidata](https://wiki.openstreetmap.org/wiki/Key:model:wikidata)

matkoniecz:
> The photos I posted above disprove your point about the information only being available through imports

well, I have not claimed "only", I claimed "typically"

vast majority of uses I have seen seems to be not based on plates like this - and I am not proposing mass removal of such keys but rather against promoting them in iD presets.

[...]

PR #769, opened 2023-01-31
Create wikidata.json

kjonosm:
See #168

### DIFF
--- a/data/fields/operator/wikidata.json
+++ b/data/fields/operator/wikidata.json
@@ -1,0 +1,9 @@
+{
+    "key": "operator:wikidata",
+    "keys": [
+        "operator:wikidata",
+        "operator:wikipedia"
+    ],
+    "type": "wikidata",
+    "label": "Operator Wikidata"
+}


tyrasd:
This PR only defines the field, but does not add it to any preset. Without adding it to any preset, the field would be essentially useless.

Did you intend to add this field to any preset which includes the `operator` field? That would have the major downside of being a lot of effort to change, as it involves changing hundreds of presets. I think we should think about a better solution, e.g. involving an improved "wikidata" field which can directly work on a tag-triplet, e.g. here `operator` + `operator:wikidata` + `operator:wikipedia`. That would require some changes to the schema, however.

kjonosm:
> Did you intend to add this field to any preset which includes the operator field?

@tyrasd: That was the idea following the already featured `subject:wikidata` field. However, not to all presets with `operator=*` info but only those where `operator:wikidata=*` is commonly used, see NSI for reference. If you don't consider merging this pull request (and others adding `*:wikidata` fields) please let me know and I am happy to delete them.

matkoniecz:
> Without adding it to any preset, the field would be essentially useless.

What about objects where this field was already added?

kjonosm:
> What about objects where this tag was already added?

The field still needs to be added to the respective presets...

matkoniecz:
@kjonosm Do you plan to fix technical issue that was mentioned in comment above?

kjonosm:
@matkoniecz 

> Do you plan to fix technical issue that was mentioned in comment above?

That depends on the decision of @tyrasd regarding the above mentioned "better solution".

Improving the wikidata field within the schema builder is unfortunately beyond my skills.

But if people are okay to adding the field to existing specific presets I am happy to update some presets where I consider it useful.

matkoniecz:
> But if people are okay to adding the field to existing specific presets I am happy to update some presets where I consider it useful.

given that changes to schema are stalled and beyond your abilities anyway (and also mine, at least right now - maybe I could spend time to learn what ideditor/schema-builder repo is doing)...

I guess I would try probing PR adding it to just some presets (maybe low number like 10 or 30?) to see whether it will get accepted? At least I would do this if I would be making such PR.

PR #771, opened 2023-01-31
Create wikidata.json

kjonosm:
`architect:wikidata=*`
See #168


tyrasd:
see https://github.com/openstreetmap/id-tagging-schema/pull/769#issuecomment-1411862885

matkoniecz:
Given no reaction and given that it is broken as-is maybe it should be closed?

I am also not enthusiast of this key, see https://github.com/openstreetmap/id-tagging-schema/issues/168#issuecomment-1199305534

PR #776, opened 2023-02-04
Allow landuse=industrial to be mapped as a point

Lonerat:
I changed to allow point as the [wiki](https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dindustrial) seems to allow mapping in node from the first revision.

### DIFF
--- a/data/presets/landuse/industrial.json
+++ b/data/presets/landuse/industrial.json
@@ -12,6 +12,7 @@
         "phone"
     ],
     "geometry": [
+        "point",
         "area"
     ],
     "tags": {


tordans:
It it really necessary to make such tagging more visible? What is the use case or issue you are trying to solve?

Lonerat:
Simply I noticed that it doesn't seem to match the wiki, and I thought mapping by node is helpful, like when the area is very complex or the boundary is not clear. Also it seems that some landuse presets already allow point.

I think the warnings would make it seem like this is the wrong way to do it, rather than that it is not preferred.
But I am not versed enough in such matters, so please close this if this change is not recommended.

Lonerat:
One question I have is the difference in landuse presets for which node is allowed or not.
For example, landuse=residential+residential=apartments preset allows node, but landuse=residential only does not.
Since both tags are about large areas of residential landuse, I don't see the difference in whether they should or should not be allowed?

arch0345:
You can do something similar to what I did with #755: make a new unsearchable preset for `landuse=industrial` that's only for point geometries

Lonerat:
I see, I didn't know that method. Thank you so much.
I will try to make it.

tyrasd:
Thanks

PS:

> What is the use case or issue you are trying to solve?

One benefit of also a non-searchable preset for lesser used, but allowed, geometry types is that it prevents the validator warning from showing up which is saying that e.g. the `landuse=industrial` point should rather be mapped as an area.

tyrasd:
PSS: While testing this, I noticed that the build scripts didn't handle the case very well when there is are both a preset for `key/value.json` and `key/_value.json` because they end up getting the same internal identifier, thus only one of them survives in the compiled presets. I solved this in b565c7d by adding a suffix to the alternate form of the preset (i.e. `key/_value_point.json`).

PR #784, opened 2023-02-07
create preset for a Depot

k-yle:
Currently, the presets don't offer a good option for mapping vehicle depots. I've seen people use `amenity=bus_station` and `landuse=garages` since they were the only suggested presets.

`industrial=depot` is already suggested by the `landuse=industrial` preset, so this PR creates a dedicated preset for that combination, and adds a field for `depot=*`

### DIFF
--- a/data/fields/depot.json
+++ b/data/fields/depot.json
@@ -1,0 +1,5 @@
+{
+    "key": "depot",
+    "type": "combo",
+    "label": "Type"
+}
--- a/data/presets/landuse/industrial/depot.json
+++ b/data/presets/landuse/industrial/depot.json
@@ -1,0 +1,36 @@
+{
+    "icon": "maki-car",
+    "fields": [
+        "name",
+        "operator",
+        "depot"
+    ],
+    "moreFields": [
+        "{landuse/industrial}"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "industrial": "depot"
+    },
+    "addTags": {
+        "landuse": "industrial",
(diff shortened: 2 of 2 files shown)


tyrasd:
for tram depots, I've seen `landuse=railway + railway=depot + depot=tram` being used (although often only the `landuse=railway` is used), e.g. https://www.openstreetmap.org/way/82083538. By extension, tagging of other rail stock (e.g. subway, light_rail, train, etc.) could make more sense to live in the `railway` name space?!

k-yle:
okay, this may be more complicated than anticipated :/ It would also be nice to call the preset "Vehicle Depot", but that wouldn't make sense for `depot=container`

Dimitar5555:
@tyrasd I think that it would be ok to ignore them. `ìndustriial=depot` has 4.1k instances, while `railway=depot` has 121 instances. There are also quite a few other options to tag vehicle depots, so having a preset would help to unify the tagging.

tyrasd:
Fair point. It's fine to have only the "generic" depot preset for now. Though, I think we should remove "tram/train/… depot" from the [terms](https://github.com/openstreetmap/id-tagging-schema/pull/784/files#diff-d86fbea192684544cefc56bfe7b6f1fdbd2fdb15b4e037b31174014ddec07f03R29) of the new preset.

Later, a dedicated tram/train/… depot preset can be added easily.

> unify the tagging

This would be a job for a proper tagging proposal :wink:

k-yle:
okay, `terms` updated

tyrasd (approved):

PR #785, opened 2023-02-07
allow advertising presets as a line & add `message` field

k-yle:
`advertising=`&shy;[`totem`](http://wiki.osm.org/Tag:advertising=totem)/[`poster_box`](http://wiki.osm.org/Tag:advertising=poster_box) are allowed as lines according to the wiki, but the presets don't allow it. This PR fixes that inconsistency.

Also added [`message=*`](https://wiki.osm.org/Key:message) (32k uses) as a field to both presets

### DIFF
--- a/data/fields/message.json
+++ b/data/fields/message.json
@@ -1,0 +1,5 @@
+{
+    "key": "message",
+    "type": "combo",
+    "label": "Message"
+}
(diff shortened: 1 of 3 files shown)


tyrasd:
I've moved the `message` field to the optional tags sections, because I believe the field could relatively easily misunderstood to mean the literal advertising message by new mappers when presented with the field.

PR #790, opened 2023-02-14
make clear that is for pet groomer

matkoniecz:
Make clear that it is primarily place for obtaining services (bring pet for grooming), not a place to buy products

native speakers consulted in https://osmus.slack.com/archives/C01BU04GW7L/p1676391128091299


ZeLonewolf:
Thanks -- "pet Grooming Store" is definitely not a term used in native US English!

PR #791, opened 2023-02-14
fix typo

matkoniecz:
some automated system pointed it out in https://github.com/openstreetmap/id-tagging-schema/pull/790/files


tyrasd:
thx

PR #792, opened 2023-02-15
Update danger_area.json

emersonveenstra:
Clearly specify that this is for military areas only, not a general tag for "something that might be dangerous"

Examples of wrong tagging: 
* https://www.openstreetmap.org/node/5732576630/history 
* https://www.openstreetmap.org/node/10205545257/history
* https://www.openstreetmap.org/node/10607792083/history
* https://www.openstreetmap.org/node/9697570179/history


tyrasd:
Hm… according to the [wiki](https://wiki.openstreetmap.org/wiki/Tag:military%3Ddanger_area), the tag is also for "civilian" danger areas like https://osm.org/relation/3311547. But I do agree that the current naming is probably too easily confused with "dangerous" everyday situations. Also, the non-military uses of the tag seem to be the exception (other than the Chernobyl NPP site I only found a few of other examples: https://www.openstreetmap.org/way/383899913, https://www.openstreetmap.org/way/219364702, https://www.openstreetmap.org/way/345511928, non military shooting ranges are a somewhat common pattern, though).

Maybe adding the adjective _restricted_ or _prohibited_ would be an alternative, as that seems to be the most distinguishing property I could identify for such areas: What do you think about `Danger Zone, Access Prohibited`?

emersonveenstra:
That would work too, i think. Also maybe only offering it on ways and relations, not nodes? The wiki says you can put it on a node, but the examples are all on ways, and most of the incorrect usage is on nodes.

tyrasd:
> not nodes

One approach we recently used in similar situations is to limit the preset to `area`s, but also add a secondary preset for `point` geometries which is unsearchable. That way existing data is still displayed correctly and the validator does not falsely show a warning that point geometries would not be allowed (which would be the case if we only removed the point geometry type from the preset).

emersonveenstra:
Okay, i think i did it correctly, based off #776

tyrasd:
Thanks @emersonveenstra!

Woazboat:
Please add a clarification to the preset description that this tag is only intended for military or weapons/explosive related areas. All the examples on the wiki are at least somewhat military related and I don't believe it was ever intended for non-military use. The Chernobyl exclusion zone is/was operated and enforced by the military, so I don't think that's a good example of non-military use.

The current preset name leads people to use the tag in situations where it's clearly misplaced, such as these areas on whistler mountain where skiing is prohibited:
https://www.openstreetmap.org/way/1174298881/history#map=17/50.06046/-122.96019

PR #795, opened 2023-02-17
Add Dialysis Center preset

### LINKED ISSUE #793: dialysis center / dialysis clinic (healthcare=dialysis)
### OSM Tag(s)

healthcare=dialysis

As a Preset

### Label

Dialysis Center

### Aliases

Dialysis Clinic

### Terms

_No response_

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:healthcare%3Ddialysis

### Status of the Tag

In Use

### Usage of the tag

815

### Replaces other Tag?

no?

### Regional Tag?

_No response_

### Further Information

discussed name for it in https://osmus.slack.com/archives/C01BU04GW7L/p1676460295550289

arch0345:
Closes #793

### DIFF
--- a/data/presets/amenity/clinic/dialysis.json
+++ b/data/presets/amenity/clinic/dialysis.json
@@ -1,0 +1,21 @@
+{
+    "icon": "fas-kidneys",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "amenity": "clinic",
+        "healthcare": "dialysis",
+        "healthcare:speciality": "dialysis"
+    },
+    "reference": {
+        "key": "healthcare",
+        "value": "dialysis"
+    },
+    "terms": [
+        "kidney dialysis",
+        "dialysis clinic"
+    ],
+    "name": "Dialysis Center"
+}
--- a/data/presets/amenity/clinic/dialysis.json
+++ b/data/presets/amenity/clinic/dialysis.json
@@ -5,10 +5,13 @@
         "area"
     ],
     "tags": {
+        "healthcare": "dialysis"
+    },
+    "addTags": {
         "amenity": "clinic",
         "healthcare": "dialysis",
-        "healthcare:speciality": "dialysis"
-    },
+        "healthcare:speciality": "nephrology"
+    }
     "reference": {
         "key": "healthcare",
         "value": "dialysis"
--- a/data/presets/amenity/clinic/dialysis.json
+++ b/data/presets/amenity/clinic/dialysis.json
@@ -11,7 +11,7 @@
         "amenity": "clinic",
         "healthcare": "dialysis",
         "healthcare:speciality": "nephrology"
-    }
+    },
     "reference": {
         "key": "healthcare",
         "value": "dialysis"
--- a/data/presets/amenity/clinic/dialysis.json
+++ b/data/presets/amenity/clinic/dialysis.json
@@ -1,5 +1,5 @@
 {
-    "icon": "fas-kidneys",
+    "icon": "maki-doctor",
     "geometry": [
         "point",
         "area"


data/presets/amenity/clinic/dialysis.json
```diff
@@ -0,0 +1,21 @@
...
+    "icon": "fas-kidneys",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "amenity": "clinic",
+        "healthcare": "dialysis",
+        "healthcare:speciality": "dialysis"
+    },
```
tyrasd, line 11: I didn't find any documentation about `healthcare:speciality=dialysis` on the wiki. It actually [recommends](https://wiki.openstreetmap.org/wiki/Tag:healthcare%3Ddialysis) to use [`healthcare:speciality=nephrology`](https://wiki.openstreetmap.org/wiki/Tag:healthcare:speciality%3Dnephrology) in combination with `healthcare=dialysis`.

Also, maybe we could use the `addTags` approach for this preset, like the following?

```suggestion
    "tags": {
        "healthcare": "dialysis"
    },
    "addTags": {
        "amenity": "clinic",
        "healthcare": "dialysis",
        "healthcare:speciality": "nephrology"
    }
```

1ec5:
The name suggestion index has [several operator entries](https://nsi.guide/index.html?t=operators&k=amenity&v=clinic&tt=dialysis) that would need to be updated with `nephrology`; otherwise, iD will wind up with competing warnings.

Regarding https://github.com/openstreetmap/id-tagging-schema/pull/795#discussion_r1109739846, the wiki used to recommend `healthcare:speciality=dialysis` but was [changed to `healthcare:speciality=nephrology`](https://wiki.openstreetmap.org/wiki/Special:Diff/2258469) without discussion last year. I suppose the idea was that `healthcare:speciality=nephrology` was already [approved](https://wiki.openstreetmap.org/w/index.php?oldid=589962) as part of the healthcare tagging scheme, but it’s only 20% as common as `healthcare:speciality=dialysis`. Some occurrences are dialysis centers, where a doctor would merely be on call, but most are doctors who explicitly specialize in nephrology.

Presumably `healthcare=dialysis` versus ` healthcare=clinic` or `healthcare=doctor` is how one would most easily distinguish these very different facilities. However, I was recently informed that at least one of the chains in NSI also has [locations](https://www.openstreetmap.org/node/10597753088) for [training dialysis center staff](https://www.davita.com/locations/ca/san-jose/725-ridder-park-dr--5258) and others for [training patients to self-administer hemodialysis at home](https://www.davita.com/locations/ca/san-jose/725-ridder-park-dr--5804). It may be necessary to make these distinctions in `healthcare` instead, leaving `healthcare:speciality` as the only way to distinguish between dialysis and nephrology more generally.

data/presets/amenity/clinic/dialysis.json
```diff
@@ -0,0 +1,24 @@
+{
+    "icon": "fas-kidneys",
```
tyrasd, line 2: unfortunately, this icon is a _pro_ icon of fontawesome, and not available in the open source version of the icon set. :unamused:

tyrasd:
I've merged the PR as is now, because it reflects the tagging suggested in the wiki. I'm not an expert in the field and can't say whether _dialysis_ should be its own value of the `healthcare:speciality` tag, but FWIW it wasn't included in the original proposal and [wikipedia](https://en.wikipedia.org/wiki/Nephrology#Scope) lists dialysis as a therapy performed in the field of nephrology.

> It may be necessary to make these distinctions in `healthcare` instead, leaving `healthcare:speciality` as the only way to distinguish between dialysis and nephrology more generally.

It may be best to discuss this outside of this repository. :nerd_face:

1ec5:
Professional training has never been a prerequisite for proposing tags formally, so we should not give that sort of deference to proposal writers by default.

I’m not a trained expert in the field. However, as a resident of California, I’ve been asked to vote on laws that are predicated on this real-world distinction for a few years in a row. I’ve also surveyed many of these facilities. That’s the extent of my qualifications on this matter. I look forward to your participation in the forthcoming discussion. 🙂

matkoniecz:
@1ec5 Can you link the discussion here once it starts?

Cj-Malone:
Should `amenity=clinic` be added with `healthcare=dialysis`? Or is it just here for Carto rendering?

PR #799, opened 2023-02-22
Add translatable strings for country fields

arch0345:
Resolves https://github.com/openstreetmap/iD/issues/7123


tyrasd (changes requested):
Could you please cross reference the strings from the `country` field in the `country_flag` field (i.e. replacing the `"strings": {…}` block with `"stringsCrossReference": "{country}")?

1ec5:
This is a lot to translate. It would be more efficient for iD to implement some custom code that populates these particular fields with values from [`Intl.DisplayNames`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames) (type `region`). If desired, this would be an opportunity to add a flag emoji to each display name for easier visual identification, as proposed in openstreetmap/iD#9508.

tyrasd:
good point @1ec5. let's handle this on the data consumer side.

Tiamogroup:
good

PR #800, opened 2023-02-25
Automated preview instances of iD for tagging PRs 

### LINKED ISSUE #289: An easy way to preview presets from PRs in iD
I find it quite abstract to judge a good preset just by the JSON file. It would be very handy, if every PR would create a preview URL that allows to preview the preset in iD.

[A long while ago](https://github.com/openstreetmap/iD/issues/6664#issue-470669563) that was possible via the URL param `?preset=ExternalUrlToPreset.json` in iD. The current API allows to specify a custom preset as part of the iD initialization https://github.com/openstreetmap/iD/blob/develop/API.md#presets. 

That doc sounds to me, that it should be possible to extend an iD instance like https://github.com/openstreetmap/iD/blob/develop/index.html with a custom URL param that would link to the JSON of a PR like https://raw.githubusercontent.com/jdhoek/id-tagging-schema/2913d690b85add4f4e651fb52d86f4e38421c7da/data/presets/amenity/parking_space.json for example ([PR](https://github.com/openstreetmap/id-tagging-schema/pull/278/files#diff-4bc9db018ae3adc40f205e0bfad0ca6212ec858063244549d99ac19f0c663176)).

Something like
```url
https://ideditor-release.netlify.app/
?idTaggingSchemaPresetPreviewUrl=https://raw.githubusercontent.com/jdhoek/id-tagging-schema/2913d690b85add4f4e651fb52d86f4e38421c7da/data/presets/amenity/parking_space.json
#background=Bing&disable_features=boundaries&map=2.00/0.0/0.0
````

At first, this setup could be documented in the README. But potentially automated via Github Actions to comment the Preview URLs for each preview file on a PR.

---

Thoughts on this idea? – Possible? Useful?

### PR DESCRIPTION
**Pull request by tyrasd**
Closes #289.
Example of how it will look like: https://github.com/tyrasd/id-tagging-schema/pull/3. 

Currently, this has some small limitations:
* Uploading changes is not possible in these test instances, because OSM's [Oauth2](https://wiki.openstreetmap.org/wiki/OAuth) setup does not allow wildcard subdomains for the redirect URI. I think that's fine, since these test instances are only meant for testing the tagging PRs, and not intended for editing.
* New strings will now show up when using a locale other than English

To do:

* [x] set secrets to deploy on netlify (`NETLIFY_AUTH_TOKEN`, `NETLIFY_SITE_ID`)
* [x] drop _action artifacts_ after deployment

PS: This is very much "inspired" by the implementation made in the [openlayers](https://github.com/openlayers/openlayers/tree/main/.github/workflows).

### CODE REVIEW

**Comment by tordans** [comment]
> New strings will now show up when using a locale other than English

Could you change [the preview URL that the bot comments](https://github.com/tyrasd/id-tagging-schema/pull/3#issuecomment-1445125712) to include `?locale=en` (this was the param, right?), so the translation issue is only visible if users force it to?

**Comment by tordans** [comment]
Also, I assume that https://github.com/openstreetmap/openstreetmap-website/pull/3955#issuecomment-1453895658 is a big `+1` that this feature is a good idea, right?

File: .github/workflows/build-preview.yml [hunk 1]
```diff
@@ -0,0 +1,47 @@
...
+
+jobs:
+  build-preview:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '18'
+      - run: npm install
```
tordans, line 19:
```suggestion
      - run: npm ci
```

I think this might be a good idea in this case. The way I understand it is, that this will skip evaluating the dependencies and just install what is provided by the package-lock, which is faster. And in this case, fast and "the same way as before" is good.

Did a quick google and articles like https://www.voorhoede.nl/en/blog/super-fast-npm-install-on-github-actions/  suggest that is true.

.github/workflows/build-preview.yml
```diff
@@ -0,0 +1,47 @@
...
+  contents: read
+
+jobs:
+  build-preview:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '18'
```
tordans, line 18: Since iD has [a .npmrc file](https://github.com/openstreetmap/iD/blob/develop/.nvmrc) which should be the single source of truth on which node to use, it might be a good idea to use something like https://github.com/osmberlin/parkraum.osm-verkehrswende.org/blob/main/.github/workflows/astro.yml#L57 which I used based on this comment https://github.com/actions/setup-node/issues/32#issuecomment-1003854758.

tordans:
This looks great and the PR you linked works as I would expect. 

The article I linked at https://github.com/openstreetmap/id-tagging-schema/pull/800/files#r1124938112 had some hints at how to maybe speed up the npm install further with some caching stuff, but I don't know if that would be relevant (enough) for this use case.

This will be a great improvement.

One thing that should likely be part of the PR is an update to the Readme on how to contribute and use this new feature.

.github/workflows/build-preview.yml
tyrasd: thanks for the suggestion. it's updated in b8ae0d7a61efd6105c6a864b29c4b614aea02cb7

.github/workflows/build-preview.yml
tyrasd: `npm ci` requires the `package-lock.json` to be committed into the repo, which has not been the case until now. I don't see a reason why we shouldn't, though. So, `clean-install` it is with 16fc132.

tyrasd: ps: the `npm install` for the iD repo is still necessary until the iD repo also includes the `package-lock.json` file.

tyrasd:
> One thing that should likely be part of the PR is an update to the Readme on how to contribute and use this new feature.

What exactly did you have in mind which should be mentioned in the Readme? As the bot automatically posts a comment directly on each PR, the contributors don't have to do anything manually and the message in the posted comment should be sufficiently "documenting" what it is about?! So, basically I don't know what exactly to put into the readme. :relaxed: If you can suggest some lines to be added to the Readme, I'd be very happy to add them, though.

PR #802, opened 2023-02-27
list some projects where it is also used

matkoniecz:
obvious and well known to people with deep knowledge of OSM but may be worth making clear to people new to it


tyrasd:
This info was already found at https://github.com/openstreetmap/id-tagging-schema/wiki/Projects-that-are-using-this-tagging-schema, but it's good to have it as a link in the project readme also!

Issue #670, opened 2022-12-05
UK post box fields


boothym:
In the UK, there are two well-used tag fields which are missing from the post box preset - namely post_box:type and royal_cypher, see https://taginfo.openstreetmap.org.uk/tags/amenity=post_box#combinations

Meanwhile the drive_through field is in the preset despite it being highly unlikely that you will find a drive through post box in the UK. It is only used 61 times out of 6,000 for something other than "no", and I'm guessing some of these will be mapping errors.

Don't know if it is possible, but could this be changed just for one country?

tyrasd:
> Don't know if it is possible, but could this be changed just for one country?

Yes, we can define are regional presets/fields using the [`locationSet`](https://github.com/ideditor/schema-builder#locationset) property.

UKChris-osm:
How would this be achieved?

Would the current [post_box.json](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/post_box.json) file need to have a locationSet exclusion, such as:

`"locationSet": {`
`"exclude": ["gb"]`
`},`

... added to it? And then a new copy of the file created with a different name and amended tagging, such as "_post_box_gb.json_" containing:

`"locationSet": {`
`"include": ["gb"]`
`},`

PR #804, opened 2023-03-01
UK specific Post Box tagging.

UKChris-osm:
#670 @boothym

Added location exclusion to the current post box file, excluding gb.
Added new file for a gb post box to promote gb specific tagging.


data/presets/amenity/post_box/post_box_gb.json
```diff
@@ -0,0 +1,45 @@
...
+        "mail dropoff",
+        "mailbox",
+        "package drop",
+        "pillar box",
+        "pillarbox",
+        "post box",
+        "postal box",
+        "postbox"
+    ],
+    "name": "Post Box"
```
tyrasd, line 44: It is possible to cross reference strings between presets, in order to avoid duplicates for translators:

```suggestion
    "name": "{amenity/post_box}"
```

data/presets/amenity/post_box/post_box_gb.json
```diff
@@ -0,0 +1,45 @@
...
+        "ref"
+    ],
+    "moreFields": [
+        "colour",
+        "indoor",
+        "level",
+        "material",
+        "support",
+        "wheelchair"
+    ],
```
tyrasd, line 22: fields can also be referenced between presets, in order to reduce duplication:

```suggestion
    "moreFields": [
        "{amenity/post_box}",
        "drive_through"
    ],
```

UKChris-osm:
I took `post_box:type` from the [the wiki](https://wiki.openstreetmap.org/wiki/Key:post_box:type), but I see it's causing a build error, is that an error on my part?

tyrasd (changes requested):
The definition of the two new fields (`post_box:type` and `royal_cypher`) are still missing in this PR. 

They should probably both by of `"type": "combo"`, I assume.

UKChris-osm:
I've tried applying the suggested changes, hopefully I have done so accurately.

data/presets/amenity/post_box/post_box_gb.json
boothym: > PS: I included `drive_through` here, as it is uses in small quantities even in the UK ([example](https://www.openstreetmap.org/node/5087338620)). Even if these were all mapping mistakes, keeping this fields as an optional field will at least not hide those potential errors in the data.

There's less than 60 of them in the UK and I've only found one other example so far of a "drive through" post box, as in one that is close enough to the road that you don't have to leave your vehicle. [example of one that was wrongly tagged as yes](https://www.bing.com/maps?cp=51.389306%7E-0.141818&lvl=17.0&v=2&sV=1&style=x&dir=87.6&pi=-8.7)

Most of them are probably tagging errors, especially as you must click twice on drive through check box to get the value of "no". So if it has to be included in this preset then it should be below all of the other fields.

tyrasd: yeah, usage is really pretty niche in the UK (and many other countries[^1]).

It's just that including it makes maintaining these regional presets so much easier if the `moreFields` are simply _"just everything this preset could ever have, even if unlikely in a given region"_. IMHO, It also doesn't hurt too much from a user experience point of view to include them, because the optional fields are only added if a user actually searches for it or the object already has a matching tag.

[^1]: I'll probably change the default preset to not include the "drive_through" field by default and create a separate regional preset for countries like the US where such post boxes are found more commonly.

tyrasd (approved):

UKChris-osm:
Thanks for helping me through this @tyrasd, sorry for my syntax errors.

tyrasd:
No worries. Thanks for your first PR to the tagging schema! :smiley:

UKChris-osm:
How long until this tagging will be usable within iD?

closes #670

tyrasd:
Not long. I've been planning to do a proper release for [6.0](https://github.com/openstreetmap/id-tagging-schema/blob/main/CHANGELOG.md#600-rc1) soon anyway.

PR #805, opened 2023-03-04
Remove postbox:type=community

kjonosm:
Although in current use in Canada, shouldn't `postbox:type=community` rather be applied to `amenity=letter_box` instead of `amenity=postbox`?

### DIFF
--- a/data/fields/post_box/type.json
+++ b/data/fields/post_box/type.json
@@ -11,8 +11,7 @@
         "options": {
             "pillar": "Freestanding Box",
             "lamp": "Attached to a Pole",
-            "wall": "Mounted in a Wall",
-            "community": "Community Mailboxes"
+            "wall": "Mounted in a Wall"
         }
     },
     "autoSuggestions": true,


1ec5:
Canada’s community mailboxes serve both purposes, so the [tagging guidelines](https://wiki.openstreetmap.org/wiki/Canada/Tagging_guidelines#Community_mailboxes) currently say to prefer `amenity=post_box`. There was a discussion about community mailboxes (called cluster mailboxes in the U.S.) on the tagging list back in [October](https://lists.openstreetmap.org/pipermail/tagging/2022-October/066214.html) with no resolution.

tyrasd:
Tricky one… As far as I can see, use with `amenity=post_box` is fine for the "outgoing" part of these community mailbox sites. Not sure whether `post_box:type=community` would imply that at these locations inbound letter boxes also exists, or whether a mapper is meant to also map a corresponding `amenity=letter_box` feature alongside.

Now, I'm not so sure anymore whether `post_box:type` is the right tag for this property, though. As the tag is intended for the physical style, [these](https://upload.wikimedia.org/wikipedia/commons/thumb/c/c4/CanadaPostCommunityMailboxes4.JPG/450px-CanadaPostCommunityMailboxes4.JPG) should probably mapped as `post_box:type=pillar`, and the fact that they share the space with community/cluster mailboxes should go into a different attribute. For a very similar reason, I omitted the value `post_box:type=meter`. I guess we should at least be consistent. :thinking:

1ec5:
I guess so. But `post_box:type` is already [next to useless](https://community.openstreetmap.org/t/examples-of-post-box-operators-please/4978/10?u=minh_nguyen) in North America, where the definition of `pillar` could describe such a wide variety of designs. If Canada Post’s community mailboxes are `pillar`s, then `pillar` would be 98% of all currently tagged [postbox types in North America](http://taginfo.geofabrik.de/north-america/keys/post_box%3Atype#values) – and that’s after I went on a furious search through the U.S. to find non-`pillar`s back in November.

Perhaps the `post_box:type` field could be replaced by `post_box:design` and `post_box:mounting` fields in regions where those keys are more meaningful?

1ec5:
#808 tracks adding `post_box:design`.

PR #806, opened 2023-03-05
Add emergency=fire_service_inlet

tiptoptom:
Maybe there is a better icon for this.


tiptoptom:
I don't know how to solve the conflicts. Can you tell me, what's missing?
https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dfire_service_inlet#Tagging

data/presets/emergency/fire_service_inlet.json
```diff
@@ -0,0 +1,22 @@
...
+        "level",
+        "survey/date"
+    ],
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "terms": [
+        "Building Hydrant Inlet",
+        "Fire Service Inlet"
```
tyrasd, line 16: Terms should be in lower case. I'd propose to add a few more entries:

```suggestion
        "building hydrant inlet",
        "charged riser",
        "dry riser inlet",
        "dry standpipe",
        "fdc",
        "fire service inlet",
        "fire sprinkler inlet",
        "riser inlet",
        "wet riser inlet"
```

data/presets/emergency/fire_service_inlet.json
```diff
@@ -0,0 +1,22 @@
+{
+    "icon": "temaki-fire_hydrant",
+    "fields": [
+        "fire_mains"
```
tyrasd, line 4: I think the `couplings` (https://wiki.openstreetmap.org/wiki/Key:couplings) field would be suitable for this preset as well.

tyrasd (changes requested):
The field `fire_mains` does not exist yet.[^1] You'd have to create the field in `data/fields/fire_mains.json`, in order to use it in the preset.


[^1]: This produces the error in the build&test script, see https://github.com/openstreetmap/id-tagging-schema/actions/runs/4335474805/jobs/7570094016

tyrasd:
PS: For reference: The tag was quite recently "approved" (see https://wiki.openstreetmap.org/w/index.php?oldid=2483791)

data/presets/emergency/fire_service_inlet.json
tyrasd: … and while we're at it maybe also `fire_sprinkler` as a checkbox field?

tiptoptom: > Maybe `ref` as well?

I don't know, if `ref` is used often? Maybe as optional tag.

tiptoptom:
I think it should be fine now.

data/fields/fire_mains.json
```diff
@@ -0,0 +1,15 @@
+{
+    "key": "fire_mains",
+    "type": "combo",
+    "label": "Type",
+    "strings": {
+        "options": {
+            "dry": "Dry Riser Inlet",
+            "wet": "Charged Riser",
+            "yes": "yes",
```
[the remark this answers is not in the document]
tyrasd: as the `yes` value is [not](https://taginfo.openstreetmap.org/keys/fire_mains#values) [used](https://wiki.openstreetmap.org/wiki/Key%3Afire_mains)) for this tag, I think we should just omit it.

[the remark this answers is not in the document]
tyrasd: //edit: oh… now I see that the [proposal](https://wiki.openstreetmap.org/w/index.php?oldid=2483791) did also introduce the `yes` option. In that case it would be best to label it _"Unknown Type"_.

data/fields/fire_mains.json
```diff
@@ -0,0 +1,15 @@
...
+            "no": "no"
```
tyrasd, line 10:
```suggestion
            "no": "No"
```

data/presets/emergency/fire_service_inlet.json
```diff
@@ -0,0 +1,32 @@
...
+    "terms": [
+        "building hydrant inlet",
+        "charged riser",
+        "dry riser inlet",
+        "dry standpipe",
+        "fdc",
+        "fire service inlet",
+        "fire sprinkler inlet",
+        "riser inlet",
+        "wet riser inlet"
```
tyrasd, line 26: some more terms from https://en.wikipedia.org/wiki/Standpipe_(firefighting):

```suggestion
        "standpipe",
        "wet riser inlet",
        "wet standpipe"
```

data/fields/fire_mains.json
```diff
@@ -0,0 +1,15 @@
+{
+    "key": "fire_mains",
+    "type": "combo",
+    "label": "Type",
```
tyrasd, line 4: I think we need to clarify that this _Type_ is only supposed to represent the non-sprinkler part of the inlet. Maybe just calling it  _Riser Inlet_ (or _Standpipe_) would be better?!

data/fields/fire_mains.json
tyrasd, line 10: Thinking about it… I believe that the terms dry/wet are probably too domain-specific for a general OSM mapper audience. I think it would be better to describe it like this: 

```suggestion
            "dry": "Dedicated Network (Normally Dry)",
            "wet": "Connected to Mains (Permanently Filled)",
            "yes": "Unknown Type",
            "no": "No"
```

data/presets/emergency/fire_service_inlet.json
tyrasd: I was thinking about large buildings which might have more than one inlet, which might be numbered. Currently, 32 out of 1,300 features with `emergency=dry_riser_inlet` have a `ref`, so I guess an optional field would be more than sufficient.

data/presets/emergency/fire_service_inlet.json
```diff
@@ -0,0 +1,32 @@
+{
+    "icon": "temaki-fire_hydrant",
```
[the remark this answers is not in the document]
tiptoptom: What do you think about the `temaki-plumber`-icon?

data/fields/fire_mains.json
tiptoptom: I think that is clear enough. `emergency=fire_service_inlet` is basically only about the inlet. So I think Riser Inlet or Standpipe would only lead to confusion.

tiptoptom: or maybe 'System-Type'?

data/presets/emergency/fire_service_inlet.json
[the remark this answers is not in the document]
k-yle: I'll leave it up to you and @tyrasd to decide, from my perspective the most important thing is that it's different to the fire hydrant icon



<details>
<summary>these are some other icons we should probably avoid since they're used by similar features:</summary>

image


</details>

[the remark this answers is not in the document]
tiptoptom: I decided me for the  `maki-water` icon.
image

data/fields/fire_mains.json
tiptoptom: I changed it in `Network-Type`. I think this should be clear now.

PR #807, opened 2023-03-05
Update deprecated.json

kjonosm:
Add approved deprecations from February 2023, see https://wiki.openstreetmap.org/wiki/Changelog


tyrasd (approved):
Thanks for keeping an eye on the tagging changelog @kjonosm, that's very much appreciated!

PS: The Lifeboat Station preset also needs to be adjusted. I will do that right after merging this PR.

PR #809, opened 2023-03-06
Create Waste Shed

mikaeldui:


tyrasd (commented):
Not a bad idea to add a preset for this kind of shed. They are also pretty common in Germany, for example. The only small issue there would be that these sheds are typically not only for waste (e.g. residual waste, organic waste), but also for some kind of recycling (e.g. paper), which would technically fall into the [`amenity=recycling`](https://wiki.openstreetmap.org/wiki/Tag%3Aamenity%3Drecycling) tagging. :thinking:

data/presets/building/waste_shed.json
```diff
@@ -0,0 +1,25 @@
...
+    ],
+    "geometry": [
+        "area"
+    ],
+    "locationSet": {
+        "include": ["EE"]
+    },
+    "tags": {
+        "building": "shed",
+        "lockable": "yes",
```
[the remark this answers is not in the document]
mikaeldui: Nice, I've fixed that.

data/presets/building/waste_shed.json
```diff
@@ -0,0 +1,25 @@
...
+        "amenity": "waste",
```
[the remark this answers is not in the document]
mikaeldui: Waste was from the community thread 😊 I've changed it to `recycling`.

mikaeldui:
It's probably mostly recycling. In Sweden they're usually called something like "recycling station". Then, maybe `recycling_type=shed` should be added?

I've tagged one of these with `recycling_type=centre`, and then simply added a `building=yes` to make it render properly on the map.

data/presets/building/waste_shed.json
```diff
@@ -0,0 +1,27 @@
...
+    "locationSet": {
+        "include": ["DE", "EE", "SE"]
+    },
+    "tags": {
+        "building": "shed",
+        "amenity": "recycling",
+    },
+    "addTags": {    
+        "access": "private",        
+        "lockable": "yes"
```
matkoniecz, line 20: Not sure whether I would assume that

data/presets/building/waste_shed.json
```diff
@@ -0,0 +1,27 @@
...
+    ],
+    "geometry": [
+        "area"
+    ],
+    "locationSet": {
+        "include": ["DE", "EE", "SE"]
+    },
+    "tags": {
+        "building": "shed",
+        "amenity": "recycling",
```
matkoniecz, line 16: typically also non-recyled trash containers are there

data/presets/building/waste_shed.json
```diff
@@ -0,0 +1,27 @@
...
+    "fields": [
+        "{building}",
+        "lockable",
+        "access"
+    ],
+    "geometry": [
+        "area"
+    ],
+    "locationSet": {
+        "include": ["DE", "EE", "SE"]
```
matkoniecz, line 12: exist also on Poland, though I mapped some with separate objects for shed, for group of recycling containers and for waste disposal

sometimes there are separate sheds for recyclables and general waste

data/presets/building/waste_shed.json
```diff
@@ -0,0 +1,27 @@
...
+    },
+    "addTags": {    
+        "access": "private",        
+        "lockable": "yes"
+    },
+    "terms": [
+        "recycling station",
+        "environment station"
+    ],
+    "name": "Waste Shed"
```
matkoniecz, line 26: Maybe "Private Waste Shed" or similar to avoid people applying it to public ones? For example on public collection points?

data/presets/building/waste_shed.json
mikaeldui: In the Estonian community we thought locked might be the default. I don't see any way to make that region specific.

data/presets/building/waste_shed.json
mikaeldui: Would `waste_disposal` be better then and cover both?

data/presets/building/waste_shed.json
mikaeldui: Was thinking maybe "community" or "neighbourhood" or "residential" or "intermediate recycling centre".

data/presets/building/waste_shed.json
mikaeldui: Maybe the `locationSet` property isn't needed.

data/presets/building/waste_shed.json
matkoniecz: In theory there could be separate one for Estonia and other one more general, reusing labels. But that seems quite overengineered.

mikaeldui: Well, I guess if you're adding hundreds of "waste sheds" then probably not all of them are locked, so it'd be better to have `lockable` as a featured field than as a field with a default value.

tyrasd:
@mikaeldui [wrote](https://github.com/openstreetmap/id-tagging-schema/pull/809#discussion_r1128416629):
> Would `waste_disposal` be better then and cover both?

IMHO this is the central issue here. The short answer is no: `amenity=waste_disposal` is for everything but recycling, while `amenity=recycling` is, well, for recycling. You'd typically be mapping recycling and waste containers as separate nodes even when they are colocated. But that approach would not work for a single "shed" containing both. And mapping it as three different objects (one shed plus recycling/waste container inside) seems excessively detailed / cumbersome to map.

The only tag which I'm aware of which would cover both cases is (`man_made=street_cabinet` +) [`street_cabinet=waste`](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dstreet_cabinet#Legacy_street_cabinet_categories) which is defined as:

> […] hosting containers for waste and recycling which are not a public amenity, i.e. those for collecting the waste per individual house. In many cases they are locked for private access.

Actually, to some extent, I like the `street_cabinet` approach better for these features, because at least most of these garbage sheds I know around here are just barely high enough for the actual waste/recycling containers (like this [example](https://imgur.com/a/L650eXz)). For a `building=shed` I would have expected that it should be large enough for a person to stand in. The obvious downside of this tagging is that many of these are not actually located on or near a street. :unamused:

tyrasd:
Yeah, I realized that as well after checking out the original community.osm.org thread again (see my edit above). Sorry for the confusion.

matkoniecz:
yes, noticed it but I decided to leave image (may be useful for others and I spend 6 minutes looking for well visible garbage shed :) )

tyrasd:
Mapping it as separate elements would also have the additional benefit that one can easily use the (e.g. recycling) preset's fields to map details like the type of waste which is being collected.

FWIW, one of my previous apartments had a similar "garbage site" which was however not roofed, but just fenced off (seen on this [mapillary image](https://www.mapillary.com/app/?pKey=887278905336319&focus=photo)). Mico-mapping it as separate objects would be directly applicable here as well (by replacing the `building=shed` with a `barrier=fence`).

SK53:
> FWIW, one of my previous apartments had a similar "garbage site" which was however not roofed, but just fenced off (seen on this [mapillary image](https://www.mapillary.com/app/?pKey=887278905336319&focus=photo)). Mico-mapping it as separate objects would be directly applicable here as well (by replacing the `building=shed` with a `barrier=fence`).

This was the main thing I was going to say - covered & uncovered (but often still lockable) - "bin stores" are common here on commercial sites. The uncovered ones usually have a solid wall.

These are a special case of a generic storage area or compound (sheds, just covered by a roof or uncovered): on my local university site there are quite a few other uses: gas cylinder stores are particularly common.

1ec5:
In the U.S., these sheds are also common at individual fast food restaurants, hotels, pharmacies, and so on – anywhere they don’t want people to lose their appetite or don’t want non-employees dumping their trash.

matkoniecz:
> tyrasd added the wontfix

is it intentional that it remains open?

matkoniecz:
I am proposing to close it intentionally, per being stuck in draft state and apparent rejection

> @[tyrasd](https://github.com/tyrasd) tyrasd added the[ wontfix ](https://github.com/openstreetmap/id-tagging-schema/issues?q=state%3Aopen%20label%3Awontfix)label [on Mar 14, 2023](https://github.com/openstreetmap/id-tagging-schema/pull/809#event-8746940274)

PR #810, opened 2023-03-09
Add preset for man_made=quay

jdhoek:
If this PR is otherwise acceptable I can see about making an icon for it.

Also referred to in #529.

-----

I'm having some issues running `npm run build`; I get the following:

```
> @openstreetmap/id-tagging-schema@6.0.0 build
> node scripts/build.js


🏗   Validating and building for development...
[…]/id-tagging-schema/data/fields/check_date.json: 
instance.type is not one of enum values: access,address,check,combo,cycleway,defaultCheck,email,lanes,identifier,localized,manyCombo,multiCombo,networkCombo,number,onewayCheck,radio,restrictions,roadspeed,roadheight,semiCombo,structureRadio,tel,textarea,text,typeCombo,url,wikidata,wikipedia
```

Any idea? That does not seem related to this PR.


data/presets/man_made/pier.json
```diff
@@ -25,7 +25,6 @@
         "jetty",
         "landing",
         "promenade",
-        "quay",
```
jdhoek, line 28: 'Quay' is not a synonym for `pier'. ('dock' however is US English for both ''pier' and 'quay', so both have it.)

tyrasd:
> I'm having some issues running npm run build; I get the following: […] Any idea?

you most likely haven't run `npm install` lately? The [`schema-builder](https://github.com/ideditor/schema-builder/blob/main/CHANGELOG.md#601) dependency needs to be updated because the new field types are only defined the newest version.

jdhoek:
@tyrasd Yeah, that did it. Thanks.

Possible icon submitted here: https://github.com/rapideditor/temaki/pull/83

Screenshot from 2023-03-10 18-53-11

data/presets/man_made/pier.json
tyrasd: Strictly speaking, the `terms` are not only for "synonyms", but for all kinds of search terms a user might type when searching for a particular thing. As piers are often found near quays, it might be acceptable to continue to include it as a search term here…

But I'm also fine with dropping it.

jdhoek:
The `mooring` field might be desirable for the `man_made=pier` preset as well:

https://taginfo.openstreetmap.org/tags/man_made%3Dpier#combinations

Let me know if that should ride along with this PR.

tyrasd:
Good idea, let's add the `mooring` field also to the Pier preset :+1:

data/presets/man_made/pier.json
jdhoek: My guess is that people might search for 'dock' or even 'pier' when they want a quay, but not the other way around ('quay' being a somewhat more difficult word in terms of vocabulary).

'dock' should definitely be in both.

tyrasd:
[[not to self: add rendering rule in iD to render side-markers for quays (like retaining wall)]]

jdhoek:
> Good idea, let's add the `mooring` field also to the Pier preset +1

Done. Could you have a look at the field values I've suggested? Those seem to be the most sensible and well-documented ones.

tyrasd:
The values for `mooring` look fine to me: yes/no is of course the most important, and private is reasonably self explaining. I personally (with almost zero sailing/boating experience) wouldn't know how to identify guest/commercial moorings, though. Also, I can't really tell if other values (`ferry`, `declaration`, …) would be necessary in some contexts?! :shrug: The wiki (https://wiki.openstreetmap.org/wiki/Key:mooring#Values) is also not super helpful on this.

jdhoek:
The wiki is indeed a bit vague there.

jdhoek:
Icon updated to quay; it was merged in temaki.

jdhoek:
The icon isn't showing up in the preview, but that's expected I think? (This PR depends on a newer temaki icon release.)

tyrasd:
yes, as this is brand new, I need to update temaki to [v5.3](https://github.com/rapideditor/temaki/blob/main/CHANGELOG.md#530) on iD before it can show up.

tyrasd:
after upgrading teamki (and a rebuild/redeploy), the preview has the new icon now. :shipit:

jdhoek:
Yep, there it is. Looks fine to me. :+1:

PR #811, opened 2023-03-10
"dog groomer"/"cat groomer" should find "pet groomer", probably

matkoniecz:
warning: untested

-----------

"dog" finds it as result of 

https://github.com/openstreetmap/id-tagging-schema/blob/ec46ce7a78c44090313e7426eba52db6e216d512/data/presets/shop/pet_grooming.json#L9

but quite low

screen07

and is easy to miss it

"dog groomer" or "dog grooming" does not find it at all.

`shop = dog_grooming` organic tag has some use

---------------------------

BTW, there is no way in preset to mark pet groomer as dog groomer/cat groomer (not checked is there a standard tag for that)

### DIFF
--- a/data/presets/shop/pet_grooming.json
+++ b/data/presets/shop/pet_grooming.json
@@ -6,7 +6,9 @@
     ],
     "terms": [
         "cat",
-        "dog"
+        "cat grooming",
+        "dog",
+        "dog grooming"
     ],
     "tags": {
         "shop": "pet_grooming"


tyrasd (approved):

tyrasd:
Thx :+1: 

> BTW, there is no way in preset to mark pet groomer as dog groomer/cat groomer (not checked is there a standard tag for that)

I haven't found anything directly on the wiki, but [`pet`](https://taginfo.openstreetmap.org/keys/pet) seems to be somewhat common for this (~300 out of 5k `shop=pet_grooming` features have this tag). But can we track that in a dedicated issue, so that it doesn't get lost here?

matkoniecz:
Asked for now at https://wiki.openstreetmap.org/wiki/Talk:Tag:shop%3Dpet_grooming#How_to_mark_that_given_pet_groomer_is_specifically_for_dogs_or_cats%3F as it seems to not be documented

tyrasd:
[`animal`](https://wiki.openstreetmap.org/wiki/Animals) was suggested as a tag for this. But it is also bit confusing as it contains a mix of values for species (like sheep, chicken, dog, cat, …) and amenities (like `animal=horse_walker`, `animal=school`, `animal=wellness`, `animal=sport` or `animal=cemetery`). :unamused:

matkoniecz:
https://wiki.openstreetmap.org/wiki/Tag:shop%3Dpet_grooming - wait, `pet=dog` is mentioned already

PR #812, opened 2023-03-11
Update sport.json

kjonosm:
Add some more sports, see #529 
Replace `sport=rugby_union` with `sport=rugby` as recommended according to OSM wiki.


tyrasd (changes requested):
:+1: for the additional values, but I'm not quite sure what best to do with the rugby values. https://wiki.openstreetmap.org/wiki/Tag:sport%3Drugby says that 

> […] a consensus has since emerged. It is advised to solely use the primary tag `sport=rugby`. From there, if you know the version of rugby played at the location you are mapping, you can add a minor tag of `rugby=union` or `rugby=league`.

Although no source for where exactly this consensus was found is given (I assume it was a discussion on the talk mailing list), the `sport=rugby` approach does seem reasonable to me.

However, the wiki is not consistent on this: The [Key:sport](https://wiki.openstreetmap.org/wiki/Key:sport) page does not list `sport=rugby` at all and the [Tag:sport=rugby_*](https://wiki.openstreetmap.org/wiki/Tag%3Asport%3Drugby_union) [subpages](https://wiki.openstreetmap.org/wiki/Tag:sport%3Drugby_league) do only slightly hint towards the `sport=rugby` tag and not list the values themselves as discouraged or deprecated. Further, [usage numbers](https://taghistory.raifer.tech/#***/sport/rugby&***/sport/rugby_union&***/sport/rugby_league) of the "old" tags do not seem to slow down. :thinking: 

Lastly, if we were going ahead with replacing `sport=rugby_union` with `sport=rugby`, I think we should also do the following:

* update the tags of the preset for Rugby League Field
* update the tags of the preset for Rugby Union Field
* add deprecation rules for `sport=rugby_league` to `sport=rugby` + `rugby=league` and for `sport=rugby_union` to `sport=rugby` + `rugby=union`

tyrasd:
I've merged this without the proposed change to the `rugby` value for now. Let's track that in #820

Issue #373, opened 2022-01-28
Consider removing deprecation of amenity=dancing_school


tyrasd:
See https://github.com/openstreetmap/iD/pull/6251#issuecomment-487589188 and following comments.

This would involve the following changes:

* drop the deprecation of `amenity=dancing_school`
* change the "Dance School" preset to use `amenity=dancing_school` as its tag
* (optional) add a field to indicate that a "Dance Hall" also offers lessons (`dance:teaching=yes`)

matkoniecz:
see also https://www.openstreetmap.org/changeset/114834704

simonpoole:
General remark: could we be somewhat more conservative with "deprecating" stuff in general (in this case it was always clearly wrong). There is no agreed on process for deprecating tags, which does not mean that it is never a good idea, but simply something snuck in to a tagging proposal should not be enough.  See also https://www.openstreetmap.org/user/Mateusz%20Konieczny/diary/398555#comment51838

mnalis:
(ref: https://community.openstreetmap.org/t/list-of-deprecated-tags-linked-to-the-new/96737)

PR #814, opened 2023-03-12
remove disputed deprecation

matkoniecz:
see say https://community.openstreetmap.org/t/list-of-deprecated-tags-linked-to-the-new/96737/5?u=mateusz_konieczny

fixes part of #373 that is actively claiming that already mapped data is wrong and should be "upgraded"

### DIFF
--- a/dist/deprecated.json
+++ b/dist/deprecated.json
@@ -122,15 +122,6 @@
     },
     {
         "old": {
-            "amenity": "dancing_school"
-        },
-        "replace": {
-            "leisure": "dance",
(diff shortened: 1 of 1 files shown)


tyrasd:
FYI: I just noticed that we had modified the wrong file for this fix (`dist` files should not be manually be changed, as they are overwritten by the build scripts). see https://github.com/openstreetmap/id-tagging-schema/issues/1177#issuecomment-2040103619 and 915ed76ac030533c058b8cc84f36ff99616c646f for a follow-up on this

PR #817, opened 2023-03-13
Add Track & Field presets

arch0345:
Adds presets for the following Track & Field events:
- Discus
- Hammer Throw
- High Jump
- Javelin
- Long Jump
- Pole Vault
- Running
- Shot Put
- Sprinting
- Steeplechase
- Triple Jump

I made the preset for `athletics=running` unsearchable as `sport=running` is more popular.

### DIFF
--- a/data/presets/leisure/pitch/_athletics.json
+++ b/data/presets/leisure/pitch/_athletics.json
@@ -18,5 +18,6 @@
     "terms": [
         "athletics"
     ],
-    "name": "Track & Field Pitch"
+    "name": "Track & Field Pitch",
+    "searchable": false
 }
--- a/data/presets/leisure/pitch/_athletics.json
+++ b/data/presets/leisure/pitch/_athletics.json
@@ -6,12 +6,12 @@
     ],
     "fields": [
         "athletics_pitch",
-        "surface"
+        "access_simple",
+        "surface",
+        "lit"
     ],
     "moreFields": [
-        "covered",
-        "indoor",
-        "lit",
+        "{leisure/pitch}",
         "ref"
     ],
     "tags": {
--- a/data/presets/leisure/pitch/_athletics.json
+++ b/data/presets/leisure/pitch/_athletics.json
@@ -5,7 +5,14 @@
         "point"
     ],
     "fields": [
-        "athletics_pitch"
+        "athletics_pitch",
+        "surface"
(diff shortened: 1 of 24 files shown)


data/presets/leisure/track/athletics/long_jump.json
```diff
@@ -0,0 +1,18 @@
...
+    "tags": {
+        "leisure": "track",
+        "sport": "athletics",
+        "athletics": "long_jump"
+    },
+    "reference": {
+        "key": "athletics",
+        "value": "long_jump"
+    },
+    "name": "Long Jump Runway"
```
1ec5, line 17: Users less familiar with track and field may search for “(discipline) track” instead of “(discipline) runway”; consider adding an alias for that terminology.

data/presets/leisure/pitch/athletics/shot_put.json
```diff
@@ -0,0 +1,17 @@
...
+    "tags": {
+        "leisure": "pitch",
+        "sport": "athletics",
+        "athletics": "shot_put"
+    },
+    "reference": {
+        "key": "athletics",
+        "value": "shot_put"
+    },
+    "name": "Shot Put Ring / Pit"
```
1ec5, line 16: Do there need to be separate aliases for “Shot Put Ring” and “Shot Put Pit”?

data/presets/leisure/pitch/_athletics.json
```diff
@@ -0,0 +1,30 @@
+{
+    "icon": "maki-pitch",
+    "geometry": [
+        "area",
+        "point"
+    ],
+    "fields": [
+        "athletics_pitch",
+        "surface"
+    ],
```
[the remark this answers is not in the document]
tyrasd: …or maybe it's better to exclude the `name`/`ref` fields:

```suggestion
    "fields": [
        "athletics_pitch",
        "access_simple",
        "surface",
        "lit"
    ],
    "moreFields": [
        "{leisure/pitch}",
        "ref"
    ],
```

tyrasd (commented):
Maybe we should keep the generic Pitch/Track fields (like `lit`, `access` ) for the track & field subpresets (see suggested change below).

We should also add `athletics` as an option to the `sport_racing_nonmotor` field, shouldn't we?

PR #822, opened 2023-03-17
Make barrier=log preset searchable

### LINKED ISSUE #615: Make `barrier=log` preset searchable
> Although `barrier=log` is an approved tag, I made it unsearchable to prevent mappers from adding fallen trees that don't obstruct a passage.

Perhaps it is better to rename it to something obvious like "Barrier (Fallen Tree)" rather than hide it from the search? As someone who surveys a lot of forests, this is a common item to add on paths.

_Originally posted by @RudyTheDev in https://github.com/openstreetmap/id-tagging-schema/issues/506#issuecomment-1276003624_

matkoniecz:
fixes #615

warning: untested

### DIFF
--- a/data/presets/barrier/log.json
+++ b/data/presets/barrier/log.json
@@ -5,6 +5,7 @@
     ],
     "geometry": [
         "vertex",
+        "point",
         "line"
     ],
     "tags": {
--- a/data/presets/barrier/log.json
+++ b/data/presets/barrier/log.json
@@ -10,5 +10,5 @@
     "tags": {
         "barrier": "log"
     },
-    "name": "Barrier (Fallen Tree)",
+    "name": "Barrier (Fallen Tree)"
 }
--- a/data/presets/barrier/log.json
+++ b/data/presets/barrier/log.json
@@ -5,7 +5,6 @@
     ],
     "geometry": [
         "vertex",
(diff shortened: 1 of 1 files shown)


matkoniecz:
maybe some icon would be nice

curiously, it works for lines but not for standalone points

tyrasd:
> curiously, it works for lines but not for standalone points

that's because the preset only defines `line` and `vertex` as geometry types:

```
    "geometry": [
        "vertex",
        "line"
],
```

I guess we could also add the `point` type to this preset.

matkoniecz:
not entirely sure is it making sense as standalone point - I was more surprised that lines are allowed, but apparently there is some micromapping of `barrier=log` placed along paths/roads

data/presets/barrier/log.json
```diff
@@ -10,6 +10,5 @@
     "tags": {
         "barrier": "log"
     },
-    "name": "Fallen Tree",
-    "searchable": false
+    "name": "Barrier (Fallen Tree)"
```
tyrasd, line 13: I'm only not quite 100% happy with this name still. I've asked on slack's #language channel for suggestions about how to name it better.

tyrasd (approved):

Issue #24, opened 2020-08-06
Remove Hobby Shop preset


maro-21:
...because no one knows what it is. See Wiki: https://wiki.openstreetmap.org/wiki/Tag:shop%3Dhobby
My hobby is OpenStreetMap - would I buy a map in a Hobby Store?:)

This preset has terms: manga, figurine, model - we already have separate presets for these ones - `shop=anime`, `shop=collector` and `shop=model`.

I suggest addding "hobby" as a term for these presets:

* `shop=anime`
* `shop=bicycle`
* `shop=boat`
* `shop=books`
* `shop=brewing_supplies`
* `shop=collector`
* `shop=craft`
* `shop=doityourself`
* `shop=fishing`
* `shop=games`
* `shop=hunting`
* `shop=model`
* `shop=music`
* `shop=musical_instrument`
* `shop=outdoor`
* `shop=scuba_diving`
* `shop=sports`
* `shop=toys`
* `shop=video`
* `shop=video_games`

So one can easily find the right preset.

tyrasd:
I think a good first step would be to make the preset "hidden" (non-searchable), so that people are more inclined towards using the more accurate shop types.

PR #823, opened 2023-03-17
add hobby term to shop=anime

matkoniecz:
see #24


matkoniecz:
maybe other shops in #24 should also match

tyrasd:
> maybe other shops in #24 should also match

done in b9f35b5556b7e9a93ce60a1ca86930c10d9d842c

Issue #824, opened 2023-03-17
Add preset for Scout Group


tiptoptom:
### OSM Tag(s)

club=scout

### How would you like this tag to see supported?

As a Preset

### Label

Scout Group

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:club%3Dscout

### Status of the Tag

In Use

### Usage of the tag

3841

PR #836, opened 2023-03-23
Add preset for Scout Group

tiptoptom:
#824


data/presets/club/scout.json
```diff
@@ -0,0 +1,21 @@
...
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "club": "scout"
+    },
+    "terms": [
+        "pathfinder",
+        "guide"
```
tyrasd, line 18: please sort the `terms` alphabetically.

also, we could add _scouting_ as a term:

```suggestion
        "guide",
        "pathfinder",
        "scouting"
```

tyrasd (changes requested):
We should omit the `scout` field for now, as it is currently only properly used 72 times globally compared to ~4k objects with `club=scout`. Further, the use of the `scout` subtag is clustered almost exclusively around two cities in New Zealand: Wellington and Auckland. As this subtag was also never properly approved nor well documented, I would not consider it as an established tag as of now.

tiptoptom:
@tyrasd what must be done to close the change request? The change is actually already incorporated.

PR #837, opened 2023-03-24
Enable tagging unspecific crossing

### LINKED ISSUE #834: Allow tagging crossing without specifying variety (just highway=crossing)
Sometimes memory or aerial is not good enough to specify exact one 0 for example presence of traffic signals may be not entirely clear.

On node within way:

screen01

matkoniecz:
fixes #834

### DIFF
--- a/data/presets/highway/crossing.json
+++ b/data/presets/highway/crossing.json
@@ -2,12 +2,24 @@
     "fields": [
         "crossing"
     ],
+    "moreFields": [
+        "tactile_paving",
+        "crossing/island",
+        "crossing/markings",
+        "crossing_raised",
+        "flashing_lights"
+    ],
     "geometry": [
         "vertex"
     ],
     "tags": {
         "highway": "crossing"
     },
-    "matchScore": 0.95,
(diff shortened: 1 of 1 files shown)


matkoniecz:
https://github.com/openstreetmap/id-tagging-schema/issues/834#issuecomment-1483330119 points out serious issue - it is also correct for cases where cycleway crosses road and maybe even where footway crosses cycleway

tyrasd:
Yeah… just calling it "Crossing" should be better here, as that accommodates both foot and bike crossings. I've adapted the PR accordingly. I've also added a few additional search terms and enabled the crossing fields as optional `moreFields` (such that if a crossing is mapped with, e.g. a `crossing:island` tag, but without `crossing=*`, the information is not hidden from the mapper).

matkoniecz:
`crossing:signals` tag seems to be not shown

> optional moreFields (such that if a crossing is mapped with, e.g. a crossing:island tag, but without crossing=*, the information is not hidden from the mapper).

What about making them appear already when not tagged (at least crossing signals one) to encourage mapper to answer them?

Issue #314, opened 2021-12-22
Document where to find icons for presets and how to add them


tordans:
For https://github.com/openstreetmap/id-tagging-schema/issues/267 I am trying to understand …

- Where do the icons come from? (Source/Overview/Usage)
- How can I add new icons? (Process)

---

Here is what I found until now…

- I could not find any documentation in this repo.
- I could not find any documentation in the 
- There is https://github.com/openstreetmap/id-tagging-schema/issues/41, which talks about icons most, but does not help solve the questions
- There is https://github.com/ideditor/schema-builder/issues/4 which hints that the Icons are part of iD
  - Which lead me to https://github.com/openstreetmap/iD/search?l=SVG&q=icon, which is a list of fontawesome icons and I checked and found one of the icons from here, there.
  - => How does it work with the other icons?
- There is https://github.com/ideditor/temaki => Which I assume would be the best place to contribute icons to?
- There is https://github.com/mapbox/maki

--- 

Some statistics:

- `maki-` found 930 times in this repo, 233 unique icons
- `temaki-` found 1190 times in this repo, 522 unique icons
- `id-` found 146 times in this repo, 38 unique icons
- `fas-` found 437 times in this repo, 231 unique icons
- `<noprefix>` found 16 times in this repo, 7 unique icons

--- 

Where do I find which icons?

- `maki-` – https://github.com/mapbox/maki
- `temaki-` – https://github.com/ideditor/temaki
- `id-` – => ?
- `fas-` – https://github.com/openstreetmap/iD/tree/develop/svg/fontawesome, which is pulled from https://github.com/openstreetmap/iD/blob/develop/scripts/build_data.js#L11-L16
- `<noprefix>` – => ?

tyrasd:
the `id-` icons should be the ones found here: https://github.com/openstreetmap/iD/tree/develop/svg/iD-sprite/presets

The non-prefixed ones are only shown on fields, which is a bit strange: While the [schema](https://github.com/ideditor/schema-builder/blob/main/schemas/field.json#L123-L125) apparently allows for an "icon" property, it is not documented and AFAIK it is also not used anywhere. :shrug: My best guess would be that an early prototype of iD once had support for icons of fields (or there were plans for such a thing), but it was dropped at some point (or never implemented). Anyway these unused icons can point to non-existing icons and nobody would notice. I've opened https://github.com/ideditor/schema-builder/issues/30 to clarify this.

tyrasd:
[...]

PR #838, opened 2023-03-25
try to document icon sources

matkoniecz:
first step toward  #314

not fully explaining what is going on, not sure is it the best place to document it but at this point I think that substandard attempt is better than nothing :)


matkoniecz:
https://github.com/openstreetmap/iD/blob/c01adb01fabe0dfdc4d9e5b88d15967f56a20859/modules/ui/preset_icon.js#L297 mentions also `roentgen-` icons added in https://github.com/ideditor/schema-builder/issues/75 

https://github.com/enzet/map-machine#r%C3%B6ntgen-icon-set

matkoniecz:
Note that https://github.com/openstreetmap/id-tagging-schema/pull/839 apparently has not worked - and I am unsure why.

Are there additional restrictions on Font Awesome? Additional steps being required? I made some typo mistake?

tyrasd (changes requested):
Recently, I had created the [ICONS.md](https://github.com/ideditor/schema-builder/blob/main/ICONS.md) page on the schema-builder repo to document this. Wouldn't it be better to only put a link to that page here in the readme, in order to not duplicate the content unnecessarily.

PR #839, opened 2023-03-25
try more dedicated icon for sign maker

matkoniecz:
see #41


matkoniecz:
Sadly, it has not worked. Not sure why.

matkoniecz:
screen

matkoniecz:
Thanks so much for the fix!

tyrasd:
I'm wondering whether https://fontawesome.com/v5/icons/sign?f=classic&s=solid would be even more fitting?

matkoniecz:
yes, seems better as it is far more likely that they will make one looking like this

PR #840, opened 2023-03-25
use snowflake icon for shop=frozen_food

matkoniecz:


matkoniecz:
This one sadly has not worked and I am unsure whether I did something wrong, icon is unavailable or would it work after merge and rebuilding iD.

data/presets/shop/frozen_food.json
```diff
@@ -1,5 +1,5 @@
 {
-    "icon": "maki-shop",
+    "icon": "far-snoflake",
```
arch0345, line 2:
```suggestion
    "icon": "fas-snowflake",
```

arch0345 (changes requested):

data/presets/shop/frozen_food.json
matkoniecz: tried fas (solid) prefix in https://github.com/openstreetmap/id-tagging-schema/pull/839 - and failed anyway

arch0345: I was trying to point out that it should be `snowflake` instead of `snoflake`

matkoniecz:
https://github.com/openstreetmap/id-tagging-schema/pull/839/commits/0103ae0548c1d22a8d5465fd7b0a354f901e1411 is missing here, probably redoing PR after the other is merged will be the best

PR #842, opened 2023-03-26
Update embassy.json

kjonosm:
Order values alphabetically


tyrasd:
The order was actually chosen on purpose in #343 to reflect the relative frequencies of the different values (see [taginfo](https://taginfo.openstreetmap.org/keys/embassy#values)). As the alphabetic ordering would be broken by the translation strings (already in English), it is of little benefit for users. I'd prefer to keep it as is.

PR #847, opened 2023-03-31
dedicated icon for craft=saddler

matkoniecz:
for #41

BTW, regenerating https://github.com/openstreetmap/id-tagging-schema/issues/41#issuecomment-568098343 would be useful as it is mostly oudated


matkoniecz:
oh, they cleverly remove "free" filter on some actions

PR #848, opened 2023-03-31
add search terms to billboard

matkoniecz:


matkoniecz:
`ad` continues to not find it, but `spam` now finds it

PR #849, opened 2023-03-31
better castle icon

matkoniecz:
more unique

also, yes, roentgen-fort seems more castle than maki-castle

part of #41


matkoniecz:
screen04

first row - new icon

used to have like one in the second row

BTW, regenerating https://github.com/openstreetmap/id-tagging-schema/issues/41#issuecomment-568098343 would be useful as it is mostly oudated - nearly everything listed there is now solved

tyrasd (approved):

PR #857, opened 2023-04-05
add some popular cuisine values

matkoniecz:
added four that I missed in my mapping

https://taginfo.openstreetmap.org/tags/cuisine=tex-mex#overview - added based on comments below

Added also Ukrainian restaurant and Georgian restaurant as POI types (I hope that noone maps every single restaurant in Italy as Italian restaurant BTW - or do the same based on nationality of owner)

Following look sort-of-reasonable but I was not confident enough to add them

https://taginfo.openstreetmap.org/tags/cuisine=international#overview
https://taginfo.openstreetmap.org/tags/cuisine=barbecue#overview (note that there is also `cuisine=grill`) - seems to make sense but wiki is not explaining clearly what is the difference between them
https://taginfo.openstreetmap.org/tags/cuisine=arab#overview
https://taginfo.openstreetmap.org/tags/cuisine=bagel#overview
https://taginfo.openstreetmap.org/tags/cuisine=african#overview
https://taginfo.openstreetmap.org/tags/cuisine=caribbean#overview
https://taginfo.openstreetmap.org/tags/cuisine=frozen_yogurt#overview


1ec5:
> https://taginfo.openstreetmap.org/tags/cuisine=barbecue#overview (note that there is also `cuisine=grill`)

Barbecue overlaps with grilling but isn’t necessarily the same thing. There’s already a preset using this tag. It would make sense for the Cuisine field to display a human-readable value after choosing a cuisine-related preset:

https://github.com/openstreetmap/id-tagging-schema/blob/39af4001dd78ac6e7abc13fae5bbaa5497d430fc/data/presets/amenity/restaurant/barbeque.json#L24-L28

matkoniecz:
For `tex-mex` I was unsure what would be a good label - I guess that `Tex-Mex` would work?

> Barbecue overlaps with grilling but isn’t necessarily the same thing.

From looking at: https://wiki.openstreetmap.org/wiki/Tag:cuisine%3Dgrill https://wiki.openstreetmap.org/wiki/Tag:cuisine%3Dbarbecue - it seems that barbecue can be about roasting/smoking without grilling, right?

1ec5:
Correct on both counts.

tyrasd (approved):

tyrasd:
@matkoniecz did you want to further amend this PR, or should I merge it as is? I just noticed that, for example, the mentioned value `tex-mex` (see https://github.com/matkoniecz/id-tagging-schema/commit/bfd8351cb4871381e34c064d27342d3ebd7d9a5e) is not yet included in the branch of this PR… :thinking:

matkoniecz:
Thanks for spotting! It should be ready now, but I will wait for build preview and test is everything working.

tyrasd (approved):

matkoniecz (commented):
seems to work, as far as I can see

PR #859, opened 2023-04-09
Add contact:facebook field

Cj-Malone:
It's my first time, please be gentle.

### DIFF
--- a/data/fields/contact/facebook.json
+++ b/data/fields/contact/facebook.json
@@ -1,0 +1,6 @@
+{
+    "key": "contact:facebook",
+    "type": "url",
+    "label": "Facebook profile URL",
+    "placeholder": "https://www.facebook.com/LocalBusiness"
+}
--- a/data/fields/contact/facebook.json
+++ b/data/fields/contact/facebook.json
@@ -1,6 +1,6 @@
 {
     "key": "contact:facebook",
     "type": "url",
-    "label": "Facebook profile URL",
+    "label": "Facebook Username or URL",
     "placeholder": "https://www.facebook.com/LocalBusiness"
 }
--- a/data/presets/amenity/animal_boarding.json
+++ b/data/presets/amenity/animal_boarding.json
@@ -9,6 +9,7 @@
         "animal_boarding"
     ],
     "moreFields": [
+        "contact/facebook",
         "email",
         "fax",
         "gnis/feature_id-US",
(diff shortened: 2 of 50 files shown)


data/fields/contact/facebook.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "contact:facebook",
+    "type": "url",
+    "label": "Facebook profile URL",
```
tyrasd, line 4: As the tag allows to contain both facebook usernames or full URLs, we should probably use a more fitting name.

PS: Field labels should be in title case

```suggestion
    "label": "Facebook Username or URL",
```

tyrasd (commented):
Hi. Thanks for your first PR!

See one suggestion below about the label of the field.

In general, I think we can add a field for this tag. :+1: 
But I'd think we could put a bit of effort into finding out which other presets would also "need" this tag in addition to shops. I would assume that restaurants, hotels and other touristy POIs, the craft and club presets, should also have the field added. It would potentially even a little bit confusing if the field was only available for shop presets, but not these other presets.

data/fields/contact/facebook.json
Cj-Malone: I think encouraging the username will result in a lot of bad data, people adding the profile title/label, not any kind of unique identifier. But we can be review it in the future anyway.

Cj-Malone:
It's a shame there isn't a nice preset category this fits into. For what it's worth [StreetComplete](https://github.com/streetcomplete/StreetComplete/blob/0c763f1fdd22999d12ebf1841a00eeb155f03284/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt#L35) ends up with similar issues.

I've added it to some more categories. I expect I'll still have missed some, but time will fix that.

tyrasd (approved):
looks great, thank you!

FYI, I've also added the field to `amenity=doctors` and `amenity=clinic`.

tyrasd:
> It's a shame there isn't a nice preset category this fits into.

What we could do is to create a dummy template preset for POIs whose only purpose is to hold the generic optional fields which should be shared across all "POI" presets, like for example the contact fields. That dummy preset could be referenced by other POI presets by using the cross-reference syntax (e.g. `"moreFields": ["{@templates/poi}", …]`). That could make maintaining and updating these preset fields much easier in the future.

But that's of course not to be done in this PR.

PR #860, opened 2023-04-10
survey_date and date:survey -> survey:date

Marc-marc-marc:
use more common, documented and following common namespace convention https://wiki.openstreetmap.org/wiki/Key%3Asurvey%3Adate

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -483,10 +483,6 @@
     "replace": {"kerb": "$1"}
   },
   {
-    "old": {"date:survey": "*"},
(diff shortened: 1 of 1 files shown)


tyrasd (commented):
The `date:survey` variant is currently only found 12 times. I think it is not necessary to add rules for such super rare, but possible "typo" of a tag, as this would only make the list unnecessarily hard to maintain. It's better to manually clean up such mistakes or using a dedicated tool.

+1 for `survey_date`: it is still pretty commonly found and listed as a deprecated version of the tag.

PR #873, opened 2023-04-18
Additional synonym and a better name for gynecologist

### LINKED ISSUE #789: Additional synonym and maybe a better name for gynecologist
1) OBGYN, ob/gyn should find gynecologist

2) maybe "obstetrician/gynecologist" would be a better name - see https://osmus.slack.com/archives/C01BU04GW7L/p1675881005275539

note that obstetric part is already in specialty:

screen03

matkoniecz:
fixes #789

Per https://github.com/openstreetmap/id-tagging-schema/issues/789#issuecomment-1454697880

> After changing it, you’d need to run npm run build to update the localizable string table.

should I do it as a part of PR?

### DIFF
--- a/data/presets/amenity/doctors/gynaecology.json
+++ b/data/presets/amenity/doctors/gynaecology.json
@@ -30,5 +30,9 @@
         "key": "healthcare:speciality",
         "value": "gynaecology"
     },
-    "name": "Obstetrician/Gynecologist"
+    "name": "Obstetrician/Gynecologist",
+    "aliases": [
+        "Gynecologist",
+        "Obstetrician"
+    ]
 }
--- a/data/presets/amenity/doctors/gynaecology.json
+++ b/data/presets/amenity/doctors/gynaecology.json
@@ -9,6 +9,8 @@
         "gynaecologist",
         "gynaecology",
         "gynecology",
+        "OBGYN",
+        "ob/gyn",
         "obstetrics",
         "ovulation",
         "reproductive",
@@ -28,5 +30,5 @@
         "key": "healthcare:speciality",
         "value": "gynaecology"
     },
-    "name": "Gynecologist"
+    "name": "Obstetrician/Gynecologist"
 }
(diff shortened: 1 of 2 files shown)


matkoniecz:
npm run build fails for me, see

```
npm run build

> @openstreetmap/id-tagging-schema@6.2.0-dev build
> node scripts/build.js

file:///home/mateusz/Documents/install_moje/OSM_software/id-tagging-schema/node_modules/@ideditor/schema-builder/lib/build.js:417
    if (field.locationSet?.include) {
                          ^

SyntaxError: Unexpected token '.'
    at Loader.moduleStrategy (internal/modules/esm/translators.js:81:18)
```

I will try figuring out why, but it seems local breakage and failure to install proper node version - not causing problems with this PR

tyrasd:
> npm run build fails for me

Which version of nodejs do you run? You need at least nodejs version 14 to use the [`?.` (optional chaining) operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility).

matkoniecz:
`node -v` gives `v13.14.0` - that explains things

data/presets/amenity/doctors/gynaecology.json
```diff
@@ -28,5 +30,5 @@
         "key": "healthcare:speciality",
         "value": "gynaecology"
     },
-    "name": "Gynecologist"
+    "name": "Obstetrician/Gynecologist"
```
tyrasd: Let's also add the individual terms as `aliases`

```suggestion
    "name": "Obstetrician/Gynecologist",
    "aliases": [
        "Gynecologist",
        "Obstetrician"
    ]
```

tyrasd (approved):

tyrasd:
perfect, thanks!

btw: strictly speaking, it is now really necessary anymore to run `npm run build` for a PR like this one, as we have a bot which does this automatically (if necessary) after each merge/push to the main branch, like this example commit: ad4dbd4aa11e386a87f728c33a9726238209eea5. But I do usually run it manually myself and is also a thing I still appreciate in PRs, as it keeps the commit history of the project a little bit more tidy overall. :relaxed:

Issue #473, opened 2022-05-28
Use sentence case in fields in English?


1ec5:
In the English localization, field names have always been in _Title Case_, and so have translatable value labels since openstreetmap/iD#8238. However, contributors frequently overlook this point, partly because of unfamiliarity with UI design, but also because there are no contributing guidelines for this project that could set this expectation. Should we use _Sentence case_ instead?

## Background

In English, user interface elements are generally in title case, especially buttons and menu items, but it’s common to use sentence case for kinds of UI elements that are likely to be more verbose, such as checkbox labels and combo box items. For reference, here are style guides for some prominent desktop platforms and Web frameworks, but there are plenty of others along these lines:

UI element | Apple (macOS) | Google | [Microsoft](https://docs.microsoft.com/style-guide/capitalization) ([Win32](https://docs.microsoft.com/en-us/windows/win32/uxguide/text-ui#capitalization)) | [GNOME](https://developer.gnome.org/hig/guidelines/writing-style.html#capitalization) | [KDE](https://develop.kde.org/hig/style/writing/capitalization/#guidelines) | [OpenStack](https://docs.openstack.org/doc-contrib-guide/ux-ui-guidelines/ui-text-guidelines/ui-text-capitalization.html)
----|----|----|----|----|----|----
Checkboxes | [Sentence](https://developer.apple.com/design/human-interface-guidelines/macos/buttons/checkboxes/) | Sentence | Sentence | Sentence | Sentence | Sentence
Combo boxes | [Title](https://developer.apple.com/design/human-interface-guidelines/macos/fields-and-labels/combo-boxes/) | Sentence | Sentence | Sentence | Sentence | Sentence
Group boxes | [Depends](https://developer.apple.com/design/human-interface-guidelines/macos/buttons/disclosure-controls/) | Sentence | Sentence | Title | Title | Sentence
Popup menu labels | [Sentence](https://developer.apple.com/design/human-interface-guidelines/macos/buttons/pop-up-buttons/) | Sentence | Sentence | Sentence | Sentence | Sentence
Popup menu items | [Title](https://developer.apple.com/design/human-interface-guidelines/macos/buttons/pop-up-buttons/) | Sentence | Sentence | Title | Title | Sentence
Push buttons | [Title](https://developer.apple.com/design/human-interface-guidelines/macos/buttons/push-buttons/) | [Title](https://material.io/components/buttons#anatomy) | Preferably sentence | Title | Title | Title
[...]

PR #874, opened 2023-04-18
harmonize spelling of titles with slashes

tyrasd:
I noticed that we had been using quite a mix of different spellings in titles which contain a slash. Sometimes they contain spaces between the words, sometimes now, e.g.: `Entrance / Exit` vs. `Barbecue/Grill`.
 
I've tried to follow the rules of the CMOS (when the slash is separating single words, use no spaces, otherwise add spaces. – see https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Punctuation/faq0096.html), resulting in the following cases:

* no spaces before and after the `/` between single word alternatives (e.g. `Borough/Suburb`)
* spaces before and after the `/` where the slash is between alternatives where at least one is consisting of multiple words (e.g. `Dojo / Martial Arts Academy`)
* no spaces before and after the `/` when a part of a multi-word term consist of a two single-worded alternatives (e.g. `Preschool/Kindergarten Grounds`)
* spaces before and after the `/` when a part of a multi word term consists of at least one non single-worded alternative (e.g. `Property Management / Leasing Office`)

It would be nice if native speakers here could confirm that this approach indeed makes sense. :innocent: //cc @1ec5, @k-yle, @arch0345, et al.


ZeLonewolf:
Huh, this is something I've never really thought about. I would tend to say there should be no spaces when it's a single-word alternative and with spaces when it's a phrase.  But this is very minor.

data/presets/railway/_halt.json
```diff
@@ -7,6 +7,6 @@
     "tags": {
         "railway": "halt"
     },
-    "name": "Train Station (Halt / Request)",
+    "name": "Train Station (Halt/Request)",
```
1ec5, line 10: How about simplifying this to “Limited-Service Train Station”?

I don’t think any dialect of American English calls these “halts” or knows what a halt is. On the other hand, the U.S. community isn’t always using `railway=halt` to indicate a request stop (flag stop); that’s what [`request_stop=yes`](https://wiki.openstreetmap.org/wiki/Tag:request_stop%3Dyes) is for. Many occurrences of `railway=halt` in the U.S. indicate a limited-service station stop, for example, only on weekends or during football games, while some entire light rail systems operate exclusively on a request-stop model, including at their major stations.

I think “limited-service” would capture the intent of both the German and North American usages of this tag. We should add a field for `request_stop` while we’re at it: #875.

data/presets/highway/motorway_junction.json
```diff
@@ -13,5 +13,5 @@
     "terms": [
         "exit"
     ],
-    "name": "Motorway Junction / Exit"
+    "name": "Motorway Junction/Exit"
```
1ec5, line 16: I think this should be simplified to just “Motorway Exit” in the American English localization. Although longtime OSM contributors may be used to seeing the phrase “motorway junction”, this is a very British term that could lead to misunderstanding among newer mappers.

In the U.S., a junction would be any place where a major road connects to another major road, even indirectly via ramps, but not the specific location where `highway=motorway_junction` needs to be placed. We call those “(highway/freeway) exits”, as opposed to “entrances”. I’ve often encountered `highway=motorway_junction` incorrectly placed at a freeway entrance, potentially throwing off routers like OSRM that count exits along a step. This preset name might be a factor in that misuse.

(I’d even go a step further and advocate for replacing “motorway” with a non-British term, but that’s a separate discussion.)

1ec5 (commented):
> I've tried to follow the rules of the CMOS

_The Chicago Manual of Style_ is a well-respected style guide, though it’s far from the only style guide for American English. It’s primarily designed for academic prose, whereas this is UI text, which sometimes has different considerations. (For instance, note that Apple has both an _Apple Style Guide_ for prose, based on CMOS, and a separate _Human Interface Design Guidelines_.) I previously explored the terrain of UI style guides in #473.

I suspect the spacing and punctuation was originally chosen for aesthetics but probably not with a set standard in mind, so later contributions just did whatever. It’s something English writers normally don’t think much about, but it does matter to UI designers. The rules you’re applying are reasonable and would reduce some ambiguity, but it would be just as reasonable to always pad the slashes with spaces if you think that visual consistency would look more aesthetically pleasing.

Broadly speaking, I think we should be wary of proliferating slashes in preset names. This was originally just a pragmatic solution for when OSM tags made distinctions that were impossible to convey in a single idiomatic word, but if we overuse slashes ambivalently, then users will be less confident in their choice of presets.

data/presets/highway/motorway_junction.json
1ec5: #893

matkoniecz:
@FloEdelmann you made some rewrite tools, maybe you would be interested in making space-padded `/` PR version?

matkoniecz:
done #1817

PR #884, opened 2023-04-23
stop assuming that all shop=organic are supermarkets

matkoniecz:
also, some may be `organic=yes` so even migrating to `shop=yes` is not viable - this requires human review

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1470,10 +1470,6 @@
     "replace": {"shop": "stationery"}
   },
   {
-    "old": {"shop": "organic"},
-    "replace": {"shop": "supermarket", "organic": "only"}
-  },
-  {
     "old": {"shop": "perfume"},
     "replace": {"shop": "perfumery"}
   },


Hufkratzer:
Another alternative would probably be [shop=convenience](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dconvenience) + organic=yes/only. Maybe iD can be enhanced to allow the user to select from two or more alternatives. This could also be used for other tags, e.g. for [sport=diving](https://wiki.openstreetmap.org/wiki/Tag%3Asport%3Ddiving), which could be scuba_diving or cliff_diving.

tyrasd:
I always read the [wiki](https://wiki.openstreetmap.org/w/index.php?title=Key:organic&oldid=2465062) that `organic=only` does not mean that literally _only_ organic products are offered:

> All or **almost all** of the products the shop or restaurant offers are organic. These institutions normally **promote themself as "organic"**

(highlights by me)

> `shop=yes organic=only` would be viable

Sure, but I would argue that such a tag change would only "hide" the underlying data quality problem we have with this tag: We still don't know which shop type this is and in general some manual research is needed to fix that. Basically, the tag deprecation / upgrade rules here are just not a good place for this type of data quality issue.

matkoniecz:
> would only "hide" the underlying data quality problem we have with this tag: We still don't know which shop type this is

`shop=yes` communicates exactly this (and triggers complaints/special support in few tools like JOSM/StreetComplete/Osmose and shows in iD as underspecified, while `shop=organic` shows like thousands of other ATYL shop types (many of them valid!)

PR #888, opened 2023-04-25
Add 'cyclestreet'-field

pietervdvn:
Cyclestreets are a special legal designation in some countries, where motorized vehicles are not allowed to overtake cyclists (see [the wiki page](https://wiki.openstreetmap.org/wiki/Key:cyclestreet)).

This PR adds a 'cyclestreet'-field and adds it as extra option to residential (and thus also unclassified) roads

This type of road is mostly in use in Belgium, the Netherlands, Germany and Finland. Ideally, this field is thus only an option in these countries but I couldn't figure out how to trigger this.

### DIFF
--- a/data/presets/highway/residential.json
+++ b/data/presets/highway/residential.json
@@ -12,6 +12,7 @@
     "moreFields": [
         "covered_no",
         "cycleway",
+        "cyclestreet",
         "flood_prone",
         "incline",
         "junction_line",
--- a/data/fields/cyclestreet.json
+++ b/data/fields/cyclestreet.json
@@ -1,0 +1,11 @@
+{
+  "key": "cyclestreet",
+  "type": "defaultCheck",
+  "label": "Cyclestreet",
+  "strings": {
+    "options": {
+      "undefined": "No",
+      "yes": "Yes"
+    }
+  }
+}
--- a/interim/source_strings.yaml
+++ b/interim/source_strings.yaml
@@ -1502,6 +1502,15 @@ en:
       cycle_network:
         # cycle_network=*
         label: Network
+      cyclestreet:
+        # cyclestreet=*
+        label: Cyclestreet
+        options:
+          # cyclestreet=undefined
+          undefined: 'No'
+          # cyclestreet=yes
+          'yes': 'Yes'
+        terms: '[translate with synonyms or related terms for ''Cyclestreet'', separated by commas]'
       cycleway:
         # cycleway:left=*, cycleway:right=*
         label: Bike Lanes


1ec5:
> This type of road is mostly in use in Belgium, the Netherlands, Germany and Finland. Ideally, this field is thus only an option in these countries but I couldn't figure out how to trigger this.

Yes, it would be a good idea to limit this field to the countries that have accepted this key, to avoid exacerbating the existing confusion between this key and `bicycle_road` and also prevent U.S. mappers from using this key for “bike boulevards”. Here’s an example of limiting a field to the U.S.:

https://github.com/openstreetmap/id-tagging-schema/blob/7617c45325036f33190895c02529511f61eebdac/data/fields/expressway-US.json#L5-L9

pietervdvn:
Thanks for the hint, I've added an include-set (and ran the linter in the meantime).

data/presets/highway/residential.json
1ec5, line 1: Is a cyclestreet guaranteed to be a `highway=residential`, or can it also be on a road classified differently, like `unclassified` or `service`?

pietervdvn: The unclassified-highway [copies all 'extra'-tags from residential](https://github.com/openstreetmap/id-tagging-schema/blob/686d8447e6f062e4207641c795965e4c007ed553/data/presets/highway/unclassified.json#L7). This new field should thus appear at `unclassified` as well.

For the other types, I ran the numbers with a global overpass-entry and a little script:

| Amount | highway-type |
|--------|--------------|
|   7864 | residential (71.2%)
|   1378 | unclassified (12.4%)
|    806 | tertiary (7.3%)
|    339 | living_street (3%)
|    152 | secondary
|    139 | service
|    115 | pedestrian
|     79 | cycleway
|     53 | track
|     33 | primary
|     25 | footway
|     21 | path
|      9 | null
|      7 | construction
|      3 | tertiary_link
|      3 | proposed
|      2 | primary_link
|      1 | street_lamp
|      1 | secondary_link

I have to admit I'm a bit surprised by the high amount of tertiary roads in there, I suspected them to be less.

Also, it doesn't really make sense on _some_ values, at least not from the perspective of a Belgian ;) (e.g. primary or 'street_lamp')

tyrasd (changes requested):
Hi. Thanks a lot for this PR. I noticed a couple of things:

* Starting with a small technicality: We use the convention that regional fields should use a file name suffix indicating which countries it applies to. Could you please rename it to `cyclestreet-BE-NL.json`?
* Germany does not use the `cyclestreet` tag, but instead the [`bicycle_road`](https://wiki.openstreetmap.org/wiki/Key:bicycle_road). Could you please remove `DE` from the `locationSet` here?
* In Finland, both `cylcestreet` and `bicycle_road` are currently used very rarely (only on 6 km and 7 km, respectively). The OSM wiki lists the corresponding road sign on the pages for both tags, which is confusing. If the [description](https://wiki.openstreetmap.org/wiki/Key:cyclestreet#Finland) is correct, the more matching tag seems to be `cyclestreet`, but it would be nice if this could be clarified before we include it in the presets. For now, we could omit `FI`.
* I think the use on `highway=tertiary` roads is not unreasonable for a street which has some local importance in the road network, but still gives preference to cyclists. I found these to be somewhat common in city centers, like [in this example](https://www.openstreetmap.org/way/150016025). Could you add the field also to the preset for `highway=tertiary`?
* Similarly, the combined use with `highway=living_street` does also seem to be possible (e.g. when a larger cyclestreet-zone contains individual roads which are signposted as living streets). I'd be inclined to allow the field also in the living street preset.

tyrasd:
One additional tricky thing is that, we eventually would also like add a field for the `bicycle_road` tag. So, I'm wondering whether we would need to choose a slightly more descriptive field label, to properly distinguish between the two tags/fields. Note that the as the term _cyclestreet_ at least in principle applies to both concepts/tags. Can it be assumed that no country will ever use both tags? :monocle_face:

pietervdvn:
Hey @tyrasd ,

I have renamed the file and added it to 'tertiary' and 'living_street'.

There is - at least in Belgium - a meaningful difference between 'cyclestreet' and 'bicycle_road'. A cyclestreet is a street where cars are allowed, but may not overtake. This is often an already existing road where cars are slowed down a bit.

As I interpreted the wiki, a bicycle_road resembles what we would tag with 'cycleway', a purpose-built, wide road where cars are not allowed.

tyrasd (approved):

tyrasd (changes requested):
Thanks for the updates @pietervdvn, however it seems like the `cyclestreet-BE-NL.json` is missing in the last commit of this PR. :wink:

tyrasd:
> As I interpreted the wiki, a bicycle_road resembles what we would tag with 'cycleway', a purpose-built, wide road where cars are not allowed.

For a bit of additional context: In Germany, most `bicycle_road=yes` roads are actually not _purpose-build_, but rather converted residential roads (where car traffic is either allowed at low speeds [example](https://www.openstreetmap.org/way/37685272) or restricted to residents [example](https://www.openstreetmap.org/way/59227112)). For these, the `highway=cycleway` tag would not be a good fit IMO.

Anyway, from the iD editor point of view, it is fine that Belgium and Netherlands uses a dedicated / more specialized tag. I'm happy to support these regional variations. I was initially only a bit worried that we might have to support both tags in parallel in some countries.

pietervdvn:
> Thanks for the updates @pietervdvn, however it seems like the `cyclestreet-BE-NL.json` is missing in the last commit of this PR. wink

Derp, I've actually added it now.

tyrasd (approved):

tyrasd:
Thanks for the PR. :bow: 

FYI: I've now also added a field for `bicycle_road` in 388871fd. Note that the two fields are sharing the same translation-strings. So when translating in transifex, you only will only find the one for the `bicycle_road`. But its translations will apply to the fields of both tags.

PR #889, opened 2023-04-25
Make `crossing` line searchable

Dimitar5555:
I've also changed the name to "Crossing" so that's in line with the changes made in #837.

### DIFF
--- a/data/presets/highway/footway/crossing.json
+++ b/data/presets/highway/footway/crossing.json
@@ -27,5 +27,5 @@
         "key": "footway",
         "value": "crossing"
     },
-    "name": "Crossing"
+    "name": "Pedestrian Crossing"
 }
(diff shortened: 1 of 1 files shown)


data/presets/highway/footway/crossing.json
```diff
@@ -8,17 +8,24 @@
...
+        "foot path crossing",
+        "pedestrian crossing"
+    ],
     "reference": {
         "key": "footway",
         "value": "crossing"
     },
-    "searchable": false,
-    "name": "Pedestrian Crossing"
+    "name": "Crossing"
```
tyrasd, line 30:
```suggestion
    "name": "Pedestrian Crossing"
```

I would leave the name as is, because contrary to the `highway=crossing` preset, here we can and need to distinguish between Foot and Bike crossings

tyrasd (changes requested):

tyrasd (approved):

tyrasd:
merged as 89f7343710668f82184a40ca239a7ee4ad20cc43

PR #891, opened 2023-04-26
Create `crossing/markings-BG.json`

### LINKED ISSUE #876: Hide irrelevant 'crossing:markings` values
**Describe the bug**
Not all values of the `crossing:markings` are used in each country. It would be good to show only ones which are used in the currently edited country. For example, in Bulgaria you will find only `dashed`, `dotted` and `zebra`

**To Reproduce**
1. Go to https://www.openstreetmap.org/node/5337183644
2. Click "Edit"
3. Select the crossing markings field

**Expected behavior**
Only relevant values should be present for each country (if that's even possible).

**Screenshots**
image

Dimitar5555:
Resolves #876

### DIFF
--- a/data/fields/crossing/markings-BG.json
+++ b/data/fields/crossing/markings-BG.json
@@ -1,0 +1,25 @@
+{
+    "key": "crossing:markings",
+    "type": "combo",
+    "label": "{crossing/markings}",
+    "strings": {
+        "options": [
+            "zebra",
+            "no",
+            "dots",
+            "dashes",
+            "yes"
+        ]
+    },
+    "icons": {
+        "zebra": "iD-crossing_markings-zebra",
+        "dots": "iD-crossing_markings-dots",
+        "dashes": "iD-crossing_markings-dashes",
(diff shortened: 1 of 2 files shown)


Dimitar5555:
@tyrasd Any ideas why it's not working correctly? The screenshot is from a crossing in Bulgaria.

image

Dimitar5555:
At least the field shows now. Not sure why it's not the right one though..
image

tyrasd:
Sorry for the late reply, but the issue you're seeing is caused by the regional field not specifying `"autoSuggestions": false`. In the `autoSuggestions=true` case (which is the default), iD will query taginfo for popular values of the tag and show those in addition to the supplied `options`.

Dimitar5555:
Thanks, it is working as intended now.
image

tyrasd:
Thanks!

PS: I noticed that the new regional field was not included in the `*/traffic_signals.json` files. I assume that was an oversight, so I just added it in 57fe16f2

PR #897, opened 2023-05-05
Add Açaí Bowl Fast Food preset

arch0345:


tyrasd:
This tag is still pretty rare with only 177 uses worldwide (of which 145 use the `amenity=fast_food` POI tag). I'd say that this is still a bit too niche to warrant its own preset.

Issue #683, opened 2022-12-10
Presets for popular Seamark tags


k-yle:
This issue tracks the progress of creating presets for popular [Seamark tags](https://wiki.osm.org/Seamarks/Seamark_Objects).

Numbers indicate tag usage as of December 2022.

## Buoys and Beacons

Currently being discussed in #512

- Buoys
  - [ ] [Cardinal Buoy](https://wiki.osm.org/Tag:seamark:type=buoy_cardinal) `10k`
  - [ ] [Isolated Danger Buoy](https://wiki.osm.org/Tag:seamark:type=buoy_isolated_danger) `600`
  - [x] [Lateral Buoy](https://wiki.osm.org/Tag:seamark:type=buoy_lateral) (a.k.a. Channel Buoy) `56k` [iD#5297](https://github.com/openstreetmap/iD/pull/5297)
  - [ ] [Safe Water Buoy](https://wiki.osm.org/Tag:seamark:type=buoy_safe_water) `1k`
  - [ ] [Special Purpose Buoy](https://wiki.osm.org/Tag:seamark:type=buoy_special_purpose) (a.k.a. Generic Buoy) `12k`
  - [x] Mooring - see below
  - [ ] Visitor Mooring - see below
- Beacons
  - [ ] [Cardinal Mark](https://wiki.osm.org/Tag:seamark:type=beacon_cardinal) `1k`
  - [x] [Isolated Danger Mark](https://wiki.osm.org/Tag:seamark:type=beacon_isolated_danger) `14k` [iD#5297](https://github.com/openstreetmap/iD/pull/5297)
  - [x] [Lateral Mark](https://wiki.osm.org/Tag:seamark:type=beacon_lateral) (a.k.a. Channel Beacon) `27k` [iD#5297](https://github.com/openstreetmap/iD/pull/5297)
  - [ ] ~~[Safe Water Mark](https://wiki.osm.org/Tag:seamark:type=beacon_safe_water) `98`~~[^1]
  - [ ] [Special Purpose Mark](https://wiki.osm.org/Tag:seamark:type=beacon_special_purpose) (a.k.a. Generic Mark) `14k`

## Other feasible presets

- [x] [Piers/Jettys/Pontoons](https://wiki.osm.org/Tag:man_made=pier) `572k`
- [x] [Mooring](https://wiki.osm.org/Tag:seamark:type=mooring) `58k` [iD#5297](https://github.com/openstreetmap/iD/pull/5297)
[...]

PR #899, opened 2023-05-09
do not suggest to add unnecessary duplicative seamark tags

matkoniecz:
seamark tagging scheme in general creates unneeded parallel tagging schema, there is no need to support it an bother user about it

### DIFF
--- a/data/presets/leisure/marina.json
+++ b/data/presets/leisure/marina.json
@@ -35,9 +35,7 @@
         "leisure": "marina"
     },
     "addTags": {
-        "leisure": "marina",
-        "seamark:type": "harbour",
-        "seamark:harbour:category": "marina"
+        "leisure": "marina"
     },
     "name": "Marina"
 }


tyrasd:
For reference: these tags were added in https://github.com/openstreetmap/iD/pull/7622. See also #683

My first impression is that while potentially duplicating some information, the `seamark` tags don't generally hurt[^1] and could be potentially useful for some data consumers like specialized renderers of water facilities. //cc @Eric-Sparks @k-yle @jsavage maybe one of you could give us some input on how these tags are used in practice?

However, from what I can see, `leisure=marina` could also be used with the value `seamark:harbour:category=marina_no_facilities`, meaning that the tag might not always be the correct one. :thinking: 

[^1]: Yes, it might be somewhat distracting to see the tag upgrade message on existing `leisure=marina` objects. But that could be prevented by replacing the `addTags` with a dedicated field for the `seamark:type` tag with an appropriate default value.

k-yle:
It would be a real shame to stop supporting `seamark` tags purely because the current design of iD's tag-upgrade warning is a bit intrusive :(

With this argument, should the warnings to add `healthcare=*` to `amenity=*` also be removed because they are two parallel tagging schemes?

There's a lot of things i don't like about seamark tags, but it's become well established and this is one of the very few cases where there is an equally popular tag in the normal tag system... There are very few marinas in the database, even in big coastal cities, so has anyone actually found this specific warning annoying?

tyrasd:
To make it clear: this was not meant to stop supporting the _seamark_ tags altogether. Only for specific cases (like for marinas), where their value could be implied by another tag might be "considered redundant".

But I think you have a very good point: Like in the healthcare case, we should consider the use-case specific tags (here: `seamark:*`) as the "main" tag of these presets, and the `leisure`/`amenity` tags should be considered as merely kept around for backwards compatibility. I wish the wiki could be more bold or explicit about this, though: Currently, the page for [`leisure=marina`](https://wiki.openstreetmap.org/w/index.php?title=Tag:leisure%3Dmarina&oldid=2429879) doesn't even mention the `seamark:*` tags at all.

You convinced me: I think, in general, keeping this as a proper _tag upgrade_ does actually make sense to keep.

Any feedback on the `marina_no_facilities` value? A way to toggle between the two possible values (`seamark:harbour:category=marina` vs. `seamark:harbour:category=marina_no_facilities`) should probably be at least available as a field, doesn't it?

Eric-Sparks:
So, to be clear, leisure=marina doesn't have a 1:1 relationship with a seamark tag.  seamark:harbour:category=marina means that the facility has services that a captain might be looking for.  seamark:harbour:category=marina_no_facilities means that the marina basically has a place to tie up.  So, it is proper to have leisure=marina _and_ the seamark at the same time since the seamark is more descriptive which is important to a maritime user.

matkoniecz:
In cases where it is not "always implies seamark:XYZ" then such "tag upgrade" (which BTW should be maybe renamed to make clear that it is automated and not very smart replacement) is damaging data.


> the seamark tags don't generally hurt

Well, in cases where there specific tag always indicates some seamark tags then in such case they are pointless duplicates and pure waste of time to add them.

seamark: especially are irritating as they appear to basically have parallel tagging schema, with `seamark:small_craft_facility:category=toilets` being a good example of a problem.

> the leisure/amenity tags should be considered as merely kept around for backwards compatibility

not really, `amenity=toilets` is in every way superior to `seamark:small_craft_facility:category=toilets` to take one example (and someone interested in only nearby toilets can just filter `amenity=toilets` rather having multiprefixed tags)

matkoniecz:
> Any feedback on the marina_no_facilities value?

Is it still `leisure=marina` when you have no facilities whatsoever? In such cases I mapped anchorage/buoy moorings/moorings but no `leisure=marina` as there was no marina there (what makes me quite suspicious about also `seamark:harbour:category=marina_no_facilities` but my sailing experience is not extensive)

matkoniecz:
> so has anyone actually found this specific warning annoying?

me, that is why I created this PR

Eric-Sparks:
> > Any feedback on the marina_no_facilities value?
> 
> Is it still `leisure=marina` when you have no facilities whatsoever? In such cases I mapped anchorage/buoy moorings/moorings but no `leisure=marina` as there was no marina there (what makes me quite suspicious about also `seamark:harbour:category=marina_no_facilities` but my sailing experience is not extensive)

Yes, it's still a marina.  A marina implies a place to dock that isn't an anchorage that is open for non-owners to use.  If there are facilities (and "facilities" has a wide meaning), then it is still a marina, it just has added amenities.

Eric-Sparks:
> not really, `amenity=toilets` is in every way superior to `seamark:small_craft_facility:category=toilets` to take one example (and someone interested in only nearby toilets can just filter `amenity=toilets` rather having multiprefixed tags)

You have to take this in context.  seamark:small_craft_facility:category=toilets specifically calls out a toilet that is built/maintained for small craft users.  It doesn't stop people from using other toilets or for non-boaters from using the toilet, but when you are looking at a sea map, you can see the facilities that are specifically for you (as a boater).

jsavage:
> marina_no_facilities

Yes!  I agree that the addition of "no_facilities" is meaningful - especially for sailors planning a route.   Certainly in the UK, it is common place for toilet facilities not be for exclusive use of the users of the marinas not the general public so appropriate that they only appear as such in OpenSeaMap. 

What I do this is a bit odd is the use of the term 'seamark' (presumably as opposed to landmark) for anything that needs to be included in OpenSeaMap that is is not in OpenStreetMap.

tyrasd:
Hm… The not a 1:1 match argument by @matkoniecz is actually a valid critique here. :thinking: On the other hand I would also like to support openseamap tagging in iD wherever reasonably possible.

It is a bit unfortunate is that the `seamark` tags are generally quite poorly defined/documented on the wiki, which makes it especially hard to evaluate this for people which are unfamiliar with the topic and/or domain specific terms. E.g., as a mapper, I wouldn't know what _facilities_ to look out for when deciding which kind of marina it is.

<hr>

What about the following solution: We keep the current tag upgrade of the `leisure=marina` preset (under the assumption that what mappers usually associate with `leisure=marina` does more commonly have "extra facilities" than not), and add a checkbox-field to de-select "has extra facilities"[^1]. At the same time we also add a preset for `seamark:harbour:category=marina_no_facilities` which is titled _Yacht Berths_ ([see wiki](https://wiki.openstreetmap.org/w/index.php?title=Key:seamark:harbour:category&oldid=2379148)) and has _marina_ as one of its search terms.

[^1]: again: a better description of which "extra facilities" are expected would be quite useful, though.

jsavage:
[tyrasd](https://github.com/tyrasd) commented that a better description of which "extra facilities" are expected would be quite useful.   I interpret this as a request to clarify what facilities should be mapped in OpenSeaMap using seamarks.

In thinking about this I do think that mappers should realise that sailors and other seafarers typically have cruising guides, pilot books, almanacs and other documents that they can refer to when planning passages.  These documents already provide details of the facilities that are available - albeit in different ways!    One popular publication is Reeds Nautical Almanac.
This describes facilities at 3 levels of detail:
**A qualitative rating for each harbour - eg  facilities available:**   
1. Possibly some domestic facilities but little else.  
3. Most domestic needs catered for but limited boatyard facilities  
3. Good facilities for vessel and crew

**For each harbour or marina a paragraph entitled Facilities:**
This describes berthing options and facilities including berthing rates
Underneath this are relevant symbols for facilities available.
 
The symbols used for every harbour and marina covered within the 1000+ pages:
  
-----
Other publications provide similar info.  Often in much less detail than Reeds.

It seems to me that in relation to facilities, seafarers need to know:
1. What facilities are available  <<<leave for the Almanacs to cover
2. Where these facilities are   **<<< Concentrate on this - and in particular the facilities that are approached by from the water**  Eg. Visitors berths.   Fuel jetty,  Refuse Barge, pump out facilities.  In small marinas if they are all within 10 metres of each other then there may be little benefit in mapping these but in larger harbours mapping these is much more likely to be useful. 
3. When (opening times)   <<<leave for the Almanacs to cover
4. How (special instructions)   <<<leave for the Almanacs to cover

 I hope the above is of some help in this discussion

NB. I will add some images later to illustrate

Eric-Sparks:
> What about the following solution: We keep the current tag upgrade of the `leisure=marina` preset (under the assumption that what mappers usually associate with `leisure=marina` does more commonly have "extra facilities" than not), and add a checkbox-field to de-select "has extra facilities"[1](#user-content-fn-1-1525c61ba90e90725b2e41a73378f580). At the same time we also add a preset for `seamark:harbour:category=marina_no_facilities` which is titled _Yacht Berths_ ([see wiki](https://wiki.openstreetmap.org/w/index.php?title=Key:seamark:harbour:category&oldid=2379148)) and has _marina_ as one of its search terms.

I think this would be a good solution.

PR #911, opened 2023-05-28
Add Length Limit and Axle Weight Limit fields

arch0345:
Labels match what [MUTCD](https://wiki.openstreetmap.org/wiki/MUTCD) signs use:
image
image

Also:
* added these fields to moreFields of appropriate highway presets
* changed maxweight field label to Weight Limit to match MUTCD signage:
image
* swapped maxweight_bridge for maxweight field for Residential Road preset as it's more common for minor roads like these to have a weight limit that's not just at a bridge


1ec5:
While you’re at it, this “Max” terminology also appears in the `maxwidth` field, which could be renamed to avoid abbreviating “maximum”:

https://github.com/openstreetmap/id-tagging-schema/blob/6fd8dc03253ca391610774f95e96515a9d5606a5/data/fields/maxwidth.json#L4

These are called “[Width Limit](https://dot.ca.gov/programs/safety-programs/sign-specs)” or “Width Restriction” in the U.S., though few signs refer to the restriction by a noun.

[Over 10 Feet Wide / Use Alternative Route](https://commons.wikimedia.org/wiki/File:Interstate_90_in_South_Dakota_%2834978356456%29.jpg)

data/fields/maxheight.json
```diff
@@ -1,6 +1,6 @@
 {
     "key": "maxheight",
     "type": "roadheight",
-    "label": "Max Height",
+    "label": "Vertical Clearance",
```
1ec5, line 4: Though this term is officially used in the U.S., it may be less familiar to some users who see these signs and think of them as simply “height limits”. Fortunately, schema-builder lets you specify `terms` for this field that will show up when searching for the field in iD. It’s probably a good idea to add all the previous labels as terms to accommodate existing users’ muscle memory.

tyrasd (approved):

PR #912, opened 2023-05-28
Add Bridge Number field

arch0345:


1ec5:
For bridges that have been mapped as `man_made=bridge` areas, the Bridge Area preset should also have a `ref` field called Bridge Number:

https://github.com/openstreetmap/id-tagging-schema/blob/6fd8dc03253ca391610774f95e96515a9d5606a5/data/presets/man_made/bridge.json#L3-L8

tyrasd (approved):

PR #917, opened 2023-06-06
Add EV Charging Stall preset

### LINKED ISSUE #896: Add support for man_made=charge_point
### OSM Tag(s)

man_made=charge_point

As a Preset

### Label

Electric Vehicle Charge Point

### Aliases

Charging Stall

### Terms

Charging Station

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dcharge_point

### Status of the Tag

Approved

### Usage of the tag

22

### Replaces other Tag?

No

### Regional Tag?

No

### Further Information

Just now a proposal based on a draft by me was approved, leading to the creation of this tag: [https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dcharge_point](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dcharge_point)
It'd be nice if this was recognized by iD. Currently there are only 22 uses, but this expected to increase very rapidly, since the proposal has only just been approved.
The only fields that would go along with this tag would be `ref=*` and those to enter details on the sockets the charge points have.

arch0345:
Closes #896 

Although usage is [still relatively low](https://taginfo.openstreetmap.org/tags/man_made%3Dcharge_point), the addition of this preset will prevent new mappers from accidentally selecting the wrong brand presets in NSI. For example, [this ALDI store](https://www.openstreetmap.org/way/1176949590) was mistakenly mapped as a `man_made=charge_point`.

### DIFF
--- a/data/presets/man_made/charge_point.json
+++ b/data/presets/man_made/charge_point.json
@@ -18,7 +18,7 @@
         "ev",
         "electric vehicle",
         "charging stall",
-        "charging station"
+        "charging station",
         "charge point"
     ],
     "name": "EV Charging Point"
--- a/data/presets/man_made/charge_point.json
+++ b/data/presets/man_made/charge_point.json
@@ -1,0 +1,33 @@
+{
+    "icon": "fas-charging-station",
+    "fields": [
+        "brand",
+        "operator",
+        "capacity",
+        "access_simple",
+        "fee",
+        "payment_multi_fee",
+        "charge_fee",
+        "ref"
+    ],
+    "moreFields": [
+        "covered",
+        "level",
+        "manufacturer",
+        "maxstay",
+        "network"
+    ],
+    "geometry": [
+        "point"
+    ],
+    "tags": {
+        "man_made": "charge_point"
+    },
+    "terms": [
+        "ev",
+        "electric vehicle",
+        "charging station",
+        "charge point"
+    ],
+    "name": "EV Charging Stall"
+}
--- a/data/presets/man_made/charge_point.json
+++ b/data/presets/man_made/charge_point.json
@@ -1,21 +1,12 @@
 {
-    "icon": "fas-charging-station",
+    "icon": "fas-plug-circle-bolt",
     "fields": [
-        "brand",
-        "operator",
         "capacity",
-        "access_simple",
-        "fee",
-        "payment_multi_fee",
-        "charge_fee",
         "ref"
     ],
     "moreFields": [
-        "covered",
         "level",
-        "manufacturer",
-        "maxstay",
-        "network"
+        "manufacturer"
     ],
     "geometry": [
         "point"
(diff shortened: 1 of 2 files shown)


data/presets/man_made/charge_point.json
```diff
@@ -0,0 +1,33 @@
...
+    "tags": {
+        "man_made": "charge_point"
+    },
+    "terms": [
+        "ev",
+        "electric vehicle",
+        "charging station",
+        "charge point"
+    ],
+    "name": "EV Charging Stall"
```
tyrasd, line 32: Isn't _charge point_ the more common term for these devices in English?

data/presets/man_made/charge_point.json
```diff
@@ -0,0 +1,33 @@
...
+    "fields": [
+        "brand",
+        "operator",
+        "capacity",
+        "access_simple",
+        "fee",
+        "payment_multi_fee",
+        "charge_fee",
+        "ref"
+    ],
```
tyrasd, line 12: I'm not so sure about the fields to be used for this preset. The [wiki page](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dcharge_point) (and [proposal](https://wiki.openstreetmap.org/wiki/Proposal:EV_Charging_Station_Mapping)) don't seem to suggest to use the tags from the main `amenity=charging_station` like `brand`, `operator`, `fee` also on the individual charging points of the station. Isn't that similar to how the benches inside a park typically don't repeat the operator of the park, or the individual `parking_space`s of a car park don't repeat the `fee` the of the parking area they are in, for example?

data/presets/man_made/charge_point.json
```diff
@@ -0,0 +1,33 @@
+{
+    "icon": "fas-charging-station",
```
tyrasd, line 2: Reusing the same icon is not great to distinguish the features. 
I'm not sure what would be the best solution here, but maybe this preset could just use the `fas-plug` icon?

tyrasd (commented):
see some comments below

data/presets/man_made/charge_point.json
arch0345: I think that charge point is more of a British term[^1][^2][^3], at least on the west coast these are usually referred to as charging stalls[^4][^5][^6]

[^1]
[^2]
[^2]

arch0345: ~~I think that charge point is more of a British term, these are usually called charge stalls on the west coast of the US~~

Never mind, it's probably best to use 'point' since 'stall' could refer to both the charger and parking space. I'll change the label to _charging point_ instead of _charge point_ to disambiguate this preset from the brand [ChargePoint](https://www.chargepoint.com/)

tyrasd (approved):

Issue #879, opened 2023-04-22
add scheme to replace parking:lane:_side_=parallel/diagonal/perpendicular


tiptoptom:
### Deprecated OSM Tag(s)

parking:lane:_side_=parallel/diagonal/perpendicular

### Replacement OSM Tag(s)

parking:_side_=lane
parking:_side_:orientation=parallel/diagonal/perpendicular

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Street_parking
old scheme: https://wiki.openstreetmap.org/wiki/Key:parking:lane

tordans:
If this is something that would be merged we can add a few others as well. 

In the meantime there is https://github.com/osmberlin/osm-tag-updater and the Zlant editor has a updater integration (well hidden behind the 🔄 icon).

tyrasd:
The implementation of tag deprecation rules here does not scale up super well if there are many versions of a tag to be considered… I'd say that adding these 9 cases would be fine, but (much) more than that would be a stretch.

@tordans Do you think you could provide the "business logic" part of the @osmberlin/osm-tag-updater as an npm module? Because, then I could quite easily build that completely into iD directly as a dedicated validation module.

tiptoptom:
So I will now make a PR for the mentioned 9 cases?

tordans:
> But, from a technical standpoint, I have to note that the implementation of tag deprecation rules here does not scale up well if there are many versions of a tag to be considered… I'd say that adding these 9 cases would be still fine, but (much) more than that would be a stretch.

The proposal lists some data at https://wiki.openstreetmap.org/wiki/Proposal:Street_parking_revision#Numbers_of_usage_of_the_current_schema

> Currently (as of the end of October 2022), about 260.000 road segments in OSM have "primary" parking lane tagging ([parking:lane:both](https://wiki.openstreetmap.org/wiki/Key:parking:lane:both)=*, [parking:lane:left](https://wiki.openstreetmap.org/wiki/Key:parking:lane:left)=* or [parking:lane:right](https://wiki.openstreetmap.org/wiki/Key:parking:lane:right)=*), (…)

All other old tags are tags are well below the 50k mark, so I think we should ignore those for deprecation rules.

Following this logic, we could add a deprecation rule for those cases:

Source: https://osmberlin.github.io/osm-tag-updater/#/list

[...]

PR #918, opened 2023-06-08
 add scheme to replace parking:lane:_side_=parallel/diagonal/perpendicular

tiptoptom:
#879

### DIFF
--- a/data/presets/piste/_piste.json
+++ b/data/presets/piste/_piste.json
@@ -37,5 +37,6 @@
         "ice",
         "skating"
     ],
-    "name": "Snowsports Trail / Piste"
+    "name": "Snowsports Trail / Piste (Unspecified Type)",
+    "searchable": false
 }


tyrasd:
@tordans could you take a look at this PR perhaps? Is the assosciation of `parking:lane:both=yes` to `parking:both=lane` fine (I saw that in https://github.com/openstreetmap/id-tagging-schema/issues/879#issuecomment-1584111430 you suggested to also use the value `yes` for the new tag)?

tyrasd:
(small ping @tordans)

data/deprecated.json
```diff
@@ -1205,6 +1205,126 @@
     "old": {"parking": "street"},
     "replace": {"parking": "lane"}
   },
+  {
+    "old": {"parking:lane:both": "yes"},
+    "replace": {"parking:both": "lane"}
```
tordans, line 1210:
```suggestion
    "old": {"parking:lane:both": "yes"},
    "replace": {"parking:both": "yes"}
```

When we build https://osmberlin.github.io/osm-tag-updater/#/manual , we made it that 
INPUT: `parking:lane:both=yes`
OUTPUT: `parking:both=yes`

Our idea being…
- it is likely `lane`, but we cannot be sure
- its better to have it a non standard `yes` so people get nudged to update it, that to auto update to something that might be wrong

data/deprecated.json
```diff
@@ -1205,6 +1205,126 @@
...
+    "old": {"parking:lane:both": "no_standing"},
+    "replace": {"parking:both": "no", "parking:both:restriction": "no_standing"}
+  },
+  {
+    "old": {"parking:lane:both": "no_stopping"},
+    "replace": {"parking:both": "no", "parking:both:restriction": "no_stopping"}
+  },
+  {
+    "old": {"parking:lane:both": "fire_lane"},
+    "replace": {"parking:both": "no", "parking:both:restriction": "no_stopping", "parking:both:restriction:reason": "fire_lane"}
```
tordans, line 1230: Side node: We do not have a preset for the `*:reason` tag, so this will become "invisible" to users that do not look at the "Tags" area. That is OK IMO. Its just a +1 to some day have a even more advanced preset system that would allow to have something like the reason more visible.

data/deprecated.json
```diff
@@ -1205,6 +1205,126 @@
...
+    "replace": {"parking:both": "lane", "parking:both:orientation": "diagonal"}
+  },
+  {
+    "old": {"parking:lane:both": "parallel"},
+    "replace": {"parking:both": "lane", "parking:both:orientation": "parallel"}
+  },
+  {
+    "old": {"parking:lane:both": "perpendicular"},
+    "replace": {"parking:both": "lane", "parking:both:orientation": "perpendicular"}
+  },
```
tordans, line 1247:
```suggestion
  {
    "old": {"parking:lane:both": "diagonal"},
    "replace": {"parking:both:orientation": "diagonal"}
  },
  {
    "old": {"parking:lane:both": "parallel"},
    "replace": {"parking:both:orientation": "parallel"}
  },
  {
    "old": {"parking:lane:both": "perpendicular"},
    "replace": {"parking:both:orientation": "perpendicular"}
  },
```

Same as with my comment above: I would think we should rather be more conservative when updating the tags that auto-fill data that might not be true. When you add `parking:lane:both= diagonal` to https://osmberlin.github.io/osm-tag-updater/#/manual we only transform the given key but show a yellow info to add the missing key manually. 
Unfortunately, the way iD works is that the orientation field will become visible, but the "parking" field does not.
It would be better to have it show like …
image

However I tried that but we are missing the capabilities to show fields conditionally depending on other tags.

Since the deprecations happen automatically (as in, the user is not made aware that things change), I would still vote to only update what we are 100% sure about.

data/deprecated.json
```diff
@@ -1205,6 +1205,126 @@
...
+    "replace": {"parking:left": "lane", "parking:left:orientation": "diagonal"}
+  },
+  {
+    "old": {"parking:lane:left": "parallel"},
+    "replace": {"parking:left": "lane", "parking:left:orientation": "parallel"}
+  },
+  {
+    "old": {"parking:lane:left": "perpendicular"},
+    "replace": {"parking:left": "lane", "parking:left:orientation": "perpendicular"}
+  },
```
tordans, line 1287: Same as above

```suggestion
  {
    "old": {"parking:lane:left": "diagonal"},
    "replace": {"parking:left:orientation": "diagonal"}
  },
  {
    "old": {"parking:lane:left": "parallel"},
    "replace": {"parking:left:orientation": "parallel"}
  },
  {
    "old": {"parking:lane:left": "perpendicular"},
    "replace": {"parking:left:orientation": "perpendicular"}
  },
```

data/deprecated.json
```diff
@@ -1205,6 +1205,126 @@
...
+    "replace": {"parking:right": "lane", "parking:right:orientation": "diagonal"}
+  },
+  {
+    "old": {"parking:lane:right": "parallel"},
+    "replace": {"parking:right": "lane", "parking:right:orientation": "parallel"}
+  },
+  {
+    "old": {"parking:lane:right": "perpendicular"},
+    "replace": {"parking:right": "lane", "parking:right:orientation": "perpendicular"}
+  },
```
tordans, line 1327: Same as above

```suggestion
  {
    "old": {"parking:lane:right": "diagonal"},
    "replace": {"parking:right:orientation": "diagonal"}
  },
  {
    "old": {"parking:lane:right": "parallel"},
    "replace": {"parking:right:orientation": "parallel"}
  },
  {
    "old": {"parking:lane:right": "perpendicular"},
    "replace": {"parking:right:orientation": "perpendicular"}
  },
```

tordans (changes requested):
Sorry for taking so long…

I suggest to being a bit more conservative with the updates.

I talked to @SupaplexOSM he was asking about auto updating 

`parking:lane:<side>:<orientation>=<position>` to `parking:<side>=lane`

That would soften the change I suggest here, since most of the time there is an update path for both _parking_ and _parking orientation._

The disadvantage of adding `parking:lane:<side>:<orientation>=<position>` would be, that those are quite a few combinations of tags (due to the very unfortunate side-orientation-key…) so we would increase the deprecation list quite a bit. However, looking at the usage, only one set is really relevant. Everything below 3k can be be ignored IMO

```csv
parking:lane:both:parallel; 20,170
parking:lane:both:parallel=on_street; 15,311

parking:lane:left:parallel; 13,569
parking:lane:left:parallel=on_street; 7.350

parking:lane:right:parallel; 17,331
parking:lane:right:parallel=on_street; 9,602

parking:lane:both:diagonal; 264
parking:lane:both:marked; 99
parking:lane:both:perpendicular; 1,038
parking:lane:left:diagonal; 1,329
parking:lane:left:marked; 182
parking:lane:left:perpendicular; 2,018
parking:lane:right:diagonal; 1,525
parking:lane:right:marked; 185
parking:lane:right:perpendicular; 2,001
```

That would mean also adding the deprecation rules …

```json
 {
    "old": {"parking:lane:both:parallel": "on_street"},
    "replace": {"parking:both": "lane"}
  },
 {
    "old": {"parking:lane:left:parallel": "on_street"},
    "replace": {"parking:left": "lane"}
  },
 {
    "old": {"parking:lane:right:parallel": "on_street"},
    "replace": {"parking:right": "lane"}
  },
```

Update: I thought about changing this to use the "update tag but keep value" version of the deprecation.json (see https://github.com/ideditor/schema-builder/pull/105/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R710-R717), but that would also preserve values like "marked" which we do not want in the new schema. So better to keep the specific version above.

data/deprecated.json
tiptoptom: But isn't it safe with parkin:**lane**:both that it is lane?

PR #931, opened 2023-06-22
Add preset for office=courier

govvin:
Add preset for office=courier, to improve tagging consistensy, and assist contributors in finding the tag for a [courier delivery service office](https://wiki.openstreetmap.org/wiki/Tag:office%3Dcourier).

Taginfo shows https://taginfo.openstreetmap.org/keys/office#values (search for "courier", "delivery") several similar values that appear to be trying to tag the same thing.

### DIFF
--- a/data/presets/office/courier.json
+++ b/data/presets/office/courier.json
@@ -7,7 +7,6 @@
     ],
     "moreFields": [
         "{@templates/contact}",
-        "{payment}",
         "{@templates/poi}"
     ],
     "geometry": [
--- a/data/presets/office/courier.json
+++ b/data/presets/office/courier.json
@@ -2,14 +2,13 @@
     "icon": "fas-people-carry-box",
     "fields": [
         "name",
-        "address",
-        "opening_hours"
+        "brand",
+        "{office}"
     ],
     "moreFields": [
-        "{@templates/poi}",
-        "air_conditioning",
-        "branch_brand",
-        "brand"
+        "{@templates/contact}",
+        "{payment}",
+        "{@templates/poi}"
     ],
     "geometry": [
         "point",
@@ -19,14 +18,15 @@
         "office": "courier"
     },
     "terms": [
+        "carrier",
         "courier",
+        "courier service",
         "delivery",
-        "express delivery",
         "delivery service",
-        "parcel",
-        "package",
+        "express delivery",
         "messenger",
-        "carrier"
+        "package",
+        "parcel"
     ],
-    "name": "Courier Service Office"
+    "name": "Courier Service"
 }
--- a/data/presets/office/courier.json
+++ b/data/presets/office/courier.json
@@ -1,0 +1,32 @@
+{
+    "icon": "fas-people-carry-box",
+    "fields": [
+        "name",
+        "address",
+        "opening_hours"
+    ],
+    "moreFields": [
+        "{@templates/poi}",
+        "air_conditioning",
+        "branch_brand",
+        "brand"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "office": "courier"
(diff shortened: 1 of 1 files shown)


data/presets/office/courier.json
```diff
@@ -0,0 +1,32 @@
...
+    "terms": [
+        "courier",
+        "delivery",
+        "express delivery",
+        "delivery service",
+        "parcel",
+        "package",
+        "messenger",
+        "carrier"
+    ],
```
tyrasd, line 30: could you please sort these alphabetically?

data/presets/office/courier.json
```diff
@@ -0,0 +1,32 @@
...
+    "fields": [
+        "name",
+        "address",
+        "opening_hours"
+    ],
+    "moreFields": [
+        "{@templates/poi}",
+        "air_conditioning",
+        "branch_brand",
+        "brand"
```
tyrasd, line 12: I'd say, for courier services, the _brand_ information is quite important. Maybe we should promote at least the `brand` field to a regular `field`?

tyrasd (commented):
Thanks! Looks good in general :+1: but please note my two minor comments below.

data/presets/office/courier.json
govvin: Fixed. Thank you for the feedback.

data/presets/office/courier.json
govvin: I agree with you! I'll fix that.

govvin:
> Thanks! Looks good in general +1 but please note my two minor comments below.

Thank you for your feedback @tyrasd !  I've incorporated them in the PR now.

1ec5:
Has there been any discussion about this tag beyond this wiki page? I agree that it’s somewhat incongruous to tag delivery services as post offices, as you pointed out in osmlab/name-suggestion-index#8290, but I’m concerned that existing usage of this tag is actually a mix of post office–like delivery services and [headquarters offices of bicycle messenger services](https://wiki.openstreetmap.org/wiki/Talk:Tag:office%3Dcourier#Bicycle_couriers).

govvin:
> Has there been any discussion about this tag beyond this wiki page? I agree that it’s somewhat incongruous to tag delivery services as post offices, as you pointed out in [osmlab/name-suggestion-index#8290](https://github.com/osmlab/name-suggestion-index/pull/8290), but I’m concerned that existing usage of this tag is actually a mix of post office–like delivery services and [headquarters offices of bicycle messenger services](https://wiki.openstreetmap.org/wiki/Talk:Tag:office%3Dcourier#Bicycle_couriers).

Hi @1ec5 ! Thanks for checking this out. 

Many of the discussions (forums, mailing list) I've seen had to to with the original proposal, but limiting myself to what is written on the current wiki page, I suppose it meets the definition?

I have a slightly longer reply wiki talk link you posted, since that's  easily accessible to more OSM contributors.

1ec5:
I hadn’t noticed that this wiki page was based on [an abandoned proposal](https://wiki.openstreetmap.org/wiki/Proposal:Office%3Dcourier), so I guess that’s something. To be clear, I don’t have any objection to splitting out a new tag for a superset of POIs that’s less specific than `amenity=post_office`; mapping these facilities as post offices has been bothering me since I first starting doing that. I was just surprised to see no mention of the tag on the tagging list, given all the activity around “[postal partners](https://wiki.openstreetmap.org/wiki/Proposal:Shop_as_post-partner)” a couple years ago.

For what it’s worth, these same companies often place [drop-off boxes](https://wiki.openstreetmap.org/wiki/Key:post_box:type#United_States) right next to the postal service’s official mail collection box. These are currently being tagged as `amenity=post_box`, by analogy with `amenity=post_office`.

PR #932, opened 2023-06-22
Create tagging schema for "Barangay Hall"

govvin:
Create tagging schema for "Barangay Hall".

A "barangay" in the Philippines is the smallest administrative government unit.

According to [taginfo](https://taginfo.openstreetmap.org/tags/townhall%3Atype=barangay#overview), "barangay" is the third most used value for `townhall:type`, and one of the most popular POI added by mappers from the Philippines.

I'm new to adding id tagging schemas, so please be gentle with me. :) Feedback is welcome!

### DIFF
--- a/data/presets/amenity/townhall.json
+++ b/data/presets/amenity/townhall.json
@@ -21,6 +21,7 @@
     ],
     "terms": [
         "village",
+        "barangay",
         "city",
         "government",
         "courthouse",
--- a/data/presets/amenity/townhall.json
+++ b/data/presets/amenity/townhall.json
@@ -20,8 +20,7 @@
         "area"
     ],
     "terms": [
-        "village",
-        "barangay",
+        "village"
         "city",
         "government",
         "courthouse",
--- a/data/presets/amenity/townhall.json
+++ b/data/presets/amenity/townhall.json
@@ -20,7 +20,7 @@
         "area"
     ],
     "terms": [
-        "village"
+        "village",
         "city",
         "government",
         "courthouse",
--- a/data/presets/amenity/townhall/barangay.json
+++ b/data/presets/amenity/townhall/barangay.json
@@ -9,7 +9,6 @@
         "area"
     ],
     "terms": [
-        "barangay hall",
         "barrior hall"
     ],
     "tags": {
--- a/data/presets/amenity/townhall/barangay.json
+++ b/data/presets/amenity/townhall/barangay.json
@@ -1,0 +1,25 @@
+{
+    "locationSet": {
+        "include": [
+            "ph"
+        ]
+    },
+    "icon": "fas-building-flag",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "barangay hall"
+    ],
+    "tags": {
+        "amenity": "townhall",
+        "townhall:type": "barangay",
+        "admin_level" : "10"
+    },
+    "reference": {
+        "key": "townhall:type",
+        "value": "barangay"
+    },
+    "name": "Barangay Hall"
+}
--- a/data/presets/amenity/townhall/barangay.json
+++ b/data/presets/amenity/townhall/barangay.json
@@ -4,13 +4,13 @@
             "ph"
         ]
     },
-    "icon": "fas-building-flag",
     "geometry": [
         "point",
         "area"
     ],
     "terms": [
-        "barangay hall"
+        "barangay hall",
+        "barrior hall"
     ],
     "tags": {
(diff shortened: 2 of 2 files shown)


data/presets/amenity/townhall.json
```diff
@@ -21,6 +21,7 @@
     ],
     "terms": [
         "village",
+        "barangay",
```
tyrasd, line 24: It's not necessary to repeat a "name" as  a term for the generic preset, if a dedicated preset exist with that name already. I'd remove it from here.

data/presets/amenity/townhall/barangay.json
```diff
@@ -0,0 +1,25 @@
+{
+    "locationSet": {
+        "include": [
+            "ph"
+        ]
+    },
+    "icon": "fas-building-flag",
```
tyrasd, line 7: Did you choose this one on purpose to be different from the icon of the "generic" town hall preset?

data/presets/amenity/townhall/barangay.json
```diff
@@ -0,0 +1,25 @@
...
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "barangay hall"
+    ],
```
tyrasd, line 14: The search `terms` should not repeat the preset's name.

data/presets/amenity/townhall/barangay.json
```diff
@@ -0,0 +1,25 @@
...
+    "tags": {
+        "amenity": "townhall",
+        "townhall:type": "barangay",
+        "admin_level" : "10"
+    },
+    "reference": {
+        "key": "townhall:type",
+        "value": "barangay"
+    },
```
tyrasd, line 23: as far as I can see, there is no wiki page dedicated to the `townhall:type=barnangay` tag yet (see https://wiki.openstreetmap.org/wiki/Tag:townhall:type=barangay).

We could either link to the documentation page of the tag key (https://wiki.openstreetmap.org/wiki/Key:townhall:type) or `amentiy=townhall` (https://wiki.openstreetmap.org/wiki/Tag:amenity=townhall) instead. I'd slightly lean towards the latter:

```suggestion
    "reference": {
        "key": "amenity",
        "value": "townhall"
    },
```

tyrasd (changes requested):
Hi. Thanks for this suggestion. It looks quite good already, only a few minor changes would be required: As this is a regional preset, the file name should have a suffix indicating the region it is valid for (i.e. `data/…/barangay-PH.json`). A few further comments are below:

data/presets/amenity/townhall/barangay.json
```diff
@@ -0,0 +1,25 @@
...
+        "point",
+        "area"
+    ],
+    "terms": [
+        "barangay hall"
+    ],
+    "tags": {
+        "amenity": "townhall",
+        "townhall:type": "barangay",
+        "admin_level" : "10"
```
tyrasd, line 18: There's a superfluous space here:

```suggestion
        "admin_level": "10"
```

tyrasd:
ah, one last suggestion: we could split up the `tags` section of the preset into two parts, like this: 

```json
    "tags": {
        "amenity": "townhall",
        "townhall:type": "barangay"
    },
    "addTags": {
        "amenity": "townhall",
        "townhall:type": "barangay",
        "admin_level": "10"
    },
```

This would have the benefit that if the `admin_level` tag is missing on an existing OSM object, iD would suggest to add it.

data/presets/amenity/townhall.json
govvin: Got that. Thank you.

data/presets/amenity/townhall/barangay.json
govvin: I did, but I misunderstood the need for this line, so I'll just remove that.

data/presets/amenity/townhall/barangay.json
govvin: That sounds reasonable, and thanks for bringing up the missing wiki page. I should write a stub to get it started. Thanks!

govvin (commented):
Revised the PR to incorporate the changes suggested by reviewers.

data/presets/amenity/townhall/barangay.json
tyrasd: I'm sorry, this was supposed to be just a question. Now, the preset has no icon at all. :see_no_evil:  I'll restore the icon while merging this PR :blush:

tyrasd:
merged as 699e10f7

PR #933, opened 2023-06-24
Add translatable strings to telecom field

kjonosm:


tyrasd (approved):
These terms are going to be tricky to translate, I fear. :sweat_smile:

Eventually, we should also add individual presets (which have the benefit that we can add alternative terms and can hotlink to the definition on the wiki). For example, we could add a preset for _Telecom Cabinet_ using the tags `man_made=street_cabinet` + `utility=telecom` + a field to select between `telecom=service_device` and `telecom=connection_point`.

PR #929, opened 2023-06-21
Create courier.json, add preset for office=courier

govvin:
Add preset for office=courier, to assist contributors in finding the right tag for a [courier delivery service office](https://wiki.openstreetmap.org/wiki/Tag:office%3Dcourier).

Taginfo shows https://taginfo.openstreetmap.org/keys/office#values (search for "courier", "delivery") several similar values that appear to be trying to tag the same thing.

PR #945, opened 2023-07-05
Add courier.json

govvin:
Add preset for `office=courier`, to improve tagging consistency, and assist contributors in finding the tag for a [courier delivery service office](https://wiki.openstreetmap.org/wiki/Tag:office%3Dcourier).

Taginfo shows https://taginfo.openstreetmap.org/keys/office#values (search for "courier", "delivery") several similar values that appear to be trying to tag the same thing.

Note: I previously submitted a [PR](https://github.com/openstreetmap/id-tagging-schema/pull/929), but deleted the branch it was on, and apparently doing so closes the PR, too.

### DIFF
--- a/data/presets/office/courier.json
+++ b/data/presets/office/courier.json
@@ -1,0 +1,32 @@
+{
+    "icon": "fas-people-carry-box",
+    "fields": [
+        "name",
+        "brand",
+        "{office}"
+    ],
+    "moreFields": [
+        "{@templates/contact}",
+        "{payment}",
+        "{@templates/poi}"
+    ],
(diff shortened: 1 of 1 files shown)


matkoniecz:
> I think the correct order of procedure would be to revive the dormant proposal, discuss it in the broader OSM community and eventually bring it to a vote.

@govvin Do you plan to do either?

(I commented for now in https://wiki.openstreetmap.org/wiki/Talk:Tag:office%3Dcourier#distinction_from_post_office - I think this tag may make sense, but current definition is confusing/weird at best)

PR #946, opened 2023-07-05
Presets for common taxi vehicles found in developing regions

govvin:
Added the following preset combos for taxis, by [`taxi_vehicle`](https://wiki.openstreetmap.org/wiki/Key:taxi_vehicle) types.

- Auto Rickshaw - `amenity=taxi` +`taxi_vehicle=auto_rickshaw`
- Cycle Rickshaw - `amenity=taxi` +`taxi_vehicle=cycle_rickshaw`
- Motorcycle Taxi Stand - `amenity=taxi` +`taxi_vehicle=motorcycle`
- Add `taxi_vehicle.json` field
- Update taxi.json, to allow selecting `taxi_vehicle`

This is my first time to do something like this, adding a preset for a combination of tags. Icons could be better.

Is this how it's done?

### DIFF
--- a/data/fields/taxi_vehicle.json
+++ b/data/fields/taxi_vehicle.json
@@ -6,7 +6,7 @@
         "options": {
             "motorcar": "Standard car",
             "motorcycle": "Motorcycle, with passengers 'riding pillon' (i.e. mototáxi)",
-            "auto_rickshaw": "Motorized rickshaws (i.e.  tuk-tuk)",
+            "auto_rickshaw": "Motorized rickshaws (i.e. tuk-tuk)",
             "cycle_rickshaw": "Pedal-powered rickshaws (i.e. padyak)"
         }
     }
--- a/data/fields/taxi_vehicle.json
+++ b/data/fields/taxi_vehicle.json
@@ -1,0 +1,13 @@
+{
+    "key": "taxi_vehicle",
+    "type": "combo",
+    "label": "Type",
+    "strings": {
+        "options": {
+            "motorcar": "Standard car",
+            "motorcycle": "Motorcycle, with passengers "riding pillon" (i.e. mototáxi)",
+            "auto_rickshaw": "Motorized rickshaws (i.e.  tuk-tuk)",
+            "cycle_rickshaw": "Pedal-powered rickshaws (i.e. padyak)"
+        }
+    }
+}
--- a/data/fields/taxi_vehicle.json
+++ b/data/fields/taxi_vehicle.json
@@ -1,13 +1,29 @@
 {
     "key": "taxi_vehicle",
-    "type": "combo",
-    "label": "Type",
+    "type": "semicombo",
+    "label": "Vehicle Type",
     "strings": {
         "options": {
-            "motorcar": "Standard car",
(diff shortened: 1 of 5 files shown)


data/fields/taxi_vehicle.json
```diff
@@ -0,0 +1,13 @@
+{
+    "key": "taxi_vehicle",
+    "type": "combo",
+    "label": "Type",
+    "strings": {
+        "options": {
+            "motorcar": "Standard car",
+            "motorcycle": "Motorcycle, with passengers 'riding pillon' (i.e. mototáxi)",
+            "auto_rickshaw": "Motorized rickshaws (e.g. tuk-tuk, bajaj, etc.)",
+            "cycle_rickshaw": "Pedal-powered rickshaws (e.g. padyak, velotaxi, etc.)"
```
tyrasd, line 10: I would refrain from listing examples in the option strings. Otherwise they just get too long. There is the possibility to add a separate `description` to the values of `combo` fields, which show up as a hover text in iD (for syntax, see this [example](https://github.com/openstreetmap/id-tagging-schema/blob/v6.3.0/data/fields/flashing_lights.json#L9)).

data/presets/amenity/taxi.json
```diff
@@ -4,6 +4,7 @@
         "name",
         "operator",
         "capacity",
+        "taxi_vehicle",
```
tyrasd, line 7: I'd move this to the `moreFields` section: as we would have dedicated presets for the "subtypes" (e.g. _Rickshaw Stand_, …), this field would only be needed to change the "type" of an existing taxi stand (which can also be done by just changing the preset), and to display the (implicit default) value of `motorcar`.

data/presets/amenity/taxi/auto_rickshaw.json
```diff
@@ -0,0 +1,31 @@
...
+        "access_simple",
+        "opening_hours",
+        "operator",
+        "wheelchair"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
```
tyrasd, line 17: add `bajaj` as a search term?

```suggestion
    "terms": [
        "bajaj",
```

tyrasd (commented):
The usage of this tag is currently still low, but it seems to be new and [rapidly](https://taghistory.raifer.tech/?#***/taxi_vehicle/) becoming more popular.

One question which came to my mind is: how common is it to see stands with mixed vehicles (e.g. `auto_rickshaw` and `motorcycle`, etc.)? If they are somewhat common, the field could be changed into a `semiCombo` type to support that as well.

I see that you reduced the number of included fields in the presets for the sub-types. Would it not make sense to keep these consistent with the "main" `taxi` preset?

See a few additional comments/suggestions below.

data/fields/taxi_vehicle.json
govvin: This is a great suggestion. Thanks!

tyrasd:
@govvin is this ready to be reviewed now?

govvin:
@tyrasd , yes, please. :)

> @govvin is this ready to be reviewed now?

tyrasd (approved):

PR #949, opened 2023-07-07
Create amenity=health_post

govvin:
Create a global preset for Community or Neighbourhood [Health Posts](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dhealth_post).


tyrasd:
In principle, this looks good to me. :+1: 

Only one question / suggestion: How "bad" would it be to use a more generic name for these amenities, something like _Community Health Post_? If that would be acceptable, we could use that name also for similar presets in [other countries](https://taginfo.openstreetmap.org/tags/amenity%3Dhealth_post#map) which have this kind of healthcare facility. And at the same time it would minimize the effort needed by our translators.

govvin:
Not hard at all. :) So, is the current name of this proposed entry going to be an alias , or one of its terms?

> In principle, this looks good to me. +1
> 
> Only one question / suggestion: How "bad" would it be to use a more generic name for these amenities, something like _Community Health Post_? If that would be acceptable, we could use that name also for similar presets in [other countries](https://taginfo.openstreetmap.org/tags/amenity%3Dhealth_post#map) which have this kind of healthcare facility. And at the same time it would minimize the effort needed by our translators.

data/presets/amenity/health_post.json
```diff
@@ -0,0 +1,38 @@
...
+    "fields": [
+        "name",
+        "operator",
+        "address",
+        "building_area_yes",
+        "opening_hours"
+    ],
+    "moreFields": [
+        "{@templates/poi}",
+        "healthcare",
```
govvin, line 12: Is it feasible to create a smaller sub-set of the healthcare fields and use that here instead? Any examples I can look at?

data/presets/amenity/health_post.json
```diff
@@ -0,0 +1,42 @@
...
+    "fields": [
+        "name",
+        "operator",
+        "address",
+        "building_area_yes",
+        "opening_hours"
+    ],
+    "moreFields": [
+        "{@templates/poi}",
+        "healthcare/health_post",
```
govvin, line 12: image

Kinda works, but not exactly what I wanted. Other healthcare values still showing up (in black), while those listed in health_post show up as (blue).

matkoniecz:

should it be considered as abandoned? or is it getting close to be worth reviewing again?

govvin:
> > This pull request is still a work in progress
> 
> should it be considered as abandoned? or is it getting close to be worth reviewing again?

Thanks for following-up. I think a review is very welcome.  

The tag usage is still growing, while the proposed equivalent healthcare=health_post seems to have stagnated.

data/presets/amenity/health_post.json
matkoniecz: I am not sure, but it seems that `"autoSuggestions": false` (see say https://github.com/openstreetmap/id-tagging-schema/pull/1412/files code) may be used to disable such automatic suggestions

see https://github.com/ideditor/schema-builder/blob/17e6bacf1b575e088d590db72da9566fe624934c/README.md#autosuggestions

govvin:
No good reason, other than kinda forgetting about it. It's been two year since this PR was drafted.

Thanks @matkoniecz ,  for the comments above, and the possible fix to the issue I mentioned.

I'm a bit busy now, but I'll look into this again in a week or two, and prepare a new PR.

matkoniecz:
It waited literally for years so I feel a bit silly pinging about it - but have you looked into it by any chance?

(I am trying to help untuck PRs to avoid problem of PRs sitting stale for years)

matkoniecz:
It is marked as draft not ready for review.

Please, mark as ready for review once it is ready.

matkoniecz:
Again, sorry for taking so long. Sadly PR and issue backlog is still quite large.

matkoniecz:
@govvin ping? (I would understand if it would be abandoned now, sadly it waited really long - but it would be really nice if it would not be)

govvin:
I thought I'd re-start a new PR #2661 , because I was getting a headache trying to identify the failing checks, but sadly I still failed to identify them.

If anyone could help me identify the error in the new PR, I'll be grateful.

PR #956, opened 2023-07-12
Presets: Two specific tree presets

tordans:
Most trees in Europe tend to fall into one of the two categories
- Broadleaved Tree (deciduous)
- Needleleaved Tree (evergreen)

There are cases, where this is not true, of course, but in my experience the majority fits into those two groups.

Those two specific presets would make it a lot easier to add a tree of those groups and apply all needed tags in one go. Variations are also easier to change, eg. when only `leaf_cycle` needs to change to `semi_evergreen`.

- https://wiki.openstreetmap.org/wiki/Tag:natural%3Dtree#How_to_map
- https://wiki.openstreetmap.org/wiki/Key:leaf_cycle
- https://wiki.openstreetmap.org/wiki/Key:leaf_type

---

My overall goal is, to be able to type "laubbaum" in GoMap and have all tree tags filled out…

---

Open questions:

- [x] How does the search priority work => OK https://github.com/openstreetmap/id-tagging-schema/pull/956#issuecomment-1633559615
- [x] Do the icons look good => OK https://github.com/openstreetmap/id-tagging-schema/pull/956#issuecomment-1633560808

### DIFF
--- a/data/presets/natural/tree/_needleleaved.json
+++ b/data/presets/natural/tree/_needleleaved.json
@@ -1,0 +1,22 @@
+{
+    "icon": "roentgen-needleleaved_tree",
+    "fields": [
+        "leaf_type_singular",
+        "leaf_cycle_singular",
+        "{natural/tree}"
+    ],
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "tags": {
+        "natural": "tree",
+        "leaf_type": "needleleaved"
+    },
+    "terms": [
+        "conifer",
+        "needleleaved",
+        "softwood"
+    ],
+    "name": "Needleleaved Tree"
+}
--- a/data/presets/natural/tree/_needleleaved.json
+++ b/data/presets/natural/tree/_needleleaved.json
@@ -18,5 +18,6 @@
         "needleleaved",
         "softwood"
     ],
-    "name": "Needleleaved Tree"
+    "name": "Needleleaved Tree",
+    "searchable": false
 }
--- a/data/presets/natural/tree/needleleaved/_deciduous.json
+++ b/data/presets/natural/tree/needleleaved/_deciduous.json
@@ -1,0 +1,24 @@
+{
+    "icon": "roentgen-needleleaved_tree",
+    "fields": [
+        "leaf_type_singular",
+        "leaf_cycle_singular",
+        "{natural/tree/needleleaved}"
+    ],
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "tags": {
+        "natural": "tree",
+        "leaf_type": "needleleaved",
+        "leaf_cycle": "deciduous"
+    },
+    "terms": [
+        "conifer",
+        "deciduous",
+        "needleleaved",
+        "softwood"
+    ],
+    "name": "Needleleaved Tree (deciduous)"
+}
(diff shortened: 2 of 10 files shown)


tordans:
Update: I need to change the icon… will look into it later.

tordans:
Update: I change to use `maki-park` for broadleaved trees since that icon looks better then roetgen and it is more consistent.

image Bildschirm­foto 2023-07-12 um 22 29 54

tordans:
New question @tyrasd : How do I get the preset to render the `leave_cycle` and `leaf_type` field for the new presets even though those are part of the `tags` object?

Right now it looks like this ([Link](https://pr-956--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&map=22.00/52.47571/13.44494))… which makes it hard to change the given settings for `leave_cycle` and `leaf_type`.

Bildschirm­foto 2023-07-12 um 22 30 17

tyrasd:
just add the fields again to the individual presets: they are only omitted in the inherited/referenced fields from other presets (and also present as `tags` of the preset), but kept if directly included in the `fields` list.

tordans:
> * [ ]  How does the search priority work

I think that is fine as is…

tordans:
> - [x] Do the icons look good

I think this looks good now…

image

tordans:
@tyrasd thanks. This is ready for review, now :)

PS: I created two PRs in the schema repo to improve the docs. https://github.com/ideditor/schema-builder/pulls/tordans

tyrasd:
I've added (for now unsearchable) presets for the not-so-common-but-still-possible combinations (needleleaved+deciduous & broadleaved+evergreen) and the leaf-type-only versions. This should make the mapping more seamless with existing data and will show the correct icon for all leaf_type-mapped trees. Also, I've changed the icon for the generic Tree preset, such that it is visually different in the presets list and on the map.

What do you think?

data/presets/natural/tree/_needleleaved.json
```diff
@@ -0,0 +1,22 @@
...
+        "natural": "tree",
+        "leaf_type": "needleleaved"
+    },
+    "terms": [
+        "conifer",
+        "needleleaved",
+        "softwood"
+    ],
+    "name": "Needleleaved Tree"
+}
```
tordans, line 22:
```suggestion
    "name": "Needleleaved Tree",
    "searchable": false
}
```

data/presets/natural/tree/needleleaved/_deciduous.json
```diff
@@ -0,0 +1,24 @@
...
+        "leaf_type": "needleleaved",
+        "leaf_cycle": "deciduous"
+    },
+    "terms": [
+        "conifer",
+        "deciduous",
+        "needleleaved",
+        "softwood"
+    ],
+    "name": "Needleleaved Tree (deciduous)"
```
tordans, line 23:
```suggestion
    "name": "Needleleaved Tree (deciduous)",
    "searchable": false
```

tordans (commented):
Thanks for the update @tyrasd.

I was holding back on adding more detailed presents to not overwhelm the translators. But given how many trees we have in OSM, adding those presents makes it a lot clearer, so that is good!

I still do not like the roentgen-tree icon inside the map in. It looks like an ice cream cone to me. However, given having the general preset different from the more precise presets is good idea. Ideally we would have a "left leaves, right needle" icon in temaki.

I saw that a few of the unsearchable presets did not have the unsearchable flat. I will add this after submitting this comment.

    "searchable": false

PR #958, opened 2023-07-13
Update birthing_centre.json, add doctors-PH.json

govvin:
Added additional terms for healthcare=birthing_centre preset. Fix typo.

Add regional variant for doctors-PH.json, including aliases.

### DIFF
--- a/data/presets/amenity/doctors-PH.json
+++ b/data/presets/amenity/doctors-PH.json
@@ -15,7 +15,7 @@
     ],
     "moreFields": [
         "{@templates/poi}",
-        "operator:type"
+        "operator/type"
     ],
     "geometry": [
         "point",
--- a/data/presets/amenity/doctors-PH.json
+++ b/data/presets/amenity/doctors-PH.json
@@ -1,0 +1,43 @@
+{
+    "icon": "maki-doctor",
+    "locationSet": {
+        "include": [
+            "ph"
+        ]
+    },
+    "fields": [
+        "name",
+        "operator",
+        "healthcare/speciality",
+        "address",
+        "building_area_yes",
+        "opening_hours"
+    ],
+    "moreFields": [
+        "{@templates/poi}",
+        "operator:type"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "rhu"
+    ],
+    "tags": {
+        "amenity": "doctors"
+    },
+    "addTags": {
+        "amenity": "doctors",
+        "healthcare": "doctor"
+    },
+    "reference": {
+        "key": "amenity",
+        "value": "doctors"
+    },
+    "name": "Doctor",
+    "aliases": [
+        "Rural Health Unit",
+        "Urban Health Center"
+    ]
+}
--- a/data/presets/healthcare/birthing_center.json
+++ b/data/presets/healthcare/birthing_center.json
@@ -22,6 +22,7 @@
         "healthcare": "birthing_centre"
     },
     "addTags": {
+        "healthcare": "birthing_centre",
         "healthcare:speciality": "maternity"
     },
     "name": "Birthing Center"
--- a/data/presets/healthcare/birthing_center.json
+++ b/data/presets/healthcare/birthing_center.json
@@ -21,5 +21,8 @@
     "tags": {
         "healthcare": "birthing_centre"
     },
+    "addTags": {
+        "healthcare:speciality": "maternity"
+    },
     "name": "Birthing Center"
 }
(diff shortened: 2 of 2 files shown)


data/presets/healthcare/birthing_center.json
```diff
@@ -6,16 +6,23 @@
...
+        "midwives",
         "pregnancy",
         "pregnant"
     ],
     "tags": {
         "healthcare": "birthing_centre"
     },
+    "addTags": {
+        "healthcare:speciality": "maternity"
+    },
```
tyrasd, line 26:
```suggestion
    "addTags": {
        "healthcare": "birthing_centre",
        "healthcare:speciality": "maternity"
    },
```

tyrasd (changes requested):
:+1: for the changes to the Birthing Center preset

But regarding the Doctor preset: Am I seeing it correctly that the only change is the country-specific `aliases`? I would have said that such aliases could be easier managed through a translation on transifex: I've just opened translation for `en_PH` _English (Philippines)_ for this: https://app.transifex.com/openstreetmap/id-editor/translate/#en_PH/presets/68222339?q=key%3Adoctors. Does that make sense here?

govvin:
Thanks for reviewing this @tyrasd .

As for the doctors-PH.json, I'm not really sure if translation if the way to go. 

The "RHU"/Rural Health Unit is a local type of field health facility operated by a government, under the supervision of a government physician, which (locally) may be then tagged as amenity=doctor/healthcare=doctor,  and similar to a doctor's office.

I wonder if it's appropriate to create a different preset to display it with a different name, rather than using translate.

data/presets/amenity/doctors-PH.json
```diff
@@ -0,0 +1,43 @@
...
+        "amenity": "doctors",
+        "healthcare": "doctor"
+    },
+    "reference": {
+        "key": "amenity",
+        "value": "doctors"
+    },
+    "name": "Doctor",
+    "aliases": [
+        "Rural Health Unit",
```
matkoniecz, line 40: so these are PH-specific names for `amenity=doctors` - without any special tagging?

govvin: It's a common health service provided by the government in rural areas, often GPs.  I'd say it's alright to add the `healthcare:speciality=general;community`

matkoniecz: If "Rural Health Unit" makes sense as dedicated entry with own tags then they should be own preset entries.

I am still unsure why dedicated PH entry is created here - is it to add this extra terms?

data/presets/healthcare/birthing_center.json
```diff
@@ -6,22 +6,24 @@
...
-    ],
-    "moreFields": [
-        "{healthcare}"
-    ],
     "tags": {
         "healthcare": "birthing_centre"
     },
+    "addTags": {
+        "healthcare": "birthing_centre",
+        "healthcare:speciality": "maternity"
```
matkoniecz, line 26: if `"healthcare:speciality": "maternity"` should be mass added to all `healthcare=birthing_centre` (seems pointless to me) then it should be done as an approved bot edit, see https://wiki.openstreetmap.org/wiki/Automated_Edits_code_of_conduct

data/presets/healthcare/birthing_center.json
```diff
@@ -6,22 +6,24 @@
...
+        "birthing home",
         "childbirth",
         "delivery",
-        "labor",
         "labour",
-        "midwive",
+        "labor",
+        "lying-in",
+        "maternity",
+        "midwife",
```
matkoniecz, line 16: are there other objects that should appear when one search with midwife/maternity term?

data/presets/healthcare/birthing_center.json
```diff
@@ -6,22 +6,24 @@
...
     "terms": [
         "baby",
+        "birthing home",
         "childbirth",
         "delivery",
-        "labor",
         "labour",
-        "midwive",
+        "labor",
+        "lying-in",
```
matkoniecz, line 14: https://en.wikipedia.org/wiki/Lying-in - I guess that it may be reasonable search term?

data/presets/healthcare/birthing_center.json
matkoniecz: @govvin ping?

data/presets/amenity/doctors-PH.json
matkoniecz: @govvin

data/presets/healthcare/birthing_center.json
govvin: That’s a good point! I don’t think it’s necessary, and can just leave the primary tag there. 

I'll drop it from the PR.

PR #962, opened 2023-07-19
Add aliases and terms to shop=outpost

westnordost:
See https://community.openstreetmap.org/t/how-to-tag-online-groceries-gorillas-getir-flink/101391 :

See [Dark Store on Wikipedia](https://en.wikipedia.org/wiki/Dark_store) and [Online Grocer on Wikipedia](https://en.wikipedia.org/wiki/Online_grocer).

### DIFF
--- a/data/presets/shop/outpost.json
+++ b/data/presets/shop/outpost.json
@@ -5,6 +5,7 @@
         "area"
     ],
     "terms": [
+        "delivery",
         "online",
         "pick up",
         "pickup"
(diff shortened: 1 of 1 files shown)


matkoniecz:
I would be quite against it, marking places where only employees can go in and use it to pickup deliveries as `shop=` seems a really bad idea to me.

westnordost:
Some kind of consensus seemed to have been reached to have a new tag for dark stores instead of using `shop=outpost` for that, see https://wiki.openstreetmap.org/wiki/Key:dark_store

So this PR is outdated.

Issue #318, opened 2021-12-26
Add preset for `memorial=stolperstein`


tordans:
We have a lot of those special memorial stones in Germany and both the Wiki(data)/Data Item and OSM Community are recording them. It think a preset would be a nice recognition of this work. 

However, the tagging seems be in a process of change, so we need to find a good preset tagging here.

## **Here is my research:**

- Wiki https://wiki.openstreetmap.org/wiki/DE:Stolpersteine + Discussion https://wiki.openstreetmap.org/wiki/DE_talk:Stolpersteine
- Should only apply for Germany since this is where most of those stones are (right?)
- The recommended tagging changed in 2018 which was discussed on the forum [and the wiki-talk-page](https://wiki.openstreetmap.org/wiki/DE_talk:Stolpersteine#Tagging_scheme_change) and the wiki page shows the result of this change. It was not a formal voting proposal process, though
- The existing tagging was never updated, which means the old tagging still has a lot more use
  - `memorial=stolperstein` 4.499 https://taginfo.openstreetmap.org/tags/memorial=stolperstein
  - `memorial:type=stolperstein` 24.784 https://taginfo.openstreetmap.org/tags/memorial%3Atype=stolperstein
  - I assume the other "unusal" tags will show similar numbers.
- There are two JOSM presets which also still follow the old tagging guideline
  - https://josm.openstreetmap.de/wiki/Presets/Stolpersteine
  - https://josm.openstreetmap.de/wiki/Presets/StolpersteineLight
- There is an active community adding this data as Wikidata Data Item, eg. https://www.wikidata.org/wiki/Wikidata:WikiProject_Stolpersteine [with a map of this data](https://query.wikidata.org/embed.html#%23defaultView%3AMap%0ASELECT%20%3FStolperstein%20%3FStolpersteinLabel%20%3Fgeographische_Koordinaten%20WHERE%20%7B%0A%20%20SERVICE%20wikibase%3Alabel%20%7B%20bd%3AserviceParam%20wikibase%3Alanguage%20%22%5BAUTO_LANGUAGE%5D%2Cen%22.%20%7D%0A%20%20%3FStolperstein%20wdt%3AP31%20wd%3AQ26703203.%0A%20%20OPTIONAL%20%7B%20%3FStolperstein%20wdt%3AP625%20%3Fgeographische_Koordinaten.%20%7D%0A%7D) => I think we should recommend Wikidata as a preset field as well

## Preset

**What tags to add to the preset:**

[...]

PR #964, opened 2023-07-21
Add preset `stolperstein` with deprecations and `memorial:addr` field

tordans:
Recently the wiki page https://wiki.openstreetmap.org/wiki/DE:Stolpersteine was updated which resolved a few of the blockers I saw in #318.

### Preset

- The fields are split of from the regular memorial preset in order to now show the `material` on the Stolperstein Preset which always brass AFAIK and none of the wiki pages does recommend it
- The name "Memorial Plaque Stolperstein" is meant to be descriptive for other language users. I think the German translations should just be "Stolperstein"
- I did not add a ref pr memorial:ref field because again, the wiki/community is not clear on what to use but mainly there is no current usage of either [in Taginfo](https://taginfo.openstreetmap.org/tags/memorial=stolperstein#combinations)
- I did not add those [`person:date_of_birth`](https://taginfo.openstreetmap.org/keys/person%3Adate_of_birth) fields because I did not want to research how to add custom date fields but also I think those special fields can be added with the "Tags" area in iD when users want to…
- Info-i: Nothing to do here; the wiki and wikidata is present https://wiki.openstreetmap.org/wiki/Tag:memorial%3Dstolperstein
- 

### `memorial:addr` field

- TODO: Lets see if this works in the branch preview…
- The wiki and talk page are not clear on which tag to use, so I added both "address" and "memorial:addr". At least for the new tagging of `memorial=stolperstein` all users agreed to use "memorial:addr" ([Taginfo](https://taginfo.openstreetmap.org/tags/memorial=stolperstein#combinations))
- The wiki is not clear on which address, so I took a risk here and clarified as "Nearest building address" which is the only think that makes sense IMO because we map what we see in OSM. Research tasks like "where did this person live" can be recorded on the Wikidata item. It is also what makes sense for the [completeness-verification process](https://wiki.openstreetmap.org/wiki/DE:Stolpersteine#Soll-Ist-Abgleich-Tool) that is described on the wiki.
- I only added the subkeys that are most often used https://taginfo.openstreetmap.org/search?q=memorial%3Aaddr#keys but also described as importent on the wiki
- Info-i: There is no wiki page ([Search](https://wiki.openstreetmap.org/w/index.php?search=memorial%3Aaddr&title=Special%3ASearch&go=Seite)) but there is an empty Wiki data item https://wiki.openstreetmap.org/wiki/Item:Q1507

### deprecations

- The wiki is now pretty clear about the update path for `memorial:type` ([usage 25k](https://taginfo.openstreetmap.org/keys/memorial%3Atype#overview) with a bit of double tagging)
- And it makes sense to also update the core info from `memorial:text` at the same time ([usage 21k](https://taginfo.openstreetmap.org/search?q=memorial%3Atext#keys))

### Regional preset?

In https://github.com/openstreetmap/id-tagging-schema/issues/318#issuecomment-1015287842 we talked about making this a regional preset or not. The wiki page on which countries/cities have Stolpersteine (https://de.wikipedia.org/wiki/Liste_der_Orte_mit_Stolpersteinen) is quite long.

Given this long list, I think we should either make it a regional preset for just the top 3 countries … or keep it as a general preset.

### TODOs

- [x] Check address-field ==> Works great
- [ ] Later: Add info on address-wikidata item
- [ ] Resolve regional preset
- [ ] Where does the empty "wikipedia" field come from? / how to remove it? It is not part of the preset…
     image

---

Closes https://github.com/openstreetmap/id-tagging-schema/issues/318


tordans:
# Testing

- https://pr-964--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=n6879076451&locale=en&map=22.00/52.47629/13.44686 is a good place to test the preset
- https://pr-964--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=n10982639457&map=20.00/52.46499/13.33045 can be used to test the deprecations (https://overpass-turbo.eu/s/1xN9)

tordans:
@tyrasd ready to review. Please have a look at the todo list above for your input. 🙏

tordans:
Thanks for your feedback & help @tyrasd. This is ready for review again.

> > Regional Preset
> 
> Given that there are (logically) very few [exceptions](https://de.wikipedia.org/wiki/Liste_der_Orte_mit_Stolpersteinen#Verlegung_auf_anderen_Kontinenten) outside of Europe, maybe limiting the scope to Europe [(code Q46)](https://ideditor.codes/?code=Q46) would be reasonable. 

Good idea! I added it like that in https://github.com/tordans/id-tagging-schema/commit/8e90f9954fa716e6df3bece0af461e1bd4731462

> > At least for the new tagging of memorial=stolperstein all users agreed to use `memorial:addr`
> 
> But the wiki page recommends to put the full address as a string into `memorial:addr=…` and that also seems to be [more common](https://taginfo.openstreetmap.org/search?q=memorial%3Aaddr#keys) than using the address subtags. 🤷 🤔

OMG this tagging is so unfortunate…
I tried adding both variants and make it that the UI picks the right one based on existing tags. However, the `type:address`  does not allow other tags that `addr:*` (it does not automatically use a different prefix), so that was a dead end.

Now, there is just an optional (moreFields) address string field…

> > Where does the empty "wikipedia" field come from? / how to remove it? It is not part of the preset…
> 
> This is the behaviour of the `wikidata` field type: It automatically tries to sync a matching wikipedia page for wikidata entries, adding the field for the wikipedia entry. This does of course not make all too much sense for wikidata entries which don't have a corresponding wikipedia entry. This must be fixed on iD's side, however.

I think that is fine for now…

tyrasd (approved):
Looks good to me.

I gave the wikipedia/wikidata field a second look and think I have a better solution for the case where there is a wikidata identifier, but no wikipedia page: see https://github.com/openstreetmap/id-tagging-schema/commit/14bd2c7b9a68bab5829df8bac1a04939325eee3e

Also, for the `memorial:addr` vs. `memorial:addr:*` tag(s): not sure what you tried, but one can use different address prefixes in the field, see #971 / 0fe03d7. Would you like to integrate this approach here before we merge?

tordans:
@tyrasd thanks for the help with the `*addr*` part. I changed it to use the street instead of housenumber since that is a good fit as well and a bit more broad IMO.

I looked at the data https://overpass-turbo.eu/s/1z4c and Hamburg uses this schema a lot, so it is worth adding…
image

## Testcase

https://pr-964--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=n2652126042&locale=en&map=19.00/53.57327/10.03401

1. First update the tags
2. Now the address-fields show up (because now the Stolpersteine preset is active)

---

@tyrasd this is ready to be merged IMO

PR #968, opened 2023-07-29
preset `traffic_sign=maxspeed`: Use direction icon same as other traffic_sign presets

tordans:
The preset for `traffic_sign=maxspeed` was the only one in the traffic_sign preset list that used the maki square icon, which is the default "no icon" icon (at least in GoMap).

This PR changes the icon to use `fas-directions` which is what the other generic icons use.

FYI, the square Icon is on the maxspeed preset [since the start of this repo](https://github.com/openstreetmap/id-tagging-schema/commit/bf95421f65018ccd7f398f8459b37462febddcba#diff-38fbf34d089c0103d54684ce1b39efe8fb2ab623361fd73283c787748e231efeR2).

**Previous:**

Bildschirm­foto 2023-07-29 um 13 15 04

### DIFF
--- a/data/presets/traffic_sign/maxspeed.json
+++ b/data/presets/traffic_sign/maxspeed.json
@@ -1,5 +1,5 @@
 {
-    "icon": "maki-square-stroked",
+    "icon": "fas-directions",
     "fields": [
         "{traffic_sign}",
         "maxspeed"


tyrasd (approved):

tyrasd:
yeah, the icon was like that from the very first version of the preset in iD (see [openstreetmap/iD@`cd8be8b`](https://github.com/openstreetmap/iD/commit/cd8be8b5b449fa77a55e2fa4f314f19fb1810144#diff-87bb531f79ad25b44278adfc62f0acf7e3fe058aa53c99eb42d3298b8b231aee)). I assume this was chosen because it somewhat resembles the shape of speed limit signs [found in](https://en.wikipedia.org/wiki/Comparison_of_MUTCD-influenced_traffic_signs#Regulatory) the US, Canada and Liberia. I guess we could have a regional copy of the preset with the square icon for these countries… :thinking:

PR #971, opened 2023-08-03
[demo] add secondary memorial:addr:* field for when the subtag schema is used

tyrasd:
for #964 (this PR is not meant to be merged, it's just to get the demo/preview of the below mentioned functionality)

> I tried adding both variants and make it that the UI picks the right one based on existing tags. However, the type:address does not allow other tags that addr:* (it does not automatically use a different prefix), so that was a dead end.

Something like this should do the trick. It's not 100% waterproof, as it relies on the `memorial:addr:housenumber` tag to determine which field to display (there might be the case where only be a `memorial:addr:housename` exists), but except for that it seems to work fine.

### DIFF
--- a/data/fields/memorial/addr.json
+++ b/data/fields/memorial/addr.json
@@ -1,5 +1,8 @@
 {
     "type": "text",
     "key": "memorial:addr",
-    "label": "Nearest Building Address"
+    "label": "Nearest Building Address",
+    "prerequisiteTag": {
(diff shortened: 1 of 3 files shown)


tordans:
Thanks a lot @tyrasd I merged this into https://github.com/openstreetmap/id-tagging-schema/pull/964. This PR can be closed IMO.

PR #973, opened 2023-08-08
Add translatable strings to "marker" field

kjonosm:


tyrasd:
I think at least some of the value labels are not really intuitive… for example _Aerial_ was not quite clear to me without reading the [wiki page](https://wiki.openstreetmap.org/wiki/Key:marker). It could be a good idea to use the `title`/`description` syntax here, to add some additional information to these values:

```json
        …
        "options": {
            "aerial": {
                "title": "Aerial",
                "description": "Designed to be seen by aerial inspection"
            },
            …
```

What do you think?

kjonosm:
@tyrasd
I added short descriptions based on OSM wiki.
Feel free to adjust the wording...

data/fields/marker.json
```diff
@@ -1,5 +1,33 @@
 {
     "key": "marker",
     "type": "typeCombo",
-    "label": "Type"
+    "label": "Type",
+    "strings": {
+        "options": {
+            "aerial": {
+                "title": "Aerial",
+                "description": "Designed to be seen by aerial inspection."
```
1ec5, line 9: This is a simpler way to say the same thing:

```suggestion
                "description": "Designed to be visible from the air."
```

data/fields/marker.json
```diff
@@ -1,5 +1,33 @@
...
+            },
+            "ground": {
+                "title": "Ground",
+                "description": "Flat marker fixed directly onto the ground."
```
1ec5, line 13: “Fixed” is a potentially confusing word, and it’s assumed for a marker anyways. Maybe something like this?

```suggestion
                "description": "Flat marker laid in the ground."
```

Or maybe:

```suggestion
                "description": "Laid flat on the ground in a fixed location."
```

1ec5:
I guess it’s fine to omit [`marker=paddle`](https://wiki.openstreetmap.org/wiki/Tag:marker%3Dpaddle). As far as we know so far, it’s only being used in one country, although in reality it’s being used in very large numbers.

tyrasd:
merged as c0bfb2812c1d0ae94cc05fa0567002c787296551

Issue #975, opened 2023-08-10
Add preset for amenity=baking_oven


AntMadeira:
### OSM Tag(s)

amenity=baking_oven

### How would you like this tag to see supported?

As a Preset

### Label

Baking Oven

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbaking_oven

### Status of the Tag

Approved

### Usage of the tag

752

### Further Information

Hi there.
I would like to request the addition of this preset to iD pluss the oven=* field.

Regards.

N-45div:
@tyrasd sir can you please check the pull request and tell me any changes required

N-45div:
@tyrasd sir i have made the necessary changes please check it...Thank You

qugebert:
I think this can be closed as completed in #983

PR #978, opened 2023-08-13
#975 Added in amenity Group Baking_oven field

N-45div:
Please tell any changes required


data/presets/baking_oven/bakehouse.json
```diff
@@ -0,0 +1,24 @@
+{
+    "icon": "Bakehouse",
```
tyrasd, line 2: Hi. We need a valid icon for the preset. See https://github.com/ideditor/schema-builder/blob/main/ICONS.md for more information about which icon sets are available.

As a last resort, we could use `fas-bread` (https://fontawesome.com/v6/icons/bread-slice?f=classic&s=solid) maybe?

data/presets/baking_oven/bakehouse.json
```diff
@@ -0,0 +1,24 @@
...
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "amenity": "baking_oven",
+        "cuisine": "Bakehouse"
```
tyrasd, line 9: where did you find the tag `cuisine=Bakehouse`? As far as I can see, this tag is not used in OSM.

data/presets/baking_oven/bakehouse.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    "reference": {
+        "key": "cuisine"
```
tyrasd, line 12: reference should be the main tag of the preset, here: `amenity=baking_oven`
```json
        "key": "amenity",
        "key": "baking_oven"
```

this can be completely omitted if the preset only has one tag in `tags` (as I would assume will be the case here).

data/presets/baking_oven/bakehouse.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    "terms": [
+        "breads",
+        "croissants",
+        "muffins",
+        "tea Cakes",
+        "bagels",
+        "nibbles",
+        "cookies"
```
tyrasd, line 21: are you sure about the _pastry_ type products as search terms for this preset? I think that typically, such baking ovens are used to only bake breads and bread-like products.

Suggestion:

```suggestion
        "beehive oven",
        "bread",
        "clay oven",
        "horno",
        "masonry oven",
        "tabun oven",
        "pizza oven"
```

data/presets/baking_oven/bakehouse.json
```diff
@@ -0,0 +1,24 @@
...
+    ],
+    "name": "Bakehouse"
```
tyrasd, line 23: This name would clash with the tag `building=bakehouse` (https://wiki.openstreetmap.org/wiki/Tag:building%3Dbakehouse). Maybe one of the following would be better:

```suggestion
    "name": "Baking Oven"
```

```suggestion
    "name": "Communal Oven"
```

tyrasd (changes requested):
Thank you for this! :+1:

Please find some suggestions for changes below. :relaxed:

N-45div:
Thank you So much Sir for your suggestions..will inculcate those suggestions and revert to you back....

data/presets/baking_oven/bakehouse.json
N-45div: I have used now Temaki-bread for the icon

data/presets/baking_oven/bakehouse.json
N-45div: Thank you so much for the suggestion implemented the same....

data/presets/baking_oven/bakehouse.json
N-45div: Thank you so much for the suggestion implemented the same....

data/presets/baking_oven/bakehouse.json
N-45div: Sir just kept the key = amenity as suggested

data/presets/baking_oven/bakehouse.json
N-45div: yes sorry didnt noticed it did the necessary changes

N-45div:
@tyrasd sir did the changes please do review it once...Thank you

data/presets/amenity/baking_oven/bakehouse.json
```diff
@@ -0,0 +1,24 @@
+{
+    "icon": "temaki-bread",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "amenity": "baking_oven",
+        "building": "Bakehouse"
```
kjonosm, line 9:
```suggestion
        "building": "bakehouse"
```

N-45div: Thank You so much for reviewing...I have committed your suggestion...Hope it gets merged...

PR #981, opened 2023-08-15
ATM: Add charge

mvl22:
Many ATMs increasingly have a charge for withdrawals, though others are free. Amounts vary widely. This is useful information if trying to avoid racking up fees.

### DIFF
--- a/data/presets/amenity/atm.json
+++ b/data/presets/amenity/atm.json
@@ -3,6 +3,7 @@
     "fields": [
         "operator",
         "network",
+        "charge",
         "cash_in",
         "currency_multi",
         "drive_through"


kjonosm:
[`charge=*`](https://wiki.openstreetmap.org/wiki/Key:charge) is used to specify the exact amount to withdraw cash.
I guess [`fee=*`](https://wiki.openstreetmap.org/wiki/Key:fee) is what you are actually looking for?

Dimitar5555:
ATMs don't charge for withdrawing money. Banks do. The charge depends on your plan (student/standard/pensioner), card issuer (visa/mastercard), card type (credit/debit/prepaid), the ATM (in-network or out-network) and your bank's terms and condition. How would you fit all of this information in that field? `fee=yes;no`?

mvl22:
> ATMs don't charge for withdrawing money.

That doesn't seem to be my experience, and it's not what is defined on the OSM wiki. Pretty sure an ATM I used today stated that the machine would charge a fee of 1.80 EUR for instance. It was listed as a machine fee, not something from the bank. The bank may also charge a VISA fee for instance.

https://wiki.openstreetmap.org/wiki/Tag:amenity=atm#Tagging

Dimitar5555:
> > ATMs don't charge for withdrawing money.
> 
> Are you sure? That doesn't seem to be my experience, and it's not what is defined on the OSM wiki. Pretty sure an ATM I used today stated that the machine would charge a fee of 1.80 EUR for instance. It was listed as a machine fee, not something from the bank. The bank may also charge a VISA fee for instance.
> 
> https://wiki.openstreetmap.org/wiki/Tag:amenity=atm#Tagging
> 
> `charge=* | Specific of the ATM. | 1.75 GBP (fixed fee)`

I assume that you've used Euronet's ATM. Euronet is not a bank, so their ATMs do charge. They are also notorious for having high transaction and exchange fees.

I was referring to ATMs of actual banks. Each bank will charge a different amount based on the conditions above. So for bank A's ATM you would have something along the lines of `charge=0 EUR @ customers of Bank A and currency=local;0.5 EUR + 0.25% @ customers of Bank A and currency=USD;1 EUR @ customers of Bank B;max(5 EUR, 0.75%) @ customers of Bank B and card=Visa Platinum|Mastercard Platinum and currency!=local;5 EUR +1.25% @ customers of non-EEA bank`. That's not even counting the existence of Bank C, Bank D and so on. For simplicity I ignored most card types and most premium and social packages but you can see where this is going.

mvl22:
Yes, Euronet are 7 EUR here. That presumably would be suitable for `charge=7 EUR`?

No, the example I gave yesterday was a cash machine on the outside wall of a Banco Sabadell branch, which is a bank. The receipt clearly states `ATM fee: 1,80 EUR`. Are you saying that 'ATM fee' is actually a banking fee and is customer-dependent?

tyrasd (changes requested):
As mentioned, the tag/field for this would be `fee`. And as also mentioned, the case where the fee _depends_ is probably the most common, so the simple `yes`/`no` checkbox field would not be sufficient here. I'd say, we can add a new field for atm fees (`fields/fee_atm.json`) with the values `yes`/`some`/`no` (as mentioned on the [wiki page](https://wiki.openstreetmap.org/w/index.php?title=Tag:amenity%3Datm&oldid=2584157#Tagging)). Does that make sense?

//Edit: and perhaps include this field only as an optional one (in `moreFields`), as the tag seems to be not [very much used](https://taginfo.openstreetmap.org/tags/amenity=atm#combinations) at the moment.

mvl22:
The wiki suggests that fee is a simpler boolean type field and that charge is the actual amount. (Seems a bit like duplication to me.) Is my reading incorrect?

See also the talk page which implies fee is not an amount field.

https://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Datm#fee=*_for_the_actual_amount_of_money_requested_to_operate_the_atm

PR #983, opened 2023-08-16
#975 Added baking_oven in preset

N-45div:
Please review it once actually it was done earlier I deleted my earlier repo due to which earlier pull request also got deleted...


N-45div:
@tyrasd sir thank you so much for reviewing and making the necessary changes...any more changes I have to do from my side please tell...Once again thank you it's very exciting contributing to the repo it was my first time but yes will involve more in it...

danieldegroot2:
Fixes #975.

data/presets/amenity/baking_oven/bakehouse.json
danieldegroot2, line 1: > data/presets/amenity/baking_oven/bakehouse.json

Rename file to `baking_oven.json`.

Adding `/bakehouse` is too specific; It defines what would be an additional preset for `amenity=baking_oven`, with some pre-filled fields (properties). However, there should probably be a base preset first, before this can be considered. Noting previous changes at https://github.com/openstreetmap/id-tagging-schema/pull/983/commits/fc9649aa4ee7cd647bd9a4393628952e902f3a0a , only the file name still needs to be simplified.

You can change the file name while in edit mode. Put the cursor between `bakehouse` and `.json`. Remove "bakehouse" using backspace. Then, by pressing backspace once more, the additional `/` will be removed, resulting in `baking_oven.json` as the final name.

Fyi You can see an example of additional presets at [`https://github.com/openstreetmap/id-tagging-schema/tree/main/data/presets/amenity/parking . Note they are in their own sub-folder.

danieldegroot2, line 1: Try adding `fields` and `moreFields`, like in https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/bbq.json

The fields mentioned there can be kept. Either [`fuel`](https://wiki.openstreetmap.org/wiki/Key:fuel) or [`oven`](https://wiki.openstreetmap.org/wiki/Key:oven) is fine (do not add both), it can be changed later if needed. Other fields could be considered later.
(sidenote: seems `oven` was overlooked in the proposal process. It could be proposed for it to be changed to `fuel`.)

danieldegroot2 (changes requested):

PR #984, opened 2023-08-17
Add "unspecified traffic sign"

tordans:
Sometimes, we map `highway=traffic_sign` without specifying the `traffic_sign=*` (which is usually the primary tag for historic reasons).

Those traffic signs are missing a preset, which I added as unsearchable.

I created a separate file for that because I think there is no way to give a preset two different, unrelated `tags` to match on.

- Wiki https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtraffic_sign
- Usage 1k+ https://taginfo.openstreetmap.org/tags/highway=traffic_sign#map, all over the world

### DIFF
--- a/data/presets/highway/_traffic_sign.json
+++ b/data/presets/highway/_traffic_sign.json
@@ -12,5 +12,5 @@
     },
     "terms": [],
     "searchable": false,
-    "name": "Unspecified Traffic Sign"
+    "name": "{traffic_sign}"
 }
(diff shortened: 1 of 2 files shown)


tordans:
# Testing

- https://pr-984--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=n10311996320&locale=en&map=19.00/52.46748/13.47302
- https://www.openstreetmap.org/node/10311996320

PR #987, opened 2023-08-30
Add fhrs:id

Cj-Malone:
We use [fhrs:id](https://wiki.openstreetmap.org/wiki/UK_Food_Hygiene_Rating_Scheme) a [fair bit](https://taginfo.openstreetmap.org/keys/fhrs%3Aid) in GB, to match to official data of "food establishments". There definition and OSMs don't line up perfectly, but I've copied a [reasonable list](https://github.com/gregrs-uk/fhodot/blob/5a1faabe5ab719c755bc9bda4403864f832b4fa9/import/osm/imposm_mapping.yml#L17) from a QA tool we use.

FSA seem to allow anything in, hence why some clothes shops are listed. FHODOT falls back and selects everything that already has `fhrs:id=*`, so I'm curious how iD will react to a poi with `shop=clothes` + `fhrs:id=123`, not that the edge case really matters.

### DIFF
--- a/data/fields/fhrs/id-GB.json
+++ b/data/fields/fhrs/id-GB.json
@@ -1,0 +1,17 @@
+{
+    "key": "fhrs:id",
+    "type": "identifier",
+    "label": "FHRS ID",
+    "urlFormat": "https://ratings.food.gov.uk/business/en-GB/{value}",
+    "pattern": "^[0-9]{1,}$",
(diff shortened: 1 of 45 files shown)


tyrasd (approved):

tyrasd (approved):

Cj-Malone:
Thank you

PR #988, opened 2023-09-01
Add new preset for fuel pump and rename current one

### LINKED ISSUE #411: "Gas Pump" preset clarifications
The Gas Pump preset automatically includes amenity=vending_machine+vending=fuel

Is this supposed to be micro-mapped for an individual gas pump inside a "Gas station" (amenity=fuel), or for a standalone automated vending machine(s) point of sales ?

https://wiki.openstreetmap.org/wiki/Tag:vending%3Dfuel
has this confusing statement: "To avoid tag conflicts, a fuel station sells fuel using a fuel vending machine, it's good to mark it as [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=[fuel](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dfuel) + [vending_machine](https://wiki.openstreetmap.org/wiki/Key:vending_machine)=yes."

Dimitar5555:
Resolves #411

### DIFF
--- a/data/presets/amenity/vending_machine/fuel.json
+++ b/data/presets/amenity/vending_machine/fuel.json
@@ -4,6 +4,12 @@
         "point",
         "vertex"
     ],
+    "fields": [
+        "ref",
+        "fuel/fuel_multi",
+        "fuel/fuel_multi-RU",
+        "operator"
+    ],
     "terms": [
         "petrol",
         "fuel",
@@ -22,6 +28,6 @@
         "key": "vending",
         "value": "fuel"
     },
-    "name": "Fuel Vending Pump",
+    "name": "Gas Vending Pump",
     "matchScore": 0.5
 }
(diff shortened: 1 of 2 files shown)


data/presets/amenity/vending_machine/fuel.json
```diff
@@ -4,6 +4,12 @@
         "point",
         "vertex"
```
Dimitar5555, line 5: Do we need to have them as vertex? I haven't seen fuel pumps that are attached to walls (or even fuel stations that have walls on the outside part).

Dimitar5555: Looks like there are (https://community.openstreetmap.org/t/differing-approaches-to-tagging-vending-machines-for-motor-fuel-dispensing-have-caused-confusion/103421/1)

PR #990, opened 2023-09-01
Correct icon names for fast food presets

danieldegroot2:
Correct icon name to be the same as in [`fast_food.json`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/fast_food.json), which has a proper rendering.


tyrasd:
this was also fixed in #1057, sorry for having overlooked this one

PR #992, opened 2023-09-03
Update speciality.json

kjonosm:
Add some more specialties and make obstetrics a dedicated specialty besides gynaecology


matkoniecz:
will close and reopen to recreate preview

data/fields/healthcare/speciality.json
```diff
@@ -31,11 +31,13 @@
             "diagnostic_radiology": "Diagnostic Radiology",
             "emergency": "Emergency Medicine",
             "endocrinology": "Endocrinology",
+            "endodontics": "Endodontics",
```
matkoniecz, line 34: I guess that it is not necessary to provide "Endodontics" in more understandable way?

kjonosm: I don't know how to interpret your question. However, first and foremost adding these strings allow for a translation and therefore indeed provide for more understandable labels for many non English users.

matkoniecz: I meant that at least to me term "Endodontics" is not clear at all and does not really explain anything - but other medical fields are also using similar labels rather than more clear ones (say "Dental pulp specialisation"), so it fits what we have already

data/fields/healthcare/speciality.json
```diff
@@ -31,11 +31,13 @@
...
+            "familiy_medicine": "Family Medicine",
```
matkoniecz: "familiy" is surely a typo?

data/fields/healthcare/speciality.json
```diff
@@ -54,17 +56,20 @@
             "neuropsychiatry": "Neuropsychiatry",
             "neurosurgery": "Neurosurgery",
             "nuclear": "Nuclear Medicine",
+            "obstetrics": "Obstetrics",
```
matkoniecz, line 59: should it be a separate value?

https://wiki.openstreetmap.org/wiki/Key:healthcare:speciality does not list it

kjonosm: Obstetrics is a medical specialty that deals with pregnancy, childbirth and the postpartum period. It covers topics such as prenatal care, fetal assessments, labor and delivery, and complications of pregnancy.

Gynaecology encompasses sexual health and diagnosing and treating health issues arising from the female reproduction system.

Both specialities are often but not always combined

data/fields/healthcare/speciality.json
kjonosm: I would leave that question to a native English speaker.  It should however not block this pr as changing the label in a separate pr later is always possible.

data/fields/healthcare/speciality.json
matkoniecz: If that makes sense then wiki docs should be changed first, potentially after discussion with osm community

matkoniecz: iD tagging schema does not get to invent/change tags even if they would make sense

if you want to change tagging please consult with community first - please create thread at https://community.openstreetmap.org/ and link it here

kjonosm: Those specialties are very often combined. But what if e.g. `amenity=doctors` only offers `obstetrics` but no `gynaecology` or vice versa? Not tag it at all? If you don't want to have this value in the list, please amend the pr according to your liking. I don't want to repeat myself but first and foremost this pr allows for a translation and therefore provide for more understandable labels for many non English users. I hope this argument is considered in your triage.

matkoniecz: I am not saying that this tag makes no sense, I am saying that iD PR is not a good place to invent brand-new tagging

> If you don't want to have this value in the list, please amend the pr according to your liking

Note that PRs where authors are willing to fix problems have higher priority and important cases from https://github.com/openstreetmap/id-tagging-schema/issues have priority

kjonosm: > to invent brand-new tagging

Current [taginfo](https://taginfo.openstreetmap.org/search?q=healthcare%3Aspeciality%3Dobstetrics) numbers show that `healthcare:speciality=obstectrics` is in reasonable use.

matkoniecz: if that is true and tagging is not controversial then maybe adding it to wiki without discussion would make sense (I have NOT investigated this)

PR #999, opened 2023-09-08
Add field for `opening_hours:drive_through=*`

arch0345:
Many establishments with drive throughs have different opening hours compared to their normal opening hours. For example, [this Krispy Kreme location](https://www.openstreetmap.org/way/170963639) has their drive through open 24/7 while the inside of their shop is only open during certain hours.

### DIFF
--- a/data/fields/opening_hours/drive_through.json
+++ b/data/fields/opening_hours/drive_through.json
@@ -1,0 +1,12 @@
+{
+    "key": "opening_hours:drive_through",
+    "type": "combo",
+    "label": "Drive Through Hours",
+    "placeholder": "Unknown",
+    "prerequisiteTag": {
+        "key": "drive_through",
+        "value": "yes"
+    },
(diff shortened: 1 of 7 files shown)


kjonosm:
@arch0345 Is it possible to only show this field when `drive_through=yes` has been added?

arch0345:
> @arch0345 Is it possible to only show this field when `drive_through=yes` has been added?

Yes, I added `drive_through=yes` as a `prerequisiteTag`:
https://github.com/openstreetmap/id-tagging-schema/blob/70eb775a9223dfd8d7af3d01d244cb771d7b7fe2/data/fields/opening_hours/drive_through.json#L6-L9

tyrasd (approved):
The tag is not quite super popular right  now, but the trajectory is pointing steeply upwards, so :+1: from my side to includes it. See one possible improvement suggestion below; let me know what you think about that.

PR #1000, opened 2023-09-09
Add ref:edubase

arrival-spring:
[ref:edubase](https://wiki.openstreetmap.org/wiki/Key%3Aref%3Aedubase) should be used on all educational establishments in England and Wales and is used to link to [official data](https://get-information-schools.service.gov.uk/) about them.


tyrasd (approved):
Thank you. :100:  FYI: I just changed the file name to follow the naming conventions for regional fields (i.e. adding the country as a suffix to the field name) and moved into the `ref` subdirectory for convenience.

PR #1001, opened 2023-09-09
Update public_bookcase.json - Add more fields

danieldegroot2:
Adds moreFields capacity, colour, height, support and moves lit from fields to moreFields.

### DIFF
--- a/data/presets/amenity/public_bookcase.json
+++ b/data/presets/amenity/public_bookcase.json
@@ -4,22 +4,25 @@
         "name",
         "public_bookcase/type",
         "operator",
-        "capacity",
-        "lit"
+        "capacity"
     ],
     "moreFields": [
         "access_simple",
         "address",
         "branch_brand",
         "brand",
+        "colour",
         "covered",
         "email",
+        "height",
         "indoor",
         "level",
+        "lit",
         "location",
         "opening_hours",
         "phone",
         "ref",
+        "support",
         "wheelchair"
     ],
     "geometry": [


data/fields/books.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "books",
+    "type": "combo",
```
tyrasd, line 3: this should probably a `semiCombo` field, for when different genres are offered

data/presets/amenity/public_bookcase.json
```diff
@@ -3,23 +3,28 @@
     "fields": [
         "name",
         "public_bookcase/type",
+        "books",
```
tyrasd, line 6: I would have said that for public bookcases, it is rather the exception to find only specific types of books.  For this reason, I would have put this field to the `moreFields` section.

data/presets/amenity/public_bookcase.json
```diff
@@ -3,23 +3,28 @@
...
         "operator",
         "capacity",
-        "lit"
```
tyrasd, line 8: The `lit` tag seems to be a relatively popular combination (see taginfo: https://taginfo.openstreetmap.org/tags/amenity=public_bookcase#combinations). Why hide it?

data/presets/amenity/public_bookcase.json
```diff
@@ -3,23 +3,28 @@
...
+        "direction"
```
tyrasd, line 9: The `direction` tag is not used very commonly on public bookcases. What's the intention behind offering the field by default for these features?

tyrasd (commented):
Usually, I use the popularity of tag combinations to determine which fields to offer by default and which to include as optional fields. See below for some instances where I noticed that the preset would not match this rule of thumb.

Also, the `books` field should pobably allow for a multi-selection (see also the inline comment below).

data/fields/books.json
danieldegroot2: Sure.

data/presets/amenity/public_bookcase.json
danieldegroot2: It is already exception they are mapped, let alone such tags being discovered by users, but it is fine to change it.

data/presets/amenity/public_bookcase.json
danieldegroot2: For clarity, it is moved to moreFields. (it is not hidden if people can easily discover it)
I think it is overall more of an exception they are lit, although relatively popular now.
Similarly, I think it is exception people will use a bookcase at night (except in areas with longer nights).
Otherwise I also don't think it is significant property of object itself, like for advertising.

data/presets/amenity/public_bookcase.json
danieldegroot2: So it can be more easily discovered.
`direction` in general is not often used, because it is not directly a object/access type.
However, it is often found on/near barrier and usually only accesible from one side.
It is recommended to add direction to such objects to prevent ambiguity with its position relative to other objects.
(like cameras, defibrillators, etc.)

tyrasd (approved):
Thanks for the feedback. I've changed the field type to allow multiple values and tweaked the label to also fit other uses of the tag (e.g. on book shops or libraries). I've also reverted the field "ordering" to match the usage/popularity of the attributes for now.

PR #1002, opened 2023-09-09
Add translatable strings to bridge field

kjonosm:


kjonosm:
Hi @tyrasd, would you be able to review this pr?

tyrasd (approved):

tyrasd:
Thank you, this looks good to me. It's only a bit unfortunate that this only applies to the _Bridge Type_ field for the preset `man_made=bridge` and not for for selecting the type of bridge in the `structure` field which is used on all highway presets for example. That would need to be addressed separately, however, as it would involve adjusting some code in iD and likely also the tagging schema's json schema.

PR #1003, opened 2023-09-09
Add translatable strings to boundary field

kjonosm:


data/fields/boundary.json
```diff
@@ -1,5 +1,22 @@
...
+            "forest": "Forest Boundary",
+            "forest_compartment": "Forest Compartment",
+            "hazard": "Hazardous Area",
+            "maritime": "Maritime Boundary",
+            "national_park": "National Park",
+            "political": "Political / Electoral Boundary",
+            "postal_code": "Postal Code Boundary",
+            "protected_area": "Protected Area",
+            "special_economic_zone": "Special Economic Zone",
+            "disputed": "Disputed Territory"
```
matkoniecz, line 19: I have doubts should we encourage mapping those

matkoniecz: @kjonosm are you sure that mapping these =disputed has support from OSM community and there is consensus it is a good idea?

data/fields/boundary.json
```diff
@@ -1,5 +1,22 @@
 {
     "key": "boundary",
     "type": "combo",
-    "label": "Type"
+    "label": "Type",
+    "strings": {
+        "options": {
+            "aboriginal_lands": "Aboriginal Lands",
+            "administrative": "Administrative Boundary",
```
k-yle, line 8: we should re-use translations where possible. Same for `hazard`

```suggestion
            "administrative": "{presets/boundary/administrative}",
```

matkoniecz:
@kjonosm Is this an abandoned PR?

matkoniecz:
considering rejection as abandoned one

Issue #960, opened 2023-07-16
Presets for extended playground 


tordans:
### OSM Tag(s)

playground=*

### How would you like this tag to see supported?

As a Preset

### Label

Playground

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Proposal:Extended_playground_equipment#Tagging:_List_of_new_device_values

### Status of the Tag

Approved

### Usage of the tag

New values have low usage

### Regional Tag?

global

### Further Information

The proposal to extend the playground equipment values was just approved. 
* [Update List of `playground=*` documented values](https://wiki.openstreetmap.org/wiki/Key:playground#Values)
* ([Proposal](https://wiki.openstreetmap.org/wiki/Proposal:Extended_playground_equipment#Tagging:_List_of_new_device_values))

What are your thoughts on …

- [ ] Adding the missing tags as separate presets
- [ ] Adding a group them in a category ([existing categories](https://github.com/openstreetmap/id-tagging-schema/tree/main/data/preset_categories)) (Ping https://github.com/openstreetmap/id-tagging-schema/issues/58)
- I will also create OSM wikidata items for the key so the info `(i)` shows a description for the given key
- I don't think there are any deprecations or renames part of the proposal

---

Some of the values have low usage as of now. But since apps like GoMap and EveryDoor work best with presets, I think we should add them anyways in order to promote the documented values of custom values that people come up with on the spot…

---

I though about adding one preset with a options list of strings + descriptions; however that would reduce the discoverability for the values a lot and also we have many values as separate presets already.

cuatim:
Happy to help! :)

There are a bunch of OSM wiki data items for `playground=...` missing. Is there an easy way to create the basic item for all those? The [normal](https://wiki.openstreetmap.org/wiki/Data_items#Item_creation_process) way is a bit tedious for this.

> Remove all presets for those presets that are unspecific like `playground=water`

IMHO, the "fallback values for devices" have legitimate use. There are devices too specific for its own tag...

cuatim:
Current state of OSM wiki data items, according this to abominable oneliner:
[...]

PR #1004, opened 2023-09-15
Add translatable strings to playground field

kjonosm:
See also #960


tordans:
I don't think this is a good solution to https://github.com/openstreetmap/id-tagging-schema/issues/960. I think we should continue to have separate presets for each playground equipment like we already have for most.

My main reasons are
- There are separate presets already, having two systems of presets is not helpful
- The list proposed here does not allow to validate the geometry of subkeys. Eg. springy should always be a node where seesaw can also be a line/area (https://wiki.openstreetmap.org/wiki/Key:playground#Values)
- This solution does not allow to suggest specific fields for each equipment

kjonosm:
This pr should not replace #960 but help to choose the right playground equipment providing a label/description in local language. It may also help to easily switch between presets once they are created. So +1 to create separate presets!

data/fields/playground.json
```diff
@@ -1,5 +1,94 @@
...
     "key": "playground",
     "type": "combo",
-    "label": "Type"
+    "label": "Type",
+    "strings": {
+        "options": {
+            "activitypanel": "Activity Panel",
+            "aerialrotator": "Aerial Rotator",
+            "archimedes_screw": "Archimedes Screw",
+            "balance": "Balancing Device (unspecified)",
```
1ec5, line 10: As a side note, the “(unspecified)” pattern is a bit unfortunate, since we otherwise aim for natural-language preset names. Maybe it should be “Generic Balancing Device” etc. But that would be a bigger change involving the presets too, not just these strings.

data/fields/playground.json
```diff
@@ -1,5 +1,94 @@
...
+            "balancebeam": "Balance Beam",
+            "ball_pool": "Ball Pool",
+            "bannister_bars": "Bannister Bars",
```
tordans:
```suggestion
            "bannister_bars": "Banister Bars",
```

> "Banister" is mainly used in US English, while "bannister" is mainly used in British English. In the US, "banister" is preferred, and in the UK, "bannister".

https://www.google.com/search?q=Bannister+or+Bannister

data/fields/playground.json
tordans: I like that it is a postfix, not a prefix, which gives the descriptive part more visibility.
We should move that into a separate discussion to maybe change it for all occurrences.

data/fields/playground.json
```diff
@@ -1,5 +1,91 @@
...
+            "water_basin": "Water Basin",
+            "water_cannon": "Water Cannon",
+            "water_channel": "Water Channel",
+            "water_seesaw": "Water Seesaw",
+            "water_sprayer": "Water Sprayer",
+            "water_stream": "Water Stream",
+            "water_wheel": "Water Wheel",
+            "wobble_bridge": "Wobble Bridge",
+            "youth_bench": "Youth Bench",
+            "zipwire": "Zipwire"
```
tordans:
```suggestion
            "zipwire": "Zipline / Zipwire"
```

Cannot hurt to have both variations in here…

> "Zipwire" and "zipline" (or "zip line") are largely synonymous terms for a cable system with a pulley that allows a person or object to slide from a higher point to a lower one due to gravity. While some sources indicate that "zipwire" is a UK term for the activity and "zipline" is more common in the US, they refer to the same type of gravity-powered cable ride used for recreation or transportation.

data/fields/playground.json
```diff
@@ -1,5 +1,91 @@
...
+            "ride_on": "Fixed Riding Device",
+            "rope_swing": "Rope Swing",
+            "rope_traverse": "Rope Traverse",
+            "rotator": "Rotating Device",
+            "roundabout": "Roundabout",
+            "sand": "Sand Play Device",
+            "sand_pulley": "Sand Pulley",
+            "sand_seesaw": "Sand Seesaw",
+            "sand_wheel": "Sand Wheel",
+            "sandpit": "Sandpit",
```
tordans:
```suggestion
            "sandpit": "Sandbox / Sandpit",
```

> A sandpit (most Commonwealth countries) or sandbox (US and Canada) is a low, wide container or shallow depression filled with soft (beach) sand in which ...

https://www.google.com/search?q=sandpit+vs.+sandbox

data/fields/playground.json
```diff
@@ -1,5 +1,91 @@
...
+            "musical_instrument": "Musical Instrument",
+            "parallel_bars": "Parallel Bars",
+            "platform": "Platform",
+            "playhouse": "Playhouse",
+            "pump": "Pump",
+            "ride_on": "Fixed Riding Device",
+            "rope_swing": "Rope Swing",
+            "rope_traverse": "Rope Traverse",
+            "rotator": "Rotating Device",
+            "roundabout": "Roundabout",
```
tordans:
```suggestion
            "roundabout": "Merry-Go-Round / Roundabout",
```

https://en.wikipedia.org/wiki/Roundabout_(play)

> A roundabout ([British English](https://en.wikipedia.org/wiki/British_English)), merry-go-round ([American English](https://en.wikipedia.org/wiki/American_English)), or carousel ([Australian English](https://en.wikipedia.org/wiki/Australian_English)), is a piece of [playground](https://en.wikipedia.org/wiki/Playground) equipment,…

data/fields/playground.json
tordans: Moved to https://github.com/openstreetmap/id-tagging-schema/issues/1706

tordans:
@1ec5 could you check the latest translation changes in the (now hidden) code review comments (I put my sources in those comments). I change some strings to use the US / Int. english variation or have both in the string.

I think we can merge this now. The PR will add the keys from https://wiki.openstreetmap.org/wiki/Key%3Aplayground except for one which has no usage. (That one is useful… but should be used at least once :-).) I did not check all usage stats, which I will to for https://github.com/openstreetmap/id-tagging-schema/issues/960; it OK to have less used strings in this list IMO.

tordans (approved):

PR #1008, opened 2023-09-20
Add translatable strings for Swimming Pool field

arch0345:
Currently, the swimming pool field is only being used for the Swimming Pool preset, however [the most common values for `swimming_pool=*`](https://taginfo.openstreetmap.org/keys/swimming_pool#values) are ones [that aren't relevant for its use in combination with `leisure=swimming_pool`](https://wiki.openstreetmap.org/wiki/Key:swimming_pool#Values) (ie `yes`, `no`, `outdoor`, `indoor`).

This PR also adds a preset for `leisure=hot_tub` and adds a deprecation rule for `leisure=swimming_pool` + `swimming_pool=hot_tub`

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -987,6 +987,10 @@
     "replace": {"landuse": "recreation_ground"}
   },
   {
+    "old": {"leisure": "swimming_pool", "swimming_pool": "hot_tub"},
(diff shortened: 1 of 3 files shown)


tyrasd (approved):
looks good to me, thanks :+1:

PR #1014, opened 2023-09-24
Add translatable strings to field seamark:beacon_isolated_danger:shape

kjonosm:


data/fields/seamark/beacon_isolated_danger/shape.json
```diff
@@ -1,5 +1,12 @@
 {
     "key": "seamark:beacon_isolated_danger:shape",
     "type": "combo",
-    "label": "Shape"
+    "label": "Shape",
+    "strings": {
+        "options": {
+            "pile": "Pile",
+            "stake": "Stake",
+            "tower": "Tower"
```
k-yle, line 9: `seamark:beacon_*:shape` is a standard key, so I would suggest defining all the possible options, not just these 3.

[Here's how the preset is defined in a fork of iD](https://github.com/k-yle/id-tagging-schema/blob/f202907/data/fields/seamark/beacon_lateral/shape.json#L6-L14) (see [here](https://kyle.kiwi/iD/#id=n10251795794&overlays=openseamap-overlay&map=18/-33.83677/151.27212) for a preview)

I would also suggest updating the field for `seamark:beacon_lateral:shape`, so that it references the same options. [Example of the code here](https://github.com/k-yle/id-tagging-schema/blob/f202907/data/fields/seamark/beacon_cardinal/shape.json#L5)

kjonosm: @k-yle

> I would also suggest updating the field for seamark:beacon_lateral:shape, so that it references the same options.

I've read the schema-builder [section](https://github.com/ideditor/schema-builder?tab=readme-ov-file#stringscrossreference) about cross referencing fields between presets. How do I reference fields from a different folder?

k-yle: @kjonosm [here's an example](https://github.com/k-yle/id-tagging-schema/blob/f202907/data/fields/seamark/beacon_cardinal/shape.json#L5)

data/fields/seamark/beacon_isolated_danger/shape.json
```diff
@@ -1,5 +1,41 @@
 {
     "key": "seamark:beacon_isolated_danger:shape",
     "type": "combo",
-    "label": "Shape"
+    "label": "Shape",
+    "strings": {
+        "options": {
+            "pile": {
+                "title": "Pile Beacon",
+                "description": "A long, heavy pile or section of metal, wood or concrete, driven into the seabed, used as a navigational aid."
```
matkoniecz: Are these meanings documented anywhere?

I checked https://wiki.openstreetmap.org/wiki/Key:seamark:beacon_isolated_danger:shape and see only few of them therre

matkoniecz (changes requested):

data/fields/seamark/beacon_isolated_danger/shape.json
```diff
@@ -1,5 +1,41 @@
...
+            },
+            "stake": {
+                "title": "Stake",
+                "description": "An elongated wood or metal stake, embedded in the bottom of the sea, used as a navigational aid."
+            },
+            "pole": {
+                "title": "Pole",
+                "description": "An elongated wood or metal pole, embedded in the bottom of the sea, used as a navigational aid."
+            },
+            "withy": {
```
matkoniecz: https://wiki.openstreetmap.org/wiki/Tag:seamark:beacon_isolated_danger:shape=withy is not existing

https://taginfo.openstreetmap.org/tags/seamark:beacon_isolated_danger:shape=withy shows zero uses

https://taghistory.raifer.tech/#***/seamark%3Abeacon_isolated_danger%3Ashape/withy shows that it was never ever used

have I messed up links, value is wrong here or is it adding support to value that never was used even once?

matkoniecz (changes requested):
I am a bit confused

k-yle (approved):
Looks good to me.

@matkoniecz the documentation is here: https://osm.wiki/Seamarks/Beacons#Shapes 

Even if there are 0 uses for some values under `…beacon_isolated_danger:*`, we still need to define all the labels, so that [5 of the 11 other `seamark:*:shape` fields](https://taginfo.openstreetmap.org/search?q=%3Ashape) can reference these labels with [`stringsCrossReference`](https://github.com/ideditor/schema-builder#stringscrossreference).

matkoniecz:
Which specific tags in iD need this references? 

Is it possible to do this without introducing presets for tags not used even once on entire planet?

k-yle:
`withy` is a weird one, it's only [used 2 times](https://taginfo.openstreetmap.org/tags/seamark%3Abeacon_lateral%3Ashape=withy). So maybe that value should be removed?

---

And yes, these translations are for `seamark:beacon_*:shape`, not just `seamark:beacon_isolated_danger:shape`.

The file you mentioned, `seamark:beacon_lateral:shape` is one example which should [crossReference](https://github.com/ideditor/schema-builder#stringscrossreference) this file. In the future, there might be [other fields](https://osm.wiki/Seamarks/Seamark_Objects#:~:text=beacon_) referencing these translations.

matkoniecz:
@k-yle Are you potentially interested in taking over this PR/submitting own inspired by it?

k-yle:
yeah, i'm happy to champion this PR. 

lets definitely remove `withy`. and should we remove all the `description`s? I think `title` is sufficient, since very few presets have `description`

matkoniecz:
I think descriptions are fine if they are clearly helpful and labels are enough to distinguish between values...

But if you, as new PR author think that these are unwanted I will not protest.

I would not avoid them solely because they are not used much, this way no new feature (or tag) would ever gain use...

data/fields/seamark/beacon_isolated_danger/shape.json
k-yle: they're documented on [Seamarks/Beacons#Shapes](https://osm.wiki/Seamarks/Beacons#Shapes) and other pages like [Key:seamark:beacon_lateral:shape](https://osm.wiki/Key:seamark:beacon_lateral:shape)

data/fields/seamark/beacon_isolated_danger/shape.json
k-yle: i removed withy

k-yle:
## Test-Documentation

<table>
<tr>
 <td>iso danger
 <td>channel markers
<tr>
 <td>image
 <td>image
</table>



### Info-`i`

<table>
<tr>
 <td>iso danger
 <td>channel markers
<tr>
 <td>image
 <td>image

</table>

k-yle (approved):
my approval is meaningless because i'm now a contributor. 

latest changes:
 - removed `withy` due to low usage
 - removed the descriptions. very few presets have descriptions, and it will just cause more hate against `seamark` tags if translators are given excessive work 
 - added `stringsCrossReference` to all other `seamark:*:shape` fields

matkoniecz:
[seamark:beacon_lateral:shape](https://wiki.openstreetmap.org/wiki/Key:seamark:beacon_lateral:shape)=perch seems to have quite many uses, but no idea should it be supported

matkoniecz:
It would be really really nice to have icons. https://pinhead.ink/ has added some canoe-focused icons maybe maintainer would also accept PR adding some seamark-related ones? (or maybe even implement some? but https://github.com/waysidemapping/pinhead/issues has more than enough ideas for now)

(this is not a blocker for this PR, just mentioning as meanings are non-obvious and you basically need to check wiki anyway)

matkoniecz:
@k-yle Maybe it is just me, maybe it is an iD bug. But say

```
seamark:beacon_lateral:shape=pile_beacon
seamark:type=beacon_lateral
```

is NOT displayed with translation for me, despite dropdown having translation, if you select this option translation is not listed in GUI. What is going on?

data/fields/seamark/beacon_isolated_danger/shape.json
matkoniecz: added link in https://wiki.openstreetmap.org/w/index.php?title=Key:seamark:beacon_isolated_danger:shape&diff=prev&oldid=3062596

matkoniecz:
Both pole and stake are listed as separate entries, while https://wiki.openstreetmap.org/wiki/Key:seamark:beacon_lateral:shape fails to distinguish between those.

Are these synonyms? If synonyms they should share label (and ideally community would deduplicate values if these are true duplicates, or document differences or at least document controversy)

matkoniecz:
screen-2026-07-26-14-13-01

info button for some provides no helpful info, I guess it is powered by that data items rather that OSM Wiki so I would not consider it to be blocker

though having some simple description (not long ones like those removed) would be nice, if possible

or those already mentioned icons

k-yle:
> maybe it is an iD bug

oh, yeah this is a schema-builder v7 bug 😭 it also affects `plant:source=*` on `power=planet`

k-yle:
~~draft because it's blocked by #2608~~

k-yle:
PR is no longer blocked.

`stake`, `post`, `pole`, and `perch` refer to the same thing. So i removed `pole`, which is used [much less](https://taginfo.openstreetmap.org/keys/seamark%3Abeacon_lateral%3Ashape#values). the [S-57 spec (page 11)](https://iho.int/uploads/user/pubs/standards/s-57/31ApAch2.pdf#page=11) also uses a single value for this concept

matkoniecz:
Do we have way to show other with the same label without duplicate values in dropdown?

Or should they be rather retagged (after discussion,  not decided here etc)

matkoniecz (changes requested):
https://github.com/openstreetmap/id-tagging-schema/pull/1014#issuecomment-5104216605

k-yle:
yeah those weird values should probably be retagged, it doesn't make sense to support 4 values with exactly the same definition. i guess it's partially the fault of the [docs for `BCNSHP`](https://docs.teledynecaris.com/s-57/attribut/def/d-bcnshp.htm).

matkoniecz:
@k-yle so per meeting discussion, it should be fine to merge without support for extra ones

matkoniecz (approved):

PR #1019, opened 2023-09-30
Add `contact:instagram` field for POIs

sbraz:
Hi, I based this on #859, hopefully I got it right :)

### DIFF
--- a/data/fields/contact/instagram.json
+++ b/data/fields/contact/instagram.json
@@ -1,0 +1,6 @@
+{
+    "key": "contact:instagram",
+    "type": "url",
+    "label": "Instagram account name or URL",
+    "placeholder": "https://www.instagram.com/openstreetmap/"
+}
--- a/data/presets/@templates/poi.json
+++ b/data/presets/@templates/poi.json
@@ -4,6 +4,7 @@
         "{@templates/contact}",
         "address",
         "contact/facebook",
+        "contact/instagram",
         "gnis/feature_id-US",
         "level",
         "not/name",
--- a/data/presets/leisure/horse_riding.json
+++ b/data/presets/leisure/horse_riding.json
@@ -9,6 +9,7 @@
     "moreFields": [
         "{@templates/contact}",
         "contact/facebook",
+        "contact/instagram",
         "gnis/feature_id-US",
         "opening_hours",
         "payment_multi"


matkoniecz:
Is there some way to get this autodeployed code preview? Why it has not appeared here?

sbraz:
Let's see if force-pushing helps.

data/fields/contact/instagram.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "contact:instagram",
+    "type": "url",
+    "label": "Instagram account name (also used on Threads) or full Instagram URL",
```
tyrasd, line 4:
```suggestion
    "label": "Instagram Account Name or URL",
```

tyrasd (changes requested):
Hey, thank you for this. It generally looks good to me. Only the field label feels a bit too long. See below for a suggestion to shorten it.

image


Btw: I had to manually trigger the build, as this is your first PR on this repository. You can now try the changes out on the preview build.

data/fields/contact/instagram.json
```diff
@@ -0,0 +1,6 @@
...
+    "placeholder": "account_name"
```
tyrasd, line 5: Full URLs seem to be slightly more popular at first glance (see [taginfo](https://taginfo.openstreetmap.org/keys/?key=contact%3Ainstagram#values)) and are probably easier to copy-paste, so maybe use something like this for the placeholder:

```suggestion
    "placeholder": "https://www.instagram.com/LocalBusiness"
```

sbraz: I _personally_ prefer using the account name because I've never seen a business advertise their Instagram with the URL, but I've updated the PR to use a URL. I took the liberty of using https://www.instagram.com/openstreetmap/ which seems like a legitimate Instagram account.

sbraz:
> You can now try the changes out on the preview build.

Thanks and thanks for the review, I think this push addresses all comments :)

Hufkratzer:
`contact:instagram` should additionally be added everywhere where `contact:facebook` is. e.g. also in [leisure=horse_riding](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/leisure/horse_riding.json). I don't know if there are more such cases.

sbraz:
I addressed @tyrasd's comments long time ago. Can someone please review this again and let me know if something is still missing?

sbraz:
The preview works as expected:
image

matkoniecz (approved):
though for the record `contact: ` prefix is really silly here

but, well, just because I can approve PR here does not mean that I get to issue deprecations decrees and usage seems sadly clear: https://taghistory.raifer.tech/#***/contact%3Ainstagram/&***/instagram/

and fixing bad `contact:` prefixes was attempted already and failed

tyrasd (approved):

pyrog:
iD don't open the instagram website for this [POI ](https://www.openstreetmap.org/node/7725333239).
(Even for Facebook if I remove the `https://www.facebook.com/` prefix).

```
contact:facebook=100063464295860
contact:instagram=atelierplongequilibre
contact:twitter=plongequilibre

```

or this [one ](https://www.openstreetmap.org/edit?node=3033091336): 
```
contact:facebook=GladinesBatignolles
contact:instagram=chez.gladines

```

matkoniecz:
if that is not-yet reported it likely should be a new issue in https://github.com/openstreetmap/iD/issues

comments in closed PRs can be easily missed

matkoniecz:
from what I see iD tagging schema cannot do much here, and it is an iD interface problem

I faintly remember it being reported for Facebook field already

k-yle:
I opened https://github.com/openstreetmap/iD/pull/12306 to address [this comment](https://github.com/openstreetmap/id-tagging-schema/pull/1019#issuecomment-4324879974)

PR #331, opened 2022-01-05
Add presets for Germany School Types based on `isced:level`

tordans:
# Why

At the moment, there is only one preset for schools, when in fact we have at least three level of schools with multiple "versions" for level 2 (Realschule, Gymnasium, Hauptschule, Gesamtschule), and level 3 (Gymnasium, Gesamtschule).

The goal is, to have precise presets that allow to search and see the more precise school.

Example in OSM: https://www.openstreetmap.org/way/47023392#map=18/52.47302/13.44481

# Tagging

Unfortunately, this topic is quite complex. The best way to tag those schools seems to be using `isced:level=1|2|2;3` (for Germany).

Some Wiki research:

* The German "School" Wiki page explains the most common values "1" and "2;3" for Germany https://wiki.openstreetmap.org/wiki/DE:Tag:amenity%3Dschool, same for the EN-Version
* The German "isced:level" Wiki page lists "Schulformen" with the corresponding recommended isced-level https://wiki.openstreetmap.org/wiki/DE:Key:isced:level#Werte_f.C3.BCr_D-A-CH
* The English "isced:level" Wiki page has good short table at the start https://wiki.openstreetmap.org/wiki/Key:isced:level#Level and a huge table afterwards, which tries to find terms per country https://wiki.openstreetmap.org/wiki/Key:isced:level#Level

## Tagging: `grades=*`

There is also the tag `grades` which we might want to add to those presets. https://wiki.openstreetmap.org/wiki/DE:Key:grades

The given example would match nicely with the new presets here. But I did not look into it too much, so that might be wrong … — Writing this down I see a first issue, since some Bundesländer don't have a year 13 anymore, right? So maybe adding a tag that might change at any year based on laws is not that great(?)

```
"isced:level": "1"-preset => "grades": "1-4"
"isced:level": "2"-preset => "grades": "5-10"
"isced:level": "3"-preset => "grades": "11-13" // 11-12 somewhere, right?
"isced:level": "2;3"-preset => "grades": "5-13" // 5-12 somewhere, right?
```

## Tagging: Out of scope

- I did not look at other countries (see above)
- I did not look at Kindergarden and Universities(…)

Both can IMO be follow up PRs.

# Scope: Germany

This PR tries to look at Germany only. Which I find complex enough. I suggest we try to get this going and add other countries afterwards. Since it is quite hard to find good (american) english terms for this, I just went and used German terms in the json files. Which is probably something we need to clean up.

# FYI: Related OSM Wiki-Work

[...]

PR #1021, opened 2023-10-06
Add specific school presets for the US

arch0345:
Based off of #331

I decided to make the field semiCombo as K-8 (and less common 6-12) schools exist, however they might not be common enough to warrant it's own option/preset


data/presets/amenity/school/3-US.json
```diff
@@ -0,0 +1,24 @@
...
+        "point"
+    ],
+    "tags": {
+        "amenity": "school",
+        "isced:level": "3"
+    },
+    "reference": {
+        "key": "isced:level"
+    },
+    "name": "High School Grounds"
```
1ec5, line 23: ISCED level 3 is officially known as “upper secondary” and covers grades 10 through 12 or, alternatively, the first three years of vocational school. This preset applies to 0.9% of public schools and 0.4% of private schools. However, 9.8% of public schools and 4.0% of private schools have “High School” or “Senior High School” (but not “Junior High School”) in their name.

1ec5 (changes requested):
`isced:level` was a mistake on multiple levels. Much of the criticism directed at the rejected [`isced:2011:level`](https://wiki.openstreetmap.org/wiki/Proposal:ISCED_2011_Level_of_Education) and [`isced_2011_programme`](https://wiki.openstreetmap.org/wiki/Proposal:ISCED_2011_Education_Programme) proposals applies equally to `isced:level` (which is not an approved tag anyhow).

The problems with `isced:level` are particularly acute for the American school system. The proposed presets have tags that should apply to at most 10% of schools in the U.S., but they have names implying that they apply to a much broader set of schools. In the vast majority of cases, mappers will unknowingly mistag `isced:level`, or they’ll change it from a correct value to an incorrect one. They may not even be able to decide on the correct value of `isced:level` because they don’t know which grades the school serves. This completely defeats the purpose of `isced:level`, which is to harmonize school classification internationally.

Moreover, this PR would apply the presets to Puerto Rico, which has never had an official mapping from ISCED to the local school system. Puerto Rico has been moving towards the American system since 2018, but a comprehensive import of schools in Puerto Rico [applied `isced:level` according to the old system](https://wiki.openstreetmap.org/wiki/Talk:Puerto_Rico_Projects/Puerto_Rico_Points_of_Interest_Import#Updating_schools).

Any school classification tagging for the U.S. [really should use `school`](https://www.openstreetmap.org/user/Minh%20Nguyen/diary/398412) instead, in conjunction with the `grades` field that was added in openstreetmap/iD@ab682380db487279f39e9bbc81822d972cdf2646.

1ec5:
> I decided to make the field semiCombo as K-8 (and less common 6-12) schools exist, however they might not be common enough to warrant it's own option/preset

There are 3,514 public K–8 schools across the country, or 3.5% of all public schools, far more than any of the sets of schools that are covered by these presets.

arch0345:
Thanks for pointing that out @1ec5, I was looking at implementing these changes a couple months ago, and that was probably why I didn't end up opening a PR. I was hoping that `isced:level` could help with the differing grade levels these types of schools serve, but it looks like this isn't the case.

Most of the common values for `grades` aren't applicable to the US, so I'm thinking of creating a new region specific field that highlights values common in the US: https://taginfo.openstreetmap.org/keys/grades#values

arch0345:
@tordans do you want me to repurpose this PR for the changes you made in https://github.com/arch0345/id-tagging-schema/pull/1 or would you prefer updating #331/creating a new PR?

1ec5:
> Most of the common values for `grades` aren't applicable to the US, so I'm thinking of creating a new region specific field that highlights values common in the US: https://taginfo.openstreetmap.org/keys/grades#values

That’s a good idea. I’d be particularly interested in omitting the “0” value, which came in from an Australian import that misunderstood a unilateral wiki edit that misunderstood the key as a different way of counting ages. Mappers have been logically assuming it refers to kindergarten in some cases.

Since there are quite a few likely ranges, I wonder if we need a new `range` field type that can accept a minimum and maximum value (which can both be nonnumeric).

tordans:
> Mappers have been logically assuming it refers to kindergarten in some cases.

JFYI because I just came across it, this is in fact how it is documented in https://wiki.openstreetmap.org/wiki/DE:Key:grades#Nach_Land

tordans:
Thanks this input! I took a step back and looked at possible alternatives to roll out in Germany, see https://community.openstreetmap.org/t/art-der-schulen-uber-isced-level-school-oder-school-de/104643

1ec5:
> > Mappers have been logically assuming it refers to kindergarten in some cases.
> 
> JFYI because I just came across it, this is in fact how it is documented in https://wiki.openstreetmap.org/wiki/DE:Key:grades#Nach_Land

Yes, that was based on some unilateral edits in 2017 that still partially remain [in this table](https://wiki.openstreetmap.org/wiki/Key:grades#2017_proposal) due to subsequent use. Originally, [the scheme](https://wiki.openstreetmap.org/wiki/Proposal:Grades) went from −5 to 28, for lifelong learning from cradle to grave. It has little basis in reality.

matkoniecz:

Do you plan on finishing it? (note: I did no review, no idea whether this PR should be merged once finished)

PR #1024, opened 2023-10-07
Improve maxspeed tagging for Germany

tordans:
There is finally a conversation about cleaning up the `maxspeed` tagging in Germany. 

Right now, this PR holds the parts of https://community.openstreetmap.org/t/bereinigung-vereinheitlichung-von-source-maxspeed-maxspeed-type-usw/102689/5 which are clearly to be cleaned up (and above 100 usage).

I will update the PR with more deprecations once they are discussed. We can then use the tag update rules in iD with a MapRoulette Challenge to quickly cleanup the data.

Ideally, we will be able to add maxspeed fields as well.

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1940,5 +1940,21 @@
   {
     "old": {"memorial:text": "*"},
     "replace": {"inscription": "$1"}
+  },
+  {
+    "old": {"zone:maxspeed": "DE:urban"},
+    "replace": {"maxspeed:type": "DE:urban"}
+  },
+  {
+    "old": {"zone:maxspeed": "DE:rural"},
+    "replace": {"maxspeed:type": "DE:urban"}
+  },
+  {
+    "old": {"zone:maxspeed": "30"},
+    "replace": {"zone:maxspeed": "DE:30"}
+  },
+  {
+    "old": {"DE:zone30": "30"},
+    "replace": {"zone:maxspeed": "DE:30"}
   }
 ]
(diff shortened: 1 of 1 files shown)


data/deprecated.json
```diff
@@ -1940,5 +1940,21 @@
...
     "old": {"memorial:text": "*"},
     "replace": {"inscription": "$1"}
+  },
+  {
+    "old": {"zone:maxspeed": "DE:urban"},
+    "replace": {"maxspeed:type": "DE:urban"}
+  },
+  {
+    "old": {"zone:maxspeed": "DE:rural"},
+    "replace": {"maxspeed:type": "DE:urban"}
```
Dimitar5555, line 1950:
```suggestion
    "replace": {"maxspeed:type": "DE:rural"}
```

data/deprecated.json
```diff
@@ -1940,5 +1940,21 @@
...
+  },
+  {
+    "old": {"zone:maxspeed": "30"},
+    "replace": {"zone:maxspeed": "DE:30"}
+  },
```
Dimitar5555, line 1955:
```suggestion
```
This would apply to all places around the world with `zone:maxspeed=30`. It would be easier to mass edit it with JOSM.

tordans:
I will close this. There was a long discussion on the forum about all those maxspeed values and it sounded like they all have some slightly different meaning. Also, the deprecations are not a good place for this kind of cleanup.

Issue #1025, opened 2023-10-08
cycleway vs. cycleway:both


mangerlahn:
The iD Tagging schema currently uses `cycleway` to define whether a given way features cycling infrastructure of any kind. 
Relating tags are `cycleway:left` and `cycleway:right`. `cycleway:both` exists alongside `cycleway`, with mostly the same meaning. `cycleway:both` is seeing wide adoption in recent years and recently surpassed usage of `cycleway` (1.1 million vs. 1 million uses). According to the [wiki entry](https://wiki.openstreetmap.org/wiki/Key:cycleway:both) for `cycleway:both`, this tag is preferable over `cycleway` as its definition is more clear and consistent. 

I would suggest that the key `cycleway` used in `data/fields/cycleway.json` should be replaced with `cycleway:both`, as it appears to be the more popular of the two by now.
This would also make it more consistent with similar behaving tags like `parking:both`.

Not sure if ways using `cycleway` should offer a migration. The tag appears to be still valid, so a deprecation seems to be premature.

Any opinion on this?

mangerlahn:
Just realized this is a dupe of #743. Argument is even stronger now that `cycleway:both` is more common than `cycleway`.

tordans:
I vote to leave this open and consider it a different solution than https://github.com/openstreetmap/id-tagging-schema/issues/743. In https://github.com/openstreetmap/id-tagging-schema/issues/743 the idea is to expand the logic to handle both.

This issue could be about switching the default-both-tag from `cycleway` to `cycleway:both`. Wicht, AFAIK, is what https://github.com/openstreetmap/id-tagging-schema/pull/1026 does with minimal changes. Once the PR-Preview is there we can test this in a few places.

tordans:
@mangerlahn personally I would happily +1 a proposal to official deprecate the "null" case in favor of the "both" case which would help a lot to add the other tagging to the deprecation file which will trigger a tag-update-notice. And we could also run MapRoulette Challenges to update the "null" case tagging.

mangerlahn:
A deprecation/migration would have to be done carefully though, as cycleway would sometimes not refer to both left and right. 
Otherwise I’m all for it!

1ec5:
> A deprecation/migration would have to be done carefully though, as cycleway would sometimes not refer to both left and right.

[...]

PR #1026, opened 2023-10-09
cycleway: Switch to `:both` as the default

tordans:
Closes https://github.com/openstreetmap/id-tagging-schema/issues/1025
More info in that issue.

### DIFF
--- a/data/fields/cycleway.json
+++ b/data/fields/cycleway.json
@@ -1,5 +1,5 @@
 {
-    "key": "cycleway",
+    "key": "cycleway:both",
     "keys": [
         "cycleway:left",
         "cycleway:right"


matkoniecz:
Is it intentional that `cycleway=lane` is not shown at all?

k-yle:
I think this change needs to be made in iD, rather than this repo. As @matkoniecz pointed out, we would still want to support the legacy tags like `cycleway` right?

Once https://github.com/openstreetmap/iD/pull/9587 is merged, iD's field will support both `cycleway` and `cycleway:both`, so it would be a one-line change to make `*:both` the default for all `directionalCombo` (or just this one)

PR #1027, opened 2023-10-11
Add "barber" search term to hairdresser.json

pkoby:
Searching for "barber" in apps using this tagging schema doesn't turn up "Hairdresser" (noticed in Every Door app). This should allow that to work.

### DIFF
--- a/data/presets/shop/hairdresser.json
+++ b/data/presets/shop/hairdresser.json
@@ -5,6 +5,7 @@
         "area"
     ],
     "terms": [
+        "barber",
         "beard"
     ],
     "fields": [


matkoniecz:
though note that alias barber shop is already there...

is Every Door app using alias field?

pkoby:
It is now. I started this PR before Zverik decided to pull aliases as well. I'm fine with closing this if it's unnecessary.

matkoniecz:
then I would say that it was ED bug, duplicating all aliases because app was buggy and was not using aliases for aliasing search terms seems unhelpful

PR #1028, opened 2023-10-12
mark former NSI wikipedia tags as discardable

k-yle:
Closes https://github.com/openstreetmap/iD/issues/9936, Closes https://github.com/openstreetmap/iD/issues/9103

`*:wikipedia` is no longer maintained by NSI, so the tag value becomes out of date when changing presets. This PR will cause the tag to be silently deleted when a user edits the feature.

Related: https://github.com/openstreetmap/iD/issues/9854

### DIFF
--- a/data/discarded.json
+++ b/data/discarded.json
@@ -5,6 +5,10 @@
   "odbl": true,
   "odbl:note": true,
 
+  "operator:wikipedia": true,
+  "brand:wikipedia": true,
+  "network:wikipedia": true,
+
   "tiger:upload_uuid": true,
   "tiger:tlid": true,
   "tiger:source": true,


arch0345:
This should also resolve https://github.com/openstreetmap/iD/issues/9103

matkoniecz:
note that it will start removing also such fields which were added manually

bhousel:
> note that it will start removing also such fields which were added manually

Yeah I was just going to mention this - if a user wants to use the tags, iD will delete them before uploading the changeset, which is maybe not what you want.

tyrasd:
Yeah, this is not really going to work as a solution for the original issue, as it would also remove valid data (e.g. in the case where an object only has a `brand:wikipedia` tag, but not also a `brand:wikidata´).

PR #1030, opened 2023-10-15
Specify crossing:markings in Poland

Zaczero:


Zaczero:
image

I strongly believe that all bicycle crossings are indeed dots. I suppose the issue arises from the interpretation of what constitutes a dot and what constitutes a dash. I believe the mapping of dashes is incorrect, but OSM documentation does not state that clearly.

image

Zaczero:
:heart:

maro-21:
Why this PR? Why is a separate set of hard-coded tags needed just for Poland?

This has resulted in the tag `crossing:markings=zebra:bicolour` not being recognized in Poland, even though it appears in the selection list for nearly every country in the world.

See https://github.com/openstreetmap/id-tagging-schema/issues/2223#issuecomment-4414554991
and https://github.com/openstreetmap/id-tagging-schema/issues/2223

Issue #1016, opened 2023-09-25
Add support for `amenity=bicycle_wash`


mcliquid:
### OSM Tag(s)

amenity=bicycle_wash

### How would you like this tag to see supported?

As a Preset

### Label

Bicycle Wash

### Aliases

Bike Wash, Bike Cleaning, Bicycle Cleaning

### Terms

bike wash, cycle wash, bicycle cleaning station, bike cleaning station, bike rinse area, cycle cleaning spot, bike wash station

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_wash

### Status of the Tag

In Use

### Usage of the tag

74

### Regional Tag?

global

### Further Information

Most of the existences I have seen so far in OSM were entered as `amenity=drinking_water`, which is very misleading at this point. That's why I documented this new tag in the wiki and would be happy about an implementation in the tagging scheme, so that apps like iD or EveryDoor can also benefit from it.
Currently usage is still very low as there has been no documentation on this facility, but this implementation should quickly increase usage as there are a number of these facilities.

PR #1032, opened 2023-10-16
#1016 Create bicycle_wash.json (amenity=bicycle_wash)

mcliquid:
Fix for #1016

First attempt to add the new tagging scheme to iD. This is my very first pull request here, so feel free to share what needs to be improved.

The current icon in the PR is from FontAwesome, as I've seen it that way in other PRs, but there would also be a custom icon that would fit better. How to mark this one?
https://wiki.openstreetmap.org/wiki/File:Bicycle_wash.svg

### DIFF
--- a/data/presets/amenity/bicycle_wash.json
+++ b/data/presets/amenity/bicycle_wash.json
@@ -1,0 +1,34 @@
+{
+    "icon": "fas-bicycle",
+    "fields": [
+        "name",
+		"fee",
+		"payment_multi_fee",
+        "capacity",
+        "access",
+		"operator",
+        "self_service"
+    ],
+    "moreFields": [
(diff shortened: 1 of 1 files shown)


tordans:
@mcliquid ideally ask to add the icon to be added at https://github.com/rapideditor/temaki/issues which makes it easy to integrate here.

tyrasd:
@mcliquid: thank you and no: this PR looks great now. :relaxed:  I'll merge this very soon, so it can be included with the upcoming release of the tagging presets.

tyrasd:
PS: it seems like the proposal already brought a bit of traction/visibility to the tag:

tyrasd (approved):

PR #1035, opened 2023-10-26
Update tree icons

tordans:
The new tree presets from #956 are great to get more detailed trees on the map a lot easier.

However, especially the unspecified tree icon can be improved. @bhousel created a new set of icons in https://github.com/rapideditor/temaki/issues/87 which this PR applies. ([Temaki release 5.6](https://github.com/rapideditor/temaki/blob/main/CHANGELOG.md#560).) 

Preview of icons: https://rapideditor.github.io/temaki/docs/ => "tree"

Let's see how they look on the map in the preview…

## Todos:
- [x] The icons don't show. ==> This requires https://github.com/openstreetmap/iD/pull/9953
- [x] The drop-down icons need Toben updated as well ==> Done

### DIFF
--- a/data/presets/natural/tree.json
+++ b/data/presets/natural/tree.json
@@ -1,5 +1,5 @@
 {
-    "icon": "roentgen-tree",
+    "icon": "temaki-tree_leafless",
     "fields": [
         "leaf_type_singular",
         "leaf_cycle_singular",
--- a/data/fields/leaf_type.json
+++ b/data/fields/leaf_type.json
@@ -11,8 +11,8 @@
         }
     },
     "icons": {
-        "broadleaved": "maki-park",
-        "needleleaved": "roentgen-needleleaved_tree"
+        "broadleaved": "temaki-tree_broadleaved",
+        "needleleaved": "temaki-tree_needleleaved"
     },
     "autoSuggestions": false,
     "customValues": false
--- a/data/presets/natural/tree/_broadleaved.json
+++ b/data/presets/natural/tree/_broadleaved.json
@@ -1,5 +1,5 @@
 {
-    "icon": "maki-park",
+    "icon": "temaki-tree_broadleaved",
     "fields": [
         "leaf_type_singular",
         "leaf_cycle_singular",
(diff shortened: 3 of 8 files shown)


data/presets/natural/tree.json
```diff
@@ -1,5 +1,5 @@
 {
-    "icon": "roentgen-tree",
+    "icon": "temaki-tree_leafless",
```
tyrasd, line 2: I'm not convinced this is a good idea: It does [not look great](https://github.com/rapideditor/temaki/issues/87#issuecomment-1781913162) IMO, could be confused with a dead tree, and is even fundamentally not a good fit for this preset: it is a representation of a tree with `leaf_cycle=deciduous` which is an attribute we do not want to directly depict with this preset.

tyrasd (commented):
I'm not really convinced that using the `temaki-tree_leafless` icon is a good idea: It does [not look great](https://github.com/rapideditor/temaki/issues/87#issuecomment-1781913162) IMO, could be confused with a dead tree, and is just fundamentally not a good fit for the "unspecified tree type" preset: it is actually more a representation of a tree with `leaf_cycle=deciduous` which is an attribute we do not want to directly depict with that preset.

But let's talk about what we could do instead:

* Option A: To be honest, Maki's _park_ icon (and this copy of it) always reminded me of a [stone pine](https://en.wikipedia.org/wiki/Stone_pine) tree (Punis pinea), which interestingly is actually a needleleaved tree (thinking about it… it is not very typical for broadleaved trees to have long exposed branches, maybe that's why). But I do of course also see the resemblance of the characteristic bushy crown of many broadleaved tree types. Maybe it was even originally the intention of Maki's designers to have a tree icon that can fit many different varieties perhaps?!
  
  So, perhaps it could be a better candidate for the generic tree icon than the one from Röntgen (which I, btw, partially chose because it also resembles both typical broadleaved trees as well as vaguely the shape of a cupressus tree). In this case, the only thing missing would be an icon for a "simple" broadleaved tree, maybe something like https://www.svgrepo.com/svg/354697/tree-deciduous-14
* Option B: I would be OK with an icon which has a bushy half and a spiky half. But I would guess that it is not easy to make such an icon look reasonably well.
* Option C: something else?

tordans:
Closing this. I don't see a good way forward for now. We will need to think about the icons more…

PR #1037, opened 2023-10-28
Add Library Book Return preset

arch0345:
Adds a preset for [`amenity=library_dropoff`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dlibrary_dropoff) which was approved back in 2022

### DIFF
--- a/data/presets/amenity/library_dropoff.json
+++ b/data/presets/amenity/library_dropoff.json
@@ -1,0 +1,28 @@
+{
+    "icon": "maki-library",
+    "fields": [
+        "operator",
+        "operator/type"
+    ],
+    "moreFields": [
+        "collection_times",
+        "colour",
+        "ref/isil",
+        "lit"
+    ],
+    "geometry": [
+        "point",
+        "vertex"
+    ],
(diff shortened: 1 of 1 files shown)


matkoniecz (changes requested):
"Library Book Return" sounds like including desk at library where books are normally returned, while https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dlibrary_dropoff defines it as "A remote dropoff location for a library."

tyrasd (approved):

tyrasd:
I think @matkoniecz has a valid point here. While _library [book returns](https://americansecuritycabinets.com/products/book-returns/)_ seems to be the correct term for these drop boxes, there could be confusion with _in person_ book returns ([example usage](https://www.lib.cam.ac.uk/using-library/zero-contact-services/book-returns)). But I think the adjective _remote_ could also be irritating, because these devices are often found just outside or even inside of a particular library building.

Would something like _Book Returns Drop Box_ potentially work, @arch0345 ?

tyrasd (approved):

PR #1038, opened 2023-10-28
Rename `vending=sweets` preset

### LINKED ISSUE #1036: `vending=sweets` preset name misleading
Currently, the preset for `vending=sweets` is named "Snacks Vending Machine". 'Snacks' covers a much broader range than that of what the wiki suggests should be tagged as `vending=sweets`: https://wiki.openstreetmap.org/wiki/Tag:vending%3Dsweets

A new Snack Vending Machine preset that uses [`vending=snacks`](https://taginfo.openstreetmap.org/tags/vending=snacks) could be created (although it doesn't have quite as many uses), while the current preset could be renamed to something like Candy Vending Machine

arch0345:
Also added a preset for `vending=snacks`

Closes #1036

### DIFF
--- a/data/presets/amenity/vending_machine/food/snacks.json
+++ b/data/presets/amenity/vending_machine/food/snacks.json
@@ -1,0 +1,21 @@
+{
+    "icon": "temaki-vending_machine",
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "terms": [
+        "candy",
+        "gum",
+        "chip",
+        "pretzel",
+        "cookie",
+        "cracker"
+    ],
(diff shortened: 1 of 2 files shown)


tyrasd (commented):
Hm… `vending=snacks` does not seem to be [documented](https://wiki.openstreetmap.org/wiki/Key:vending), nor used very much.

If I'm not mistaken, the _"correct"_ way to map this would be to use `vending=food` + `food=snacks`?? I'm very surprised that there is no proper established tag for snack vending machines.

tyrasd (approved):

PR #1039, opened 2023-10-28
Add Mailroom preset

arch0345:
Adds a preset for [`amenity=mailroom`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dmailroom), which was approved back in 2022

### DIFF
--- a/data/presets/amenity/mailroom.json
+++ b/data/presets/amenity/mailroom.json
@@ -1,0 +1,31 @@
+{
+    "icon": "fas-mail-bulk",
+    "fields": [
+        "name",
+        "operator",
+        "address",
+        "building_area",
+        "phone"
+    ],
+    "moreFields": [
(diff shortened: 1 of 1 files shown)


tyrasd:
The number of (currently) only 169 uses is quite a bit lower than what I would typically have in mind as a minimum for considering to add a tag. I see that the tag seems to be (relatively) a bit more common in the US. Maybe we could include it as a regional preset for now?

matkoniecz:
>  wontfix-niche 

Is it intentional that PR remained open if that got wontfixed?

> Maybe we could include it as a regional preset for now?

Or is it unintentional to label it as wontfix?

tyrasd:
let's close the PR for now and reconsider should usage grow a bit more (let's say at least 300+ or so).

PR #1040, opened 2023-10-28
Add `crossing:whistle` field

arch0345:
Adds a field for [`crossing:whistle=*`](https://wiki.openstreetmap.org/wiki/Key:crossing:whistle) for `railway=crossing` and `railway=level_crossing` presets. This key was approved back in June


data/fields/crossing/whistle.json
```diff
@@ -0,0 +1,15 @@
+{
+    "key": "crossing:whistle",
+    "type": "combo",
+    "label": "Train Horn",
```
tyrasd, line 4: IMO it would be important to indicate that this tag should only be used in situations of _non-standard train horn operations_. :thinking:

matkoniecz: @tyrasd What is the plan for dealing with abandoned PRs? Maybe warning author and then closing it? Maybe just closing it if it is stale and tuck on ignored comment?

tordans: The question for us is IMO: If we merge this as is (with the updated data item description), will this improve the data because the field is exposed or will it create unnecessary or even wrong data because the specifics of the definition cannot be conveyed properly in the fields we have.

In the end, only time will tell, but which bet do we want to make?

I would prefer from some mappers who know about train mapping to give their opinion on this…

data/fields/crossing/whistle.json
```diff
@@ -0,0 +1,14 @@
...
+    "strings": {
+        "options": {
+            "yes": "Required",
+            "optional": "No Train Horn",
```
matkoniecz, line 8:
```suggestion
            "optional": "Used If Necessary",
```

matkoniecz: on any crossing horn may be used (and basically always will be used if someone is in front of train)

matkoniecz, line 8:
```suggestion
            "optional": "Only as Needed",
```

matkoniecz: I added also such suggestions

data/fields/crossing/whistle.json
tordans: Those look like good changes to me. We now have a field that cannot be filled out every time … without adding semi false data because there is no "not applicable"-like option. However, that would be more of an issue for apps like SC than iD … as long as users retain themselves from filling the field with the closest option.

Are we ready to merge this now?

matkoniecz:
I will close and reopen to get review page back

data/fields/crossing/whistle.json
matkoniecz: Thanks for improving, I am considering this one as resolved!

data/fields/crossing/whistle.json
1ec5: This appears to have been resolved in 352036651c6ee3c145b9fa84fbf425899b485b12.

matkoniecz:
lets regenerate it

data/fields/crossing/whistle_optional.json
```diff
@@ -0,0 +1,19 @@
...
+        "options": {
+            "yes": "Mandatory",
+            "wayside": "Wayside/Stationary Horn"
+        }
+    },
+    "autoSuggestions": false,
+    "customValues": false,
+    "locationSet": {
+        "include": [
+            "150"
```
1ec5, line 16: This field isn’t showing up in Europe. I tested iD somewhere in France. This is surprising, since iD normally does respect UN&nbsp;49 codes like 150 for fields and presets.

matkoniecz: is it possible that it is related to PR starting in 2023 and possibly not including changes made since then?

tordans: The preview should always use the latest packages AFAIK. 

But we had this discussion to only use country code for this because those are more widely supported by other editors. I think there is a thread in the schema repo or here related to the "Stolpersteine" preset.

data/fields/crossing/whistle_yes.json
```diff
@@ -0,0 +1,20 @@
...
+            "wayside": "Wayside/Stationary Horn"
+        }
+    },
+    "autoSuggestions": false,
+    "customValues": false,
+    "locationSet": {
+        "include": [
+            "ca",
+            "us"
+        ]
```
1ec5, line 18: [This study](https://www.sciencedirect.com/science/article/abs/pii/S0003687021000806) finds that Australia is extremely inconsistent in horn usage but notes that sounding the horn is the default. Unsurprisingly [New Zealand](https://forums.dovetailgames.com/threads/how-often-do-trains-use-the-horn-where-you-live.47372/#post-420044) has the same default.

data/fields/crossing/whistle_optional.json
1ec5, line 1: Should we have a preset worldwide that presents all three options and has no default? The whole reason for splitting this preset is that European mappers were concerned about proliferating tags that are redundant according to regional rules, but other parts of the world might not be as clear.

data/fields/crossing/whistle_optional.json
1ec5, line 16: Japan is [reportedly](https://www.trainorders.com/discussion/read.php?6,575895) another region where horns are usually not required.

1ec5: I’m a bit surprised it’s an issue, since NSI has tons of presets that are scoped to UN M49 regions, mostly 001 (world), but also a bunch of 150. Maybe there’s something different between how iD handles presets versus fields. If we have a third, more generic preset for the rest of the world, then we can limit this preset to a more conservative list of specific countries where we’re sure `optional` would be the default.

matkoniecz:
@arch0345 do you plan on getting it ready for review?

Or should it be closed as abandoned?

(see https://github.com/openstreetmap/id-tagging-schema/pull/1040#issuecomment-3419349725 - I will close it soon, though it can be reopened if anyone is author will be interested)

PR #1041, opened 2023-10-28
Add field for `kerb:approach_aid`

arch0345:
Adds a field for [`kerb:approach_aid`](https://wiki.openstreetmap.org/wiki/Key:kerb:approach_aid) for the Raised Curb preset, which was approved back in April. Each option of this field would benefit from having their own icons similar to those shown on the wiki article so editors have a better understanding of what this field is about.


tordans:
@arch0345 a few thinks I notices:

**Remove tags**
When I switch from raised to lowered, the curb shape tag should be removed, IMO. This should be possible using https://github.com/ideditor/schema-builder#removetags

**Wording**
I associated the term "Concave" with the corner of the curb that I would fall on (which would be a duplicate to `kerb=rolled`), not the one where water would flow. Is there a term to describe that it is about the lower part of the curb? Otherwise…

**Image**
I feel like this field would benefit a lot from field images. That would make the wording less important. Maybe we can suggest two new images to https://rapideditor.github.io/temaki/docs/ that show the two types?

([My testcase](https://pr-1041--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=n5396229957&locale=en&map=20.36/52.47449/13.44705))

arch0345:
> **Wording** I associated the term "Concave" with the corner of the curb that I would fall on (which would be a duplicate to `kerb=rolled`), not the one where water would flow. Is there a term to describe that it is about the lower part of the curb? Otherwise…
> 
> Update: Also, why not use the term "approach aid" for the key? Maybe "approach aid shape"?

I tried to avoid using approach aid as this could be confused for curb cut-outs (`kerb=lowered`).

After doing a bit of research, I found a more technical term for what this key is referring to (https://en.wikipedia.org/wiki/Kassel_kerb). I'll change this to a check field and change the label to just 'Kassel Curb' to avoid any ambiguity.

tyrasd (changes requested):
I did read the wiki page such that this tag would be an attribute of `public_transport=platform` objects. IMO, it does not make much sense to use this tag on "generic" raised curbs, e.g. one at a pedestrian crossing.

Perhaps it would be useful to specify `kerb=raised` as a [`prerequisiteTag`](https://github.com/ideditor/schema-builder/#prerequisitetag) for this field.

matkoniecz:

Do you plan on finishing it? (note: I did no review, no idea whether this PR should be merged once finished)

PR #1042, opened 2023-10-28
Add field for `line_arrangement`

arch0345:
Adds a field for [`line_arrangement=*`](https://wiki.openstreetmap.org/wiki/Key:line_arrangement), which was approved back in July. This PR also adds presets for [`power=insulator`](https://wiki.openstreetmap.org/wiki/Tag:power%3Dinsulator) and [`power=terminal`](https://wiki.openstreetmap.org/wiki/Tag:power%3Dterminal)


data/presets/power/insulator.json
```diff
@@ -0,0 +1,20 @@
+{
+    "icon": "temaki-power",
+    "fields": [
+        "ref",
+        "operator",
+        "line_attachment"
+    ],
+    "moreFields": [
+        "height",
+        "line_arrangement",
```
tyrasd, line 10: FWIW,  the combination `power=insulator` + `line_arrangement`  has (currently) zero uses in OSM: https://overpass-turbo.eu/s/1DGp

:shrug:

data/fields/line_arrangement.json
```diff
@@ -0,0 +1,17 @@
+{
+    "key": "line_arrangement",
+    "type": "combo",
+    "strings": {
+        "options": {
+            "horizontal": "Horizontal",
+            "semi_horizontal": "Semi Horizontal",
+            "vertical": "Vertical",
+            "semi_vertical": "semi-vertical",
```
tyrasd, line 9:
```suggestion
            "semi_vertical": "Semi Vertical",
```

data/fields/line_arrangement.json
```diff
@@ -0,0 +1,17 @@
...
+            "square": "Square",
+            "triangular": "Triangular",
+            "delta": "Delta",
```
tyrasd, line 12: As both of these options are triangular shapes, I think we should clarify the difference:

```suggestion
            "triangular": "Triangular (scalene)",
            "delta": "Triangular (isosceles)",
```

tyrasd (commented):
Please find some inline comments below.

Generally speaking, I'm struggling a little bit to see the benefit these domain-specific attributes/tags provide in a general purpose OSM editor. Especially when it comes to translatable values there might also be downsides:

* for volunteers translating on transifex (which I assume are typically not trained electrical engineers), it takes valuable time and effort to fully comprehend these tagging schemas and potentially even more time and effort to find proper translations
* it would be easy to introduce a few small mistakes when translating those terms, which could result in a large number of mappers inadvertently choosing wrong tag values
* if the fields were left "raw", it would be implicitly a bit more obvious that one should consult the wiki before mapping these attributes

What do you think?

data/fields/line_management.json
```diff
@@ -1,6 +1,17 @@
...
     "type": "combo",
+    "strings": {
+        "options": {
+            "straight": "Straight",
+            "branch": "Branch",
+            "split": "Split",
+            "transpose": "Transpose",
+            "cross": "Cross",
+            "transition": "Transition",
+            "termination": "Termination"
```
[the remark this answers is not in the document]
flacombe: "straight": "Line continues", no need to have a single line

data/presets/power/insulator.json
flacombe: Yes but it's a valid combination

matkoniecz:
> @[arch0345](https://github.com/arch0345) arch0345 marked this pull request as draft [2 years ago](https://github.com/openstreetmap/id-tagging-schema/pull/1042#event-11015642198)

do you plan on continuing it or should it be closed as an abandoned PR?

PR #1044, opened 2023-10-28
Refactor crossing presets

arch0345:
This PR is an attempt to simplify crossing presets. As explained in [this thread](https://osmus.slack.com/archives/CBK3JLUJU/p1697130675944029?thread_ts=1697105393.431999&cid=CBK3JLUJU) on the [OSMUS Slack](https://openstreetmap.us/slack), the current options for crossings can be quite confusing.

I've made specific crossing presets with `crossing=*` tags unsearachable, with the only searchable options now being Crossing, Pedestrian Crossing, Cycle Crossing, and Cycle & Foot Crossing. I've also introduced a preset for `path=crossing`

### DIFF
--- a/data/fields/crossing/signals.json
+++ b/data/fields/crossing/signals.json
@@ -1,0 +1,5 @@
+{
+    "key": "crossing:signals",
+    "type": "check",
+    "label": "Crosswalk Signals"
+}
--- a/data/presets/highway/crossing.json
+++ b/data/presets/highway/crossing.json
@@ -1,9 +1,9 @@
 {
     "fields": [
-        "crossing",
         "tactile_paving",
         "crossing/island",
         "{@templates/crossing/markings}",
+        "crossing/signals",
         "crossing_raised"
     ],
     "moreFields": [
--- a/data/presets/highway/crossing/_traffic_signals.json
+++ b/data/presets/highway/crossing/_traffic_signals.json
@@ -2,27 +2,25 @@
     "icon": "temaki-railway_signals",
     "fields": [
         "crossing",
-        "surface",
         "tactile_paving",
         "crossing/island",
         "{@templates/crossing/markings}",
         "crossing_raised",
         "button_operated",
         "traffic_signals/sound",
-        "traffic_signals/vibration",
-        "access"
+        "traffic_signals/vibration"
     ],
     "moreFields": [
+        "kerb",
         "traffic_signals/arrow",
         "traffic_signals/countdown",
         "traffic_signals/minimap"
     ],
     "geometry": [
-        "line"
+        "vertex"
     ],
     "tags": {
-        "highway": "footway",
-        "footway": "crossing",
+        "highway": "crossing",
         "crossing": "traffic_signals"
     },
     "reference": {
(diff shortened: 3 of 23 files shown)


tordans:
Note to self: Read up on what the matchScore does when the preset is unsearchable? I assume it can be removed...

arch0345:
I believe `matchScore` is also used to tell which preset name should show up in the feature menu when two presets have similar tags (ie Cycle Crossing and Cycle & Foot Crossing)

1ec5:
For context, every now and then a less experienced mapper asks what the difference is between a “Pedestrian Crossing” and a “Marked Crossing”, especially since there’s also “Pedestrian Crossing With Traffic Signals”.

This change will continue the progression in #837 and #889. Reducing the number of “Crossing” presets will allows mappers to rely on the recent preset list without each slight variant of a crossing pushing another preset off the list.

For those who have become accustomed to the more specific presets, the loss of these presets may add to the number of clicks required to fully tag a crossing, but I don’t expect it to lead to a noticeable decline in tagging crossing details, because iD already has rendering rules and validation rules that encourage applying crossing subkeys. There has never been a fully complement of specific presets for styles of railway crossings, but `crossing:*` is very commonly tagged on them, maybe even more commonly than on pedestrian and cyclist crossings.

With this change, it’ll also be more straightforward to create regional variants of a smaller number of presets. #1030 only had to touch some templates, but some regions have well-known names for specific crossing configurations that go well beyond markings. It’s always been difficult to support these configurations because of the sheer number of crossing presets.

tordans:
@Bonkles I think I remember rapid has a shortcut to cycle through crossing presets. (Or maybe this is https://github.com/facebook/Rapid/issues/1154?) — Will this update to the preset affect this feature?

bhousel:
> Will this update to the preset affect this feature?

Yes this is  https://github.com/facebook/Rapid/issues/1154 / https://github.com/facebook/Rapid/pull/1160
It is still under development, but when finished it will cycle through whatever the presets are, as identified by their presetID.

If the presets change, we can always override them, per https://github.com/facebook/Rapid/issues/1182 - don't worry, I'm paying attention and thinking ahead 👍

tordans:
@bhousel 

> when finished it will cycle through whatever the presets are, as identified by their presetID.

Great, [it looks like](https://github.com/facebook/Rapid/pull/1160/files#diff-a73f3960a744eb36ff0328d1fa96a7fe838495c7620caa52d707c72522a167dbR21-R44) those can be specified regardless of their searchable status so the changes here will not conflict.

tordans:
FYI the next iD Community Meeting is earmarked to talk about the crossing presets https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2024-02-28

Can you join @arch0345 @1ec5 so we can maybe get topic untangled?

tordans:
One part of this PR was to add the `crossing:signals` to the presets which I did in a follow up PR https://github.com/openstreetmap/id-tagging-schema/pull/1203 of the now merged crossing refactoring. ==> Please have a look at https://github.com/openstreetmap/id-tagging-schema/pull/1203 and the open questions given there.

PR #1047, opened 2023-10-31
Change "Draft Beer" to "Sold Beer Brands"

mcliquid:
See also: https://wiki.openstreetmap.org/wiki/Key:brewery

### DIFF
--- a/data/fields/brewery.json
+++ b/data/fields/brewery.json
@@ -1,8 +1,12 @@
 {
     "key": "brewery",
     "type": "semiCombo",
-    "label": "Draft Beers",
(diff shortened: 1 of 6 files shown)


tyrasd (approved):
Thanks :+1: 

One remark, though: translations are fetched from transifex, and in general any of your changes in the `dist` directory would be overridden by the automated build scripts. Because of that, I have now adjusted the respective changes in this PR. See the [contributing](https://github.com/openstreetmap/id-tagging-schema/blob/main/CONTRIBUTING.md#Translating) guide about how to update the translation.

PR #1059, opened 2023-11-20
Add switchgear preset

arch0345:


data/presets/power/switchgear.json
```diff
@@ -0,0 +1,14 @@
+{
+    "icon": "temaki-power",
```
tyrasd, line 2: what about an icon like `temaki-power_switch` which would represent the _switching_ aspect of these facilities?

PR #1068, opened 2023-12-07
add presets for `cycleway=traffic_island`

### LINKED ISSUE #950: Add 2 more presets for refuge islands
### OSM Tag(s)

cycleway=traffic_island

As a Preset

### Label

"Cycle & Foot Refuge Island" and "Cycle Refuge Island"

### Aliases

Shared Path Refuge Island; Shared Path Traffic Island etc.

### Terms

_No response_

### Link to OSM Wiki page

https://osm.wiki/Tag:cycleway=traffic_island + https://osm.wiki/Tag:footway=traffic_island

### Status of the Tag

Approved

### Usage of the tag

655

### Replaces other Tag?

_No response_

### Regional Tag?

_No response_

### Further Information

This is a table of the current presets. Besides from the (contentious?) crossing presets, the obvious outlier is the "refuge island" preset, which has no equivilant for cycleways or shared path. 

|                    | Foot                    | Cycle             | Cycle & Foot               |
| ------------------ | ----------------------- | ----------------- | -------------------------- |
| **Roadside Path**      | ✅ Sidewalk             | ✅ Cycle Path     | ✅ Cycle & Foot Path     |
| **Generic Crossing**   | ✅ Pedestrian Crossing  | ✅ Cycle Crossing | ✅ Cycle & Foot Crossing |
| **Specific Crossings** | ✅ Unmarked Crossing<br />✅ Marked Crossing<br />✅ Uncontrolled Crossing<br />✅ Zebra Crossing<br />✅ Crossing With Traffic Signals | ✅ Unmarked Cycle Crossing<br />✅ Marked Cycle Crossing<br />✅ Uncontrolled Cycle Crossing<br />❌<br />✅ Cycle Crossing With Traffic Signals | ❌<br />❌<br />❌<br />❌<br />❌ |
| **Traffic Island** | ✅ Refuge Island  | ❌ | ❌ |

k-yle:
Closes #950

No one commented on that issue so I assume this is non-controversial... 

Here's the updated table, the 3 entries marked as `🆕` are the ones added in this PR:

|                        | Foot                    | Cycle             | Cycle & Foot (International) | Cycle & Foot (EU)       |
| ---------------------- | ----------------------- | ----------------- | ---------------------------- | ------------------------|
| **Roadside Path**      | ✅ Sidewalk             | ✅ Cycle Path      | ✅ Cycle & Foot Path         | ✅ Cycle & Foot Path     |
| **Generic Crossing**   | ✅ Pedestrian Crossing  | ✅ Cycle Crossing  | ✅ Cycle & Foot Crossing     | ✅ Cycle & Foot Crossing |
| **Specific Crossings** | ✅ Unmarked Crossing<br />✅ Marked Crossing<br />✅ Uncontrolled Crossing<br />✅ Zebra Crossing<br />✅ Crossing With Traffic Signals | ✅ Unmarked Cycle Crossing<br />✅ Marked Cycle Crossing<br />✅ Uncontrolled Cycle Crossing<br />❌ <s>Zebra Crossing</s><br />✅ Cycle Crossing With Traffic Signals | ❌ <s>Unmarked Crossing</s><br />❌ <s>Marked Crossing</s><br />❌ <s>Uncontrolled Crossing</s><br />❌ <s>Toucan Crossing</s><br />❌ <s>Crossing With Traffic Signals</s> | ✅ Unmarked Cycle & Foot Crossing<br />✅ Marked Crossing<br />✅ Marked Cycle & Foot Crossing<br />❌ <s>Toucan Crossing</s><br />✅ Cycle & Foot Crossing With Pedestrian Signals |
| **Traffic Island**     | ✅ Refuge Island        | 🆕 Cycle Refuge Island | 🆕 Cycle & Foot Refuge Island | 🆕 Cycle & Foot Refuge Island |

### DIFF
--- a/data/presets/highway/cycleway/traffic_island.json
+++ b/data/presets/highway/cycleway/traffic_island.json
@@ -1,14 +1,14 @@
 {
-    "icon": "temaki-pedestrian",
+    "icon": "fas-biking",
     "geometry": [
         "line"
     ],
     "tags": {
-        "highway": "footway",
-        "footway": "traffic_island"
+        "highway": "cycleway",
+        "cycleway": "traffic_island"
     },
     "reference": {
-        "key": "footway",
+        "key": "cycleway",
         "value": "traffic_island"
     },
     "terms": [
@@ -19,5 +19,5 @@
         "Traffic Island",
         "Pedestrian Island"
     ],
-    "name": "Refuge Island"
+    "name": "Cycle Refuge Island"
 }
--- a/data/presets/highway/cycleway/traffic_island_shared.json
+++ b/data/presets/highway/cycleway/traffic_island_shared.json
@@ -1,0 +1,35 @@
+{
+    "locationSet": {
+        "exclude": [
+            "fr",
+            "lt",
+            "pl",
+            "de",
+            "il",
+            "ps"
+        ]
+    },
+    "icon": "temaki-pedestrian_and_cyclist",
+    "geometry": [
+        "line"
+    ],
+    "tags": {
+        "highway": "cycleway",
+        "cycleway": "traffic_island",
+        "foot": "designated",
+        "bicycle": "designated"
+    },
+    "reference": {
+        "key": "cycleway",
+        "value": "traffic_island"
+    },
+    "terms": [
+        "crossing island"
+    ],
+    "aliases": [
+        "Crosswalk Island",
+        "Traffic Island",
+        "Pedestrian Island"
+    ],
+    "name": "Cycle & Foot Refuge Island"
+}
--- a/data/presets/highway/path/traffic_island_shared.json
+++ b/data/presets/highway/path/traffic_island_shared.json
@@ -1,0 +1,30 @@
+{
+    "locationSet": {
+        "include": [
+            "fr",
+            "lt",
+            "pl",
+            "de",
+            "il",
+            "ps"
+        ]
+    },
+    "icon": "temaki-pedestrian_and_cyclist",
+    "geometry": [
+        "line"
+    ],
+    "tags": {
+        "highway": "path",
+        "path": "traffic_island",
+        "foot": "designated",
+        "bicycle": "designated"
+    },
+    "reference": {
+        "key": "path",
+        "value": "traffic_island"
+    },
+    "fields": [
+        "{highway/cycleway/traffic_island_shared}"
+    ],
+    "name": "{highway/cycleway/traffic_island_shared}"
+}


data/presets/highway/cycleway/traffic_island_shared.json
```diff
@@ -0,0 +1,36 @@
...
+    "locationSet": {
+        "exclude": [
+            "fr",
+            "lt",
+            "pl",
+            "de",
+            "il",
+            "ps"
+        ]
+    }
```
tordans, line 35: I think this should be moved to the top to sync it with data/presets/highway/cycleway/bicycle_foot.json.

tordans (commented):
I think we should also add a (unsearchable) preset for `path=traffic_island` https://wiki.openstreetmap.org/wiki/Tag:path=traffic_island which has low usage ATM but it is a pain to hot have all those crossing presents on all common highway classes and for some tagging the `hw=path+bicycle=designated+foot=designated` is very common.

tordans:
@k-yle I think now that #1201 is merged it would be worth restarting this to add the missing presets. Do you want to look into it?

k-yle:
@tordans done, and I've updated the table in the PR description.

Dimitar5555:
Having **both** `foot=designated` and `bicycle=designated` seems to break the preset for `Cycle & Foot Traffic Island`.
|Both |`foot=designated` |`bicycle=designated` |
|--|--|--|
|image | image | image |

data/presets/highway/path/traffic_island_shared.json
```diff
@@ -0,0 +1,31 @@
...
+            "il",
+            "ps"
+        ]
+    },
+    "icon": "temaki-pedestrian_and_cyclist",
+    "geometry": [
+        "line"
+    ],
+    "tags": {
+        "path": "traffic_island"
```
tordans, line 17: Same as above; taken from https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/highway/path/bicycle_foot.json#L19-L23

```suggestion
        "path": "traffic_island",
        "foot": "designated",
        "bicycle": "designated"
```

data/presets/highway/cycleway/traffic_island_shared.json
```diff
@@ -0,0 +1,36 @@
...
+            "il",
+            "ps"
+        ]
+    },
+    "icon": "temaki-pedestrian_and_cyclist",
+    "geometry": [
+        "line"
+    ],
+    "tags": {
+        "cycleway": "traffic_island"
```
tordans, line 17: This need to be more precise so we don't trigger on every `cw=traffic_island`, just like https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/highway/cycleway/bicycle_foot.json#L27-L30
```suggestion
        "cycleway": "traffic_island",
        "foot": "designated"
```

data/presets/highway/path/traffic_island_shared.json
```diff
@@ -0,0 +1,31 @@
...
+    },
+    "addTags": {
+        "highway": "path",
+        "path": "traffic_island"
+    },
```
tordans, line 22: Same as above, IMO

```suggestion
    "removeTags": {
        "crossing:island": "*",
    },
```

data/presets/highway/cycleway/traffic_island_shared.json
```diff
@@ -0,0 +1,36 @@
...
+    },
+    "addTags": {
+        "highway": "cycleway",
+        "cycleway": "traffic_island"
+    },
```
tordans, line 22: **`addTags`**
I don't think that we need `addTags` because all that we need can be part of `tags`.
Docs: https://github.com/ideditor/schema-builder?tab=readme-ov-file#addtags

The `cycleway/bicycle_foot` preset uses `addTags` to force the `bicycle=designated` but I am not sure we need to force this, see #1411.

**`removeTags`**
The docs at https://wiki.openstreetmap.org/wiki/Tag%3Acycleway%3Dtraffic_island and https://wiki.openstreetmap.org/wiki/Tag:footway%3Dtraffic_island go into details on the `crossing:island`. My understand is, that it is likely that a way has a `crossing:island` when it is migrated to a `*=traffic_island` and in this case, we want to remove all `crossing:island` because those only belong on `crossing=*` ways.

```suggestion
    "removeTags": {
        "crossing:island": "*",
    },
```

data/presets/highway/path/traffic_island_shared.json
```diff
@@ -0,0 +1,31 @@
...
+    "reference": {
+        "key": "path",
+        "value": "traffic_island"
```
tordans, line 25: We should add this wiki page as a stub at least…

data/presets/highway/cycleway/traffic_island.json
```diff
@@ -0,0 +1,26 @@
...
+    "geometry": [
+        "line"
+    ],
+    "tags": {
+        "cycleway": "traffic_island"
+    },
+    "addTags": {
+        "highway": "cycleway",
+        "cycleway": "traffic_island"
+    },
```
tordans, line 12: I don't think this is needed, the `tags` is enough. We could move the `hw=cycleway` to `tags` but the again this should always match with `cycleway=*`.

```suggestion
```

tordans (changes requested):
Thanks for this, I think it is great to have all those presets for alle the way types ready.

I left a few inline comments.

In addition to those:

a. I think we should update the  footway preset to follow the same "rythmn" that these new files have and add the changes from here there as well https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/highway/footway/traffic_island.json#L9-L12

b. Did anyone check the wiki data entries already?

c. I think we schon start cleaning up the wiki pages. Ideally, we would have a good wiki page setup, but I am not sure if we should make this a requirement for this PR.
- https://wiki.openstreetmap.org/wiki/Tag:footway%3Dtraffic_island looks good
- https://wiki.openstreetmap.org/wiki/Tag%3Acycleway%3Dtraffic_island is there but a bit of a stub and not that well structured
- https://wiki.openstreetmap.org/wiki/Tag%3Apath%3Dtraffic_island is missing

The think that bothers me is, that most of the "how to" on those pages is the same, but we don't have a good way to write this into wiki pages that don't share a primary key. We could just make up a wiki page "Tag:*=traffic_sign" and redirect the other 3 there…
Or we could write the cycleway+path pages with a strong reference to the footway page…

Anyone up to tackling this?

data/presets/highway/path/traffic_island_shared.json
k-yle: I'm a bit confused here. Is `crossing:island` meant to be used on ways with `*=traffic_island`? [The wiki page](https://osm.wiki/Key:crossing:island)doesn't mention this. 

There are also [344 ways](https://overpass-turbo.eu/s/1VZg) with `*=traffic_island`+`crossing:island=no`, which seems like a contradiction. Most that I checked were the result of crossing ways being split, and the crossing tags being accidentally carried over to the island.

data/presets/highway/cycleway/traffic_island_shared.json
k-yle: _[please ignore previous deleted comment 🤦]_

I think these `removeTags` belong on the crossing presets, not the traffic island presets right? because we want `crossing:*` to be removed when changing from crossing to traffic_island

data/presets/highway/path/traffic_island_shared.json
k-yle: lets discuss this one in [the other thread](#discussion_r1894376539)

data/presets/highway/path/traffic_island_shared.json
k-yle: done. for some reason the wiki UI for linking data items to pages is completely broken, but I think i've managed it...

data/presets/highway/cycleway/traffic_island_shared.json
k-yle: regarding `bicycle=designated` should we keep it for consistency and change it later if #1411 is accepted? 

Otherwise it might be confusing if only the refuge island preset matches, but not the connected paths. [Example here](https://pr-1068--ideditor-presets-preview.netlify.app/id/dist/#id=w1289722394).

tordans: @k-yle I agree, lets keep the presets in sync and update them later, if needed.

tordans: > [please ignore previous deleted comment 🤦]

JFY thanks for this, I lead the comments via mail and this helps because updates don't get emailed :)

tordans: > I think these removeTags belong on the crossing presets, not the traffic island presets right? because we want crossing:* to be removed when changing from crossing to traffic_island

Let's leave this out of V1 of this addition and maybe add it later… It makes this PR more complex that it has to be.

The use case I was thinking of is… Users use `crossing:island=yes` on ways when they don't split of the `traffic_island` part of the way. Then, at some point, users want to add details and split off the `traffic_island` part. Which means that a ways that already has `crossing:island=yes` gets transformed into a `…=traffic_island` which would be when this `removeTags` gets triggered (right?).

But again, this should be done separately…

tordans (approved):
I think this looks great now.

PR #1074, opened 2023-12-19
Change name `Traffic island` -> `Traffic calming (island)`

ireun:
Hello. 

I disagree with the name 'Traffic Island' as it does not include any reference to `traffic_calming`.  The tag `traffic_calming` should only be used only in places where the structure somehow makes the cars go slower - not every 'traffic island' does that.

https://wiki.openstreetmap.org/wiki/Key:traffic_calming

People are making mistakes due that name. Practical example of what issue this raises is that OsmAnd navigation has a feature to alert drivers of incoming `traffic_calming` nodes on their route. In this situation it would alert people in places where they should not be alerted.

Of course I'm open for any other name suggestion.

### DIFF
--- a/data/fields/traffic_calming.json
+++ b/data/fields/traffic_calming.json
@@ -18,7 +18,7 @@
             "dip": "Dip",
             "double_dip": "Double Dip",
             "hump": "Speed Hump",
-            "island": "Traffic calming (island)",
+            "island": "Traffic Calming Island",
             "mini_bumps": "Mini Speed Bumps",
             "rumble_strip": "Rumble Strip",
             "table": "Speed Table"
--- a/data/fields/traffic_calming.json
+++ b/data/fields/traffic_calming.json
@@ -18,7 +18,7 @@
             "dip": "Dip",
             "double_dip": "Double Dip",
             "hump": "Speed Hump",
-            "island": "Traffic Island",
+            "island": "Traffic calming (island)",
             "mini_bumps": "Mini Speed Bumps",
             "rumble_strip": "Rumble Strip",
             "table": "Speed Table"
--- a/data/presets/traffic_calming/island.json
+++ b/data/presets/traffic_calming/island.json
@@ -16,5 +16,5 @@
     "tags": {
         "traffic_calming": "island"
     },
-    "name": "Traffic Island"
+    "name": "Traffic Calming Island"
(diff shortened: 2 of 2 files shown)


1ec5:
Thanks for bringing this up. It always bothered me that traffic islands were lumped under `traffic_calming`, since the effect is usually the opposite as you point out. If this tag is only meant for a subset of traffic islands that calm traffic, do you have any reference materials on this kind of device (such as traffic engineering manuals) that could inform our naming of the preset?

In the absence of an established term for these devices, I guess “Traffic Island Obstacle” would get the point across. I’d probably steer clear of a parenthetical, because translations of that might not capture the meaning in the way that you intend.

Also, do we know if the tag has already been [skunked](https://wiki.openstreetmap.org/wiki/Skunked_tags) due to this misnomer? The [`traffic_calming=painted_island`](https://wiki.openstreetmap.org/wiki/Tag:traffic_calming%3Dpainted_island) tag seems to suggest that some have viewed `island` as applying to all traffic islands, such as refuge islands, to the extent of putting other non-calming features under the same key.

ireun:
>  do you have any reference materials
Sadly I don't, i live in non-English speaking country so I don't even know where I could find such thing. :)

> since the effect is usually the opposite as you point out
Indeed, traffic islands whose sole purpose is to direct traffic to flow better do not slow down traffic, they do the opposite. 
To calm the traffic usually the turn would be sharp. [Example](https://commons.wikimedia.org/wiki/File:Reduced_radius_corner.jpg).

I think that for traffic islands that do not impose 'calming' effect `area:highway=traffic_island` should be used instead. Despite it's from an [inactive proposal](https://wiki.openstreetmap.org/wiki/Proposal:Street_area) it still has almost 18k usages.

> because translations of that might not capture the meaning in the way that you intend
Isn't what context informations should be used for? 
image

But I'm still open in that regard, maybe just 'Traffic calming island' would se sufficient.

tyrasd (approved):
Let's go with _Traffic Calming Island_, as that's also how the [wiki calls these features](https://wiki.openstreetmap.org/w/index.php?title=Tag:traffic_calming%3Disland&oldid=2632839).

1ec5:
Oh good call, that seems to be what they’re called in the wild too.

PR #1077, opened 2023-12-28
kerb: Update name of general preset

tordans:
The preset that represents an "unspecified" kerb looks the same as the more specified presets.

We started using some term to signal that this is the general preset but there are more precise once, but I could not find it just now… — Do we have a way to add wording to those general presets, already?

We could also change the icon, but I don't see anything obvious and easy to do.

This is what it looks ATM.

image

### DIFF
--- a/data/presets/barrier/kerb.json
+++ b/data/presets/barrier/kerb.json
@@ -17,5 +17,5 @@
         "barrier": "kerb"
     },
     "matchScore": 0.5,
-    "name": "Curb"
+    "name": "Curb (unspecified)"
 }


tyrasd:
> Do we have a way to add wording to those general presets, already?

We used `… (Unspecified Type)` in a few places (e.g. `barrier/-yes.json`).

In general, in such situations it would be better to create a dedicated icon for the generic case, as that would also be show up distinctively on the map. Maybe the existing icon with a question mark instead of the arrow would be an option?

tyrasd:
ok, thinking about it again: I'd say that adding a "suffix" to the preset name is not adding too much value: I would say that most data consumers treat unspecified kerbs as full height ones, and mappers are asked to choose a specific type[^1] in the preset field. Also, we currently only use the `… (Unspecified Type)` for unsearchable presets of tags which are generally discouraged to use (like the mentioned `barrier=yes`).

In general, a dedicated icon would be the optimal solution. I've opened https://github.com/rapideditor/temaki/pull/90 for that. Feedback welcome. :wink: 


[^1]: what would help here is that the `kerb` field would have translatable names and show the icons for the documented types, though.

tyrasd:
89139bb  switches to the new temaki icon. I'd say this addresses the original issue sufficiently well that we can close this PR. :blush:

PR #1083, opened 2024-01-19
Add `ref` and `railway:track_ref` fields to `railway=rail` and `=switch` presets

gy-mate:
- `railway=rail`
	- Create [`ref=*`](https://wiki.openstreetmap.org/wiki/OpenRailwayMap/Tagging#Tracks) (Line Number) and [`railway:track_ref=*`](https://wiki.openstreetmap.org/wiki/Key:railway:track_ref) (Track Number) fields
	- Add these fields to its preset
- `railway=switch`
	- Create [`ref=*`](https://wiki.openstreetmap.org/wiki/Tag:railway%3Dswitch#:~:text=switch%20number) (Switch Number) field
	- Add this field to its preset

### DIFF
--- a/data/presets/railway/rail.json
+++ b/data/presets/railway/rail.json
@@ -2,12 +2,12 @@
     "icon": "temaki-railway_track",
     "fields": [
         "{railway/light_rail}"
-        "ref_rail"
+        "railway/track_ref"
     ],
     "moreFields": [
         "{railway/light_rail}",
         "highspeed"
-        "railway/track_ref"
+        "ref_rail"
     ],
     "geometry": [
         "line"
--- a/data/presets/railway/rail.json
+++ b/data/presets/railway/rail.json
@@ -2,10 +2,12 @@
     "icon": "temaki-railway_track",
     "fields": [
         "{railway/light_rail}"
+        "ref"
     ],
     "moreFields": [
         "{railway/light_rail}",
         "highspeed"
+        "railway:track_ref"
     ],
     "geometry": [
         "line"
--- a/data/presets/railway/rail.json
+++ b/data/presets/railway/rail.json
@@ -2,12 +2,12 @@
     "icon": "temaki-railway_track",
     "fields": [
         "{railway/light_rail}"
-        "ref"
+        "ref_rail"
     ],
     "moreFields": [
         "{railway/light_rail}",
         "highspeed"
-        "railway:track_ref"
+        "railway/track_ref"
     ],
     "geometry": [
         "line"
--- a/data/presets/railway/rail.json
+++ b/data/presets/railway/rail.json
@@ -1,12 +1,12 @@
 {
     "icon": "temaki-railway_track",
     "fields": [
-        "{railway/light_rail}"
+        "{railway/light_rail}",
         "railway/track_ref"
     ],
     "moreFields": [
         "{railway/light_rail}",
-        "highspeed"
+        "highspeed",
         "ref_rail"
     ],
     "geometry": [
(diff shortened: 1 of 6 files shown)


data/presets/railway/switch.json
```diff
@@ -1,5 +1,8 @@
 {
     "icon": "temaki-junction",
+    "fields": [
+        "ref_switch"
+    ],
```
tordans, line 5: Side note: 

It looks like `railway:switch` is as commonly used (numbers: https://taginfo.openstreetmap.org/tags/railway%3Dswitch#combinations) as `ref`. Maybe something for a follow up PR? 

https://wiki.openstreetmap.org/wiki/DE:Tag:railway%3Dswitch

data/presets/railway/rail.json
```diff
@@ -2,10 +2,12 @@
     "icon": "temaki-railway_track",
     "fields": [
         "{railway/light_rail}"
+        "ref_rail"
     ],
     "moreFields": [
         "{railway/light_rail}",
         "highspeed"
+        "railway/track_ref"
```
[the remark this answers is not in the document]
gy-mate: @tordans You're totally right, thanks for the suggestion! Fixed in [8d637b7](https://github.com/openstreetmap/id-tagging-schema/pull/1083/commits/8d637b7d10b25e4017cc1d9b295f4426eb0370c3).

data/presets/railway/switch.json
gy-mate: @tordans Good idea, thanks! Follow-up PR: #1084

tordans (approved):
LGTM But disclaimer: I know litte about railway-mappging :-).

data/presets/railway/rail.json
```diff
@@ -2,10 +2,12 @@
     "icon": "temaki-railway_track",
     "fields": [
         "{railway/light_rail}"
```
tyrasd, line 4: syntax error (missing comma at end of the line here)

```suggestion
        "{railway/light_rail}",
```

data/presets/railway/rail.json
```diff
@@ -2,10 +2,12 @@
...
+        "railway/track_ref"
     ],
     "moreFields": [
         "{railway/light_rail}",
         "highspeed"
```
tyrasd, line 9:
```suggestion
        "highspeed",
```

tyrasd (commented):
Apart from a few minor syntax errors (see below), this looks fine for me as well.

Only one question: the `ref` field seems to be quite a bit more common on railway=rail objects (see [taginfo](https://taginfo.openstreetmap.org/tags/railway=rail#combinations)). What was your reasoning behind putting the `ref` field into `moreFields`? Are the track numbers usually easier to map?

tyrasd:
One more suggestion: As switches only have a single reference number, it would be sufficient to use the generic field for the `ref` tag, wouldn't it? This would have the benefit of being slightly less effort to translate and maintain. Or is _Switch Number_ a technical term?

gy-mate:
> What was your reasoning behind putting the `ref` field into `moreFields`? Are the track numbers usually easier to map?

@tyrasd Yeah, @tordans [suggested](https://github.com/openstreetmap/id-tagging-schema/pull/1083#discussion_r1458608101) this and I agree with him:

> `railway:track_ref` would be commonly known but `ref` (`ref_rail`) is something that only external data or professionals would know

tyrasd:
Ok, that makes sense now. :+1:

gy-mate:
> As switches only have a single reference number, it would be sufficient to use the generic field for the `ref` tag, wouldn't it? This would have the benefit of being slightly less effort to translate and maintain. Or is _Switch Number_ a technical term?

@tyrasd Well it's not really an essential technical term. You're right, there is no need to overcomplicate, it's still obvious without `ref_switch`. Reverted those commits.

tyrasd (approved):

Issue #1089, opened 2024-01-28
Thai restaurant icon is wrong


louwers:
The icon for a Thai restaurant is a bowl with chopsticks, but the Thai don’t use chopsticks.

It seems to be defined here: https://github.com/openstreetmap/id-tagging-schema/blob/1aedde053e3dcfd0a4cd7e30a0a35d6c1a878fba/data/presets/amenity/restaurant/thai.json#L24

The best option may be be an icon with a spoon and fork? But just `maki-restaurant` is already better.

louwers:
Related? #41

PR #1090, opened 2024-01-28
Fix icon Thai restaurant

louwers:
Closes https://github.com/openstreetmap/id-tagging-schema/issues/1089

Thai do not use chopsticks. Change icon to a spoon instead. Still trying to find a fork & spoon icon. But this is infinitely better in the meantime.

https://en.wikipedia.org/wiki/Thai_cuisine#Serving


tyrasd:
Thx!

PR #1092, opened 2024-01-30
add option "only" to drive_through field

michalgwo:
Added yes, no, only options for [drive_through](https://wiki.openstreetmap.org/wiki/Key:drive_through) key.

### DIFF
--- a/data/fields/drive_through.json
+++ b/data/fields/drive_through.json
@@ -10,6 +10,6 @@
             "only": "Drive Through Only"
         }
     },
-    "autoSuggestions" : false,
-    "customValues" : false
+    "autoSuggestions": false,
+    "customValues": false
 }
--- a/data/fields/drive_through.json
+++ b/data/fields/drive_through.json
@@ -1,5 +1,15 @@
 {
     "key": "drive_through",
-    "type": "check",
-    "label": "Drive-Through"
+    "type": "combo",
+    "label": "Drive-Through",
+    "placeholder": "Yes, No, Drive Through Only...",
+    "strings": {
+        "options": {
+            "yes": "Yes",
+            "no": "No",
+            "only": "Drive Through Only"
+        }
+    },
+    "autoSuggestions" : false,
+    "customValues" : false
 }


tyrasd (commented):
I'm a bit unsure about this change. Yes, the value is documented on the wiki, but usage is [very rare](https://taginfo.openstreetmap.org/keys/drive_through#values) compared to the usual yes/no cases. Adding the third case makes the field significantly less easy to use compared to the simple checkbox.[^1]

Also, at least in some situations, this tag seems to have the the same downsides as the [`<status>=yes`](https://wiki.openstreetmap.org/wiki/Comparison_of_life_cycle_concepts#%3Cstatus%3E_=_yes) life cycle tagging approach, which is AFAICT generally discouraged. For example: on a general purpose map, one would not be able to distinguish a drive-tru-only supermarket from a regular supermarket.

I'd be happy to hear more opinions on this tag. :pray: 

[^1]: an alternative solution would be to create special presets for drive-thru-only amenities, e.g. drive thru ATMs, drive thru [item pick-up locations](https://wiki.openstreetmap.org/wiki/Tag:shop%3Doutpost), etc.

Dimitar5555:
> Adding the third case makes the field significantly less easy to use compared to the simple checkbox.

Somebody somewhere suggested a field that has 2 or 3 radio buttons next to each other. There was also a screenshot but I can't find the PR/issue.

michalgwo:
> Yes, the value is documented on the wiki, but usage is [very rare](https://taginfo.openstreetmap.org/keys/drive_through#values) compared to the usual yes/no cases.

The usage of this value is indeed significantly lower than yes/no, but I wouldn't say 436 is too little to include in presets. It could be low because it's not shown in the iD and some people may not know about this option and select 'yes' instead.

> Also, at least in some situations, this tag seems to have the the same downsides as the [`<status>=yes`](https://wiki.openstreetmap.org/wiki/Comparison_of_life_cycle_concepts#%3Cstatus%3E_=_yes) life cycle tagging approach, which is AFAICT generally discouraged. For example: on a general purpose map, one would not be able to distinguish a drive-tru-only supermarket from a regular supermarket.

Having only yes/no option doesn't really improve this. Assuming there is a drive-thru-only supermarket and someone wants to map it. If they have only 2 options yes/no, they will probably set yes, which is only partially true, now on the map that doesn't support the drive-thru tag, the effect will be the same, on the other hand, on the map supporting drive_through tag, they will see the difference - if this would be tagged as 'yes', they would see partially wrong information.

tyrasd:
> Somebody somewhere suggested a field that has 2 or 3 radio buttons next to each other. There was also a screenshot but I can't find the PR/issue.

That would not be hard to implement and look somewhat like this: 


IMO this is not much better than the combobox solution: The field is still significantly more complex and harder to scan than a basic checkbox.

tyrasd:
I also noticed that the `only` value was only very recently documented on the wiki: https://wiki.openstreetmap.org/w/index.php?title=Key:drive_through&diff=2647903&oldid=2427522 and I was not able to find a proper discussion around this change, except for a very limited [discussion in the wiki](https://wiki.openstreetmap.org/wiki/Talk:Key:drive_through)[^1].

Given that this is in fact a rather major change to the semantics of the `drive_through` tag[^2], I will not further consider it until there has been a proper discussion in the community, for example in form of a tag [proposal](https://wiki.openstreetmap.org/wiki/Proposal_process) on the wiki.

[^1]: and even for the given examples there, I come more and more to the conclusion that for these situations the tag is indeed a kind of "troll tag" [similar to `delivery=only`](https://wiki.openstreetmap.org/wiki/Talk:Key:delivery#delivery=only_considered_trolltag?_(please_avoid_using_it)), because these are more like click-and-collect locations, for which there already exist a quite established tag for: [`shop=outpost`](https://wiki.openstreetmap.org/wiki/Tag:shop%3Doutpost).
[^2]: which btw has also not even been propagated to wiki pages which suggest the usage of the `drive_though` tag in combination, e.g. [here](https://wiki.openstreetmap.org/w/index.php?title=Tag:amenity%3Dfast_food&oldid=2610926#Tags_to_use_in_combination) or [here](https://wiki.openstreetmap.org/w/index.php?title=Tag:amenity%3Datm&oldid=2610190#Tagging)

PR #1102, opened 2024-02-06
Deprecate `leisure=maze` (-> `attraction=maze`)

westnordost:
[`attraction=maze`](https://wiki.openstreetmap.org/wiki/Tag:attraction%3Dmaze) and [`leisure=maze`](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dmaze) are documented to mean exactly the same thing.

Since 2014, `attraction=maze` is the slightly more popular tag:
https://taghistory.raifer.tech/#***/leisure/maze&***/attraction/maze

In 2018, `attraction=maze` was made an iD preset. At the time of writing, it is now used 7 times more than `leisure=maze`.

In mid 2023, a note was added to the wiki page of `leisure=maze` that it has the same meaning as the more popular `attraction=maze` and that the latter should be used instead.

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -987,6 +987,10 @@
     "replace": {"leisure": "ice_rink", "sport": "ice_skating"}
   },
   {
+    "old": {"leisure": "maze"},
+    "replace": {"attraction": "maze"}
+  },
+  {
     "old": {"leisure": "recreation_ground"},
     "replace": {"landuse": "recreation_ground"}
   },


tyrasd:
I'm slightly hesitant to include this one. The tag is (at least at the moment) not quite described as _deprecated_ in the wiki. And as the tag upgrade suggestions in iD (originating from this repo) can be seen as a kind of semi-automated edit to OSM, I'm pretty careful when it comes to inclusions to the list.

Also, the current "advantage" of the `attraction=maze` tag might be in large parts because this repo has support for it as a preset, but not for `leisure=maze`. So one could argue that iD's presets have skewed the numbers and created the precedent for this deprecation in the first place.

On the other hand, it really does not make sense to have two active tags for this kind of map feature. I'm going to open a discussion thread on the forum and see if there is a consensus to consider `leisure=maze` deprecated.

Issue #1085, opened 2024-01-19
amenity=charging_station -> Which vehicles can be charged here?


qugebert:
### OSM Tag(s)

amenity=charging_station

### How would you like this tag to see supported?

As a Field

### Label

Supported vehicles

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station

### Status of the Tag

Approved

### Usage of the tag

bicycle=* 12 283

### Further Information

Sommething like the activity-field

qugebert:
> maybe it's also worth to additionally create new vehicle dependent presets (e.g. bike charging station, car charging station), for better discoverability while searching and also to allow for vehicle dependent fields (e.g. different charging ports on cars vs. bikes).

sounds good. Maybe i will try this on my own the next days.
Is it neccesary to find new icons or can both of these presets have the same/existing icon?

tyrasd:
> Is it neccesary to find new icons or can both of these presets have the same/existing icon?

Ideally yes, but I would say don't worry too much about it: if you cannot find one, it is still fine to reuse the icon from the generic preset.

FYI: I've just created https://github.com/ideditor/schema-builder/issues/135 for a possible enhancement to the preset schema to make the icon-ification of such presets potentially much easier.

mvl22:
It is good that this preset now exists, but it really does need the vehicle type: whether car or bicycle for instance.

Example of bicycle-only charging points:
https://overpass-turbo.eu/s/21ex

Currently this key information has to be added manually in the tags section:

![Image](https://github.com/user-attachments/assets/30204c2e-b311-4ae7-ad5a-9f1a91947d72)

matkoniecz:
see https://community.openstreetmap.org/t/amenity-charging-station-with-access-yes/141225/

matkoniecz:
> now that schema-builder v7 is released, this can be done with `"type": "access"`, similar to #2245
>
> (assuming that we want to allow other values like `=designated`, and `=private`, not just `=yes` and `=no`)

https://github.com/openstreetmap/id-tagging-schema/issues/2573#issuecomment-5033904037

PR #1106, opened 2024-02-07
amenity=charging_station: add field for supported vehicles

qugebert:
This is the first part of #1085 

I used the list of possible vehicles from https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station#Vehicles


data/fields/electric_vehicle.json
```diff
@@ -0,0 +1,24 @@
...
+        "hgv",
+        "bus",
+        "small_electric_vehicle"
+    ],
+    "type": "manyCombo",
+    "label": "Supported vehicles",
+    "strings": {
+        "options": {
+            "moped": "Mopeds",
+            "bicycle": "E-Bikes",
```
1ec5, line 16: Consider expanding this to “Electric Bicycles”, since the UI strings tend not to use informal terminology.

data/fields/electric_vehicle.json
```diff
@@ -0,0 +1,24 @@
...
+            "motorcar": "Cars",
+            "motorcycle": "Motorcycles",
+            "hgv": "Trucks",
+            "bus": "Buses",
+            "small_electric_vehicle": "Small electric vehicles"
```
1ec5, line 21: This is a tricky one to word in American English (the default locale): “small electric vehicles” is a legal term in “[many European countries](https://wiki.openstreetmap.org/wiki/Key:small_electric_vehicle)”, but to an American it would sound like a description of, say, a Smart car, because things the size of a bicycle are usually called “devices” rather than “vehicles” here.

Also, this PR translates `hgv` as “trucks” rather than “heavy goods vehicles”, the UK legal term, which I agree with, but it makes the user even less likely to notice that “small electric vehicles” is a legal term.

Assuming it’s OK to use this value even in the U.S., consider either “personal mobility devices” or “micromobility devices”. These terms can also apply to unpowered devices like kick scooters, but I don’t think we need to be concerned about that ambiguity in the context of this field.

Regardless, this string should be in Title Case.

qugebert: Thank you for your feedback. It's sometimes hard for me as a non-native-speaker to find the correct wordings.

I had another look to the [wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station) and now i'm even more confused than before.
While in the [vehicles-section](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station#Vehicles) the tagging ```small_electric_vehicles=yes``` is used, the tagging ```scooter=yes``` is used in the examples-section (and in the german wiki page). 
And for trucks, the english wiki says ```hgv=yes``` whilst the german wiki says ```truck=yes```

Should i omit these two from the list as there is a quite low usage anyway or should we try to get a clarification for a unique tagging from the community forum?

Thank you also for the hint to "Title Case" convention.

data/fields/electric_vehicle.json
qugebert: Sure, i will change this.

data/fields/electric_vehicle.json
1ec5: Wow, that is confusing! A discussion in the forum would be a good idea, just to be on the safe side and raise awareness.

SupaplexOSM: Don't care about the German Wiki. The quoted information is from the first version of the article from 10 years ago and was no longer up to date. I have just updated the wiki at this point. The term "scooter" in OSM refers to more "moped" like vehicles (which is also how it was written in the translation) rather than "kick scooters" (which are `small_electric_vehicles` if equipped with an electric motor). `truck` should explicitly not be used and replaced with `hgv`.

tyrasd: > personal mobility device

I think this term might be easily confused with [_personal mobility aid_](https://agis.com.sg/personal-mobility-aidpma/) devices (i.e. "fancy" electric wheelchairs).

The [wiki](https://wiki.openstreetmap.org/wiki/Key:access) uses "_Electric scooter_" as a description for the `small_electric_vehicle` tag. Maybe that's good enough, especially if we're also thinking about simplifying `hgv` to _Truck_?!

matkoniecz: > Wow, that is confusing! A discussion in the forum would be a good idea, just to be on the safe side and raise awareness.

Is anyone planning to do so? Or did? Or has a good reason to believe it is not needed?

matkoniecz: I will close this PR as abandoned in about 70 days if that will remain unanswered and not acted on.

1ec5: > The [wiki](https://wiki.openstreetmap.org/wiki/Key:access) uses "Electric scooter" as a description for the small_electric_vehicle tag. Maybe that's good enough, especially if we're also thinking about simplifying hgv to Truck?!

Yes, as long as that’s the intended scope of the key. It wasn’t clear based on earlier discussion.

matkoniecz:
my summary of linked discussion is that it has complaints about (mis)use of legal access tagging here, but is not mentioning alternatives and this tagging is widely used and is documented

I commented there for now

I think that we can proceed with this PR given de facto tagging situation? Or is anyone signing up for deprecation attempt?

matkoniecz:
screen-2026-08-23-08-57-10

one annoying part is that `designated` appears like worse than `yes`

data/fields/electric_vehicle.json
```diff
@@ -0,0 +1,24 @@
...
+    "keys": [
+        "bicycle",
+        "moped",
+        "motorcar",
+        "motorcycle",
+        "hgv",
+        "bus",
+        "small_electric_vehicle"
+    ],
+    "type": "manyCombo",
```
k-yle, line 11:
```suggestion
    "type": "access",
```

we could use `type=access` here, right? similar like #2245 

the UX would be better, and you can constrain the `options` to only `yes` and `designated`

k-yle (changes requested):

data/fields/electric_vehicle.json
matkoniecz: @qugebert have you seen that?

PR #1107, opened 2024-02-07
Seperate presets for ```amenity=charging_station``` (for ```bicycle=yes``` and ```motorcar=yes```) 

qugebert:
See https://github.com/openstreetmap/id-tagging-schema/issues/1085#issuecomment-1918998696

This adds two presets for charging stations depending on the vehicle they are intended for.

### DIFF
--- a/data/presets/amenity/charging_station/bicycle.json
+++ b/data/presets/amenity/charging_station/bicycle.json
@@ -13,7 +13,7 @@
     ],
     "tags": {
         "amenity": "charging_station",
-        "bicycle": "yes"    
+        "bicycle": "yes"
     },
     "terms": [
         "EV",
--- a/data/presets/amenity/charging_station/bicycle.json
+++ b/data/presets/amenity/charging_station/bicycle.json
@@ -17,13 +17,13 @@
     },
     "terms": [
         "EV",
-        "Electric bicycle",
-        "Pedelec charging",
+        "Electric Bicycle",
+        "Pedelec Charging",
         "E-Bike"
     ],
     "reference": {
         "key": "amenity",
         "value": "charging_station"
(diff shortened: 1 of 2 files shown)


tyrasd:
Hey, thanks for the PR. Unfortunately, when suggesting this in [my comment](https://github.com/openstreetmap/id-tagging-schema/issues/1085#issuecomment-1918998696), I forgot about one thing: The way the preset matches OSM objects makes it such that these two presets would apply to not only the single-vehicle-type sites we would want it to apply to (e.g. `amenity=charging_station` + `bicycle=yes`), but also to multiple-vehicle-type sites where a partial match exists (e.g. `amenity=charging_station` + `bicycle=yes` + `motorcycle=yes`, etc.).

So, for now this does not work in the tagging schema (see https://github.com/ideditor/schema-builder/issues/25 for a suggestion to fix this, potentially). I'm sorry for the inconvenience. :bow:

qugebert:
> Hey, thanks for the PR. Unfortunately, when suggesting this in [my comment](https://github.com/openstreetmap/id-tagging-schema/issues/1085#issuecomment-1918998696), I forgot about one thing: The way the preset matches OSM objects makes it such that these two presets would apply to not only the single-vehicle-type sites we would want it to apply to (e.g. `amenity=charging_station` + `bicycle=yes`), but also to multiple-vehicle-type sites where a partial match exists (e.g. `amenity=charging_station` + `bicycle=yes` + `motorcycle=yes`, etc.).
>
Oh, is see. This would be kinda confusing.
 
> I'm really sorry for not thinking about this in the first place! 🙇

It's fine. I'm still learning how this projects works and with that I also learned a little bit about it.

PR #1108, opened 2024-02-08
add emergency=disaster_response preset

andrewharvey:
recently approved https://wiki.openstreetmap.org/wiki/Tag:emergency=disaster_response with existing usage https://taginfo.openstreetmap.org/tags/emergency=disaster_response#overview

the two terms added are the abbreviations used in Australia and Germany, although not sure if we should omit these and instead rely on the translations to localise their naming :shrug: 

https://github.com/osmlab/name-suggestion-index/pull/9221/ adds brand presets for Australia.

### DIFF
--- a/data/presets/emergency/disaster_response.json
+++ b/data/presets/emergency/disaster_response.json
@@ -15,8 +15,11 @@
         "area"
     ],
     "terms": [
-        "SES",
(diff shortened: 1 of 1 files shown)


tyrasd:
Thanks!

PR #1109, opened 2024-02-08
Deprecate ```amenity=emergency_service```,```emergency=ses_station```,```emergency_service=technical``` -> ```emergency=disaster_response```

qugebert:
See also #1108 

The [approved proposal (approved: 2024-01-18)](https://wiki.openstreetmap.org/wiki/Proposal:Emergency%3Ddisaster_response) said, the three combinations should be converted:
* ```amenity=emergency_service```
* ```emergency=ses_station```
* ```emergency_service=technical```

The four wiki pages have them marked as deprecated too
https://wiki.openstreetmap.org/wiki/Tag:emergency=disaster_response
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Demergency_service
https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dses_station
https://wiki.openstreetmap.org/wiki/Tag:emergency_service%3Dtechnical

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -120,6 +120,10 @@
     "replace": {"emergency": "water_rescue"}
   },
   {
+    "old": {"amenity": "emergency_service"},
+    "replace": {"emergency": "disaster_response"}
+  },
+  {
     "old": {"amenity": "notice_board"},
     "replace": {"advertising": "board"}
   },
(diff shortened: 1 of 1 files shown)


tyrasd:
Oh, I only just noticed after mergin this PR that according to the wiki, the old tags are apparently to quite [fully deprecated](https://wiki.openstreetmap.org/w/index.php?title=Tag:emergency%3Ddisaster_response&oldid=2655932), quote:

> `amenity=emergency_service` and `emergency_service=technical` were/are used for the "Regionalstellen" and "Landesverbände" of the German THW. These facilitys do not match `emergency=disaster_response`.

I don't know the details, but I assume that these particular sites do only administrative work and no direct/practical disaster response. The [wiki](https://wiki.openstreetmap.org/w/index.php?title=Tag:emergency%3Ddisaster_response&oldid=2655932) also says this about the new tag:

> It is the place where the members start a mission after being alerted to an incident. 

This would confirm that the new tag is not useful for these kind of facilities.[^1]

[^1]: Side note: To me it seems that the tag `amenity=emergency_service` is also not really a good fit such a facility, and perhaps some `office` tag should be used instead in the future, but that's a different discussion.

In any case, it seems like we cannot simply upgrade the old tags in these situations. Therefore, I have to partially revert this PR, unfortunately.

PR #1115, opened 2024-02-11
Use more common name for `man_made=water_works`

arch0345:
Changes the label for `man_made=water_works` to a more common term used in American English

### DIFF
--- a/data/presets/man_made/water_works.json
+++ b/data/presets/man_made/water_works.json
@@ -15,5 +15,9 @@
     "tags": {
         "man_made": "water_works"
     },
-    "name": "Water Works"
+    "aliases": [
+        "Potable Water Treatment Plant",
+        "Water Works"
+    ],
+    "name": "Water Treatment Plant"
 }


data/presets/man_made/water_works.json
```diff
@@ -15,5 +15,9 @@
     "tags": {
         "man_made": "water_works"
     },
-    "name": "Water Works"
+    "aliases": [
+        "Potable Water Treatment Plant",
+        "Water Works"
+    ],
+    "name": "Water Treatment Plant"
```
1ec5, line 22: Interesting, I’m personally more familiar with the term “water works” than “water treatment plant”. At least in my corner of the Midwest, we’d call it a “water works” even if its official name includes “Water Treatment Plant”, and even if [hasn’t been one since the 1950s](https://www.openstreetmap.org/way/23766551). “Water treatment plant” would be easily misunderstood as a wastewater treatment plant.

[Google Ngrams Viewer](https://books.google.com/ngrams/graph?content=water+works%2Cwater+treatment+plant&year_start=1800&year_end=2019&corpus=en-US-2019&smoothing=3) says “water treatment plant” has always been obscure but “water works” had a moment around the turn of the last century. Maybe we’re missing another common term, or maybe it’s one of those concepts with hopelessly regional terms.

tyrasd: I can't comment on which terms is more common in US English, but assuming that there's no clearly preferred one, I would have said, let's just keep _Water Works_ as the default label (as it is less likely to be confused with waste water treatment plant) and have _Water Treatment Plant_ as the alias. What do you think?

PS: I also found _Drinking Water Treatment Plant_ to be more common on google search results compared to _Potable Water Treatment Plant_, maybe those two could be search `terms`?

matkoniecz: I tried doing this in #1256

PR #1116, opened 2024-02-14
feat: Add menstrual products

moan0s:
I hope I did this correctly, if not I'm happy to correct!
Key Documentation: https://wiki.openstreetmap.org/wiki/Key:toilets:menstrual_products

### DIFF
--- a/data/fields/toilets/menstrual_products.json
+++ b/data/fields/toilets/menstrual_products.json
@@ -8,7 +8,7 @@
             "no": "No"
         }
     },
-    "label": "Free Menstrual Products available",
+    "label": "Free Menstrual Products Available",
     "prerequisiteTag": {
         "key": "toilets",
         "valueNot": "no"
--- a/data/fields/toilets/menstrual_products.json
+++ b/data/fields/toilets/menstrual_products.json
@@ -1,0 +1,16 @@
+{
+    "key": "toilets:menstrual_products",
+    "type": "radio",
+    "strings": {
+        "options": {
+            "yes": "Yes, in all stalls",
+            "limited": "Limited to some stalls",
+            "no": "No"
+        }
+    },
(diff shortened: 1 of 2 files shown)


tyrasd:
Hey. Thank you for the suggestion. The tag was only quite recently introduces, that's why the number of uses is still quite low. I would however say that it would be good to have a field for this tag in the editors already now.

This pull request is however still missing the json file for the field of the new tag: You would have to create a new file in `data/fields/toilets/menstrual_products.json` and fill in the required details (e.g. a label for the field, the possible options, etc.). You might use the field for the [`wheelchair`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/wheelchair.json) tag as a template. You could use something like _Free Menstrual Products_ as a label. I'm currently not quite sure how to properly describe the option for the `limited` tag value… Maybe something like _Limited to Some Stalls_ could work?!

An alternative would be to add this tag as a checkbox field (`"type:" "check"`) for now and monitor how the tag is going to be used in practice over time.

moan0s:
Thank you so much for the feedback, I added the required file. I liked the *Limited to Some Stalls* option and used that and kept the radio type.

tyrasd (approved):

moan0s:
I tried the changes in the preview and `toilet:menstrual_products` is correctly suggested. On the other hand the values are not whereas `toilet:wheelchair` works as expected. So there is no dropdown menu to select a value. Is that expected or did I do something wrong?

tyrasd:
You might have misunderstood how the [`prerequisiteTag`](https://github.com/ideditor/schema-builder?tab=readme-ov-file#prerequisitetag) is interacting with a field: here it had prevented the field from being included in the list of suggested additional fields. I've changed that in d18c7cc, and AFAICS it should work properly now.

Issue #1099, opened 2024-02-06
Add support for noname=yes, disable name field when set


RudyTheDev:
### OSM Tag(s)

noname=yes

### How would you like this tag to see supported?

Other (please indicate below)

### Label

Unnamed

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:noname%3Dyes

### Status of the Tag

De Facto

### Usage of the tag

780,454

### Further Information

iD shows the `name` field prominently. However, if one were to add `noname=yes`, the name field display remains exactly the same. A user who doesn't check or have tags expanded would presume the name is missing, even though it may actually have been verified that the feature does not have a name. For example, official street names versus what someone might assume is the street layout without consulting sources.

I assume iD doesn't want all the new users to start adding `noname=yes` to everything, so I propose to show a "no name" checkbox under the name field if and only if the `noname=yes` is set and then disable/gray out the name field.

tyrasd:
The name field could be automatically hidden by using a [`prerequisiteTag`](https://github.com/ideditor/schema-builder?tab=readme-ov-file#prerequisitetag). There are only a few (rare) corner cases of when a feature is tagged with `noname=no` and no `name` (here's an example: https://www.openstreetmap.org/node/8859106172 where the mapper indicated that the feature does have a name, but they don't know it). There are currently less than 100 such cases in OSM. Properly handling these would require extending the tagging schema to support more complex prerequisite tag handling (see https://github.com/ideditor/schema-builder/issues/137 for that). An alternative option would be to hardcode this in iD for now…

maro-21:
Duplicate of https://github.com/openstreetmap/iD/issues/7589.

PR #1117, opened 2024-02-14
hide empty name field if `noname` tag is present

tyrasd:
partially addresses #1099

to test (might not work as expected): add a checkbox field for the noname tag (only) if it is present, such that mappers are not irritated by the lack of name field, and be able to uncheck it to be able to add a name.

### DIFF
--- a/data/fields/name.json
+++ b/data/fields/name.json
@@ -7,5 +7,8 @@
     "terms": [
         "label",
         "title"
-    ]
+    ],
+    "prerequisiteTag": {
+        "keyNot": "noname"
+    }
 }
--- a/data/fields/noname.json
+++ b/data/fields/noname.json
@@ -1,0 +1,11 @@
+{
+    "key": "noname",
+    "type": "check",
+    "label": "Nameless",
+    "universal": true,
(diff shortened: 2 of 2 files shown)


matkoniecz:
`noname=no` is rare and dubious, but I still feel uncomfortable with hiding names in such cases

matkoniecz:
@tyrasd now that schema builder changes are easier - are you interested in making it more reliable and supporting also `noname=no` silliness?

data/fields/name.json
```diff
@@ -7,5 +7,8 @@
     "terms": [
         "label",
         "title"
-    ]
+    ],
+    "prerequisiteTag": {
+        "keyNot": "noname"
```
Dimitar5555, line 12: I think this should work as expected.

```suggestion
        "key": "noname",
        "valuesNot": "yes"
```

tyrasd: No, this will exclude the `name` field from showing up in all cases, except when the feature has a tag with `key` `noname` and value different from `yes`. But we want to only exclude the `name` field if the tag is present and not `no`.

PR #1123, opened 2024-02-18
Add access to suggested tags on parkingspace

watmildon:
The third most common value of the `parking_space` tag is `private`. However, this makes is impossible to specify a `disabled` spot that is not accessible to the public. A group of editors have been working through the ~10,000 instances in the db and migrating it to the `access` tag. It would be great for editors to show this as a possible combination.


tyrasd (approved):

MapSpot:
As of sending this message, there are no longer any instances of parking_space=private.

Another reasonable cleanup will involve manually migrating the 1700 instances of parking_space=customers to the more common access=customers where appropriate.

PR #1126, opened 2024-02-22
add `via` tag to ferry routes

k-yle:
Ferry routes aren't necessarily point-to-point. They can stop along way, so this PR adds the `via` field. Ferry routes are now consistent with all other modes of transport.

### DIFF
--- a/data/presets/route/ferry.json
+++ b/data/presets/route/ferry.json
@@ -28,6 +28,7 @@
         "ref_route",
         "reservation",
         "roundtrip",
+        "via",
         "vhf",
         "wheelchair"
     ],


tyrasd (commented):
In general this is ok, I think, but maybe adding it in `moreFields` would be fine, as usage is relatively rare it seems (246 uses today compared to 3.7k for `from/to`). 

Also, it would be good if the [wiki page](https://wiki.openstreetmap.org/wiki/Tag:historic%3Dcannon) would mention the usage of the `via` for this.

k-yle:
sure, wiki and PR have been updated

Issue #1073, opened 2023-12-19
On adding archaeological_site=fortification, add fortification_type field


danieldegroot2:
### OSM Tag(s)

fortification_type

### How would you like this tag to see supported?

As a Field

### Label

Fortification Type

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Key:fortification_type

### Status of the Tag

De Facto

### Usage of the tag

17.874

### Further Information

When adding fortification, it would be useful if fortification_type field was automatically added, so the user does not have to add this manually for every single one (usually fortification type is known).

PR #1129, opened 2024-02-22
add field for `fortification_type`

k-yle:
Added a field for `fortification_type=*`, which is a popular subtag for `archaeological_site=fortification`.

80% of `archaeological_site=fortification` have `fortification_type=*`


tyrasd (approved):
:+1: this should also fix #1073

PR #1130, opened 2024-02-22
fix LPG/LNG confusion

k-yle:
There seems to be a misunderstanding about [LPG](https://en.wikipedia.org/wiki/Liquefied_petroleum_gas) vs [LNG](https://en.wikipedia.org/wiki/Liquefied_natural_gas). They are two separate fuels. 

Currently, there is only 1 option, called "Liquefied **N**atural Gas (L**P**G)". This is very confusing and has possibly caused incorrect information to be entered into the database.


1ec5:
Yikes, hopefully the inconsistency within the label itself gave mappers pause before choosing this option.

tyrasd (approved):
Thanks for noticing!

PR #1135, opened 2024-02-23
Create preset for kitchen garden

imagoiq:
There are presets for "residential" and "community" garden:type (https://github.com/openstreetmap/id-tagging-schema/pull/714). This one adds a preset for kitchen garden. Like the others, it's a de facto tag with some usages (> 1200), see the [original proposal.](https://wiki.openstreetmap.org/wiki/Proposal:Garden_specification)

### DIFF
--- a/data/presets/leisure/garden/kitchen.json
+++ b/data/presets/leisure/garden/kitchen.json
@@ -1,0 +1,21 @@
+{
+    "icon": "fas-carrot",
+    "geometry": [
+        "area"
+    ],
+    "tags": {
+        "leisure": "garden",
+        "garden:type": "kitchen"
+    },
+    "reference": {
+        "key": "garden:type"
+    },
+    "terms": [
+        "vegetable garden",
+        "potager garden",
+        "culinary garden",
+        "edible garden",
+        "cook's garden"
+    ],
+    "name": "Kitchen Garden"
+}


tyrasd (commented):
This looks ok. I'd only like to see this tag value been described on the wiki; at least as an entry on the table for the [`garden:type`](https://wiki.openstreetmap.org/wiki/Key:garden:type) tag key. Once that's done, this preset can be added here as well.

imagoiq:
Thanks, you're right, I've added the type to the table.

tyrasd (approved):

imagoiq:
@tyrasd I double check the preview and there are two fields which doesn't make sense in this case ("Allowed access" and "Fee"). Is there a way to remove them by default?

image

simonpoole:
May I point out that _kitchen_ was/is a (semi-documented) value for _garden:style_ not _garden:type_. This PR would seem to be creating unnecessary divergent tagging.

data/presets/leisure/garden/kitchen.json
k-yle, line 1: Is there a reason why this preset has `"matchScore": 1.1`? This causes the preset to be the first suggestion when you search for `Garde…`, while all the other garden types are at the bottom of the list

PR #1139, opened 2024-02-25
Add `changing_table` as optional field to several presets

### LINKED ISSUE #1060: Add `changing_table` field to nodes with `toilets` field
### OSM Tag(s)

changing_table

As an Option of a Field

### Label

Changing Table

### Aliases

_No response_

### Terms

_No response_

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Key:changing_table

### Status of the Tag

Approved

### Usage of the tag

39000

### Replaces other Tag?

_No response_

### Regional Tag?

_No response_

### Further Information

The tag is useful for finding whether a given place offers a changing table, and where. A field is currently being offered on separately mapped toilets. However, it does not appear on other points of interest where toilets (and often changing tables) are common. I would propose to add `changing_table` as an optional field to some objects that also offer the `toilets` field:
- amenity=cafe
- amenity=restaurant
- highway=rest_area
- tourism=xyz
- shop

mangerlahn:
Fixes #1060.
I hope I catched all sensible presets. I follow mostly the `toilets` field, but skipped a few like `pub` or `camp_site`.


tyrasd (approved):

tyrasd:
thanks!

Issue #1056, opened 2023-11-19
Add (hidden?) present for `crossing=informal`


tordans:
### OSM Tag(s)

crossing=informal

### How would you like this tag to see supported?

As a Preset

### Label

Informal crossing

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:crossing%3Dinformal

### Status of the Tag

In Use

### Usage of the tag

7,600

### Regional Tag?

global

### Further Information

StreetComplete started using the tag `crossing=informal` ([Example](https://www.openstreetmap.org/node/8364259198)) which is not recognized by iD.

Westnordost wrote about it in https://github.com/streetcomplete/StreetComplete/issues/4944#issuecomment-1507010957 and added the tag in https://github.com/streetcomplete/StreetComplete/commit/8e99dd589129e02e99402dea7d803a4fbe842307 in April 2023.

Right now, iD does not show any indication that some kind of crossing is tagged except for the little dot on the node. However the preset panel than does not suggest any special tagging. Only the "(All) Tags" section reveals the tagging.

<img width="757" alt="image" src="https://github.com/openstreetmap/id-tagging-schema/assets/111561/651ca2fe-2c32-4f38-b04a-327ac686bdc7"> [Example](https://www.openstreetmap.org/edit?editor=id&node=8364259198#map=20/52.37506/13.63278)

I suggest we add at least an unsearchable present to give the tagging some kind of visibility.

---

On a different node, I think that the user chose the wrong crossing type in this case. [Mapillary shows a lowered kerb](https://www.mapillary.com/app/?z=17&lat=52.375111799972&lng=13.6323612&pKey=1074033653411074&focus=photo&x=0.5295094535546165&y=0.5963598760979361&zoom=0). But to make it easier to read this ticket, lets leave it like this for a bit and only fix it later…

---

Update: Another example ([OSM](https://www.openstreetmap.org/node/8378263476)) that fits better ([Mapillary](https://www.mapillary.com/app/?z=17&lat=52.369115299972&lng=13.639327299972&pKey=5077122972307353&focus=photo&x=0.2343944225161529&y=0.5368733382684019&zoom=0)).

tordans:
Closed by https://github.com/openstreetmap/id-tagging-schema/pull/1140

PR #1140, opened 2024-02-26
Add `crossing=informal` as unsearchable preset for point and line geometries

mikaeldui:
I'm currently adding a lot of informal paths, and just stumbled upon one that crosses a large road. I thought I'd add a crossing for the point but there was no preset for informal crossings. I almost made it an *unmarked crossing* but after browsing the wiki it seems *informal crossing* is correct. However, there's currently no preset for informal crossings which makes it hard to add, even harder as they shouldn't have the highway tag.

Informal crossings are well documented on the Wiki: https://wiki.openstreetmap.org/wiki/Tag:crossing%3Dinformal

### DIFF
--- a/data/presets/highway/crossing/_informal.json
+++ b/data/presets/highway/crossing/_informal.json
@@ -7,7 +7,6 @@
         "vertex"
     ],
     "tags": {
-        "highway": "crossing",
         "crossing": "informal"
     },
     "terms": [
--- a/data/presets/highway/crossing/_informal.json
+++ b/data/presets/highway/crossing/_informal.json
@@ -7,16 +7,18 @@
         "vertex"
     ],
     "tags": {
+        "highway": "crossing",
         "crossing": "informal"
-    },
-    "reference": {
-        "key": "crossing",
-        "value": "informal"
     },
     "terms": [
         "informal crosswalk",
         "informal foot path crossing",
         "informal pedestrian crossing"
     ],
-    "name": "Informal Crossing"
+    "reference": {
+        "key": "crossing",
+        "value": "informal"
+    },
+    "name": "Informal Crossing",
+    "searchable": false
 }
--- a/data/presets/highway/crossing/informal.json
+++ b/data/presets/highway/crossing/informal.json
@@ -1,0 +1,22 @@
+{
+    "icon": "temaki-pedestrian",
+    "fields": [
+        "crossing"
+    ],
+    "geometry": [
+        "vertex"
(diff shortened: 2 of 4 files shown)


tyrasd:
This would address / fix #1056

tordans (approved):

tordans:
I ran into this again the other day and think we should merge this as an unsearchable preset for now.

Out in the wil:
Since SC uses the tag, there are crossings that just don't show up right ATM, eg (from GoMap):
> image
> image

I think we should support this since SC supports it.

However, it should be unsearchable IMO, because…
- the think the wiki page is quite bad for this tag https://wiki.openstreetmap.org/wiki/Tag:crossing%3Dinformal
- the reasoning why it is used is not easy to find, apparently somewhere in this forum thread at https://github.com/streetcomplete/StreetComplete/issues/4944
- I did not see any of those crossings in the wild which I would actually consider informal crossings. The wiki shows a good example, but the cases I saw in cities are all `unmarked` IMO.

I changes a few things in this PR:
1. I made the preset unsearchable
2. I added a corresponding line based preset to give users a way to use the same logic when mapping sidewalks as separate geometries. Those are only use 2k times for now, but keeping line and point data in sync is important IMO.
  In my example below the used did not add any sub tag "crossing type" at all, which is not better IMO https://www.openstreetmap.org/way/1079416103. Also, once we have https://github.com/openstreetmap/iD/issues/10757 in place point and line data should have a match.
    > image


One thing I did not add is a translated value for the `crossing=*` field which means this value still shows up as a raw value
> image

Given the state of the tagging described above I think we should keep it this way. It would be very hard to find a 2-3 word name for those crossings that will be easy to understand for users. SC has more context that it can use to guide users and still there are miss matches. I fear that promoting the tag further is too early.


---

**Testing:**

https://pr-1140--ideditor-presets-preview.netlify.app/id/dist/#locale=en&disable_features=boundaries&map=20.30/52.46062/13.53049&background=Bing&id=n9894826209

**Data item** looks good:
> image

**Icon** looks good:
It uses the same as `unmarked` which is fitting.
> image


---

Closes https://github.com/openstreetmap/id-tagging-schema/issues/1056
Checks one box in https://github.com/openstreetmap/id-tagging-schema/issues/1410

tordans:
@tyrasd I merged this now but you might want to double check this. My reasoning is in https://github.com/openstreetmap/id-tagging-schema/pull/1140#issuecomment-2817249038

PR #1143, opened 2024-02-27
Add `oneway` to `moreFields` for path like highways

tordans:
During our Mapping event for bike infrastructure we ran into the issue that on a `hw=path` (eg. with `bicycle=yes|designate+foot=designated`) the `oneway` field is missing.

This PR add the field as "moreFields" option to path like highways which did not have it, yet.

This way the value becomes visible if set, but is not requested by default (because usually it is not needed).

### DIFF
--- a/data/presets/highway/crossing.json
+++ b/data/presets/highway/crossing.json
@@ -8,7 +8,8 @@
     ],
     "moreFields": [
         "flashing_lights",
-        "kerb"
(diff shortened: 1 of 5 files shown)


tyrasd:
merged as 3e5e99ff881d88f4e8761e64356c830875bc8437 (`moreFields` are usually sorted alphabetically)

PR #1145, opened 2024-02-28
add additional values to Vending field

CornerSyrup:
Adding keys from [wiki](https://wiki.openstreetmap.org/wiki/Key:vending) and [taginfo](https://taginfo.openstreetmap.org/keys/vending#values) to dictionary. Only coping those familiar and commonly used.

Changes:

- Add missing but common keys
- Sort to organise

### DIFF
--- a/data/fields/vending.json
+++ b/data/fields/vending.json
@@ -13,6 +13,7 @@
             "books": "Books",
             "bottle_return": "Bottle Return",
             "bread": "Bread",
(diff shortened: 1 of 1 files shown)


tyrasd:
like in #1146: would you mind explaining the reasoning behind which options you selected for this PR? For example, `"alcohol"` does not seem to be documented in the wiki, is not really "in use" according to taginfo…

PR #1147, opened 2024-02-29
Introduce `cycleway:both=*` and make `cycleway:NULL=*` unsearchable

tordans:
This is another take at the same issues as https://github.com/openstreetmap/id-tagging-schema/pull/1137.

The goal is, to work around
- a deprecation of the `:NULL` variant that used to represent "both" or sometimes "the primary side"
- the need to add another feature to the files which would allow to handle both keys at the same time.

The idea is, to have the legacy value kind of unsearchable and only show up when the `cycleway:NULL` key is present. However, that is hard ATM because this key is also used for `crossing`, `asl` and such (https://taginfo.openstreetmap.org/keys/cycleway#values) and the `prerequisiteTag` does not allow any arrays as values, yet.

Lets see how the preview behaves … but I think that the limitation of `prerequisiteTag` will make this a nonstarter as well…

### DIFF
--- a/data/fields/cycleway/_cycleway_legacy.json
+++ b/data/fields/cycleway/_cycleway_legacy.json
@@ -1,5 +1,9 @@
 {
     "key": "cycleway",
+    "prerequisiteTag": {
+        "key": "cycleway",
+        "valueNot": "valueNot"
+    },
     "keys": [
         "cycleway:left",
         "cycleway:right"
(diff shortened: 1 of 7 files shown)


tordans:
I will close this. The field now supports both variations and the discussion on which one to recommend in iD is ongoing https://wiki.openstreetmap.org/wiki/Proposal:Make_cycleway:both_the_default_to_indicate_both_sides

PR #1150, opened 2024-03-07
more terms for amenity=public_bookcase

matkoniecz:
based on attempts to find while mapping with StreetComplete (it is using iD presets)


matkoniecz:
before:
screen01

after:
screen02

PR #1151, opened 2024-03-07
shot-put appears to be preferred

matkoniecz:
see https://wiki.openstreetmap.org/wiki/Tag:sport%3Dshot-put

thanks to https://www.openstreetmap.org/changeset/148005491 

note: untested, I will test once bot deploys it

### DIFF
--- a/data/presets/leisure/pitch/athletics/shot_put.json
+++ b/data/presets/leisure/pitch/athletics/shot_put.json
@@ -7,7 +7,7 @@
     "tags": {
         "leisure": "pitch",
         "sport": "athletics",
-        "athletics": "shot_put"
+        "athletics": "shot-put"
     },
     "reference": {
         "key": "athletics",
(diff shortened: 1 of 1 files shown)


matkoniecz:
Seems that at least one other place needs to be changed to get sport label working

screen01
screen

tyrasd (dismissed):

matkoniecz:
Wait. There is https://wiki.openstreetmap.org/wiki/Tag:athletics%3Dshot_put but https://wiki.openstreetmap.org/wiki/Tag:sport%3Dshot-put

matkoniecz:
So iD is right, though tagging situation is silly.

Maybe it is viable to get it better but that would not happen on this issue tracker.

tyrasd:
yeah, was just about to write the same. It's quite confusing that the two values have different spelling, though. :see_no_evil: 

for reference, #817 added these presets

PR #1153, opened 2024-03-10
add ```highway=ladder``` 

### LINKED ISSUE #1054: Add newly approved `highway=ladder` preset
### OSM Tag(s)

highway=ladder

As a Preset

### Label

Highway

### Aliases

_No response_

### Terms

_No response_

### Link to OSM Wiki page

- https://wiki.openstreetmap.org/wiki/Proposal:Ladders
- https://wiki.openstreetmap.org/wiki/Tag:highway%3Dladder

### Status of the Tag

Approved

### Usage of the tag

10

### Replaces other Tag?

From the Proposal: > This deprecates the documented practise to combine highway=path with the property ladder=yes"

### Regional Tag?

global

### Further Information

The newly approved proposal https://wiki.openstreetmap.org/wiki/Proposal:Ladders solves very confusing and potentially dangerous tagging topic on the highway network.

I think even with low usage it is worth adding the preset now because there is no good way to tag this data now.

**Deprecation:**

I also suggest to add a deprecation for the combination of `highway=path+ladder=yes` which was also approved as part of the proposal.

**Geometry:**

We should have a preset for ways and one for notes.

**Icons:**
- Ways: A line with a steps-Icon above? 
   image
- Node: A steps icon?

Temaki does not have a ladder Icon yet; did not check the other sources, yet.

---

This Issue super seeds https://github.com/openstreetmap/id-tagging-schema/issues/494.

qugebert:
Adds ```highway=ladder``` and deprecates ```highway=path + ladder=yes``` 
Closes #1054 

grafik
grafik

### DIFF
--- a/data/presets/highway/ladder.json
+++ b/data/presets/highway/ladder.json
@@ -19,7 +19,6 @@
     ],
     "geometry": [
         "vertex",
-        "point",
         "line"
     ],
     "tags": {
--- a/data/presets/highway/ladder.json
+++ b/data/presets/highway/ladder.json
@@ -1,0 +1,32 @@
+{
+    "icon": "temaki-crossing_markings-ladder",
+    "fields": [
+        "name",
+        "incline_steps",
+        "access_simple",
+        "handrail",
+        "step_count",
+        "surface",
+        "width",
+        "height"
+    ],
+    "moreFields": [
+        "covered_no",
+        "level_semi",
+        "lit",
+        "oneway",
+        "ref"
+    ],
+    "geometry": [
+        "vertex",
+        "point",
+        "line"
+    ],
+    "tags": {
+        "highway": "ladder"
+    },
+    "terms": [
+        "ladder"
+    ],
+    "name": "Ladder"
+}
--- a/data/presets/highway/ladder.json
+++ b/data/presets/highway/ladder.json
@@ -6,7 +6,7 @@
         "access_simple",
         "handrail",
         "step_count",
-        "surface",
+        "material",
         "width",
         "height"
     ],
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -791,6 +791,10 @@
     "replace": {"ford": "*"}
   },
   {
+    "old": {"highway": "path", "ladder": "yes"},
+    "replace": {"highway": "ladder"}
+  },
+  {
     "old": {"highway": "platform"},
     "replace": {"highway": "platform", "public_transport": "platform"}
   },


data/presets/highway/ladder.json
```diff
@@ -0,0 +1,32 @@
...
+        "covered_no",
+        "level_semi",
+        "lit",
+        "oneway",
+        "ref"
+    ],
+    "geometry": [
+        "vertex",
+        "point",
+        "line"
```
tordans, line 23:
```suggestion
        "vertex",
        "line"
```

I think we should only allow points on lines and lines. Not single points.

tordans:
I was wondering how the direction of up/down would look like on the vertex point:

image

I wonder if the forward/backward UI that indicates the direction would be a good fit image. But then again its only[ 28 point ladders (vs 31 ways),](https://taginfo.openstreetmap.org/tags/highway=ladder#overview) yet, so that can come later…

tordans (approved):

data/presets/highway/ladder.json
qugebert: sure, you're right

qugebert:
> I was wondering how the direction of up/down would look like on the vertex point:
> But then again its only[ 28 point ladders (vs 31 ways),](https://taginfo.openstreetmap.org/tags/highway=ladder#overview) yet, so that can come later…

so should i make a seperate preset for vertex without the incline?

tordans:
> so should i make a seperate preset for vertex without the incline?

I think it is all fine. 
- The definition of the vertex as I understand is, that any directional tags are to be considered to use the direction of the attached way. By adding the direction element, we make this explicit to users so they can validate their choices. Right now the user has to check this manually, which makes it hard but not impossible
- Given that the tag is not used too much, yet I don't think splitting it up is necessary ATM as it does increase the overall complexity

data/presets/highway/ladder.json
```diff
@@ -0,0 +1,31 @@
+{
+    "icon": "temaki-crossing_markings-ladder",
+    "fields": [
+        "name",
+        "incline_steps",
+        "access_simple",
+        "handrail",
+        "step_count",
+        "surface",
```
tyrasd, line 9: I'd replace `surface` with `material`, as `surface` does not make much sense conceptually for ladders, and the offered values would be silly (e.g. `unpaved`, `sand`, `ground`, `asphalt`, etc. are pretty unlikely to fit for a ladder :stuck_out_tongue_winking_eye: ).

```suggestion
        "material",
```

tyrasd (approved):

PR #1159, opened 2024-03-13
drop "Portable Toilets" preset

### LINKED ISSUE #1148: "Portable Toilets" preset doesn't match the wiki
**Describe the bug**
[The preset for Portable Toilets](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/toilets/portable.json) adds the tag `portable=yes`, but that tag is not documented and has never been recommended on the [amenity=toilets](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dtoilets) page, so it should probably be changed to `toilets:disposal=tank` to match the wiki page.

tyrasd:
Fixes #1148 and in addition changes the following strings:

* rename generic amenity=toilets preset to "Toilets" (was: "Restroom"), as _restroom_ is typically only used for rooms in a building, but the tag can also be used for outdoor toilets
* rename field for "toilets" tag to "Restroom" (for the opposite reason as above)
* move alias "Water Closet" to "Flush Toilets" preset


tordans (commented):
This looks good to me. I have the open question (last paragraph) if we should remove the portable field…
---
In https://github.com/openstreetmap/id-tagging-schema/issues/1148#issuecomment-2002009354 the language and changes where approved.

---

> rename generic amenity=toilets preset to "Toilets" (was: "Restroom"), as restroom is typically only used for rooms in a building, but the tag can also be used for outdoor toilets

When a toilet is mapped separately inside the building shape, it also get this updated name, now. Which is something we have to live with, since we cannot target the surrounding shape. I think this is fine. Example https://pr-1159--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=n5938578259&locale=en&map=20.00/47.84838/12.07912
- Overpass https://overpass-turbo.eu/s/1MvD
---
Example "flush" https://pr-1159--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=n10284044581&locale=en&map=20.00/47.84891/12.09018
- Overpass https://overpass-turbo.eu/s/1MvE
---
Example "portable" https://pr-1159--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=n9124438348&locale=en&map=20.00/48.11587/12.34760
- Overpass https://overpass-turbo.eu/s/1MvF
- Why do we still have the field image. It did sound like we should rather remove it? Maybe even add a deprecation to it?

matkoniecz:
@tyrasd Can you resolve conflicts that apparently appeared since then?

If that PR is stuck - maybe splitting it into parts and merging them separately would be a good idea?

matkoniecz:
created #1816 for term tweaks, without dropping portable toilet

matkoniecz:
I will close this one as part of change that was extracted to https://github.com/openstreetmap/id-tagging-schema/pull/1816 was protested there and deleting portable toilet file will be easier than resolving conflict here (and new PR can be easily opened for that)

PR #1163, opened 2024-03-16
A few dedicated icons suggestions

qugebert:
See #41 

I found some icons that could replace multiple used icons in some cases 
They are not always a better fit, but none of them is used in another preset so far, so they at least reduce the number of multiple used icons.

Here is a summary of all suggestions
grafik

### DIFF
--- a/data/presets/advertising/column.json
+++ b/data/presets/advertising/column.json
@@ -1,5 +1,5 @@
 {
-    "icon": "temaki-storage_tank",
+    "icon": "roentgen-advertising_column",
     "fields": [
         "{advertising}"
     ],
--- a/data/presets/amenity/food_court.json
+++ b/data/presets/amenity/food_court.json
@@ -1,5 +1,5 @@
 {
-    "icon": "maki-restaurant",
+    "icon": "roentgen-food_court",
     "fields": [
         "name",
         "operator",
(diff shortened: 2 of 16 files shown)


tordans:
I am not sure about the changes. I think we should collect a view example links. 
The old icons are very explicit for their category. 
The new don't describe the category quite that well/quick. 
We gain something from more specific icons if they are used with other icons of the same kind _and_ the category is clear somehow from other features. 
Example: AFAIK for the fitness stations those are often used without a surrounding are AFAIK. So now the category is unclear.

tyrasd:
> The old icons are very explicit for their category. The new don't describe the category quite that well/quick.

This is a valid concern, but it kind of opposes the premise of #41. :see_no_evil: It reminded me of https://github.com/ideditor/schema-builder/issues/135: Maybe the optimal solution for these cases would be to use a slightly generic "category" icon in combination with a more precise specific icon for the sub-type (e.g. for the fitness station: the running person icon in combination with the rings)… :thinking:

qugebert:
> Thanks for these suggestions! I think most of them are reasonable improvements. My only objection would be the [`leisure=swimming_area`](https://wiki.openstreetmap.org/wiki/Tag:leisure=swimming%20area) icon: this tag is meant for for swimming locations in natural bodies of water (e.g. a lake or an ocean), where ladders are not typically found. Can you revert this one back to the previous icon please?

You're right. That icon doesn't really fit, I did a revert for that

tyrasd (approved):

PR #1165, opened 2024-03-17
Create barber.json

### LINKED ISSUE #1161: Add preset for barbershops
### OSM Tag(s)

shop=hairdresser + hairdresser=barber

As a Preset

### Label

Barber

### Aliases

barbers, barbershop

### Terms

_No response_

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:hairdresser%3Dbarber

### Status of the Tag

Approved

### Usage of the tag

300

### Replaces other Tag?

_No response_

### Regional Tag?

_No response_

### Further Information

Recently approved but already in use tagging for barbershops: `shop=hairdresser` + `hairdresser=barber`.

More commonly tagged in OSM using `shop=hairdresser` plus gender access tags (i.e., `male=yes` and optionally `female=no`) but this is problematic for the reasons discussed in the approved proposal.

kjonosm:
closes #1161

### DIFF
--- a/data/presets/shop/hairdresser/barber.json
+++ b/data/presets/shop/hairdresser/barber.json
@@ -36,7 +36,7 @@
         "Barbershop"
     ],
     "reference": {
-    "key": "hairdresser",
-    "value": "barber"
+        "key": "hairdresser",
+        "value": "barber"
     }
 }
--- a/data/presets/shop/hairdresser/barber.json
+++ b/data/presets/shop/hairdresser/barber.json
@@ -1,0 +1,42 @@
+{
+    "icon": "temaki-beauty_salon",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "beard",
+        "moustache",
+        "shave",
+        "shaving",
+        "groom",
+        "grooming",
+        "razor",
+        "hairdresser",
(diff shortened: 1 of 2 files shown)


data/presets/shop/hairdresser/barber.json
```diff
@@ -0,0 +1,42 @@
+{
+    "icon": "temaki-beauty_salon",
```
tyrasd, line 2: optimally, this would have a dedicated icon, but I couldn't find a suitable one in the current icon sets. Maybe something like a [razor](https://www.svgrepo.com/svg/300806/razor-barber) could work…

Dimitar5555: Or a [barber pole](https://stock.adobe.com/292296536)?

tordans (changes requested):
@kjonosm I think this looks quite good pending the search for a better icon (see https://github.com/openstreetmap/id-tagging-schema/pull/1165/files#r1541076843). 

Please add an example link to the two presets using the preview link and check the info-`i` text that is sourced from the osm wikidata items and should describe the features well.

kjonosm:
image

tordans (approved):

Issue #1155, opened 2024-03-10
Sync house preset description between StreetComplete <> iD <> Wiki (DE lang)


tordans:
This is a stub of an issue and more of something to discuss, I guess.
It is only about the DE translations (for now)

I noticed that the new building overlay in StreetComplete uses different description for the building types that I was used to and that iD/the translations from the presets presents different wordings. The wiki is a bit different as well.

I think we should aim to bring those presets more in line so users don't tag something one way with iD and different in SC.

---

The main thing I stumbled on was …

### `building=house`

* SC: <img width="383" alt="image" src="https://github.com/openstreetmap/id-tagging-schema/assets/111561/2b8210ec-23ae-4d8c-858e-b0926b8a7eed">
* iD: <img width="400" alt="image" src="https://github.com/openstreetmap/id-tagging-schema/assets/111561/76080823-1f65-4807-8a5a-4f7601760ffa">
* Wiki: <img width="661" alt="image" src="https://github.com/openstreetmap/id-tagging-schema/assets/111561/7a6c3420-57ef-4073-8788-8207847d6094">

According to SC most houses I mapped as `building=house` with iD should actually be `building=detached`.

1ec5:
Apparently there’s also an alternative style for mapping detached houses, `building=house` `house=detached`, which the House preset even supports via a translatable field.

matkoniecz:
> According to SC most houses I mapped as building=house with iD should actually be building=detached.

Note that tagging them as `building=house` is not wrong, it is only losing a bit of precision.

matkoniecz:
> Suggestions: (1) Add a dashed blue line around the building icons to indicate that those are one geometry representing multiple attached houses. I see the icon and think to map single attached houses like it.

Is suggestion for SC? If yes then I would raise it at https://github.com/streetcomplete/StreetComplete/issues

tordans:
> > According to SC most houses I mapped as building=house with iD should actually be building=detached.
> 
> Note that tagging them as `building=house` is not wrong, it is only losing a bit of precision.

That is not how I read the definition in all three places (iD, SC, Wiki). It is not described as a generic value that can gain precision by using different values but rather a specific type of house.

> Is suggestion for SC? If yes then I would raise it at https://github.com/streetcomplete/StreetComplete/issues

We have an issue that spans platforms. My goal is to use this one place to discuss it and then split it in tasks/issues for those platforms. I don't see anything gained by discussing them in separate issues.

[...]

PR #1167, opened 2024-03-18
remove downranking building=house entry

matkoniecz:


matkoniecz:
see https://github.com/openstreetmap/id-tagging-schema/issues/1155#issuecomment-2003768510

matkoniecz:
Not sure why this downranking was added and what it solved.

matkoniecz:
@DujaOSM It fixes what you reported but tracking down why downranking was added could be useful. Maybe there was a good reason for that?

bhousel:
All buildings presets have `"matchScore": 0.5` and have had it for a long time.

This is because in OSM (unfortunately) buildings tags are often combined with more interesting tags, and the preset system is expected to match the more interesting tag.

For example `building=house`+`craft=plumber`

DujaOSM:
Wow that was quick... 
For specifically residential buildings (house, detached, residential, apartments),  I don't see a reason to downgrade matchScore (much), since those are most commonly used standalone. However, apart from the name, I do not know how exactly matchScore is used: if it only affects string matching, I don't see why we would downgrade it when matching "hou" or "res"...

By the way, the common preset "residential area" also seems to suffer from the same disease...

bhousel:
Off the top of my head, I don't know what will happen.
If you change the matchScores around without understanding what they do, you can probably expect some surprising preset matching behavior.

Yes, you might get "House" to rank higher on the list before "Houseware Store", but you wouldn't want "Church Building" match before "Church" or "Stadium Building" match before "Stadium".  The building tags are weird.

Maybe a better solution would be to prioritize shorter names in the preset list, but I don't know how that would affect different languages.

matkoniecz:
As is it is not ready.

I will try to investigate more here (or in other iD preset area) once https://github.com/openstreetmap/id-tagging-schema/pulls/matkoniecz are processed

PR #1176, opened 2024-03-23
Add presets for ski jump tagging

olafkryus:
Added most important presets for ski jumping venue tagging – the in-run structure, the piste and the take-off edge.

### DIFF
--- a/data/presets/man_made/ski_jump.json
+++ b/data/presets/man_made/ski_jump.json
@@ -1,6 +1,9 @@
 {
     "icon": "temaki-ski_jumping",
-    "fields": [],
+    "fields": [
+        "name",
+        "operator"
+    ],
     "moreFields": [
         "building"
     ],
--- a/data/presets/man_made/ski_jump.json
+++ b/data/presets/man_made/ski_jump.json
@@ -17,12 +17,15 @@
         "in-run",
         "inrun",
         "ski jump structure",
-        "ski jumping hill structure",
-        "ski jump in-run",
-        "ski jumping hill in-run tower"
+        "ski jump in-run tower",
+        "ski jumping hill tower"
     ],
     "tags": {
         "man_made": "ski_jump"
     },
-    "name": "Ski Jumping Hill Structure / In-Run Tower"
+    "name": "Ski Jumping Tower",
+    "aliases": [
+        "Ski Jumping Hill Structure",
+        "Ski Jumping Hill In-Run Tower"
+    ]
 }
--- a/data/presets/man_made/ski_jump.json
+++ b/data/presets/man_made/ski_jump.json
@@ -1,0 +1,25 @@
+{
+    "icon": "temaki-ski_jumping",
+    "fields": [],
+    "moreFields": [
+        "building"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "jump",
+        "ski jump",
+        "in-run",
+        "inrun",
+        "ski jump structure",
+        "ski jumping hill structure",
+        "ski jump in-run",
+        "ski jumping hill in-run tower"
+    ],
+    "tags": {
+        "man_made": "ski_jump"
+    },
+    "name": "Ski Jumping Hill Structure / In-Run Tower"
+}
--- a/data/fields/piste/type.json
+++ b/data/fields/piste/type.json
@@ -13,7 +13,8 @@
             "ice_skate": "Ice Skate",
             "snow_park": "Snow Park",
             "playground": "Playground",
-            "connection": "Connection"
+            "connection": "Connection",
+            "ski_jump": "Ski Jump"
         }
     },
     "autoSuggestions": false,
--- a/data/fields/sport.json
+++ b/data/fields/sport.json
@@ -51,6 +51,7 @@
             "shooting": "Shooting",
             "skateboard": "Skateboard",
             "skiing": "Skiing",
+            "ski_jumping": "Ski Jumping",
             "soccer": "Soccer",
             "softball": "Softball",
             "speedway": "Motorcycle Speedway",
--- a/data/presets/piste/ski_jump.json
+++ b/data/presets/piste/ski_jump.json
@@ -1,0 +1,23 @@
+{
+    "icon": "temaki-ski_jumping",
+    "fields": [
+        "name",
+        "piste/type",
+        "ref",
+        "oneway_yes",
+        "lit"
+    ],
+    "geometry": [
+        "line",
+        "area"
+    ],
+    "tags": {
+        "piste:type": "ski_jump"
+    },
+    "terms": [
+        "jump",
+        "ski jump",
+        "ski jumping hill"
+    ],
+    "name": "Ski Jumping Piste"
+}
--- a/data/presets/piste/take_off.json
+++ b/data/presets/piste/take_off.json
@@ -1,0 +1,21 @@
+{
+    "icon": "temaki-ski_jumping",
+    "fields": [
+        "height"
+    ],
+    "geometry": [
+        "vertex"
+    ],
+    "tags": {
+        "piste:takeoff": "yes"
+    },
+    "terms": [
+        "jump",
+        "ski jump",
+        "take-off",
+        "takeoff",
+        "ski jump take-off",
+        "ski jumping hill take-off"
+    ],
+    "name": "Ski Jumping Hill Take-Off Point"
+}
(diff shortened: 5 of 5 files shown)


data/presets/man_made/ski_jump.json
```diff
@@ -0,0 +1,25 @@
...
+        "inrun",
+        "ski jump structure",
+        "ski jumping hill structure",
+        "ski jump in-run",
+        "ski jumping hill in-run tower"
+    ],
+    "tags": {
+        "man_made": "ski_jump"
+    },
+    "name": "Ski Jumping Hill Structure / In-Run Tower"
```
tyrasd, line 24: maybe this could be slightly shortened to the following?

```suggestion
    "name": "Ski Jumping In-Run/Tower"
```

data/presets/man_made/ski_jump.json
```diff
@@ -0,0 +1,25 @@
...
+        "in-run",
+        "inrun",
+        "ski jump structure",
+        "ski jumping hill structure",
+        "ski jump in-run",
+        "ski jumping hill in-run tower"
+    ],
+    "tags": {
+        "man_made": "ski_jump"
+    },
```
tyrasd, line 23: would it make sense to add `sport=ski_jumping` to these features?

```suggestion
    "tags": {
        "man_made": "ski_jump"
    },
    "addTags": {
        "man_made": "ski_jump",
        "sport": "ski_jumping"
    },
```

data/presets/man_made/ski_jump.json
```diff
@@ -0,0 +1,25 @@
+{
+    "icon": "temaki-ski_jumping",
+    "fields": [],
```
tyrasd, line 3: Perhaps these two fields would make sense?

```suggestion
    "fields": [
        `name`,
        `operator`
    ],
```

data/presets/man_made/ski_jump.json
tyrasd: at second glance, I guess it that tag would be mostly superfluous (as it can be considered implied from the primary tag).

data/presets/man_made/ski_jump.json
olafkryus: While in most of the cases it is true that man-made part is equal to the in-run tower.
It is also not uncommon for the man-made structure to include more than the in-run itself if the terrain is too flat (e.g.: [Holmenkollen](https://upload.wikimedia.org/wikipedia/commons/b/b0/Holmenkollbakken_under_construction_aerial_2.jpg)).
There are also ski jumps on slopes that happen to be just right (e.g. as seen [here](https://upload.wikimedia.org/wikipedia/commons/c/c0/Buskerud_Vikersund_ski_jumping_hills_2010-03-13_K45_K65_0cr.jpg)) and whether they really qualify as "man-made" (or "tower") is debatable.
If shortening of this (quite long I agree) name is desired, I would rather opt for `Ski Jumping Hill Structure`.

data/presets/man_made/ski_jump.json
olafkryus: I wonder whether the structure itself will actually ever have a name of it's own… 😅 But probably better those two fields than nothing. Committing.

olafkryus:
> Thanks, this looks quite good already. See a few suggestions inline.

Thanks for the feedback. \
I've applied one patch, and provided more info on the other – please be the judge ;)

> In addition, could you also add the option for `ski_jumping` to the `sport` field please?

Sure. Done: e2f42db

tordans:
I just reran the actions to the preview is there again.

List of examples cases: https://overpass-turbo.eu/s/1NO4

The last open issue from above was the naming. I agree it should be shorter. It is two lines now and hard to understand:
> image

Please move relevant search terms to the `aliases` or `terms` https://github.com/ideditor/schema-builder?tab=readme-ov-file#preset-schema

Please also check if the OSM Wikidata item can be improved to better explain the tag usage (its in the screenshot).
Here: https://wiki.openstreetmap.org/wiki/Item:Q16841

One thing I am unclear about is the tagging as area vs. line.

- This example seems to work as intended https://pr-1176--ideditor-presets-preview.netlify.app/id/dist/#background=EsriWorldImagery&disable_features=boundaries&id=w187922757&locale=en&map=19.94/47.89814/8.11134

However when mapped as a line…
- there is an error to remove tags
- the preset does not apply
- the help text and icon is different

image
 
Examples:
- https://pr-1176--ideditor-presets-preview.netlify.app/id/dist/#background=EsriWorldImageryClarity&disable_features=boundaries&id=w270557257&locale=en&map=20.00/48.73621/10.10972
- https://pr-1176--ideditor-presets-preview.netlify.app/id/dist/#background=swisstopo_swissimage&disable_features=boundaries&id=w794427489&locale=en&map=20.00/46.50609/7.68285

Another thing is, why the `piste:type` is not a field when using like https://pr-1176--ideditor-presets-preview.netlify.app/id/dist/#background=EsriWorldImagery&disable_features=boundaries&id=w187922757&locale=en&map=19.94/47.89814/8.11134 – Is this tagging wrong or is the field missing?

tordans (changes requested):
(Formal PR status; see last comment.)

olafkryus:
Thanks for the feedback.

> The last open issue from above was the naming. I agree it should be shorter. It is two lines now and hard to understand

I've changed it now to a simple "Ski Jumping Tower". I hope the additional info will be enough for the more unusual cases.

(I've also went with it and shortened "Ski Jumping Hill Take-Off Point" to just "Ski Jumping Take-Off")

> Please move relevant search terms to the `aliases` or `terms`

I've made some changes to that, I hope they are in-line with what was expected.

> Please also check if the OSM Wikidata item can be improved to better explain the tag usage

I've reworded descriptions for both [Q16841](https://wiki.openstreetmap.org/wiki/Item:Q16841) and [Q21202](https://wiki.openstreetmap.org/wiki/Item:Q21202), I hope they are a bit more clear now.

> One thing I am unclear about is the tagging as area vs. line

> Another thing is, why the piste:type is not a field when using like https://pr-1176--ideditor-presets-preview.netlify.app/id/dist/#background=EsriWorldImagery&disable_features=boundaries&id=w187922757&locale=en&map=19.94/47.89814/8.11134 – Is this tagging wrong or is the field missing?

As for the tagging:

* `piste:type=ski_jump` is supposed to span the whole of the "piste", the path the athlete travels, so that would be from the very top of the in-run to the very end of the out-run at the bottom. A piste can be tagged either as a line following the "path" of travel, or as an outline of the whole area available to the athlete.
* `man_made=ski_jump` is supposed to cover only the outline of the man-made structure. So most of the time it will cover only the outline of the in-run tower, or it will be omitted completely if the ski jumping hill is "natural" (the in-run follows the curve of a fortunately shaped slope)

Therefore, it follows that, in almost all of the cases, the above two tags should be applied to separate elements.

So, the examples provided above seem to be wrongly tagged, and that seems to be the source of the mentioned problems.

I hope the link below will serve as a better example (and a more interesting, informative one, since the situation is more complex as the out-run doubles as a football pitch in the summer):

https://pr-1176--ideditor-presets-preview.netlify.app/id/dist/#background=EsriWorldImagery&disable_features=boundaries&id=w1223388311&locale=en&map=18/47.32155/11.17741

tordans:
> As for the tagging: (…)
> Therefore, it follows that, in almost all of the cases, the above two tags should be applied to separate elements.
> So, the examples provided above seem to be wrongly tagged, and that seems to be the source of the mentioned problems.

Thanks for explaining, @olafkryus. I wonder what the best way to handle this would be. Would you be up for creating a MapRoulette Challenge for those mistagged features, so the data gets fixed? My concern is that iD will display this very prominent message, but the actions on the message are wrong. And AFAIK, it is not easy to change those messages because they are autogenerated…

olafkryus:
> I wonder what the best way to handle this would be. Would you be up for creating a MapRoulette Challenge for those mistagged features, so the data gets fixed? My concern is that iD will display this very prominent message, but the actions on the message are wrong. 

@tordans: Fortunately, there were "only" about a few dozen of such cases worldwide, therefore I decided to fix them all on my own on a case-by-case basis.

So, it shouldn't be an issue anymore: https://overpass-turbo.eu/s/1NRb

tordans (approved):

tordans:
> Fortunately, there were "only" about a few dozen of such cases worldwide, therefore I decided to fix them all on my own on a case-by-case basis.

Lucky me to pick two of them right away :-D. Thanks for the cleanup!

---

I think this looks good now and ready to be merged.

PR #1179, opened 2024-03-26
try to make description less misleading for cycleway contraflow traffic

matkoniecz:
1) cycleway=opposite may be used also where there is no dedicated separate bicycle lane (but contraflow traffic is still allowed)

2) cycleway=opposite_lane is nowadays treated as inferior to tagging oneway:bicycle=no + (for example) cycleway:left=lane so it should not be phrased in way suggesting it to be preferable for cases where dedicated contraflow lane exists

Opening in draft as I would really want native English speaker to look at it (@1ec5 @zelonewolf ?)

### DIFF
--- a/data/fields/cycleway.json
+++ b/data/fields/cycleway.json
@@ -37,12 +37,12 @@
                 "description": "A bike lane shared with a bus lane"
             },
             "opposite_lane": {
-                "title": "Opposite Bike Lane",
-                "description": "A bike lane that travels in the opposite direction of traffic"
+                "title": "Contraflow Bike Traffic",
+                "description": "Cyclisys may travel in both directions on a one-way street"
             },
             "opposite": {
-                "title": "Contraflow Bike Lane",
-                "description": "A bike lane that travels in both directions on a one-way street"
+                "title": "Contraflow Bike Traffic",
+                "description": "Cyclisys may travel in both directions on a one-way street"
             },
             "separate": {
                 "title": "Cycleway Mapped Separately",
--- a/data/fields/cycleway.json
+++ b/data/fields/cycleway.json
@@ -38,11 +38,11 @@
             },
             "opposite_lane": {
                 "title": "Contraflow Bike Traffic",
-                "description": "Cyclisys may travel in both directions on a one-way street"
+                "description": "Cyclists may travel in both directions on a one-way street"
             },
             "opposite": {
                 "title": "Contraflow Bike Traffic",
-                "description": "Cyclisys may travel in both directions on a one-way street"
+                "description": "Cyclists may travel in both directions on a one-way street"
             },
             "separate": {
                 "title": "Cycleway Mapped Separately",


data/fields/cycleway.json
```diff
@@ -37,12 +37,12 @@
...
-                "title": "Opposite Bike Lane",
-                "description": "A bike lane that travels in the opposite direction of traffic"
+                "title": "Contraflow Bike Traffic",
+                "description": "Cyclisys may travel in both directions on a one-way street"
             },
             "opposite": {
-                "title": "Contraflow Bike Lane",
-                "description": "A bike lane that travels in both directions on a one-way street"
+                "title": "Contraflow Bike Traffic",
+                "description": "Cyclisys may travel in both directions on a one-way street"
```
ZeLonewolf, line 45: Spelling error - "cyclists"

data/fields/cycleway.json
```diff
@@ -37,12 +37,12 @@
                 "description": "A bike lane shared with a bus lane"
             },
             "opposite_lane": {
-                "title": "Opposite Bike Lane",
-                "description": "A bike lane that travels in the opposite direction of traffic"
+                "title": "Contraflow Bike Traffic",
+                "description": "Cyclisys may travel in both directions on a one-way street"
```
ZeLonewolf, line 41: Spelling error - "cyclists"

ZeLonewolf:
In US English "contraflow" is normally used when automobile traffic is routed on a stretch of road that normally goes in a different direction.  In some cities, there are "contraflow lanes" that are opened at high-traffic times - for example, an expressway with 4 lanes on each side, might have 5 lanes inbound in the morning and 3 outbound -- the one lane that got shifted would be the contraflow lane.

ZeLonewolf:
In this context, I would find "two-way traffic" to be a more understandable way to say "bicycles can travel in both directions".

ZeLonewolf:
It does seem that "contra-flow" in the way it's used here is valid and accurate terminology though, based on what I'm finding online:

https://nacto.org/publication/urban-bikeway-design-guide/bike-lanes/contra-flow-bike-lanes/

matkoniecz:
Yep, "contraflow" should be fine. I was worried about Traffic part in "Contraflow Bike Traffic"

ZeLonewolf:
"Cyclists may travel in both directions on a one-way street" makes sense then.  In fact, the old description is wrong because it implies that cyclists can only ride towards car traffic.

data/fields/cycleway.json
matkoniecz: thanks for spotting!

matkoniecz:
`oneway:bicycle=no` means that cyclists can travel in both directions on this road

it does not mean that they can travel in both direction on every single lane. For example, there may be only a contraflow lane.

Note that `highway=tertiary cycleway=track oneway=yes oneway:bicycle=no` may be also used to map as one object oneway road with cycleway along it

> oneway needs to be scoped to the lane as well: cycleway:left:oneway=no

why you think so?

> should allow biking on the centerline

what you mean by that?

> Unfortunately the wiki is now tell maintained in this area.

I do not understand this sentence

tordans:
@matkoniecz FYI: I deleted my comment to not move this discussion away from this PR and also don't have the energy to look into it more. I suggest you hide your comment so your reply is not confusing because of my deleted comment…

matkoniecz:
I have hidden it and I am fine with them being deleted.

tordans:
Update: There is https://wiki.openstreetmap.org/wiki/Proposal:Deprecate_cycleway%3Dopposite_family being discussed which would make this PR obsolete. Please comment over there if you think we should keep those values for some reason.

tordans:
@matkoniecz I suggest we close this and work on https://github.com/openstreetmap/id-tagging-schema/issues/1271 instead. Please ping me if I should reopen.

PR #1181, opened 2024-03-28
Add terms to `historic = wayside_cross` preset

FloEdelmann:
In German, "Wayside Cross" is translated to "Wegkreuz", so the preset is not findable with the query "Kreuz". Adding a translatable term "cross" would mitigate that. See https://github.com/streetcomplete/StreetComplete/issues/5557.

### DIFF
--- a/data/presets/historic/wayside_cross.json
+++ b/data/presets/historic/wayside_cross.json
@@ -19,9 +19,9 @@
         "historic": "wayside_cross"
     },
     "terms": [
-        "cross", 
-        "Christian", 
-        "Jesus", 
+        "cross",
+        "Christian",
+        "Jesus",
         "Christus"
     ],
     "name": "Wayside Cross"
--- a/data/presets/historic/wayside_cross.json
+++ b/data/presets/historic/wayside_cross.json
@@ -19,10 +19,10 @@
         "historic": "wayside_cross"
     },
     "terms": [
+        "christ"
+        "christian",
         "cross",
-        "Christian",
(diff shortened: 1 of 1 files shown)


tordans:
Is "Wegkreuz", "Kreuz" part of the German translations already?

data/presets/historic/wayside_cross.json
```diff
@@ -18,5 +18,8 @@
     "tags": {
         "historic": "wayside_cross"
     },
+    "terms": [
+        "cross"
```
rugk, line 22:
```suggestion
        "cross", 
        "Christian", 
        "Jesus", 
        "Christ"
```

These should alow be added as indicated in the issue? As at least in Germany it's usually a Jesus statue being at the cross.

rugk (changes requested):

tordans:
German translations belong in the translation system, not the repo, see https://github.com/openstreetmap/id-tagging-schema/blob/main/CONTRIBUTING.md#translating for more. Short version: Only EN is part of the repo, all other languages are managed in the translations system. The repo should only have EN (with very few exceptions if any for country specific presets/fields).

FloEdelmann:
But the terms I added are not German or German-specific. They are added in English and should be translated to all languages via Transifex. Do you disagree with that, or did I miss something?

tyrasd:
I changed [Christus](https://dict.leo.org/englisch-deutsch/Christus) to Christ. lgtm

Issue #1177, opened 2024-03-23
Ballet studios / dancing schools


FloEdelmann:
I searched for "Ballettstudio" (in German) and nothing was found. In a [forum thread](https://community.openstreetmap.org/t/ballett-schule/86169) I found that I should rather search for dancing school ("Tanzschule" in German) and add ballet as a dancing style there. This worked. So **first suggestion:** Can we add "ballet studio" as a term for the Dancing School preset?

However, the Dancing School preset (`amenity = dancing_school`) was marked as deprecated and suggested replacing it with the Dance Hall preset preset (`leisure = dance` + `dance:teaching = yes`), which was fine for me. It seems wrong to have a suggested preset that is marked as deprecated though. So **second suggestion:** Should we remove the Dancing School preset and add its terms to the Dance Hall preset?

Note also that there is an issue about a wrong deprecation of `amenity = dancing_school`:

* https://github.com/openstreetmap/id-tagging-schema/issues/373

So maybe the deprecation is wrong at all?

simonpoole:
This is a leftover of the previous reign of maintainers were the policy was "we tag everything in our myopic view and basta".  There is unluckily still a lot of legacy from that period left (for example office=coworking).

bhousel:
> This is a leftover of the previous reign of maintainers were the policy was "we tag everything in our myopic view and basta". There is unluckily still a lot of legacy from that period left (for example office=coworking).

It's really uncool of you to write stuff like this.   Be better, Simon.

For the curious, here's a link to the thing you are still mad about:  https://github.com/openstreetmap/iD/pull/6251

TLDR - 5 years ago, @westnordost sent me a PR to deprecate a thing based on what the wiki said, and I said "sure".   Somehow that makes me a tag tyrant in Simon's view.

simonpoole:
That was not about you falling indirectly for a notorious wiki fiddler, but you refusing to fix the error once it was pointed out to you.

tyrasd:
@FloEdelmann 

> second suggestion

Yeah, there should never be a preset for a tag which is also in the "deprecated" tags list.

[...]

PR #1182, opened 2024-03-28
Add terms to Dance Hall / Dance School presets

FloEdelmann:
For **first suggestion** of #1177.

### DIFF
--- a/data/presets/leisure/dancing_school.json
+++ b/data/presets/leisure/dancing_school.json
@@ -16,14 +16,13 @@
         "area"
     ],
     "terms": [
+        "ballet",
+        "ballet studio",
+        "dance teaching",
         "jive",
         "swing",
         "tango",
-        "waltz",
-        "ballet",
-        "dance teaching",
-        "ballet studio",
-        "dance studio"
+        "waltz"
     ],
     "tags": {
         "amenity": "dancing_school"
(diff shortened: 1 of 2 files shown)


data/presets/leisure/dancing_school.json
```diff
@@ -20,7 +20,10 @@
         "swing",
         "tango",
         "waltz",
-        "dance teaching"
+        "ballet",
+        "dance teaching",
+        "ballet studio",
+        "dance studio"
```
FloEdelmann, line 26: Since there is an alias called "Dance Studio" here, does it make sense to add it as a term as well?

tyrasd: No, it's not necessary to repeat the aliases here. Let's drop it here.

tyrasd (changes requested):
the additional search terms look reasonable to me. However, the `terms` should be sorted alphabetically. Can you please change that?

data/presets/leisure/dancing_school.json
FloEdelmann: Thanks for clarifying. Removed in [`6cf3fe4`](https://github.com/openstreetmap/id-tagging-schema/pull/1182/commits/6cf3fe4611f48a3ddcfee8e1e8c2b24a6fc7d0eb).

tyrasd (approved):

PR #1184, opened 2024-03-29
_emergency.json: Added check_date to fields

jpds:
This is used by consumers such as OpenAEDMap.

### DIFF
--- a/data/presets/_emergency.json
+++ b/data/presets/_emergency.json
@@ -1,6 +1,7 @@
 {
     "fields": [
-        "emergency_combo"
+        "emergency_combo",
+        "check_date"
     ],
     "geometry": [
         "point",


tyrasd:
The `check_date` field has the `"universal": true` flag, and is therefore already available for all presets. This change would only affect AED features which do not have this tag already (then the universal field is "hidden" under "more fields"). Was is the intent of this PR to make it visible also if unset?

I'm just wondering why this preset should be handled differently compared to other features, like other emergency infrastructure like fire hydrants for example.

jpds:
> Was is the intent of this PR to make it visible also if unset?

Yes, I hadn't thought about the other emergency things - do I just move it into `_emergency.json` ?

tyrasd:
good idea. let's try adding it to the generic `_emergency.json` preset

matkoniecz (approved):

matkoniecz (changes requested):
on testing it does not actually seem to change things?

if AED is not tagged with check date (see say https://www.openstreetmap.org/edit?editor=id#map=24/50.02552718/19.90116060 ) it can be added as extra field but is not shown by default

I the same behavior after this change, and while AED with check date tagged already will show its field the same applies now, see https://pr-1184--ideditor-presets-preview.netlify.app/id/dist/#locale=en&disable_features=boundaries&map=21.87/50.06995/20.04129&background=Bing&id=n12141941802

what exactly is expected to be changed by this PR?

matkoniecz:
what exactly is expected to be changed by this PR?

can anyone point out how behaviour changes compared to current iD behaviour? Or is it influencing other editors?

jpds:
This was so long ago, but the whole point was to display the check date on these entries by default so that users would fill it in.

matkoniecz:
when I tested in on AED it was still not shown by default if it was empty
screen-2026-01-02-13-07-08

matkoniecz:
@jpds ping?

matkoniecz:
Closing as abandoned and not actually working

PR #1193, opened 2024-04-06
Do not enforce Cycle & Foot Path or Cycle & Foot Crossing in Norway

balchen:
Do not enforce Cycle & Foot Path or Cycle & Foot Crossing in Norway (NO) due to the NO community not applying bicycle=designated on highway=cycleway. See https://wiki.openstreetmap.org/wiki/No:Map_Features#Sykkel for reference.

Also fix an apparent bug in that DE is excluded from Cycle & Foot Path, but not from Cycle & Foot Crossing.

### DIFF
--- a/data/presets/highway/cycleway/crossing/bicycle_foot.json
+++ b/data/presets/highway/cycleway/crossing/bicycle_foot.json
@@ -4,8 +4,10 @@
             "fr",
             "lt",
             "pl",
+            "de",
             "il",
-            "ps"
+            "ps",
+            "no"
         ]
     },
     "icon": "temaki-ped_cyclist_crosswalk",
--- a/data/presets/highway/cycleway/bicycle_foot.json
+++ b/data/presets/highway/cycleway/bicycle_foot.json
@@ -6,7 +6,8 @@
             "pl",
             "de",
             "il",
-            "ps"
+            "ps",
+            "no"
         ]
     },
     "icon": "temaki-pedestrian_and_cyclist",


tordans:
FYI we discussed an update path to the current crossing presets during the last iD community meetup. This hat not been properly documented nor implemented, though. I think we should first work on this cleanup before adding more complexity on top of it.

balchen:
> FYI we discussed an update path to the current crossing presets during the last iD community meetup. This hat not been properly documented nor implemented, though. I think we should first work on this cleanup before adding more complexity on top of it.

I'm sorry, I'm not sufficiently involved in the iD community to understand what you wrote. 

However, the change I'm proposing doesn't appear to add any complexity. I'm just adding another country to an existing list of countries. Could you explain the added complexity?

balchen:
It's a relatively straight-forward change, so I imagine the review will be quick and simple. I don't know the process to move this forward. Is there something that needs to be done?

data/presets/highway/cycleway/crossing/bicycle_foot.json
```diff
@@ -4,8 +4,10 @@
             "fr",
             "lt",
             "pl",
+            "de",
```
tordans, line 7: Good catch! I created https://github.com/openstreetmap/id-tagging-schema/pull/1204 to improve this situation

balchen: Great. Does that mean there's progress on this PR as well?

balchen:
Any progress on approving this? It's quite simple, and it seems it's already been reviewed.

tordans:
For me, this is [the kind of PR](https://github.com/openstreetmap/id-tagging-schema/pull/1230/files#diff-87a0c4623bffa64c60b62a39eeab1c40bd0f329cebc177f5e4492555fb5d42aeR57-R73) I would like @tyrasd to review and merge. I was also hoping for https://github.com/openstreetmap/id-tagging-schema/pull/1201 to be merged first and then apply this change on top of it. But the other way around is possible to, of course.

balchen:
I don't mind who reviews it. Having tagged someone who can or should will hopefully speed this up. Thank you.

tordans:
@tyrasd since this is kind of blocking https://github.com/openstreetmap/id-tagging-schema/pull/1384 could you have a look at this. I don't have the time ATM to look into it in detail. A view notes that I did not investigate futher
- I did not find a PR preview (comment); did not understand why there is none / how to trigger it; so no preview is possible
- I am really sure this PR does what balchen described in the description. If the goal is to remove the `bicycle=designated`, which disable the whole preset? Doesn't that make the situation worse because now no presets are present at all? Or… which preset is triggered instead? – I guess we need example ways for that.

The bigger question here might be, if we should continue to add the `bicycle=designated` on `highway=cycleway`. AFAIK there is consensus that it is the implied default. We make it more explicit in those cases where foot+bike is involved by I don't think routers or data consumers actually need it. So maybe we should remove that alltogether… would that resolve the situation?

tordans:
Once the PR preview is there, please use the testing template to document testcases https://github.com/openstreetmap/id-tagging-schema/blob/main/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L24-L70

balchen:
The goal is not to remove `bicycle=designated`, but to not enforce it for combined footways and cycleways for NO.

For combined footways and cycleways, i.e. `highway=cycleway` with `foot=designated`, there is not currently a consensus that `bicycle=designated` is superfluous.

Some countries have decided not to add `bicycle=designated` to combined footways and cycleways, and I assume these countries are in the exclusion list, except NO.

I don't see that NO is any different from the other countries. Does DE's presence on the exclusion list make the situation worse for DE?

tordans:
@balchen I still might be misunderstanding this, sorry if this is the case. But the wiki and your commend make me think that this PR is not what we want to do…

I rebased this PR in order to get the preview going, so we can look at the results of this PR.
Eg https://pr-1193--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w963473070&locale=en&map=16.96/59.05554/6.65107

- [Now…](https://pr-1193--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w963473070&locale=en&map=16.96/59.05554/6.65107)  ==> The way is marked as cycle way
- [Before…](https://ideditor.netlify.app/#background=Bing&disable_features=boundaries&id=w963473070&locale=en&map=16.96/59.05554/6.65107) ==> The way is a marked as foot + cycle way

My understanding is, that you only want the `bicycle=designated` to not be used. But this change makes the whole preset disappear and as such all shared foot+bike ways being presented as bike ways.

Those regional presets are used in countries, where the recommendation is to use `highway=path` instead of `cycleway` for shared ways. (And even then having the regional presets is not ideal because the tagging is still valid, just not recommended … and will now show in iD ATM, which I is what I thinking about in https://github.com/openstreetmap/id-tagging-schema/issues/1205)

I think removing the shared preset altogether makes for a worse user experience and will lead to less consistent tagging of infrastructure.

balchen:
## Test-Documentation

https://pr-1193--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w41372170&locale=en&map=18.85/58.93795/5.75685

### Search

Available cycle presets for way in NO:

image

### Info-`i`

Not relevant.

### Wording

Not relevant.

balchen:
> My understanding is, that you only want the `bicycle=designated` to not be used. But this change makes the whole preset disappear and as such all shared foot+bike ways being presented as bike ways.

The goal of the NO community is to avoid that iD users apply a preset that creates tagging which is not according to schema for NO.

> Those regional presets are used in countries, where the recommendation is to use highway=path instead of cycleway for shared ways.

Which regional presets are you referring to?

balchen:
Ref #1205, it occurs to me that the ideal solution is for a preset to read/apply different tags in different regions. That would allow for a unified user experience while maintaining regional differences. The concept would probably be applicable to many other cases.

tordans:
@balchen but now there is no preset for this case from your wiki at all 

https://wiki.openstreetmap.org/wiki/No:Map_Features#Sykkel

> image

I don't think that is something that we should be doing. Removing a whole preset just to prevent a superfluous tag – which, btw is currently used a lot in Norway (maybe due to Id, but still). The one superfluous does less harm to the data than having no preset for this case so iD users have to come up with their own interpretation of how to tag this.

> > Those regional presets are used in countries, where the recommendation is to use highway=path instead of cycleway for shared ways.

> Which regional presets are you referring to?

Shared bicycle and footway have different recommended tagging in different countries. In some its `highway=path + foot=designated + bicycle=designated` ([Preset missing](https://github.com/openstreetmap/id-tagging-schema/pull/1384/files#diff-26070eb517fce91f8cacd55ea982dd48c2aab33fd87c2e13d6ac0cc1cd05ecc6R21-R23)). In some it `highway=cycleway + foot=designated + (bicylce=designated)` ([Preset](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/highway/cycleway/bicycle_foot.json)).

tordans:
I created https://github.com/openstreetmap/id-tagging-schema/issues/1411 to discuss the solution that would IMO resolve this issue in a better way.

balchen:
You're right in that there is no preset that matches the desired tagging. If this is desirable, we'd need to create a new preset for this case, then. Do you know how this solved for DE? Or is this an issue there as well?

tordans:
> Do you know how this solved for DE? Or is this an issue there as well?

DE is one of the countries that recommend the hw=path tagging pattern, so this hw=cycleway preset is hidden and the other one is there. (But the other one is incomplete, which is a different story/PR.)

---

I will close this now, since it does not have the desired effect. Lets look see what feedback is given to  #1411 .

PR #1194, opened 2024-04-06
emergency/bleed_control_kit.json: Added.

jpds:


tyrasd (commented):
See wiki: https://wiki.openstreetmap.org/wiki/Tag:emergency=bleed_control_kit

Tag usage is still very low with 74 occurrences in OSM. Probably mostly because these are still quite a rare thing to find on the ground. :thinking:

tyrasd:
Thinking about it: Are these not just fancy [first aid kits](https://wiki.openstreetmap.org/wiki/Tag:emergency=first_aid_kit)? :thinking:

jpds:
They are fancy first aid kits, which of course have a slightly different purpose:

- https://www.theguardian.com/uk-news/2021/sep/14/call-for-bleed-control-kits-to-be-rolled-out-across-uk
- https://turtledefibcabinets.co.uk/product-category/bleed-control/

tyrasd:
Hmm… It's still hard for me to understand what exactly the distinction is between these two tags. Would a mapper have to look into the contents of a "emergency response kit" to know which tag to use? The wiki page(s) don't really answer this at the moment.

It seems like this tag was just silently introduced by a few mappers in the UK. Which is fine! But the inclusion in this repo would be endorsing it a bit too much for the current status of this tag. Maybe the best next step would be to create a tagging proposal for it and see if it results in a good definition of the tag(s).

Issue #1015, opened 2023-09-24
Add key community_centre:for=*


AntMadeira:
### OSM Tag(s)

community_centre:for=*

### How would you like this tag to see supported?

As a Field

### Label

For

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Key:community_centre:for

### Status of the Tag

De Facto

### Usage of the tag

14,000

### Replaces other Tag?

No

### Regional Tag?

global

### Further Information

I request adding this key to the `community_centre` preset.

tiptoptom:
@tyrasd would you accept a PR?

tordans:
[Based on search](https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema+%22%3Afor%5C%22%22&type=code): We are using it in one place and we are using the corresponding tag `social_facility:for` in multiple places so I don't see why we would not. Unless we find some dispute on the wiki (…) to not use it in some situations…

Where would you want to add it?

tiptoptom:
As far as I can see, the wiki does not restrict the use of `community_centre:for=*`.
These are the most common `community_centre=*` values in combination with `community_centre:for=*`.
```
values           count
youth_club           9
social_centre       30
senior_centre       76
events_venue        81
meeting_room       101
family_centre      126 *
village_hall       156
cultural_centre    193 **
parish_hall        319
community_hall     383 *
club_home          836 **
youth_centre      1590 *
```
 \* [`community_centre:for=*` is recommended by the wiki](https://wiki.openstreetmap.org/wiki/Key:community_centre#Possible_values)
\** For those I would also find `community_centre:for=*` useful, as the target group is otherwise unclear.

Issue #605, opened 2022-09-30
Feature Request: Add cultural_centre preset


EvanCarroll:
### Description

Currently OSM supports [Cultural Centers](https://en.wikipedia.org/wiki/Cultural_center) with

```
amentiy=community_centre
community_centre=cultural_centre
```

You can see [tag:community_center](https://wiki.openstreetmap.org/wiki/Key:community_centre) in the wiki.

With the further optional qualification 

```
community_center:for=*
```

I think it would be great to implement this one.

matkoniecz:
actually, implemented already

https://github.com/openstreetmap/id-tagging-schema/blob/93113eccdddcedb502d951dc206c7add951b1d68/data/presets/amenity/community_centre/cultural_centre.json#L16

PR #1200, opened 2024-04-24
Community centre

tiptoptom:
CC: #1015 #605


tordans (changes requested):
@tiptoptom in general this looks good. Could you please add some context to your PR to make it easer to review and also add some example links to those places using the preview that was linked in the PR by the github action. Please add usage stats on the given presets as well.

For example, why did you add separate presets per center-type? They all use the same fields and icon of their parent, so we might as well just use the parent + the new field?

One specific think I wonder is: You reference https://github.com/openstreetmap/id-tagging-schema/issues/1015 which talks about https://wiki.openstreetmap.org/wiki/Key:community_centre:for but your PR does not use this tag. Is this because you consider this something to be added later or are there other reasons?

Thanks for this PR!

---

Update: Two more things we need to check
- What does the search look like with those presets, are there duplicated results?
- Are there good Wikidata items already, so we get a nice info-(i) text? (Please add screenshots) (See https://github.com/openstreetmap/id-tagging-schema/pull/1239/files#diff-18813c86948efc57e661623d7ba48ff94325c9b5421ec9177f724922dd553a35R13)

tiptoptom:
> For example, why did you add separate presets per center-type? They all use the same fields and icon of their parent, so we might as well just use the parent + the new field?

I have made various presets, as some of the names have little to do with "community center". For example, a club home of a sports club would not be called a community center but a club home.

tiptoptom:
> One specific think I wonder is: You reference #1015 which talks about https://wiki.openstreetmap.org/wiki/Key:community_centre:for but your PR does not use this tag. Is this because you consider this something to be added later or are there other reasons?

I just wanted to put them in copy, as it is something similar.

tiptoptom:
> add some example links to those places using the preview that was linked in the PR by the github action

community_centre = village_hall
https://pr-1200--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w428990719&locale=en&map=20.00/46.58519/14.41819

community_centre = club_home
https://pr-1200--ideditor-presets-preview.netlify.app/id/dist/#background=nrw_vdop_wms&disable_features=boundaries&id=n6840341676&map=20.00/50.95895/5.37044

community_centre = cultural_centre
https://pr-1200--ideditor-presets-preview.netlify.app/id/dist/#background=nrw_vdop_wms&disable_features=boundaries&id=n8879633245&map=20.00/52.05062/8.59420

tiptoptom:
> Please add usage stats on the given presets as well.


https://taghistory.raifer.tech/#***/amenity/community_centre&***/community_centre/club_home&***/community_centre/cultural_centre&***/community_centre/community_hall&***/community_centre/family_centre&***/community_centre/parish_hall&***/community_centre/village_hall

---

> What does the search look like with those presets, are there duplicated results?

image

tordans:
@tiptoptom do you know more about the values that the "most used values" feature of the dropdown shows?
Eg. is `club_house` the same as the `club_home`. 
It is always a bit confusing for users if those values are visible in the list but not explained/translated but others are.

image

tordans:
One thing we should do before merging is improving the info-`i` text:

Eg image links to https://wiki.openstreetmap.org/wiki/Item:Q6740 
But it should link to the specific sub tag.

You might need to add new wikidata items for those. Or add an explicit reference https://github.com/ideditor/schema-builder?tab=readme-ov-file#reference to the sub presets.

tordans:
@tiptoptom what are your thoughts on adding the sub-group-dropdown to the sub-presets as well?
I always find it a bit weird when I come from a preset, select a sub-preset via dropdown and then the dropdown disappears. It is hard to understand how to go back than. – This is not an issue if the primary usage will be selecting the sub-preset via search. However your graph showed a lot of usage without sub tags, so the first case is one that will happen a lot, right?

Right now it looks like this. The additional dropdown could be the first item on this fieldset
image

tiptoptom:
> @tiptoptom do you know more about the values that the "most used values" feature of the dropdown shows? Eg. is `club_house` the same as the `club_home`. It is always a bit confusing for users if those values are visible in the list but not explained/translated but others are.

I think I just took the values that have their own wiki page. Should we perhaps add the values from [this table](https://wiki.openstreetmap.org/wiki/Key:community_centre#Possible_values)?

tiptoptom:
> You might need to add new wikidata items for those.

How can I do this?

> Or add an explicit reference https://github.com/ideditor/schema-builder?tab=readme-ov-file#reference to the sub presets.

Do I understand correctly that this would be the alternative to a newly created wikidata item?

tiptoptom:
> @tiptoptom what are your thoughts on adding the sub-group-dropdown to the sub-presets as well?

Sounds reasonable :+1:

tordans:
> > You might need to add new wikidata items for those.
> 
> How can I do this?

There are docs on this worked on in https://github.com/openstreetmap/id-tagging-schema/pull/1257/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R52 

> 
> > Or add an explicit reference https://github.com/ideditor/schema-builder?tab=readme-ov-file#reference to the sub presets.
> 
> Do I understand correctly that this would be the alternative to a newly created wikidata item?

The reference is needed to tell the editor which key to use when a user clicks on the info-i.
The data that is then shown is that of the wikidata item.
So in this case, where the lookup should use the secondary not the primary key, I think both is needed.

tiptoptom:
I think I've worked through almost everything now and only the following wikidata items are missing:

- [ ] community_center=club_home
- [ ] community_center=community_hall
- [ ] community_centre=family_centre

kjonosm:
I've added data items for [`community_center=community_hall`](https://wiki.openstreetmap.org/wiki/Item:Q22571) and [`community_centre=family_centre`](https://wiki.openstreetmap.org/wiki/Item:Q22570)

tordans:
This looks good, thanks for the updates!

The one thing that is not ideal, still, is the club stuff.

1. The fact, that we list `club_house` as a raw values is not ideal, see  https://github.com/openstreetmap/id-tagging-schema/pull/1200#issuecomment-2155816194
However, users should still be nudged to the `club_home` since that is translated.

2. The fact that the wiki page for `club_home` https://wiki.openstreetmap.org/wiki/Tag:community_centre=club_home is strongly talking about some dispute. However, the main page https://wiki.openstreetmap.org/wiki/Key:community_centre does not mention this dispute, so I think this is fine as is.
`club_home` is also the third most used values ATM https://taginfo.openstreetmap.org/keys/community_centre#values, so I think it is fine to document it as we do now.

data/presets/amenity/community_centre/lgbtq.json
```diff
@@ -21,5 +28,8 @@
         "lgbtq": "primary",
         "community_centre:for": "lgbtq"
     },
+    "reference": {
+        "key": "community_centre:for"
+    },
```
tordans, line 33: It looks like this is not ideal, yet.

Looking at this example https://pr-1200--ideditor-presets-preview.netlify.app/id/dist/#background=nrw_vdop_wms&disable_features=boundaries&id=n1791368627&map=20.00/52.47851/13.42123 the help text does not fit well

image

Overpass query: https://overpass-turbo.eu/s/1OL3

I did not research in the Wiki but I assume there should be a better page that describes this preset? Or, if not, lets remove the reference and fall back to the main key.

tordans (changes requested):
I found one thing that I thing we should look at. Happy to merge afterwards.

data/presets/amenity/community_centre/lgbtq.json
tiptoptom: Thanks, you're right. I deleted the reference because the wiki page of amenity=community_centre should be the right one for this preset.

tordans (approved):

data/presets/amenity/community_centre/lgbtq.json
tordans: Lets tackle this in a separate issue. Maybe there is some wiki page – or one can be created – that is more clear on this because https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcommunity_centre does not have any search terms of "lgb"… 

I agree that the "no reference" case is OK here, because it is unrelated.

tordans: It is now tracked in https://github.com/openstreetmap/id-tagging-schema/issues/1303

PR #1201, opened 2024-04-24
Crossing refactoring

tordans:
This is a first follow up to the [community meeting 2024-02-28](https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2024-02-28) where we discussed how to improve the crossing presets.

The commits in this PR are mostly refactoring by streamlining the presets to use `@template` fields lists.
Each commits documents possible side effects.

This PR also adds _unsearchable_ crossings for `highway=path+path=crossing`, which is missing is missing ATM.


package.json
```diff
@@ -12,7 +12,7 @@
   "license": "ISC",
   "main": "build.js",
   "scripts": {
-    "lint": "prettier --check data",
+    "lint": "prettier --check 'data/**/!(*.md)'",
```
tordans, line 15: The alternative to this would be to change the prettier config to only apply the json-stringify on .json files and use regular prettier for other files.

However, our prettier config is still somewhat different from what I am used to and the resulting READM.md formatting is not great, so I think this is the better solution.

```diff
 % git diff
diff --git a/.prettierrc b/.prettierrc
index 27505a24..78e7d6df 100644
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,6 +1,13 @@
 {
-    "parser": "json-stringify",
     "proseWrap": "never",
     "bracketSpacing": true,
-    "endOfLine": "lf"
+    "endOfLine": "lf",
+    "overrides": [
+        {
+            "files": "*.json",
+            "options": {
+                "parser": "json-stringify"
+            }
+        }
+    ]
 }
```

tordans:
@tyrasd this is ready for review. The testing part of the PR descriptions has links that make comparing the current and new setup easy.

This is quite the undertaking… but I think it looks much better now!

data/presets/highway/path/_crossing.json
```diff
@@ -0,0 +1,30 @@
...
+    "addTags": {
+        "highway": "path",
+        "path": "crossing"
+    },
+    "reference": {
+        "key": "path",
+        "value": "crossing"
+    },
+    "searchable": false,
+    "name": "Cycle & Foot Crossing"
```
tordans, line 29: Thinking about this again: Is this really the right name? It is not wrong, to describe a path as Bike+Foot infrastructure. However this wording was inspired by https://github.com/openstreetmap/id-tagging-schema/pull/1201/files#diff-3b8f616654b5d2845f63eb05ce32b27741d3945ad0a71e55033de12a656d8b56R48 which behaves differently from this preset because it also has the `addTags` part about `foot+bicycle=designated`.

Since the preset is unsearchable, no one would be confused by this when searching for presets.

And also, I thing that even if people do not add the explicit `access` keys, routers would still consider the generic `path` bike + foot infrastructure (which is why some regions prefer this tagging for the bike+foot-case in the first place).

So my opinion is this is OK … or can be changed in follow up PRs when other cases are added or reevaluated.

tyrasd (approved):
Looks good to me after a first look. I'll will perform a double-check before merging

data/presets/highway/path/crossing/_traffic_signals.json
```diff
@@ -0,0 +1,30 @@
...
+    "tags": {
+        "highway": "path",
+        "path": "crossing",
+        "crossing": "traffic_signals"
+    },
+    "reference": {
+        "key": "crossing",
+        "value": "traffic_signals"
+    },
+    "name": "Crossing Cycle & Foot With Pedestrian Signals",
```
k-yle, line 28:
```suggestion
    "name": "Cycle & Foot Crossing With Pedestrian Signals",
```
I think this would sound a bit more natural

data/presets/highway/crossing/_marked.json
```diff
@@ -23,11 +26,6 @@
         "key": "crossing",
         "value": "marked"
     },
-    "terms": [
```
k-yle, line 26: did you intentionally delete these `terms`? just double checking because I don't see them in any templates

k-yle (commented):
the end result seems much more organised than before, and I like the idea of putting READMEs in subfolders, since we don't allow comments in the presets themselves

data/presets/highway/path/crossing/_traffic_signals.json
tordans: Thanks!

data/presets/highway/crossing/_marked.json
tordans: Yes. My understanding is that `terms` are ignored for unsearchable presets, so no reason to keep them around. I assume this was a copy paste error at some point.

tyrasd:
Let's get this merged. Sorry for needlessly letting this sit around. :bow:

PR #1203, opened 2024-04-28
Add `crossing:signals` field to `crossing/*traffic_signals` presets

### LINKED ISSUE #1118: `crossing:signal`
StreetComplete recently added support for `crossing:signal` https://github.com/streetcomplete/StreetComplete/pull/5471 which signals to me that we should consider adding support as well.

Maybe something for the https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2024-02-28 which is earmarked to talk about the crossing tagging.

tordans:
This is based on top of https://github.com/openstreetmap/id-tagging-schema/pull/1201 ~~and should only be merged after that was merged. For that reason it is marked as a draft for now. I will rebase here in case https://github.com/openstreetmap/id-tagging-schema/pull/1201 changes.~~ — <ins>Update: Since #1201 was merged this is now rebased and ready for review.</ins>

---

This PR adds the `field` and add the `addTags` part.

This follows the conversation during the iD Community Meetup to consider adding the tag due to the usage in StreetComplete, Routers and the general impression that the `crossing:signals=yes` tag is widely "in use" and accepted.

---

Closes https://github.com/openstreetmap/id-tagging-schema/issues/1118

Note: This change was also part of https://github.com/openstreetmap/id-tagging-schema/pull/1044/files#diff-b8505be38bbc0be40512bf63c04d79f36f8826808ba4f0c06f186527eddcc7b4

### DIFF
--- a/data/fields/crossing/signals.json
+++ b/data/fields/crossing/signals.json
@@ -1,0 +1,8 @@
+{
+    "key": "crossing:signals",
+    "type": "defaultCheck",
+    "label": "Crosswalk Signals",
+    "terms": [
+        "traffic signals"
+    ]
+}
--- a/data/fields/crossing/signals.json
+++ b/data/fields/crossing/signals.json
@@ -1,6 +1,6 @@
 {
     "key": "crossing:signals",
-    "type": "defaultCheck",
+    "type": "check",
     "label": "Crosswalk Signals",
     "terms": [
         "traffic signals"
--- a/data/presets/highway/crossing/traffic_signals.json
+++ b/data/presets/highway/crossing/traffic_signals.json
@@ -19,6 +19,9 @@
         "highway": "crossing",
         "crossing": "traffic_signals"
     },
+    "addTags": {
+        "crossing:signals": "yes"
+    },
     "reference": {
         "key": "crossing",
         "value": "traffic_signals"
--- a/data/presets/@templates/crossing/traffic_signal.json
+++ b/data/presets/@templates/crossing/traffic_signal.json
@@ -1,5 +1,6 @@
 {
     "fields": [
+        "crossing/signals",
         "button_operated",
         "traffic_signals/sound",
         "traffic_signals/vibration"
--- a/data/presets/highway/cycleway/crossing/traffic_signals.json
+++ b/data/presets/highway/cycleway/crossing/traffic_signals.json
@@ -20,6 +20,9 @@
         "cycleway": "crossing",
         "crossing": "traffic_signals"
     },
+    "addTags": {
+        "crossing:signals": "yes"
+    },
     "reference": {
         "key": "crossing",
         "value": "traffic_signals"
--- a/data/presets/highway/footway/crossing/traffic_signals.json
+++ b/data/presets/highway/footway/crossing/traffic_signals.json
@@ -20,6 +20,9 @@
         "footway": "crossing",
         "crossing": "traffic_signals"
     },
+    "addTags": {
+        "crossing:signals": "yes"
+    },
     "reference": {
         "key": "crossing",
         "value": "traffic_signals"
--- a/data/presets/highway/path/crossing/_traffic_signals.json
+++ b/data/presets/highway/path/crossing/_traffic_signals.json
@@ -21,6 +21,9 @@
         "path": "crossing",
         "crossing": "traffic_signals"
     },
+    "addTags": {
+        "crossing:signals": "yes"
+    },
     "reference": {
         "key": "crossing",
         "value": "traffic_signals"


tordans:
## Testing

Looks good.

For a crossing node with just `crossing=traffic_signals + highway=crossing` the addTags dialogue shows up as expected: 
> image

## Follow up

Cleanup the `addTags` to only include the tags that are not already part of `tags`. Which makes it a lot easier to read and maintain.

### Open question…

- [ ] Do we actually want a field for this? Because the field allows to remove the "yes" value, which triggers the yellow update box (again). I think we shoudl remove the field and rely on addTags to add a tag that is only visible in the full tags part of the sidebar.
   - image

tordans:
### Next steps

- Wait for #1201 to be ready/merged
- Double check with https://github.com/openstreetmap/id-tagging-schema/pull/1044
- Decision on the open question (to not include a specific field, just use `addTags`)

data/fields/crossing/signals.json
```diff
@@ -0,0 +1,8 @@
+{
+    "key": "crossing:signals",
+    "type": "defaultCheck",
```
tordans, line 3:
```suggestion
    "type": "check",
```

See https://github.com/ideditor/schema-builder?tab=readme-ov-file#checkboxes

Dimitar5555:
If a crossing is already tagged as `crossing=traffic_signals`, this implies that there are signals present at the crossing. Adding `crossing:signals=yes` alongside `crossing=traffic_signals` feels redundant, as it doesn't provide any additional valuable information.

If the goal is to eventually phase out `crossing=traffic_signals` and replace it with a combination of tags such as `highway=crossing + crossing:signals=yes`, then this should be clearly stated and discussed in a proposal.

In my opinion, the focus should be on maintaining clean and efficient tagging, and `crossing=traffic_signals` alone already conveys the necessary information about signal presence. Adding another tag for the same purpose could cause confusion and dilute the clarity of the tagging scheme.

tordans:
Just for reference: The draft of the proposal on this tag https://wiki.openstreetmap.org/wiki/Proposal:Crossing_signalization#Rationale

---

Lets see if we find any reason to go along with this at this point … – I am happy with benching it for the time being.

matkoniecz:
@tordans 

> I am happy with benching it for the time being.

would that mean you are fine with closing this PR?

@1ec5 do you plan to go forward with https://wiki.openstreetmap.org/wiki/Proposal:Crossing_signalization ?

1ec5:
>> I really want to take it through the proposal process, but I’m not confident that I’ll have time for it in the next couple months.
> 
> has it changed now, few months later?

The only update I have for now is that there was another discussion recently in OSMUS Slack about replacing `separate` and `shared` with less confusing keywords. I left a comment on the talk page to that effect.

To reiterate, `crossing:signals=yes`/`no` is already quite well-established. It’s easily the third most popular crossing subkey despite the lack of support in id-tagging-schema. The proposal is about _extending_ the set of values to clarify a particular regional edge case. If it goes through, this PR would need some changes: the Crossing Signals field would go from a checkbox to a combo box, and some of the presets would need to potentially account for more specific values than `yes`. This wouldn’t be so complicated if we would stick to a single generic Pedestrian Crossing preset without so many specialized crossing presets.

data/presets/highway/crossing/traffic_signals.json
```diff
@@ -19,6 +19,9 @@
         "highway": "crossing",
         "crossing": "traffic_signals"
     },
+    "addTags": {
+        "crossing:signals": "yes"
```
matkoniecz, line 23: is there a good reason to mass-migrate this via manual clicking rather than as a scripted bot edit?

Even if global bot edit would not be accepted, maybe `crossing=traffic_signals` can be initially eliminated and replaced with `crossing:signals=yes` (or amended with such duplicate tag) in some bot-happy communities like US or Poland?

If there would be acceptance for such bot edit I would be happy to run it. (yes, it is not optimal for me from some perspectives but I genuinely consider it as a superior solution)

tordans:
Next step:

- Dont merge this because it would trigger a "missing tag" warning in iD for all existing traffic signal crossing
   - we could solve this with https://github.com/ideditor/schema-builder/issues/251 but we dont have this now
   - we could solve this with a discussed bot edit but that is a different story
- We could change the preset system similar to what we did with `education`
  - new traffic signals have the new tag: `tags: { crossing: traffic_signals, crossing:signals=traffic_signal }`
  - old once dont get the message and stay as is - as unsearchable presets `tags: { crossing: traffic_signals }`
  - we need another unsearchable triggers when only the traffic:signals-tag is preset `tags: { crossing:signals=traffic_signal }`
- Personally I consider this tag "in use" enough do make this change

Discussed with @matkoniecz

tordans:
Lets close this now https://github.com/openstreetmap/id-tagging-schema/pull/1203#issuecomment-4852271593 do a new PR about this

1ec5:
Seems like this PR was originally about adding a field but then got derailed by concerns about the validator. Should we still add the field independently of any validation rule? Mappers could benefit from seeing when a crossing is already tagged `crossing:signals=*`, especially if it’s a signalized-but-unmarked crossing that doesn’t have an obvious `crossing=*` value.

tordans:
> Seems like this PR was originally about adding a field but then got derailed by concerns about the validator. Should we still add the field independently of any validation rule? 

The current setup we have makes it hard to do the one (add field) without running into the other ("missing tag alerts"), so the plan in https://github.com/openstreetmap/id-tagging-schema/pull/1203#issuecomment-4852271593 is the workaround to add the field without the warnings. More or less what we did for education.

matkoniecz:
One more problem/documentation issue for [crossing:signals](https://wiki.openstreetmap.org/wiki/Key:crossing:signals):

https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtraffic_signals describes it as "whether pedestrians need to obey a light when crossing" (so either wiki is wrong there or it must be excluded at least from United Kingdom)

https://wiki.openstreetmap.org/wiki/Key:crossing:signals has "controlled by a traffic signal" with a similar problem

https://wiki.openstreetmap.org/wiki/Proposal:Crossing:signals has no such problem as far as I can see

1ec5:
> The current setup we have makes it hard to do the one (add field) without running into the other ("missing tag alerts"), so the plan in [#1203 (comment)](https://github.com/openstreetmap/id-tagging-schema/pull/1203#issuecomment-4852271593) is the workaround to add the field without the warnings. More or less what we did for education.

What I mean is that it should be possible to at least define the Crosswalk Signals field and add it to the relevant presets, without changing `addTags` or anything of that sort.

> One more problem/documentation issue for [crossing:signals](https://wiki.openstreetmap.org/wiki/Key:crossing:signals):

Originally, `crossing:signals=*` was documented as indicating whether the crossing user “must obey” signals, based on common sense language in the U.S., but it was later watered down to accommodate other countries like the UK, where jaywalking is tolerated and the signals are less strict. By definition, signals are a form of traffic control device everywhere in the world. Please feel free to suggest alternative wording if you feel that this is unintuitive.

matkoniecz:
And I guess it affects also label in iD field...

"signal on crossing directed to pedestrian" ? Though that may be too inclusive

matkoniecz:
> should be possible to at least define the Crosswalk Signals field and add it to the relevant presets

what it should show on crossings with `crossing=traffic_signals` without that new field? I guess that ideally it would be hidden or show filled

1ec5:
> what it should show on crossings with `crossing=traffic_signals` without that new field? I guess that ideally it would be hidden or show filled

If `crossing:signals=*` is untagged, I assume the field would show “Assumed to be Yes”, just like the One Way field on the Motorway preset. Conceptually, the set of ways that would be candidates for `crossing:signals=yes` is a strict superset of the ways that are currently tagged `crossing=traffic_signals`.

> "signal on crossing directed to pedestrian" ? Though that may be too inclusive

“Pedestrian Signals” is what they’re usually called, though “Crossing Signals” would allow us to avoid a separate version of the field on the cycleway and foot-and-cycleway crossing presets. Looking at how we handled the Railway-Road Crossing preset’s fields, we can probably rely on some common sense without spelling out the exact relationship between the signals and the crossing.

PR #1206, opened 2024-04-28
fields/kerb: Make `directionalCombo` to allow optional left/right tagging

tordans:
When the kerb is tagged on the `highway=crossing` node, it is assumed to have the same kerb left an right of the node. If that is not true, the tagging is `kerb:left|right=*`. 

This PR changes the `kerb` fields for crossings to a `directionalCombo` in order to represent those `left|right` cases.

Current usage: 

- `crossing=*` has 340 208 cases with `kerb`
- 2 690 with `kerb:right`
- 2 660 with `kerb:left`

https://taginfo.openstreetmap.org/keys/crossing#combinations

### DIFF
--- a/data/fields/kerb.json
+++ b/data/fields/kerb.json
@@ -1,6 +1,13 @@
 {
     "key": "kerb",
-    "type": "combo",
+    "keys": [
+        "kerb:left",
+        "kerb:right"
+    ],
+    "reference": {
+        "key": "kerb"
+    },
+    "type": "directionalCombo",
     "label": "Curb",
     "strings": {
         "options": {


tordans:
## Testing

Overpass turbo query: `https://overpass-turbo.eu/?R&Q=[out:json][timeout:25];node["highway"]["kerb:left"]({{bbox}});out body;>;out skel qt;`

node/9165437876

- https://pr-1206--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=n9165437876&locale=en&map=20.00/52.47598/13.44462
  - 
image

tordans:
## Conclusion

Looking at the numbers some more I don't think this added complexity in the UI is worth it. 
I will close this PR for that reason.

Issue #1013, opened 2023-09-23
Add support for Bicycle Roads in Germany


tordans:
### OSM Tag(s)

bicycle_road=yes

### How would you like this tag to see supported?

As a Preset

### Label

Bicycle Roads

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/DE:Tag:bicycle_road%3Dyes

### Status of the Tag

De Facto

### Usage of the tag

15,000+

### Regional Tag?

Germany? The tag is used all over the place https://taginfo.openstreetmap.org/tags/bicycle_road%3Dyes#map but the preset might be best scoped to Germany to start with

### Further Information

Example: https://www.openstreetmap.org/way/374152382 is shows as <img width="371" alt="image" src="https://github.com/openstreetmap/id-tagging-schema/assets/111561/8e43b38e-111a-4c41-ac33-c354331f7dba"> when in fact the bicycle road at least as important when describing the road with it's default values and expected tags.

I think we should add read only presents for now that represent a combination of the most used highway-tags + bicycle_road. Both tags together describe the infrastructure.

- Preset `highway=residential+bicycle_road=yes`, [combination used 8,370 times](https://taginfo.openstreetmap.org/tags/highway=residential#combinations)
- Preset `highway=cycleway+bicycle_road=yes`, [combination used 2,000+ times](https://taginfo.openstreetmap.org/tags/highway=cycleway#combinations)
- Preset `highway=service+bicycle_road=yes`, [combination used 1,750+ times](https://taginfo.openstreetmap.org/tags/highway=service#combinations)
- Preset `highway=path+bicycle_road=yes`, [combination used 1,810+ times](https://taginfo.openstreetmap.org/tags/highway=path#combinations)

For Germany, the preset could also require/add the tag `traffic_sign=DE:244.1`.

PR #1207, opened 2024-04-29
Introduce bicycle_road presets with special access field

tordans:
Ping https://github.com/openstreetmap/id-tagging-schema/issues/1013

- Add presets for cycleway and residential as searchable presets
- Add presets for other common highway classes path and service unsearchable
- Add searchable presets to a category
- Add a new restriction field to work around the fact that the access-field does not support the `vehicle` case – See todos.
- 
--

This is a draft for now until I had time to test it in the preview.

--

**Other todos**

- [ ] Check if the field `cyclestreet-BE-NL` needs further consideration…
- [x] Rework access field. Either as dropdown. Or copy the existing access field and add `vehicle` to the list.
    - https://github.com/openstreetmap/id-tagging-schema/issues/1208


tordans:
## Documentation

This is what the first version looked like

| Road type | Part1 | Part2 |
|--------|--------|---|
| `cycleway` | Bildschirmfoto 2024-04-29 um 21 45 07 | Bildschirmfoto 2024-04-29 um 21 45 16 |
| `residential` | Bildschirmfoto 2024-04-29 um 21 45 26 |  Bildschirmfoto 2024-04-29 um 21 45 22 |

tordans:
## Documenting Test 2

This is the version with a copy of the access field. This version requires to duplicate the `fields` from the parent preset in order to get rid of the original `access` field. But that is OK IMO.

- image
- image https://pr-1207--ideditor-presets-preview.netlify.app/id/dist/#background=Berlin-2023&disable_features=boundaries&id=w1128065475&locale=en&map=20.00/52.48337/13.44718

tordans:
### Testing 

There is a bug when changing an existing way to one of those presets. It looks like the highway tags get removed so the style and preset break. – Need to create a propper test case.

tordans:
### Testing

Need to look into making this preset country specific, so moving `cyclestreet-BE-NL ` to the preset level.

tordans:
@olafkryus after your successful https://github.com/openstreetmap/id-tagging-schema/pull/1384 maybe you are interested in helping getting this of the ground :)

matkoniecz:
@tordans Have you considered completing it?

tordans:
> @tordans Have you considered completing it?

Yes, but it does not make sense to work on this ATM because I would need Martin to review it IMO. I consider it something new from what we do in the presets so I don't consider it a simple change (see our guidelines) that I could sign of on myself (which I cannot do anyways). I would prefer leaving is as is so we find it again once there is a time that allows for a quicker back-and-forth conversation on topics like this.

matkoniecz (changes requested):
it is not a full review and I have not convinced yet @tyrasd to make decision either way (so no point in implementing it right now)...

but as mentioned in

> Need to look into making this preset country specific

it definitely would need to be country-specific

tyrasd:
Sorry for the late reply.

My gut feeling is that it's probably better to not model those as separate presets, because a) we don't do the same for the similar `motorroad=yes` / `expressway=yes` tag, b) the added complexity of maintaining different variants `bicycle_road` vs. `cyclestreet`, c) that it would be very hard to translate the preset name to languages where the concept does not exist legally (yet).

I think it is probably sufficient to keep mapping those streets with a field?!

tordans:
It certainly is one of the more annoying tagging situations in OSM. We have a property `bicycle_road=yes` that change the primary tag `highway=residential|cycleway|service` (most relevant values) to something that is legally and in usage very different (…in some countries). To make it more complex, this is not just about giving visibility to the `bicycle_raod` attribute but once that is there, other rules apply to the road – most importantly the access rules which often time lead mappers to disallow all bikes from the road because they forget to set `bicycle=designated`. We even have a campaign to monitor those cases at https://radinfra.de/kampagnen/missing_access_tag_bicycle_road/.

> b) the added complexity of maintaining different variants bicycle_road vs. cyclestreet

Agreed. But also managable IMO.

> c) that it would be very hard to translate the preset name to languages where the concept does not exist legally (yet)

Agreed. We would make them regional presets, though.

Issue #1164, opened 2024-03-17
Collection of tags that StreetComplete supports but the id-tagging-schema doesn't, yet


tordans:
StreetComplete is already the best tool for mobile mapping for beginner Mappers. This usage will grow once the iOS version is there.
iD is the best tool for beginner mappers once something on the desktop is needed and/or full editor capabilities are needed.

During mapping campaigns I notice an issue when users are used to map certain things in SC but cannot map those in iD.

**Let's use this ticket as a collection of tags that are used in SC but not yet present in the id-tagging-schema**

Let's create separate tickets to discuss if/how to add them here.

- [ ] `cycleway:lane=exclusive` and `cycleway:lane=advisory` https://wiki.openstreetmap.org/wiki/Key:cycleway:lane
- [ ] https://github.com/openstreetmap/iD/issues/8882 – a great surface/smoothness UI
- [ ] All kind of `sidewalk:*` tagging on the centerline –> To be split up more detailed
- [ ] All kind of `check_date`-Tagging –> To be split up more detailed

tordans:
@westnordost @matkoniecz, could you take a look at this and perhaps add a comment with some tags that come to mind off the top of your head? (No need for research or completeness, just what you can think of right away…). 🙏

matkoniecz:
I remember https://github.com/streetcomplete/StreetComplete/blob/f7c0cc61a3d6f37fe2bea6fa5242d741d2aa463e/app/src/main/java/de/westnordost/streetcomplete/osm/Things.kt#L256 having something not yet supported by iD, or not put there as iD support was missing.

Though I think that #529 is far more useful for mining tags.

matkoniecz:
Issues I created:

show material field if material=* is tagged #922 

consider supporting shop=gold_buyer #974  (not strictly supported, but SC support is missing as it is not in presets)

Show brand:wikidata and brand:wikipedia fields if tagged #552 

crossing=no support #548 

Support addr:place addresses #353

PR #1209, opened 2024-04-29
Test cycleway_lane tagging with separate fields

tordans:
Test out support for https://wiki.openstreetmap.org/wiki/Key:cycleway:lane.

This is part of https://github.com/openstreetmap/id-tagging-schema/issues/1164.

We might want to use a directionalCombo for this. However, that would suggest to tag data for cycleway:left:lane=* when in fact there is no cycleway:left=lane. This tries to work around this by adding separate fields with re-conditions.

Another solution for this would be https://github.com/ideditor/schema-builder/issues/153.


tordans:
## Testing

### `"cycleway:right:lane"="advisory"`

- https://overpass-turbo.eu/s/1KTM
- `way/172243640`
- https://pr-1209--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w172243640&locale=en&map=17.55/52.42423/13.45225
- image

### `"cycleway:both:lane"="advisory"`

- https://overpass-turbo.eu/s/1KTN
- `way/1228412242`
- https://pr-1209--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w1228412242&locale=en&map=20.00/52.45548/13.58080
- image the cycleway:both=lane is not visible due to https://github.com/openstreetmap/id-tagging-schema/pull/1137

### More fields

- Fields are not present here; probably due to the `allowedTags` condition
- image

tordans:
## Next steps

- Solve the `cyclway:both=*` and `cycleway=*` issue in the `prerequisiteTag`; do we need to add another field for this?
- Probably add a readme to the folder that explains the weird setup.
- Move from `moreFields` to `fields` so users are nudged to add the data
- Add "undefined" option to dropdown
- Double check highway presets (any missing?)
- Extract into @template from highway presets
- Update wording "Left Bike Lane Type" (because "Bike Lanes" is the `directionalCombo`)

tordans:
In one of the meetups we talke about this or something similar for parking where we said that the logic to only show a sub-tag (`cycleway:SIDE:lane`) to show up when a specific tag is given (`cycleway:SIDE=lane`) should be part of the logic in iD that handles the directional combo component.

I wanted to look into this. However, given that my PRs in iD are still pending on the sidewalk part (which also touches the directional combo) I will not do so until there is activity over there. 

We can, however, close this draft as something that is likely not the right approach.

PR #1214, opened 2024-05-06
Add preset man_made=paifang

SteveLz:
Wiki page: [man_made=paifang](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dpaifang)

### DIFF
--- a/data/presets/man_made/paifang.json
+++ b/data/presets/man_made/paifang.json
@@ -1,0 +1,22 @@
+{
+    "icon": "",
+    "fields": [
+        "name",
+        "inscription",
+        "height",
+        "material",
+        "paifang:gates",
+        "paifang:posts",
+        "paifang:gables"
+    ],
+    "geometry": [
+        "point",
+        "vertex",
+        "line",
+        "area"
+    ],
+    "tags": {
+        "man_made": "paifang"
+    },
+    "name": "Paifang"
+}


SteveLz:
The svg below can be used as an icon if appropriate:
Paifang

tordans:
Hey @SteveLz the test in your PR [failed](https://github.com/openstreetmap/id-tagging-schema/actions/runs/8968458145/job/24627898890?pr=1214#step:5:11) because you added new fields to the new presets that do not exist, yet. You need to add those ad separate file. However, please check if they are crucial and used a lot, they look like very detailed micromapping to me that we might want to not include in the first version of this preset.

The draft docs at https://github.com/openstreetmap/id-tagging-schema/pull/1229/files#diff-22e674f3d1da91695fbafd4d2c81f91246cfabab3d16dcf6492e6024797b4da5R1 give some guidance on which fields/tags to add.

Please also check the draft at https://github.com/openstreetmap/id-tagging-schema/pull/1239 and add some testing links and screenshots once the preview of this PR was created. Thanks!

matkoniecz:
See https://wiki.openstreetmap.org/wiki/Proposal:Man_made%3Dceremonial_gate and https://community.openstreetmap.org/t/rfc-feature-proposal-historic-ceremonial-gate/130717 that tries to solve it more generally, to avoid separate value for every single ceremonial gate variant

(I have no opinion on proposal itself, but it makes sense to wait and see whether it will succeed, I think)

PR #1216, opened 2024-05-08
Accept units in Elevation field

### LINKED ISSUE #5658: imperial units silently removed from ele tag
I was adding [a peak](https://www.openstreetmap.org/node/6166375481#map=19/55.23718/-6.93624) from old OOC maps which have all elevation data in feet. I therefore entered 561 ft in the elevation field, so was somewhat surprised to find it rendered as 561 m. Testing shows that iD silenty strips any non-numeric text (including ",") from what is entered. Although ele:imperial is probably the most widely used tag for non-metric units, the ele tag with an indication of units (e.g., ft, feet, or "'") has always been in use. I would expect at least a warning, because there is a big difference between 561 feet & metres.

1ec5:
Changed the Elevation field to be of type `roadheight`, so that the user can explicitly specify a unit other than meters when appropriate without having to convert the quantity in their head. In the United States, the unit field will default to “ft”, but the user will be able to switch it to “m” if they happen to observe a sign that (unusually) states the elevation in meters. Meanwhile, anywhere else in the world, the editor will behave exactly the same as before: the unit field will show “m” by default, but the resulting tag will omit the unit. If `ele=*` is already tagged, the field shows the currently tagged units, instead of “Unknown” in the case of feet and inches. Since units other than meters are now accepted, the field’s name no longer needs to specify “(Meters)”.

This behavior is consistent with the Height Limit field, which also allows feet and inches. Technically, an elevation is not a road height, but elevations happen to be measured in the same units as road heights in both the United States and elsewhere – even in famously messy Puerto Rico, near as I can tell. Once ideditor/schema-builder#15 is implemented, `ele` would map to either the `vehicle` or `person-height` usage categories.

This change supports [a proposal to formally document](https://wiki.openstreetmap.org/wiki/Proposal:Ele_with_units) `ele=*` as yet another key that accepts units, instead of being a prominent outlier. This draft PR helps to demonstrate that the change to documentation would have only minor impact on editors.

<table>
<thead>
<tr>
<th>Before</th><th>After</th>
</tr>
</thead>
<tbody>
<tr><th colspan="2" scope="col"><a href="https://www.openstreetmap.org/way/161301851">A childcare center</a> imported from <abbr title="Geographic Names Information System">GNIS</abbr>, which records elevations in both units, on the campus of <abbr title="National Institute of Standards and Technology">NIST</abbr>, which strongly prefers meters</th></tr>
<tr>
<td>Elevation (Meters)</td><td>Elevation</td>
</tr>
<tr><th colspan="2" scope="col"><a href="https://www.openstreetmap.org/node/5914822978">A sledding hill</a> mapped by hand at a former Voice of America broadcast complex, with the elevation signposted in feet</th></tr><tr>
</tr>
<td>Elevation (Meters): Unknown</td><td>Elevation: 945 ft</td>
</tbody>
</table>

Fixes openstreetmap/iD#5658.

### DIFF
--- a/data/fields/ele.json
+++ b/data/fields/ele.json
@@ -1,7 +1,7 @@
 {
     "key": "ele",
-    "type": "number",
-    "label": "Elevation (Meters)",
+    "type": "roadheight",
+    "label": "Elevation",
     "geometry": [
         "line",
         "area",
--- a/data/fields/ele_node.json
+++ b/data/fields/ele_node.json
@@ -1,6 +1,6 @@
 {
     "key": "ele",
-    "type": "number",
+    "type": "roadheight",
     "label": "{ele}",
     "universal": true,
     "geometry": [


1ec5:
As seen in the screenshot above, the “Height (Meters)” field could use the same treatment, since that key is already documented to accept units other than meters.

matkoniecz:
> This change supports [a proposal to formally document](https://wiki.openstreetmap.org/wiki/Proposal:Ele_with_units) ele=* as yet another key that accepts units, instead of being a prominent outlier. This draft PR helps to demonstrate that the change to documentation would have only minor impact on editors.

do you plan to proceed with proposal? Or intended to make this change mergeable?

if it is solely demonstration (for now), I think it should be closed to not be mixed with PRs intended to be mergeable now or soon

matkoniecz:
from what I see https://wiki.openstreetmap.org/wiki/Key:ele has "ele=* specifies elevation (height above sea level) of a point, in metres" and https://wiki.openstreetmap.org/wiki/Key:ele#Local_units

matkoniecz:
I have no idea whether wiki should be changed based on de facto usage or not, but if usage is not clear enough to change wiki it is also not clear enough to add it to iD presets

1ec5:
Yeah, maybe I made a mistake by going the proposal route. Should I just edit the documentation page? 🤔

matkoniecz:
I guess it depends on whether it is de facto use at least in USA? If it is clearly dominating and say 99% of peaks in USA carry feet unit then it should at least get extra section about de facto use in some regions.

Though if usage differs between regions maybe we should get US-specific `ele` field? And keep m-only in other areas?

Are other countries affected?

I guess one may write interesting q-lever/postpass query to get answers

1ec5:
> If it is clearly dominating and say 99% of peaks in USA carry feet unit then it should at least get extra section about de facto use in some regions.

No, it doesn’t dominate in absolute terms. The vast majority of peaks were imported from GNIS. For better or worse, the import script preferred the meter field over the feet field. That was long ago, before OSM accepted units anywhere. However, a nontrivial number of manually edited features (not just peaks) have feet, whether explicitly or implicitly. The values implicitly in feet are what has worried the U.S. community enough to clamor for this change. It’s the same situation we had with `maxspeed=*` before the global community warmed up to the idea of expressing it in `mph`.

> Though if usage differs between regions maybe we should get US-specific ele field? And keep m-only in other areas?

The field defaults to meters outside the U.S., so the user would have to go out of their way to tag it in feet. As far as I know, we don’t customize the Speed Limit field to prevent the user from tagging `mph` in, say, Italy.

matkoniecz:
> No, it doesn’t dominate in absolute terms. The vast majority of peaks were imported

I would say that dominating in new entries currently being manually added/ edited and ignoring imports would be fine

matkoniecz:
Note that I am not saying that either way is a bad decision (initially I was convinced that it should stay 100% pure metric, but nowadays I am far less sure).

But either way change should NOT be lead by iD presets.

1ec5:
Yes, it’s a game of chicken at this point. No one wants to be the first to move. I’m leaning towards just updating the documentation without going through the formalities, like when we introduced `mph` as an option for speed limits. Multiple RfCs showed more enthusiasm for this idea than I have for it from the one mapping community where it even matters, and most developers were on board. This PR was originally intended to demonstrate that editors could accommodate the change too.

matkoniecz:
I definitely have run into deciding whether to risk proposal getting 78.9% and getting "rejected" stamp or waiting for "de facto". And doubts when "de facto" can be judged as reached.

Especially made more complicated where I was proposing/pushing it in the first place.

matkoniecz:
(this is not rejection of such tagging schema or opinion how it should be tagged: just that OSM Wiki change should be changed first, maybe after discussion/consultation/research)

(in general iD tagging schema is supposed to follow established tagging not invent new one - and maintainers, especially me, try to delegate such research decisions rather than operating this project as leader in tagging schema changes)

PR #1220, opened 2024-05-12
Add field `website:orders`

Arthur-GYT:
All is in the title ^^.

https://wiki.openstreetmap.org/wiki/Key:website:orders


Arthur-GYT:
I was thinking, some restaurants have several links to order (like uber eat, deliveroo, etc...). Wouldn't a list of urls be better?

PR #1223, opened 2024-05-14
laboratory as search term for sample collection

matkoniecz:
many things described/branded as laboratories are in fact collection points only

### DIFF
--- a/data/presets/healthcare/sample_collection.json
+++ b/data/presets/healthcare/sample_collection.json
@@ -19,6 +19,7 @@
     ],
     "terms": [
         "blood sample collection",
+        "laboratory",
         "urine sample collection"
     ],
     "tags": {


rhhsm:
Maybe include "medical" as a search term, resulting in these collection points, real labs, various medical specialisations, etc. to appear. Or is that an iD preset matter?

matkoniecz:
> Or is that an iD preset matter?

This is iD preset PR on iD preset repository

> Maybe include "medical" as a search term

Maybe this is also worth adding? I will wait for https://github.com/openstreetmap/id-tagging-schema/pulls/matkoniecz to be processed either way

tordans (approved):
This looks good to me.

---

This is what it looks like in the search

> image

---

Also, we don't use the term too much https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema%20laboratory&type=code

PR #1228, opened 2024-05-20
Add wheelchair field to bakery shop

Arthur-GYT:
All is in the title ^^.

### DIFF
--- a/data/presets/shop/bakery.json
+++ b/data/presets/shop/bakery.json
@@ -13,7 +13,8 @@
         "rolls"
     ],
     "moreFields": [
-        "fhrs/id-GB"
+        "fhrs/id-GB",
+        "wheelchair"
     ],
     "name": "Bakery"
 }
--- a/data/presets/shop/bakery.json
+++ b/data/presets/shop/bakery.json
@@ -13,8 +13,8 @@
         "rolls"
     ],
     "moreFields": [
-        "fhrs/id-GB",
-        "wheelchair"
+        "{shop}",
+        "fhrs/id-GB"
     ],
     "name": "Bakery"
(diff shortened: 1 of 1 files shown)


Arthur-GYT:
@tordans I've changed the line ^^. BTW, I can't reopen this PR ^^'

tordans:
Quick update on …

> I think the field should be a `moreFields` already given that `shop/bakery` inherits from https://github.com/openstreetmap/id-tagging-schema/blob/e0a51350b148bfdd62607aaf70e09c94afa2a3b1/data/presets/shop.json which uses https://github.com/openstreetmap/id-tagging-schema/blob/e0a51350b148bfdd62607aaf70e09c94afa2a3b1/data/presets/%40templates/poi.json#L14 which already has the field.

This was not true, see https://github.com/ideditor/schema-builder?tab=readme-ov-file#fieldsmorefields
> If fields or moreFields are not defined, the values of the preset's "parent" preset are used. For example, shop/convenience automatically uses the same fields as shop.

However, in this case there are fields defined, so nothing is inherited.

tordans:
Test case:
- this preview https://pr-1228--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=n11796220975&locale=en&map=20.00/48.02510/12.54112
- current setup https://ideditor.netlify.app/#background=Berlin-2018&disable_features=boundaries&id=n11796220975&locale=en&map=20.00/48.02510/12.54112

tordans (approved):
I think this looks good now. @tyrasd am I overlooking something in term of the hierarchy/inheritance of fields/presets for shops?

Issue #357, opened 2022-01-18
Consider adding a dedicated "en-US" locale & mark "en" as international English


tyrasd:
Currently, the strings in the `en` locale play a special double role: On the one hand, they are the basis for all other translations and need to work for people who have set their computers to "generic" English everywhere on the planet. On the other hand it also needs to work as the locale for users who explicitly have set American English (`en-US`) as their language.

Usually, this is not a problem because most differences are relatively minor (e.g. spelling of _Color_ vs. _Colour_) and people using the generic `en` don't normally care. But there are also edge cases where it doesn't work so well. For example, see https://github.com/openstreetmap/id-tagging-schema/issues/287#issuecomment-1014929030 or https://github.com/openstreetmap/id-tagging-schema/pull/288#issuecomment-985517274.

I think we can improve the situation by introducing a dedicated `en-US` translation on transifex and declare the `en` locale to be a generic "international" English. That way Americans can get dedicated texts in some cases (just like `en-UK` provides such an optimized user experience for British English speakers). While at the same time, international users and translators would get more broadly applicable labels.

westnordost:
What is "international English"? I was under the impression that en-US in the international English, at least on the internet.

bhousel:
I think changing the way translation has been done for almost 10 years just so you can have a better name for a "Layby" preset is the wrong thing to do.

westnordost:
By the way regarding `layby`, StreetComplete translated "de:Parkbuchten" to "Adjacent to roadway only in dedicated spaces" because in en-US there is no good short word for it (IIRC).

Hufkratzer:
Related issue with the so-called **data items** (in the wiki): 
[British vs American / US English](https://wiki.openstreetmap.org/wiki/Talk:Data_items#British_vs_American_.2F_US_English)

bhousel:
We could even just make it so the parking combo field that is local to UK includes the `layby` option, and the parking combo field outside the UK doesn't include the option.

bhousel:
> Then `drink:soft_drink` could be labeled as _Soft Drink_ in "international" `en` English, while the `en-US` American English version would be able to use _Soda Pop_ for this tag.

No, the label "Soft Drink" would be perfectly understood by everyone in en-US, and "Soda Pop" is a more regional phrase.

westnordost:
@tyrasd I see, I get the idea, sounds good. 

[...]

PR #1229, opened 2024-05-24
Docs: Extended guidelines for presets, fields, deprecations

tordans:
@tyrasd following our conversation on the last iD community call I created this draft of a separate document that outlines the decisions that guide adding presets, fields and so on.

I suggest we collect some feedback but the start using it which will be the best way to improve it over time.

---

This is heavily inspired by https://github.com/streetcomplete/StreetComplete/blob/master/QUEST_GUIDELINES.md – thanks a lot

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -1,0 +1,59 @@
+# Guidelines for the tagging schema
+
+Do you have an idea for a new preset or field? Read this!
+
+## 1. Decide whether the idea works out with the tagging schema project
+
+Adding a preset or field to the tagging schema is a responsibility.
+We have to make sure that new and expert users will understand the presets and fields
+and contribute high quality data to OpenStreetMap.
+
+Consider the following:
+
+### General guidelines
+
+- 📋 **Established Documentation**: The tagging schema will only consider tags that are well documented on the OSM wiki. The documentation should be established and without ambiguity.
+- 🏷️ **Established tags only**: No new or unestablished tags should be introduced through Presets. Establishing tags must remain a community process and not be dictated by software implementation.
+- ✅ **Proposal or accepted**: A tagging is considered established when it has completed the [proposal process](https://wiki.openstreetmap.org/wiki/Proposal_process) or can in some other way be considered _accepted_ by the OpenStreetMap community. Other factors are how long a tag is _in use_, how many times it is used and if the usage over time signals growing usage.
+- 🤷 **Useful purpose**: Especially for tags that are not that well established yet. Presets and fields should have some application. As by the design of OpenStreetMap, there are countless things that _could_ be collected, such as the brightness of street lamps, etc. and sometimes things like these are even documented on the wiki (because it is a wiki, obviously). That does not mean that it makes sense to nudge users to collect this information via a preset or field.
+- 🕓 **Effort vs impact**: Consider if it is worth the effort when compared to the impact the preset and field would have. For how many elements would this quest type apply? This point is especially valid if you don't plan to contribute the code changes yourself through a PR.
+
+### User experience
+
+No preset and field is isolated, they are always presentend next to other preset and fields in the different user interfaces that present the tagging schema.
+
+- 🔦 **Easy to pick**: Users have to be able to understand and pick the right preset given the limited information that the user interfaces can provide. Good presets help guide the user with the best name and good additional documentation `(i)`.
+- 🔎 **Easy to search**: When searching, similar presets will be presented next to one another. Consider and test typical search cases. You might have to adjust the name and documentation of other presets so users can make their best decision.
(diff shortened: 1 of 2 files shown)


GUIDELINES.md
```diff
@@ -0,0 +1,63 @@
...
+thereby contributing high-quality data to OpenStreetMap (OSM).
+
+Consider the following:
+
+### General Guidelines
+
+- 📋 **Established Documentation**: The tagging schema will only consider tags that are well-documented on the OSM wiki. The documentation should be clear and unambiguous.
+- 🏷️ **Established Tags Only**: No new or unestablished tags should be part of presets. Establishing tags must remain a community-driven process, not dictated by software implementation.
+- ✅ **Proposal or Accepted**: A tag is considered established when it has completed the [proposal process](https://wiki.openstreetmap.org/wiki/Proposal_process) or is otherwise accepted by the OSM community. Factors include the tag's duration of use, frequency of use, and whether its usage is increasing over time.
+- 🤷 **Useful Purpose**: Especially for less established tags, presets and fields should have a practical application. OSM allows for the collection of a wide variety of data, but not all of it is useful. For example, the brightness of street lamps might be documented, but it doesn't necessarily warrant a preset or field.
```
1ec5, line 18: One of these bullet points could mention support in mainstream data consumers as evidence that a tag is either established or useful. After all, this is one factor for classifying tags as _de facto_ on the wiki. (Mappers have occasionally been discouraged from starting a vote on a _de facto_ tag, due to the risk that a rejection might raise backwards compatibility concerns.)

GUIDELINES.md
```diff
@@ -0,0 +1,63 @@
...
+- The deprecation is documented in the wiki and is either official (resulting from a proposal process) or long-standing (about a year).
+- There is a significant drop in usage compared to current numbers, with a negative trend ([visible in the graph](https://taghistory.raifer.tech/)).
+- Usage of the deprecated tag remains stagnant for a longer period (about a year).
+
+## 2. Design the Preset
+
+The user interface must be clear, concise, and easy to use, leaving no room for misunderstandings.
+
+- Define the tags required on an object to trigger the preset.
+- Choose a name, category, and define a list of English search terms.
```
1ec5, line 50: Unless there’s a breakthrough in #357, this could point out that American English is the development language for this repository. CONTRIBUTING.md sort of mentions it already, but GUIDELINES.md is where the contributor will be looking for this information.

1ec5, line 50: Another matter of style is that we’re using Title Case in preset and field names: #473.

tordans:
```suggestion
- Choose a name, category, and define a list of American English search terms.
```

tordans: @1ec5 good point. It is just name and alias for Title Case, right? terms are all lowercased?

```
- Choose a name, category, and define a list of American English search terms.
- Use Title Case for the the `name`, `aliases` and `label` property. Use lower case for `terms`. And Title Case or sentences for `strings`-`options`.
```

This is not ideal, yet. We might want to move this to https://github.com/ideditor/schema-builder or to https://github.com/openstreetmap/id-tagging-schema/blob/main/CONTRIBUTING.md or both…

Maybe only mention the most important cases here (name, aliases, terms) and reference one of the other docs?

1ec5: Sure, that sounds fine, as long as it’s mentioned somewhere discoverable.

GUIDELINES.md
tordans: It now reads
> - ✅ **Proposal or Accepted**: A tag is considered established when it has completed the [proposal process](https://wiki.openstreetmap.org/wiki/Proposal_process) or is otherwise accepted by the OSM community. Factors include the tag's duration of use, frequency of use, and whether its usage is increasing over time.

Asteliks:
I suggest enhancing the guidelines by recommending that new presets inherit fields from more commonly used presets. For example, when creating a specific shop type, inherit fields from the general shop preset and then add specific fields as needed. This ensures consistency, reduces redundancy, and simplifies maintenance. When the general preset is updated, the specific presets automatically inherit these updates, keeping everything current and consistent.

Additionally, it might be helpful to include best practices for testing new presets to ensure they function correctly within the tagging schema.

Other than that is all seems good to me.

tyrasd (approved):
This is a great summary of the current maintenance practices of this data set. Thanks a lot @tordans!

tyrasd:
>  new presets inherit fields

This is currently documented on the `schema-builder` repo where all other more technical details are defined. This could definitely be improved, but I think it is rather something that does not need to live in this _guidelines_ document.

> include best practices

:+1: this would definitely be useful to be included. But let's still merge the doc as it is for now and improve upon it in follow up PRs. :nerd_face:

PR #1231, opened 2024-05-24
Update bicycle_repair_station.json

pietervdvn:
Add 'compressed air' as some people search for this term when they want to add a bicycle pump

### DIFF
--- a/data/presets/amenity/bicycle_repair_station.json
+++ b/data/presets/amenity/bicycle_repair_station.json
@@ -28,7 +28,8 @@
         "bike tools",
         "cycle pump",
         "cycle repair",
-        "cycling"
+        "cycling",
+        "compressed air"
     ],
     "tags": {
         "amenity": "bicycle_repair_station"


tordans (approved):

tordans:
Thanks for the addition.

---

This is what it looks like in the search:

> image

---

A few notes from my research
- I wonder if we need better keywords for "bicycle air pump" which is what the translations gives me for "Fahrradluftpumpe". But our presets don't handle those terms well: The compressed air shows up very low and the repair station not at all.
- https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcompressed_air gives a good explanation on why this keyword is a good fit for this tag
    > While [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=compressed_air was historically used for both car and bicycle pumps, but usage for bicycle pumps has shifted to [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=[bicycle_repair_station](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_repair_station) with [service:bicycle:pump](https://wiki.openstreetmap.org/wiki/Key:service:bicycle:pump)=[yes](https://wiki.openstreetmap.org/w/index.php?title=Tag:service:bicycle:pump%3Dyes&action=edit&redlink=1) - the latter having 9400 instances as of may 2024 - whereas [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=compressed_air with [bicycle](https://wiki.openstreetmap.org/wiki/Key:bicycle)=* has only about 100 entries.
    > Compressed air for bicycles is often offered as part of [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=[bicycle_repair_station](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_repair_station), which has more services than just air. Since [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=compressed_air is usually oriented towards cars, some people prefer to map cycle pumps as a bicycle repair station with just a pump and no other services.

pietervdvn:
> This is what it looks like in the search:

Looks perfect; although a preset for "bicycle pump" (with `amenity=bicycle_repair_station; service:bicycle:pump=yes` might be useful as well to counter your first note)

> https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcompressed_air gives a good explanation on why this keyword is a good fit for this tag 

It is me who added this paragraph about historic usage...

PR #1232, opened 2024-05-25
Add cycle_barrier fields

k-yle:
> [!NOTE]  
> ~~This is a draft PR because it's blocked by https://github.com/ideditor/schema-builder/pull/87. Once that PR is merged, the temporary override to package.json needs to be removed.~~ Edit: the new syntax has been removed and can be added once https://github.com/ideditor/schema-builder/pull/87 is released

This PR adds a field to the [`barrier=cycle_barrier`](https://osm.wiki/Tag:barrier=cycle_barrier) preset to specify the type of barrier, and adds the [`wheelchair`](https://osm.wiki/Key:wheelchair) tag.

Depending on which barrier type is chosen, other fields like [`maxwidth:physical`](https://osm.wiki/Key:maxwidth:physical) are shown. The logic is documented on [this wiki page](https://osm.wiki/Key:cycle_barrier).


k-yle:
The preview link above won't work, since it's using the package.json from the `main` branch, for security reasons. An alternative preview [is available here](https://kyle.kiwi/iD).

Zverik:
After #1385, I see an issue with `spacing`, `overlap`, and `opening` fields: they are meant to contain only metric values, so their type should be `number`, and they need `(m)` in labels.

Zverik (approved):

PR #1234, opened 2024-05-25
add maxweight to helipad preset

k-yle:
the `maxweight` of a helipad is often stenciled into the ground in huge block letters, so it's easy to map from aerial imagery. 

The preset would benefit from having the `maxweight` field.

### DIFF
--- a/data/presets/aeroway/helipad.json
+++ b/data/presets/aeroway/helipad.json
@@ -5,6 +5,7 @@
         "ref",
         "operator",
         "surface",
+        "maxweight",
         "lit"
     ],
     "moreFields": [


1ec5:
When we get around to implementing a `measurement` field type for ideditor/schema-builder#15, one wrinkle will be that helipads in the United States are rated in [(thousands of) pounds](https://commons.wikimedia.org/wiki/File:Hospital_Heliport_Markings.jpg) rather than short tons, and this is what’s visible in aerial imagery. iD will need to add `lbs` as a third unit option alongside `st` and `kg`. We should do this for mass measurements anyways, since weight limits on roads in some U.S. states are posted exclusively in pounds.

tordans (commented):
**Testcase:**

- https://overpass-turbo.eu/s/1M0M
- https://pr-1234--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w26444075&locale=en&map=19.73/52.50830/13.36018

Right now this shows as a dropdown field. Is that the best option we have ATM? Do we want to keep the dropdown options?

image

k-yle:
> Right now this shows as a dropdown field. Is that the best option we have ATM? Do we want to keep the dropdown options?

I'll let someone else answer this, it's been like that [for a long time](https://github.com/openstreetmap/iD/blame/v2.19.6/data/presets/fields/maxweight.json#L3). Not sure if anyone finds the suggestions useful...

tordans:
> > Right now this shows as a dropdown field. Is that the best option we have ATM? Do we want to keep the dropdown options?
> 
> I'll let someone else answer this, it's been like that [for a long time](https://github.com/openstreetmap/iD/blame/v2.19.6/data/presets/fields/maxweight.json#L3). Not sure if anyone finds the suggestions useful...

Yes, makes sense to keep it like it was. At the very least the suggestions are an indicator on how to format the values. And once https://github.com/ideditor/schema-builder/issues/15 lands this will be improved anyway.

tordans (approved):

tordans:
Thanks for the addition. I just merged this and added Minh's comment in https://github.com/ideditor/schema-builder/issues/15#issuecomment-2150737564 so we don't forget.

PR #1235, opened 2024-05-25
Create preset for footwear_decontamination

k-yle:
A preset was recently added for `amenity=bicycle_wash` - see #1032. 

This PR creates a preset for the similar approved tag [`man_made=footwear_decontamination`](https://osm.wiki/Tag:man_made=footwear_decontamination). 

It is currently limited to Australia and New Zealand, since I'm not aware of anywhere else in the world where these features exist. However, I see that `man_made=carpet_hanger` (#452) is a global preset, despite only existing in a very small number of countries. So maybe it's okay to make this a global preset?


tordans:
I think starting with presets locally is better. It makes it easier to have a good UX when less relevant options are not present for other countries (and easy to fix later). I would rather think we should limit the other preset to countries that are relevant.

tordans:
@k-yle looking at our new draft docs, I wonder if this is merge ready or fits our criteria already. A few thoughts…

- Icon: The bike wash icons tells a good story of wash + bike https://wiki.openstreetmap.org/wiki/File:Bicycle_wash.svg – The shoe icon in this PR is only used once but for shops https://github.com/openstreetmap/id-tagging-schema/blob/85b2dbb52a39177e9dc184c7d8b656e1c4cd182e/data/presets/shop/shoes.json#L2 which makes me think we should not reuse it to not confuse users.
- Usage: It's only 160 elements ATM which is really low given https://github.com/openstreetmap/id-tagging-schema/pull/1229/files#diff-22e674f3d1da91695fbafd4d2c81f91246cfabab3d16dcf6492e6024797b4da5R16-R17. However, its also a regional preset, so that does not have to be a blocker. Would you argue that it is an important data item that we should include now (vs. later…). What are the usage cases now and in the near future?

matkoniecz:
> However, I see that man_made=carpet_hanger (https://github.com/openstreetmap/id-tagging-schema/pull/452) is a global preset, despite only existing in a very small number of countries.

Maybe `man_made=carpet_hanger` should become limited?

screen12
screen13

data/presets/man_made/footwear_decontamination-AU-NZ.json
```diff
@@ -0,0 +1,33 @@
+{
+    "icon": "maki-shoe",
```
tordans, line 2:
```suggestion
    "icon": "temaki-footwear_decontamination",
```

tordans (approved):

tordans:
Is this ready to merge from a technical point of view? @k-yle @tyrasd ?

I resolved conflicts that happened after merging #1237 and updated the icons.json that was generated during one of the npm run commands. The is released on https://rapideditor.github.io/temaki/docs/ so that should work as well.

I am wondering where the updated icon will be fetched. Temaki is not in the package.json of this repo and not in the [schema builder repo](https://github.com/ideditor/schema-builder/blob/main/package.json). So it looks like the consumers of this repo need to handle adding the icon libraries themselves [like in id](https://github.com/openstreetmap/iD/blob/develop/package.json#L82).

So from my point of view this can be merged. Any objections?

k-yle:
> Is this ready to merge from a technical point of view?

I believe so, but iD will need to update temaki before the next release. I was hoping @dependabot would do it, but the bot seems to limit itself to 5 open PRs. So I've created https://github.com/openstreetmap/iD/pull/10280

> I am wondering where the updated icon will be fetched. [...]

Yes, in iD's case, runnin[g `npm run dist:svg:temaki` ](https://github.com/openstreetmap/iD/blob/49492a0e06aef71f90e96ab3ed50098df0f4ce93/package.json#L34)will bundle every icon from temaki into a single file:  
[dist/img/temaki-sprite.svg](https://github.com/openstreetmap/openstreetmap-website/blob/master/vendor/assets/iD/iD/img/temaki-sprite.svg), which is loaded when you open iD

k-yle:
> Is this ready to merge from a technical point of view?

Now that https://github.com/openstreetmap/iD/pull/10280 is merged, I believe this is ready to be merged

PR #1236, opened 2024-05-25
prefer `surface=woodchips` over `surface=bark`

k-yle:
In some English dialects, we use the word 'bark' insead of 'woodchips'. 

Mappers often add `surface=bark` to features, then someone else comments on the changeset asking about the `surface` tag, and then eventually someone changes it to `woodchips`.

This has happened on several occasions. To save everyone time, this PR will encourage users to use `surface=woodchips` when they manually enter  `surface=bark`.

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1677,6 +1677,10 @@
     "replace": {"amenity": "restaurant"}
   },
   {
+    "old": {"surface": "bark"},
+    "replace": {"surface": "woodchips"}
+  },
+  {
     "old": {"theatre:type": "amphitheatre"},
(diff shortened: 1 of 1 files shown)


tordans:
We have a issue somewhere about adding a different kind of notice that is more of a suggestion than an error. IMO this is a case for that and not a good fit for a full on deprecation. Unless there is some consensus for a full deprecation documented somewhere?

k-yle:
Okay, I'll close this and follow issue #850

I've also just realised that we can translate `Woodchips` into `Bark / Woodchips` in the `en-AU` and `en-NZ` locales, which might be a good short-term solution, to help users find the right tag

PR #1237, opened 2024-05-25
add preset for `man_made=flare`

k-yle:
add preset for [`man_made=flare`](https://osm.wiki/Tag:man_made=flare)

### DIFF
--- a/data/presets/man_made/flare.json
+++ b/data/presets/man_made/flare.json
@@ -1,0 +1,25 @@
+{
+    "icon": "temaki-chimney",
+    "fields": [
+        "height"
+    ],
+    "moreFields": [
+        "ref",
+        "visibility"
+    ],
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "terms": [
+        "flare stack",
+        "gas",
+        "oil",
+        "fire",
+        "burn"
+    ],
+    "tags": {
+        "man_made": "flare"
+    },
+    "name": "Gas Flare"
+}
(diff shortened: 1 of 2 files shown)


data/presets/man_made/flare.json
```diff
@@ -0,0 +1,25 @@
...
+    "fields": [
+        "height"
+    ],
+    "moreFields": [
+        "ref",
+        "visibility"
+    ],
+    "geometry": [
+        "point",
+        "vertex"
```
tordans, line 12: What would the vertex case be?

data/presets/man_made/flare.json
```diff
@@ -0,0 +1,25 @@
+{
+    "icon": "temaki-chimney",
+    "fields": [
+        "height"
```
tordans, line 4: The wiki talks about https://wiki.openstreetmap.org/wiki/Key:structure but the description does not fit well IMO. What are your thoughts?

tordans (approved):

data/presets/man_made/flare.json
k-yle: perhaps it could be connected to a pipeline, since the gas must come from somewhere? 

I'm not sure if it's commonly mapped as a vertex, but I didn't want to limit the preset given that I just opened #1233 😄

Happy to remove it if you think this is uncommon

data/presets/man_made/flare.json
tordans: Looked at a few examples and still don't see how the structure key adds value, so I close this.

data/presets/man_made/flare.json
tordans: Looked at 5-10 examples and found one connected to a line, so lets keep it https://www.openstreetmap.org/node/5492815582

tordans:
I think we can merge this now and add other fields later – see https://github.com/openstreetmap/id-tagging-schema/pull/1237#issuecomment-2150720633 for thoughts on that.

Issue #148, opened 2021-03-12
pipeline=substation does not guarantee an area?


cicku:
I created an area with substation (default select)
![image](https://user-images.githubusercontent.com/686438/110985810-e2b84100-833a-11eb-913d-86ed201d7b3a.png)

Then I changed the tag from
![image](https://user-images.githubusercontent.com/686438/110985855-f2378a00-833a-11eb-9496-ca9e1197f717.png)

to

![image](https://user-images.githubusercontent.com/686438/110985951-1004ef00-833b-11eb-8d88-3e5c8d9f71f3.png)

So unless I specify area=yes, iD will not map it as an area.

flacombe:
I second @cicku comment and confirm both power=substation and pipeline=substation are facilities that are mostly mapped as areas.

Additionally, that shouldn't be necessary to look for power=substation to map a pipeline substation.

@cicku Accordin to aerial picture, you probably can change substation=yes to substation=transmission depending of this particular place's role.

tyrasd:
There simply does not yet exist a preset for `pipieline=substation`. That's why it defaults back to being a basic "line".

qugebert:
Is this already closed with https://github.com/openstreetmap/id-tagging-schema/commits/2f24b915dfcb8126ee6dc16c54357f2a84e95fe1/data/presets/pipeline/substation.json ?

qugebert:
@tyrasd i think this can be closed as completed in #[2f24b91](https://github.com/openstreetmap/id-tagging-schema/commits/2f24b915dfcb8126ee6dc16c54357f2a84e95fe1/data/presets/pipeline/substation.json)

PR #1239, opened 2024-05-26
Docs: Add a pull request template

tordans:
We can use a pull request template to nudge contributors to add information to the PRs that make it easier to review them.

This template is based on https://github.com/twbs/bootstrap/blob/main/.github/PULL_REQUEST_TEMPLATE.md?plain=1 but change to fit the tagging schema.

My idea is, to nudge contributors to first add more context to PRs like links. But then after the preview was generated, to get back , edit the description and add more links and screenshots that make it super fast to review the PRs.

---

Here are the docs on adding those templates. In general they are much less fancy than the issue templates, so it's just a Markdown template.
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -17,7 +17,7 @@
 <!-- E.g., Numbers from Taginfo https://taginfo.openstreetmap.org/ and maybe local Taginfo https://taginfo.geofabrik.de/ -->
 <!-- E.g., a link to https://taghistory.raifer.tech -->
 
-### Checklist and Test-Documentation template.
+### Checklist and Test-Documentation Template
 
 Once the preview is generated, **copy the following template and create a comment** to add the info.
 This will help you check your PR and make it a lot faster to test your changes for the maintainers.
@@ -63,4 +63,3 @@ If this is your first contribution to this project, the preview will not happen 
 - [ ] `name`, `aliases` (if present) use Title Case
 - [ ] `terms` (if present) use lower case, sorted A-Z
 <!-- Learn more in https://github.com/openstreetmap/id-tagging-schema/blob/main/GUIDELINES.md#2-design-the-preset -->
-```
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -40,7 +40,7 @@ If this is your first contribution to this project, the preview will not happen 
      Add relevant **screenshots** of the sidebar of those examples. -->
 
 <!-- FYI: What we will check:
-     - Is the icon well chosen.
+     - Is the [icon](https://github.com/ideditor/schema-builder/blob/main/ICONS.md) well chosen.
      - Are the fields well-structured and have good labels.
      - Do the dropdowns (etc.) work well and show helpful data. -->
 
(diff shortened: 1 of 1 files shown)


qugebert:
I like this idea as it helps us newbies to provide all necessary informations.

Maybe we could add a hint in the "Related issues"-Section to use the ```closes #Issue-Number``` Syntax because sometimes issues stay open after the related PR was merged ([Example 1](https://github.com/openstreetmap/id-tagging-schema/issues/148), [Example 2](https://github.com/openstreetmap/id-tagging-schema/issues/25)) and afaik by using ```closes #...``` these Issues are closed automatic after a successful PR.

tordans:
Note to self: New users don't get the preview (and checks) automatically. Once a PR was merged, the checks run for that user automatically.

tyrasd (dismissed):
(testing branch protection configuration)

tyrasd (dismissed):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,59 @@
...
+- …
+
+**Relevant tag usage stats:**
+> …
+<!-- E.g., Numbers from [Taginfo](https://taginfo.openstreetmap.org/) or [local Taginfo](https://taginfo.geofabrik.de/) -->
+<!-- E.g., A link to https://taghistory.raifer.tech -->
+
+### Checklist and Test-Documentation template.
+
+Once the preview is generated, **copy the following template and create a comment** to add the info.
```
k-yle, line 22: Alternatively, we could encourage people to come back later and edit the PR description, instead of copy-pasting this template into a separate comment. 

The reason I suggest this is because, it may lead to PRs that contain the empty template in the PR description, and the completed template in a separate comment. 

But this is just a hypothetical problem, if it actually causes confusion we could revisit this in the future

k-yle (approved):

.github/PULL_REQUEST_TEMPLATE.md
tordans: Agreed. I wrote it like this because we will get notifications on new comments … but not on changes to the existing description. But we might need to change that if it does not work…

tordans: FYI: I reworked this so the initial message is less cluttered.

tordans:
This is a test comment to test the copy-paste template

---

## Test-Documentation

### Search



### Info-`i`



### Wording

- [ ] American English

tordans:
@tyrasd  this is ready to be merged from my point of view. It will make new PRs a lot easier to review…
I will iterate on it if needed once we learn how people use / understand it.

Asteliks:
Personally, I prefer the description to be in an issue and not duplicated, but I know that this will be annoying for the contributors and maintainers (these groups usually make PRs without issues). As such, my recommendation is that this template should only be used when there isn't an issue linked to the PR

tordans:
> Personally, I prefer the description to be in an issue and not duplicated,

Agreed. However I would assume people are lazy and will just leave this part blank or reference the issue for the first section. 
While reviewing the PRs recently I found that the issue did not really help a lot, which is why I ask for more context here.
The issues are sometimes created by users that only know the wiki link of a tag but don't know how to do more research on it. My goal for the PR is to ask people to do this research now …

I think we can see how it goes with this one and modify if needed.

Asteliks:
> I think we can see how it goes with this one and modify if needed.

I fully agree let's see how it will work in the wiled

tyrasd (approved):
For a lot of PRs it will not be necessary and perhaps could even be borderline a waste of time to send screenshots, etc.… Hopefully contributors are not deterred by that.

But let's try it anyway and see whether this makes the maintenance work a little bit easier.

PR #1246, opened 2024-06-05
Reuse labels for `traffic_calming`

### LINKED ISSUE #1152: Duplicated entries in Transifex for `traffic_calming`
Almost every item from https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/traffic_calming.json is duplicated by it's standalone preset. Would it be possible to merge them? Ideally there would be an automated validation that warns when there are two strings with the same source text.

Dimitar5555:
Resolves #1152

Note: I haven't tested it which is why I've marked it as a draft.

### DIFF
--- a/data/fields/traffic_calming.json
+++ b/data/fields/traffic_calming.json
@@ -10,18 +10,18 @@
     ],
     "strings": {
         "options": {
-            "bump": "Speed Bump",
-            "chicane": "Traffic Chicane",
-            "choker": "Traffic Choker",
+            "bump": "{traffic_calming/bump}",
+            "chicane": "{traffic_calming/chicane}",
+            "choker": "{traffic_calming/choker}",
             "choked_table": "Choked Speed Table",
-            "cushion": "Speed Cushion",
-            "dip": "Dip",
+            "cushion": "{traffic_calming/cushion}",
+            "dip": "{traffic_calming/dip}",
             "double_dip": "Double Dip",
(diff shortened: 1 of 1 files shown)


Dimitar5555:
It looks like it's not working. Will adding `presets/` in the beginning fix the problem?

image

tordans:
@Dimitar5555 I think the way to do this is https://github.com/ideditor/schema-builder?tab=readme-ov-file#stringscrossreference
When I look through the code https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema+%22strings%22&type=code&p=2 I see cases like https://github.com/openstreetmap/id-tagging-schema/blob/87d7b90f93c822c879392466016c55b0e7932905/data/fields/crossing/markings_yes-PL.json#L14
But looking at https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema+%22options%22&type=code&p=5 there is no case that looks like what you are doing.

In general, this reference feature only works in very specific cases, not always.

Dimitar5555:
`stringsCrossReference` is for the options in a field. It's not related to the field's label. Currently it's not possible to cross reference labels between presets and fields. This is a limitation of the schema builder.

Blocked by https://github.com/ideditor/schema-builder/issues/162

tordans:
@Dimitar5555 thanks for opening the other issue. Can we close this PR until that is  resolved?

PR #1249, opened 2024-06-06
medical as search term for sample collection

matkoniecz:
follow https://github.com/openstreetmap/id-tagging-schema/pull/1223#issuecomment-2110555992 idea

### DIFF
--- a/data/presets/healthcare/sample_collection.json
+++ b/data/presets/healthcare/sample_collection.json
@@ -20,7 +20,8 @@
     "terms": [
         "blood sample collection",
         "laboratory",
-        "urine sample collection"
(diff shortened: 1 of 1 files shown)


matkoniecz:
And for why it makes sense: sample collection points are often branded/titled as laboratories. Someone planning to type "medical laboratory" may end with just single word, notice match and be happy about it. It would be nicer to show him this value.

Disclaimer: I wrote proposal introducing this tag value.

tordans:
This is what the preview would look like

image

---

I don't see anythings negative with adding this, so LGTM.

tordans (approved):

PR #1250, opened 2024-06-06
Update prettier to 3.3.1

tordans:
https://github.com/openstreetmap/id-tagging-schema/pull/1242 only changed the lock file, so I checke `npx taize major --write` which updated the package as well and updated to 3.3.1.

Again, a non-changing change, so I think we might as well merge it.


tordans:
Done in https://github.com/openstreetmap/id-tagging-schema/pull/1252

PR #1253, opened 2024-06-06
remove matchScore=1.1 from Kitchen Garden

k-yle:
This is a follow up to https://github.com/openstreetmap/id-tagging-schema/pull/1135#discussion_r1615146725, it seems like `matchScore` is not required here and is causing unexpected behaviour (see that comment for details)

### DIFF
--- a/data/presets/leisure/garden/kitchen.json
+++ b/data/presets/leisure/garden/kitchen.json
@@ -31,6 +31,5 @@
         "edible garden",
         "cook's garden"
     ],
-    "matchScore": 1.1,
     "name": "Kitchen Garden"
 }


k-yle:
the preview confirms that this works, and you can still select a specific `garden:type`. If you select a value with a dedicated preset, then that preset takes over

tordans (approved):

tordans:
Thanks for the screenshots, I agree with this.

A follow up could be to add translated strings to the garten type field so they have more of a match between them.

tordans:
Mh… merging this causes errors at https://github.com/openstreetmap/id-tagging-schema/actions/runs/9403688023

Habe to look into this later.

k-yle:
Regarding the GitHub Actions issue, I've opened #1254

PR #1254, opened 2024-06-06
fix broken CI script

k-yle:
See https://github.com/JamesIves/github-pages-deploy-action/discussions/1589 for more info. The 'deploy' pipeline has been broken [for the past 2 weeks](https://github.com/openstreetmap/id-tagging-schema/commits/main)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -31,7 +31,7 @@ jobs:
         run: npm run build
 
       - name: Deploy 🚀
-        uses: JamesIves/github-pages-deploy-action@v4.5.0
+        uses: JamesIves/github-pages-deploy-action@v4.6.1
         with:
           BRANCH: main # The branch the action should deploy to.
           FOLDER: . # The folder the action should deploy.


tordans (approved):
Looks good. Only way to test it is to merge, right? So lets try this…

tordans:
> Looks good. Only way to test it is to merge, right? So lets try this…

Well, not really :-D
> [This job was skipped](https://github.com/openstreetmap/id-tagging-schema/actions/runs/9411838504/job/25925766954)

Maybe something checks if the right files where changed and only then allows the Action to run. But  re-running the last action https://github.com/openstreetmap/id-tagging-schema/actions/workflows/deploy.yml does not work either, it shows the same error (I thought it would use the current code to run but maybe it uses the specific commit…).

=> We will just have to wait until we merge something else, I guess.

tordans:
@k-yle the prettier merge just now ran the script in https://github.com/openstreetmap/id-tagging-schema/actions/runs/9411968617 and it still fails.

k-yle:
@tordans it's a different error now, because pushing directly to the `main` branch is no longer allowed:

image

One solution is to make the bot automatically opens a PR instead of committing to `main`, but @tyrasd should probably make a decision about what to do. 

For completeness: another solution would be to move store the `dist` files in a dedicated branch, instead of a folder on the `main` branch. This is the more standard, but it would be a major breaking change

tordans:
> pushing directly to the main branch is no longer allowed

Martin can probably add the Action to an allow list that allows it to push to main and bypass the branch protection.

k-yle:
> Martin can probably add the Action to an allow list that allows it to push to main and bypass the branch protection.

I'm interested to know if this is possible, because last time I checked, it was only possible to make exceptions for Users/Bots/GitHub _Apps_, but not GitHub _Actions_[^1][^2], since actions use [the repository's `GITHUB_TOKEN`](https://docs.github.com/en/actions/security-guides/automatic-token-authentication).

[^1]: https://github.blog/changelog/2022-05-17-consistently-allow-github-apps-as-exceptions-to-branch-protection-rules/
[^2]: [community/discussions#13836](https://github.com/orgs/community/discussions/13836)

tyrasd:
> Martin can probably add the Action to an allow list that allows it to push to main and bypass the branch protection.

I did a bit of research and as far as I see, there is no good way to make this work (see also this short [discussion](https://github.com/JamesIves/github-pages-deploy-action/discussions/1546) on the repo of the github action we use in the workflow), at least not easily. :shrug: 

We could either disable the branch protection rules again (and manually check that our "branch/PR rules" are followed), or otherwise make sure that the necessary "deploy" routine is performed when each PR is merged.

My quick opinion would be that the second option might be more error-prone in the long term than the first. What do you think?

tyrasd:
A third option would be to do the "deploy" of the `interim` folder into a separate branch, which would not have to be protected. Then we only need to adjust the respective scripts in iD and transifex to point to the new branch. That's probably the better solution here.

PR #1256, opened 2024-06-07
add new terms to water works

matkoniecz:
inspired by #1115 - and replaces and closes #1115

### DIFF
--- a/data/presets/man_made/water_works.json
+++ b/data/presets/man_made/water_works.json
@@ -12,6 +12,10 @@
         "point",
         "area"
     ],
+    "terms": [
+        "Potable Water Treatment Plant",
+        "Water Treatment Plant"
+    ],
     "tags": {
         "man_made": "water_works"
     },


matkoniecz:
screen16
screen17

is there way to have per-term priorities?

tordans:
> is there way to have per-term priorities? Drinking water is far more expected term here :/

AFAIK there is not. You could try fiddling with the general priority, but that will have side effects with regular searches for sure.

matkoniecz:
Maybe boosting drinking water would make sense? But that in turn can result in weird things where something is decorational fountain and provides drinking water and is tagged with feature tag...

I guess I would close this one or remove that Potable part?

data/presets/man_made/water_works.json
```diff
@@ -12,6 +12,10 @@
         "point",
         "area"
     ],
+    "terms": [
+        "Potable Water Treatment Plant",
+        "Water Treatment Plant"
+    ],
```
tordans, line 17:
```suggestion
    "terms": [
        "water treatment plant"
    ],
```

Lets keep the "potable" out of here; its not mentioned https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dwater_works or https://wiki.openstreetmap.org/wiki/Item:Q6468

And lower case terms, AFAIR.

tordans (changes requested):
LGTM, see comment.

matkoniecz:
screen-2025-11-03-16-40-47

matkoniecz:
so, from my side it is tested and seems to work fine

tordans (approved):

PR #1257, opened 2024-06-08
Docs: Add section on info-i to CONTRIBUTING.md

tordans:
I think we should have one place that explains how to create good info-`i` data for presets. I want to link to this section from https://github.com/openstreetmap/id-tagging-schema/pull/1239.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,6 +49,13 @@ To make a change, update the corresponding file within the `data` folder: The `p
 
 Icons from different sources (_icon sets_) can be used in the tagging schema. Head over to the [dedicated page](https://github.com/ideditor/schema-builder/blob/main/ICONS.md#icons) about how to use them.
 
+### Info-`i`
+
+iD and other tools provide users with a way to learn more about the main tag of a preset. It is important to provide good information in this information panel. Here are a few notes on how to do this:
+- Does your tag have a Wikidata entry? Click the small pencil icon next to the text to open the Wikidata item on the OSM wiki. Improve this wording if needed. If the Wikidata item is missing, [learn more about how to add it in "Current methods for creating new items"](https://wiki.openstreetmap.org/wiki/Data_items#Item_creation_process).
+- Does your tag have a Wiki page with a good image?
+- Your preset might need [a `reference` property](https://github.com/ideditor/schema-builder?tab=readme-ov-file#reference) to force the system to use a specific tag for the information section.
+
 ### Code Style
 
 The input files are JSON files which use 4-space indentation. You can use the `npm run lint` command to check whether your files match the expected code style and run `npm run lint:fix` to reformat them if they don't do so.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -49,6 +49,13 @@ To make a change, update the corresponding file within the `data` folder: The `p
 
 Icons from different sources (_icon sets_) can be used in the tagging schema. Head over to the [dedicated page](https://github.com/ideditor/schema-builder/blob/main/ICONS.md#icons) about how to use them.
 
+### Info-`i`
+
+iD and other tools provide users with a way to learn more about the main tag of a preset. It is important to provide good information in this information panel. Here are a few notes on how to do this:
+- Does your tag have a Wikidata entry? Click the small pencil icon next to the text to open the Wikidata item on the OSM wiki. Improve this wording if needed. If the Wikidata item is missing, [learn more about how to add it in "Current methods for creating new items"](https://wiki.openstreetmap.org/wiki/Data_items#Item_creation_process).
```
k-yle, line 55: I think it might be helpful to mention where the "information panel" is, before saying _"Click the small pencil icon next to the text"_. 

For someone who is not familiar with data items, it might not be clear what this paragraph is about. A [screenshot like this](https://github.com/openstreetmap/id-tagging-schema/assets/111561/13549318-cd7c-4dd1-9948-7a2d84662f04) would also help, but it might look out of place, since there are no other images in this document.

CONTRIBUTING.md
k-yle, line 1: another thought: taginfo actually [has an API](https://taginfo.osm.org/api/4/wikidata/all) that would let us easily automate this, perhaps with [GitHub's Check Annotations](https://github.com/openstreetmap/id-tagging-schema/assets/16009897/57bc3571-beab-4ef0-ac58-6f9bd2c8cee8). This would flag any PRs that add or edit a preset with no documentation in certain languages

tordans: FUI: Taginfo never switched over to use the data items and there is still opposition to rolling them out in other places. It's an unfortunate mess ATM.

CONTRIBUTING.md
tordans: (Is this screenshot URL permanent / can I use it in the readme?)

k-yle: l grabbed that URL from your earlier comment, and I believe GitHub preserves all uploaded images if they're used in at least one comment

tordans: Thanks. I added the image. It is a bit too big and also not ideal from what it shows. But we can update that later IMO, once we know this link will work mid term… 
Lets again in a few days.

CONTRIBUTING.md
1ec5: On the other hand, MediaWiki does have an API for grabbing the data item associated with a tag and its description in each language. Take a look at what iD uses to [get the documentation](https://github.com/openstreetmap/iD/blob/b2b3009a649ecc048060240c6e7a21024755167f/modules/services/wikidata.js#L156) for the ℹ️ button.

tyrasd (approved):

PR #1261, opened 2024-06-09
Add alternative terms to bottle return

rugk:
"refund" is the best term I found to be translated to "Pfand" in German, because this is usually what one may search for in Germany and some other EU countries. You get money back on the packaging. All other terms do not imply any financial reward.

Sources (coming from German):
* https://dict.leo.org/englisch-deutsch/flaschenpfand
* https://liveingermany.de/pfand-in-germany-bottle-deposits/

### DIFF
--- a/data/presets/amenity/vending_machine/bottle_return.json
+++ b/data/presets/amenity/vending_machine/bottle_return.json
@@ -9,7 +9,9 @@
         "vertex"
     ],
     "terms": [
-        "bottle return"
+        "bottle return",
+        "bottle deposit",
+        "refund"
     ],
     "tags": {
         "amenity": "vending_machine",


tordans:
Do you thing this term is importantly for English speaker? If this was added just to improve the German translations: those are decoupled and we can manage separate terms.

1ec5:
The usual terms in the United States are _refund_ and _redemption_; these terms will only make sense if it’s clear that the field relates to a bottle return program. Cans purchased anywhere in the country are inscribed with the initialism “CRV”, for _cash redemption value_, the term that California uses to refer to the refund amount.

https://en.wikipedia.org/wiki/Container_deposit_legislation_in_the_United_States

rugk:
Well German may be managed separately, but you need an entry/field to translate, don't you?

Also given that comment, maybe we should add "bottle refund" and "bottle redemption"?

tordans:
> Well German may be managed separately, but you need an entry/field to translate, don't you?

I am very sure that the terms field can be filled with as many terms per language as you like. It's not separate translations  per entry in the file, but one field for all 'terms'. Meaning you can remove all terms for one translation and keep them for the other or have more for one and less for the other and so on.
Please check the translation system to confirm; links are in the .md file, section in translations.

rugk:
Hmm, I searched for "bottle" in "presets" and I only find the canonical translation:
grafik

I can [search for "terms"](https://app.transifex.com/openstreetmap/id-editor/translate/#de/presets/179720965?q=text%3Aterms), but this just brings a lot of unrelated stuff and ~~I am somehow unable to search for things with spaces (i.e. "bottle terms")~~ [I can search for `text:'bottle terms'`](https://app.transifex.com/openstreetmap/id-editor/translate/#de/presets/484547783?q=text%3A'bottle+terms') if I quote it, [and I actually found it](https://app.transifex.com/openstreetmap/id-editor/translate/#de/presets/179607087?q=text%3A'bottle+terms'). It says: `Pfandautomat,Pfandrückgabeautomat,Pfandflaschenrückgabeautomat,Pfandflaschenautomat`, so it is a comma-separated list, that is good then! (Just very hard to find.) And BTW, it already includes the German term `Pfand`.
Also [the translation doc](https://github.com/openstreetmap/id-tagging-schema/blob/main/CONTRIBUTING.md#translating) does not mention that possibility of synonyms or so.

tordans:
> Is there anything left to do here?

@rugk I think this PR is obsolete now that we found the way to use the translation system to update the terms, so I will close it. Please ping me if I should repopen.

PR #1264, opened 2024-06-11
Added missing place presets

michalgwo:
Some values of `place=*` tag (administratively declared places) were missing, so I decided to add them. [Wiki](https://wiki.openstreetmap.org/wiki/Key:place) for reference.

### DIFF
--- a/data/presets/place/municipality.json
+++ b/data/presets/place/municipality.json
@@ -1,0 +1,17 @@
+{
+    "icon": "maki-triangle-stroked",
+    "fields": [
+        "name",
+        "population",
+        "population/date",
+        "source/population"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "place": "municipality"
+    },
+    "name": "Municipality"
+}
--- a/data/presets/place/country.json
+++ b/data/presets/place/country.json
@@ -1,0 +1,16 @@
+{
+    "icon": "maki-triangle-stroked",
+    "fields": [
+        "name",
+        "population",
+        "population/date",
+        "source/population"
+    ],
+    "geometry": [
+        "point"
+    ],
+    "tags": {
+        "place": "country"
+    },
+    "name": "Country"
+}
--- a/data/presets/place/country.json
+++ b/data/presets/place/country.json
@@ -7,7 +7,8 @@
         "source/population"
     ],
     "geometry": [
-        "point"
+        "point",
+        "area"
     ],
     "tags": {
         "place": "country"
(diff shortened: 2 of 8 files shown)


data/presets/place/municipality.json
```diff
@@ -0,0 +1,17 @@
...
+        "source/population"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "place": "municipality"
+    },
+    "name": "Municipality"
```
1ec5, line 16: “Municipality” would cause a lot of confusion in the default American English locale. “Municipality” is the proper translation of something like a _municipio_ in various Spanish-speaking countries – a kind of “space-filling” administrative area that completely partitions a larger administrative area, not necessarily urban. But to an American English speaker, “municipality” is normally a catch-all term for an incorporated city, town, or village, corresponding to a populated place.

“Township” might be a better name for this preset. In the United States, we use `place=municipality` for civil townships in the Midwest and towns in New England, New York, and Wisconsin, all of which have the same semantics. But “township” means something completely different in other English-speaking countries like South Africa. Probably we can’t responsibly add this preset until we split it up into multiple presets tailored to each region’s legal constructs.

tyrasd: Would a possible solution be to just split this preset into two _regional_ ones: One with `name=Township` for the US, and one with `name=Municipality` for everywhere else (or perhaps just the [countries](https://taginfo.openstreetmap.org/tags/place=municipality#map) like Italy, Germany, Sweden, Thailand(?) where it is commonly used in OSM).

tyrasd (changes requested):
The wiki defines [`place=country`](https://wiki.openstreetmap.org/w/index.php?title=Tag:place%3Dcountry&oldid=2551826), [`place=district`](https://wiki.openstreetmap.org/w/index.php?title=Tag:place%3Ddistrict&oldid=2576495) and [`place=county`](https://wiki.openstreetmap.org/w/index.php?title=Tag:place%3Dcounty&oldid=2573192), etc. to be only used on nodes.[^1] Let's limit the geometry to points for these to `point` geometry.

[^1]: some of these (e.g. `place=district`, `place=region`) are also used somewhat commonly on admin boundary relations, but that usage seems to be discouraged, and for these features the preset for administrative boundaries already exists and should be sufficient.

michalgwo:
> @michalgwo Do you plan to update this PR?

Sorry, it's a really old PR and I don't remember what I wanted to achieve here, so I'm not planning to update it, so I'm closing it.

PR #1273, opened 2024-06-26
Add field bridge:name=* and tunnel:name=*

SteveLz:


tordans:
@SteveLz can you please add more context to your PR following the PR template that is drafted in https://github.com/openstreetmap/id-tagging-schema/pull/1239/files
Thanks

matkoniecz:
No, please please no.

`bridge:name` is not a good tagging, bridge name should go on `man_made=bridge` object

SteveLz:
* We may need to be consistent with the `bridge:ref` that already exists in the id-tagging-schema.
* The area of ​​a tunnel is difficult to verify, and it is more common for mappers to use `tunnel:name` instead of drawing an area of ​​indeterminate shape.
* Compared to writing the names of bridges and tunnels in `bridge:name` and `tunnel:name`, it is even more unacceptable to change the `name` of highways, railways, and waterways to bridges and tunnels. We need to remind people to use `man_made=bridge` instead of rejecting the tag.
> unusual tags

Does this mean that the number of named `man_made=bridge` is slightly less than `bridge:name`?

>pointless complexity

For places where the bridge lines have not yet been fully identified and the railways have not yet been fully double-tracked, mapping the bridge area is indeed a labor-intensive micro-mapping. This is again difficult to accomplish for places where up-to-date imagery is not yet available.

matkoniecz:
> The area of ​​a tunnel is difficult to verify, and it is more common for mappers to use tunnel:name instead of drawing an area of ​​indeterminate shape.

note that I have not commented at all about `tunnel:name`

matkoniecz:
> Does this mean that the number of named man_made=bridge is slightly less than bridge:name?

1) number of objects where their name is tagged on another object rather in `name` tag is low compared to usage of `name` tag in general

2) note that `bridge:name` is counted multiple times in many cases for a single bridge

1ec5:
> bridge:name is not a good tagging, bridge name should go on man_made=bridge object in name key rather than some unusual tags that is just pointless complexity

These tagging styles aren’t mutually exclusive, are they? I’ve often tagged both since the redundancy didn’t seem to be that harmful.

For what it’s worth, #215 would also mitigate some of the tendency to overload `name=*`. (Hopefully you don’t look at the date on that PR and despair for this PR’s prospects. 😬)

tyrasd (changes requested):
I would say, that even though the mapping of bridges as separate objects can be considered the "state of the art" mapping technique, the `bridge:name` tag is still used often enough to warrant to be represented in the tagging schema (at the very least to  make it easy to move the name to it's own feature when newly drawing a bridge's polygon).

@SteveLz what's missing in this PR is the inclusion of the new fields in the respective (road and railway ) presets. I would prefer to include `bridge:name` only as an optional (`"moreFields"`) field, while `tunnel:name` might make sense to be included in the list of regular fields.

tyrasd (approved):
Sorry for the delay. Let's queue this for the release after the next (`v6.11`): there's some new strings to translate and some minor adjustments to be implemented in iD first

matkoniecz:
> These tagging styles aren’t mutually exclusive, are they? I’ve often tagged both since the redundancy didn’t seem to be that harmful.

They are not exclusive, but that does not seems to be a tagging scheme worth encouraging

> We may need to be consistent with the bridge:ref that already exists in the id-tagging-schema.

I would support its removal

PR #1274, opened 2024-06-30
Add preset for leisure=bathing_place (#1217)

### LINKED ISSUE #1217: Add `leisure=bathing_place`
### OSM Tag(s)

leisure=bathing_place

As a Preset

### Label

Swim spot

### Aliases

Bathing place, bathing spot, swimming spot, wild swimming spot, dip spot

### Terms

_No response_

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dbathing_place

### Status of the Tag

In Use

### Usage of the tag

1,129

### Replaces other Tag?

No

### Regional Tag?

global

### Further Information

Böhmerheide_-_Badestelle_Landseite
([source](https://commons.wikimedia.org/wiki/File:B%C3%B6hmerheide_-_Badestelle_Landseite.jpg))

A bathing place (German: Badestelle, Swedish: badplats) is a designated spot where you can swim or 'bathe'. In Europe these are often signposted by the local authority, as these are the spots where swimming is officially allowed or encouraged, water quality monitored, etc.

The English translation 'bathing place' is a bit awkward but I think the concept of a designated or widely known spot for outdoor swimming is a global one. I have tried to suggest a few synonyms. Maybe 'swim spot' would be most commonly used in American English, but I don't have strong views on which one should be the primary label in iD!

[`leisure=bathing_place`](https://taginfo.openstreetmap.org/tags/leisure=bathing_place#chronology) has existed for many years as a tag, but bathing places have also been mapped using various other tags that don't really fit e.g. `leisure=water_park` and I think that's partly because a preset for `leisure=bathing_place` isn't more widely available.

The tagging scheme does already contain a preset for `leisure=swimming_area`, for areas only. In a recent [community forum discussion](https://community.openstreetmap.org/t/clarification-on-the-mapping-of-bathing-places/105017) it was agreed that `leisure=swimming_area` is for the part of the water that has been roped off or otherwise clearly designated for swimming, whereas a `leisure=bathing_place` doesn't usually have a clearly defined boundary so it will usually be mapped as a node. It's a POI - in some languages, people say the equivalent of "let's meet at the bathing place".

osmuser63783:

Böhmerheide_-_Badestelle_Landseite
([source](https://commons.wikimedia.org/wiki/File:B%C3%B6hmerheide_-_Badestelle_Landseite.jpg))

A bathing place (German: Badestelle, Swedish: badplats) is a designated spot where you can swim or 'bathe'. In Europe these are often signposted by the local authority, as these are the spots where swimming is officially allowed or encouraged, water quality monitored, etc.

The English translation 'bathing place' is a bit awkward but I think the concept of a designated or widely known spot for outdoor swimming is a global one. I have tried to suggest a few synonyms.

[`leisure=bathing_place`](https://taginfo.openstreetmap.org/tags/leisure=bathing_place#chronology) has existed for many years as a tag, but bathing places have also been mapped using various other tags that don't really fit e.g. `leisure=water_park` and I think that's partly because a preset for `leisure=bathing_place` isn't more widely available.

The tagging scheme does already contain a preset for `leisure=swimming_area`, for areas only. In a recent [community forum discussion](https://community.openstreetmap.org/t/clarification-on-the-mapping-of-bathing-places/105017) it was agreed that `leisure=swimming_area` is for the part of the water that has been roped off or otherwise clearly designated for swimming, whereas a `leisure=bathing_place` doesn't usually have a clearly defined boundary so it will usually be mapped as a node. It's a POI - in some languages, people say the equivalent of "let's meet at the bathing place".

### Related issues

Closes #1217.

### Links and data

**Relevant OSM Wiki links:**
- [`leisure=bathing_place`](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dbathing_place)

**Relevant tag usage stats:**
- 1,469 uses globally ([Taginfo](https://taginfo.openstreetmap.org/tags/leisure=bathing_place))
- [Tag history](https://taghistory.raifer.tech/#***/leisure/bathing_place)

### DIFF
--- a/data/presets/leisure/bathing_place.json
+++ b/data/presets/leisure/bathing_place.json
@@ -1,0 +1,45 @@
+{
+    "aliases": [
+        "Bathing spot",
+        "Dip spot",
+        "Dipping spot",
+        "Swim spot",
+        "Swimming spot",
+        "Wild swimming spot"
+    ],
+    "icon": "fas-water-ladder",
+    "fields": [
+        "name",
+        "access_simple",
+        "shower",
+        "toilets",
+        "drinking_water_available",
+        "picnic_table"
+    ],
+    "moreFields": [
+        "opening_hours",
+        "operator",
+        "fee",
+        "payment_multi_fee",
+        "charge_fee",
+        "supervised",
+        "lifeguard_check",
+        "lit"
+    ],
+    "geometry": [
+        "point"
+    ],
+    "terms": [
+        "bathing",
+        "dipping",
+        "swimming",
+        "swimming hole",
+        "ladder",
+        "pier",
+        "water"
+    ],
+    "tags": {
+        "leisure": "bathing_place"
+    },
+    "name": "Bathing place"
+}
--- a/data/presets/leisure/bathing_place.json
+++ b/data/presets/leisure/bathing_place.json
@@ -7,7 +7,7 @@
         "Swimming Spot",
         "Wild Swimming Spot"
     ],
-    "icon": "fas-water-ladder",
+    "icon": "maki-water",
     "fields": [
         "name",
         "access_simple",
--- a/data/presets/leisure/bathing_place.json
+++ b/data/presets/leisure/bathing_place.json
@@ -1,33 +1,25 @@
 {
     "aliases": [
-        "Bathing spot",
-        "Dip spot",
-        "Dipping spot",
-        "Swim spot",
-        "Swimming spot",
-        "Wild swimming spot"
+        "Bathing Spot",
+        "Dip Spot",
+        "Dipping Spot",
+        "Swim Spot",
+        "Swimming Spot",
+        "Wild Swimming Spot"
     ],
     "icon": "fas-water-ladder",
     "fields": [
         "name",
         "access_simple",
+        "informal",
         "shower",
         "toilets",
         "drinking_water_available",
         "picnic_table"
     ],
-    "moreFields": [
-        "opening_hours",
-        "operator",
-        "fee",
-        "payment_multi_fee",
-        "charge_fee",
-        "supervised",
-        "lifeguard_check",
-        "lit"
-    ],
     "geometry": [
-        "point"
+        "point",
+        "area"
     ],
     "terms": [
         "bathing",
@@ -41,5 +33,5 @@
     "tags": {
         "leisure": "bathing_place"
     },
-    "name": "Bathing place"
+    "name": "Bathing Place"
 }
--- a/data/presets/amenity/public_bath.json
+++ b/data/presets/amenity/public_bath.json
@@ -1,5 +1,5 @@
 {
-    "icon": "maki-water",
+    "icon": "fas-water-ladder",
     "fields": [
         "name",
         "bath/type",


Hufkratzer:
Shouldn't the preset also be allowed for areas? Some are mapped as an area, e.g. https://www.openstreetmap.org/way/1203484522, https://www.openstreetmap.org/way/56313252

data/presets/leisure/bathing_place.json
```diff
@@ -0,0 +1,45 @@
+{
+    "aliases": [
+        "Bathing spot",
+        "Dip spot",
+        "Dipping spot",
+        "Swim spot",
+        "Swimming spot",
+        "Wild swimming spot"
```
tordans, line 8: Could you double check the convention of `aliases` in other presets. I think they are either all lower case or "Camel Case".

data/presets/leisure/bathing_place.json
```diff
@@ -0,0 +1,45 @@
...
+    ],
+    "icon": "fas-water-ladder",
```
tordans, line 10: "ladder" does not sound ideal for a place that is often unofficial and will likely not have a ladder. What would be other options? 

Do we use ladder for other presets as well? (Might this cause confusion?)

tordans, line 10: "ladder" does not sound ideal for a place that is often unofficial and will likely not have a ladder. What would be other options? 

Do we use ladder for other presets as well? (Might this cause confusion?)

data/presets/leisure/bathing_place.json
```diff
@@ -0,0 +1,45 @@
...
+    "fields": [
+        "name",
+        "access_simple",
```
tordans, line 13: What are the wikis recommendations or conventions on how to tag "this is an unofficial vs. official (signed) place"? Is it access or supervised? Maybe move supervised up?

data/presets/leisure/bathing_place.json
```diff
@@ -0,0 +1,45 @@
...
+        "picnic_table"
+    ],
+    "moreFields": [
+        "opening_hours",
+        "operator",
+        "fee",
+        "payment_multi_fee",
+        "charge_fee",
+        "supervised",
+        "lifeguard_check",
```
tordans, line 26: Is this tag used for kind of official places as well? Or do they become something else once they are paid and supervised?

tordans (commented):
Thanks for working on this!

I did a quick review and write down the questions I would like to check out during a proper review. Maybe someone can help with answers.

It would also be great to have a few screenshots on the search so we can see that there are no weird results where users would now pick a wrong answer due to the new preset.

data/presets/leisure/bathing_place.json
osmuser63783: Good spot. That would be `informal` which I forgot to add to the preset.

data/presets/leisure/bathing_place.json
osmuser63783: That would normally make it an `amenity=public_bath` or possibly a `leisure=beach_resort`. So although a few dozen `bathing_place`s have `fee`, `opening_hours`, `supervised` and/or `lifeguard`, the answer is almost always `no` and `24/7`. That's why I put it in `moreFields`, not in `fields`. But I am not very familiar with the conventions of id-tagging-schema - let me know if this means that, for example, it should not be in the preset at all.

data/presets/leisure/bathing_place.json
osmuser63783: Thanks! So
- `public_bath` uses maki-water which is just a drop of water, and the same icon is used by a lot of other water-related features e.g. lake and spring, so probably a poor choice
- `beach_resort` uses [fas-umbrella-beach](https://fontawesome.com/icons/umbrella-beach?f=classic&s=solid) which is probably worse because it doesn't even show water
- `swimming_area` uses fas-swimmer which I can't find on the FAS website but looks identical to [person-swimming](https://fontawesome.com/icons/person-swimming?f=classic&s=solid)
- I suggested [fas-water-ladder](https://fontawesome.com/icons/water-ladder?f=classic&s=solid) because I was under the impression that it was not used in this project (the search didn't return anything) but I have since realised that it's actually being used in the swimming pool preset where it's called fas-swimming-pool!

data/presets/leisure/bathing_place.json
osmuser63783: Camel Case! Will update

osmuser63783:
> It would also be great to have a few screenshots on the search so we can see that there are no weird results where users would now pick a wrong answer due to the new preset.

Thanks, how do I get the bot to trigger that lets me review it? I can't spot any obvious syntax errors

tordans:
> Thanks, how do I get the bot to trigger that lets me preview it? 

Sorry, did not catch that this was not triggered yet. The setup is that for first contributors, someone with permissions has to trigger the checks. On Action will be to create the preview and comment on the PR with the link. That will happen in a few minutes. For your next PR, this will run automatically.

data/presets/leisure/bathing_place.json
tordans: I think about it this way…
- add fields/moreFields that are already used a lot in the data
- add fields/moreFields that are very helpful to have and documented/"in use"
- don't add fields/moreFields that we don't want new Users to add

So in this case the question seems to be if we want to give visibility to data that is present but maybe wrong because it should be a different primary tag. I think we should not add those fields. Experienced users will see the data regardless. But less experienced users will not add "false" data. Experienced users can change the preset and then the data will be visible.

osmuser63783:
I’ve added the informal field, removed the moreFields, added area as possible geometry and used Camel Case for the aliases and name.

This leaves the icon. I can’t find a better one among the existing icon sets I’m afraid but maybe someone else has more luck?

Here are some screenshots.

Screenshot 2024-07-02 at 23 43 46
Screenshot 2024-07-02 at 23 41 32
Screenshot 2024-07-02 at 23 45 32

tordans:
Thanks @osmuser63783 for the update.

[On the topic of the Icon](https://github.com/openstreetmap/id-tagging-schema/pull/1274/files#r1662278038): How about this for a quick fix…

1. We change https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/public_bath.json#L2 to the Ladder+Water Icon you use now
Reading [the wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dpublic_bath) this is a good fit, because there will likely be some kind of formal facility (and a ladder)

2. We then use the water drop for the new preset "Bathing Place". 
Usually I would say this icon is not good for bathing at all… but given the fact that we used it for the other preset for a while now, it seems to work well enough. And the fact that it is more generic / more plain water, fits the more inoffical style of `leisure=bathing_place`.

tordans (approved):
LGTM. Lets switch the icon and then we are good to merge, IMO.

osmuser63783:
Thanks, I've swapped the icons!

tordans:
Thanks for the update @osmuser63783 I think this is ready to be merged, now.

---

Two things I notices that I wanted to document here:

**Wiki defintions:** The german wiki page and wikidata entry was different from englisch, in that it said that the tag should only be applied to official and public places. I modified wikidata and added a callout to the wiki page https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dbathing_place . I think it is OK to merge this as is and see if any feedback comes up.

**Rendering conflict:** The area example from above https://www.openstreetmap.org/way/1203484522 has non-ideal rendering in iD. It does pick the right preset, but the rendering is based on the landuse=grass tag. I would consider this a non ideal tagging, so I think it is not an issue for this ticket.

osmuser63783:
Thanks! Yes, the German Wiki hasn't been updated yet following the international discussion that led to the English page looking the way it does.

matkoniecz:
@osmuser63783 would https://pinhead.ink/v25/person_in_water.svg ( https://pinhead.ink/#person_in_water ) icon be fitting for this preset?

PR #222, opened 2021-08-20
add "sidewalk=none" to list of deprecated tags

ZeLonewolf:
This PR standardizes the tag for streets that don't have sidewalks to `sidewalk=no`, which is the tagging used by both JOSM and iD.

Issue #1110, opened 2024-02-08
Deprecate value 'seperate' (typo) in favour of "separate".


danieldegroot2:
### Deprecated OSM Tag(s)

*:*=seperate

### Replacement OSM Tag(s)

*:*=separate

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Key%3Acycleway

### Further Information

Deprecate value 'separate' (typo) in favour of "separate".
https://taginfo.openstreetmap.org/search?q=seperate#values

Deprecation is not documented (yet). I will personally not document it, as I focus on mapping.

tyrasd:
not that it is not possible to use wildcard deprecations on just a particular tag value, and it's also not . We'd have to maintain a list of tag keys for which this typo is common, see https://taginfo.openstreetmap.org/search?q=seperate#values.

matkoniecz:
Bot edit was reviewed and is running https://www.openstreetmap.org/changeset/166247704

it will rerun from time time

> We'd have to maintain a list of tag keys for which this typo is common, see https://taginfo.openstreetmap.org/search?q=seperate#values.

Bot edit is running on

```
        {
            "key": "cycleway:right",
            "accepted_bot_replacements": {
                'seperate': 'separate',
            },
        },
        {
            "key": "cycleway:left",
            "accepted_bot_replacements": {
                'seperate': 'separate',
            },
        },
        {
            "key": "cycleway:both",
            "accepted_bot_replacements": {
                'seperate': 'separate',
            },
        },
        {
            "key": "cycleway",
            "accepted_bot_replacements": {
                'seperate': 'separate',
            },
        },
        {
            "key": "sidewalk:right",
            "accepted_bot_replacements": {
                'seperate': 'separate',
            },
        },
        {
            "key": "sidewalk:left",
            "accepted_bot_replacements": {
                'seperate': 'separate',
            },
        },
        {
            "key": "sidewalk:both",
            "accepted_bot_replacements": {
                'seperate': 'separate',
            },
        },
        {
            "key": "sidewalk",
            "accepted_bot_replacements": {
                'seperate': 'separate',
            },
        },
        {
            "key": "parking:right",
            "accepted_bot_replacements": {
                'seperate': 'separate',
            },
        },
        {
            "key": "parking:left",
            "accepted_bot_replacements": {
                'seperate': 'separate',
            },
        },
        {
            "key": "parking:both",
            "accepted_bot_replacements": {
[...]

PR #1278, opened 2024-07-08
Add deprecation rules for typos in sidewalk-related tags

waldyrious:
- `sidewalks=*` → `sidewalk=*`
- `sidewalk:*=none` → `sidewalk:*=no` (subkey variants in addition to the existing rule from #222)
- `sidewalk=seperate` → `sidewalk=separate` (plus subkey variants)

~~p.s. - Also fix indentation errors in `data/deprecated.json`.~~ Extracted to #1282

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -687,10 +687,6 @@
     "replace": {"sidewalk": "both"}
   },
   {
-    "old": {"sidewalk": "none"},
-    "replace": {"sidewalk": "no"}
-  },
-  {
     "old": {"footway": "crossing", "highway": "cycleway"},
     "replace": {"cycleway": "crossing", "highway": "cycleway"}
   },
@@ -1551,6 +1547,42 @@
   {
     "old": {"shop": "winery"},
     "replace": {"craft": "winery"}
+  },
+  {
+    "old": {"sidewalks": "*"},
+    "replace": {"sidewalk": "$1"}
+  },
+  {
+    "old": {"sidewalk": "none"},
+    "replace": {"sidewalk": "no"}
+  },
+  {
+    "old": {"sidewalk:left": "none"},
+    "replace": {"sidewalk:left": "no"}
+  },
+  {
+    "old": {"sidewalk:right": "none"},
+    "replace": {"sidewalk:right": "no"}
+  },
+  {
+    "old": {"sidewalk:both": "none"},
+    "replace": {"sidewalk:both": "no"}
+  },
+  {
+    "old": {"sidewalk": "seperate"},
+    "replace": {"sidewalk": "separate"}
+  },
+  {
+    "old": {"sidewalk:left": "seperate"},
+    "replace": {"sidewalk:left": "separate"}
+  },
+  {
+    "old": {"sidewalk:right": "seperate"},
+    "replace": {"sidewalk:right": "separate"}
+  },
+  {
+    "old": {"sidewalk:both": "seperate"},
+    "replace": {"sidewalk:both": "separate"}
   },
   {
     "old": {"sloped_curb": "0"},


tordans:
@waldyrious thanks for the PR and the indirect ping on https://github.com/openstreetmap/id-tagging-schema/pull/222 which I just merged.

Here are a few unsorted thoughts on this PR:

- The indentation fixes make it harder to review. I know this is annoying, but it is a lot easier to review when those commits are separate PRs.
- Please rebase so the merge of #222 is added.

But most importantly…
- I am not sure what our threshold or guidelines are on deprecations. How often do those typos happen? Sometimes it is better to create a MapRoulette Challenge to cleanup the typos than to have a deprecation which might never trigger.
I am also not sure if adding many small deprecations has any performance impacts (in the long run).

Lets wait for input from others on this first. And lets look at some numbers as well.

tordans:
Thanks for the update. I don't see any relevant reason to not merge this. Will wait for a bit for some feedback and merge in a few days.

---

There might be better solutions for the `seperated` case that would also cover `bicycle*`. We could even consider silently changing this, which iD does in one or two situations AFAIK. But all this would need to happen in iD, which we could still do later …

tordans (approved):

tyrasd:
I don't think the addition of `sidewalks` to `sidewalk` was a good idea. The only two values this tag currently has are `1.70` and `1,70`: how is it better to have `sidewalk=1,70` instead of `sidewalks=1,70`? This tag is not even mentioned in the list of [possible tagging mistakes](https://wiki.openstreetmap.org/wiki/Key:sidewalk#Possible_tagging_mistakes) for the sidewalk tag. :shrug:

tyrasd:
Similarly, the rule for the value `seperate` value is questionable: it is rather rare, [occuring](https://taginfo.openstreetmap.org/search?q=seperate#values) less than a dozen times each in OSM.

IMO, the list of deprecated tags is not meant to include every possible typo imaginable (otherwise the list would be unbearably long).

matkoniecz:
> IMO, the list of deprecated tags is not meant to include every possible typo imaginable (otherwise the list would be unbearably long).

See https://codeberg.org/matkoniecz/OpenStreetMap_cleanup_scripts/src/branch/master/recurrent_bot_edits/surface_fix_bad_values.py or https://codeberg.org/matkoniecz/OpenStreetMap_cleanup_scripts/src/branch/master/recurrent_bot_edits/shops_fix_bad_values.py for how it would end looking like :)

(approved bot edits, currently not running due to oauth2, all this tags existed in OSM at some point)

waldyrious:
> IMO, the list of deprecated tags is not meant to include every possible typo imaginable (otherwise the list would be unbearably long).

Just FYI, I too don't think listing every typo imaginable is a good strategy. My motivation to mark "seperate" as a typo was to (hopefully) prevent iD from suggesting it to people, because it was precisely common enough that it was indeed being suggested in the dropdown for the sidewalk tag.

> the rule for the value seperate value is also questionable: it is rather rare, [occuring](https://taginfo.openstreetmap.org/search?q=seperate#values) less than a dozen times each in OSM.

I'm not sure I understand what you mean by "less than a dozen times each". Here are the top entries that I see in [taginfo](https://taginfo.openstreetmap.org/search?q=seperate#values) at the time of writing:

| Count | Key | Value |
| ----- | --- | ----- |
| 148 | sidewalk:right | seperate |
| 102 | ramp | seperate |
| 46 | barrier | seperate |
| 40 | sidewalk:left | seperate |
| 32 | sidewalk | seperate |
| 32 | cycleway | seperate |
| 27 | crossing | seperate |
| 27 | parking:both | seperate |
| 21 | parking:left | seperate |
| 20 | parking:right | seperate |
| 19 | sideway | seperate |
| 15 | ramp:wheelchair | seperate |
| 15 | cycleway:both | seperate |
| 12 | crossing:signals | seperate |
| 11 | cycleway:left | seperate |

I suspect things might have been cleaned up shortly before when you looked at the stats — which, if that's the case, shows how prevalent that typo is given that it's already sprung back up this much since your comment and reinforces (IMHO) the need for a deprecation rule.

If this is not the right mechanism to at least prevent iD from encouraging popular typos to spread out even further, can you point me in the right direction to ensure certain key-value combinations are not shown in the tag dropdowns?

tordans:
@waldyrious iD/Rapid will always show a list of most used taginfo values. The cut of rules for that are specified in the iD codebase AFAIK. The only way to work around this AFAIK is to set `autoSuggestions:false`
- docs https://github.com/ideditor/schema-builder?tab=readme-ov-file#strings
- current usage https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema%20autoSuggestions&type=code

But this will only work when there is a preset and field specified which we don't have for many of those cases.

AFAIK there is no warning system for typos in tag keys and tag values in general. I suggest to open a issue in the iD repo so we can collect a cases that would work for this, like maybe `color`.

matkoniecz:
I have https://codeberg.org/matkoniecz/OpenStreetMap_cleanup_scripts/src/branch/master/script_assisted_cleanup/obviously_mistagged_tags_using_trivial_tag_fixes.py and https://codeberg.org/matkoniecz/OpenStreetMap_cleanup_scripts/src/branch/master/script_assisted_cleanup/obviously_mistagged_tags_using_wrong_keys.py that detect some obvious typos

So far they do it using iD presets, but found more typos than I can process in my free hobby time.

If anyone would be interested in helping (and by "helping" I do not mean "blindly mass retag", rather "carefully investigate and retag or make bot edit where it is obvious") and what is in the repo is not making clear how it can be done then feel free to ping me via codeberg issues at that repo or by https://www.openstreetmap.org/message/new/Mateusz%20Konieczny

Or just steal ideas from there :)

waldyrious:
@tordans Thanks for the context and pointers! That said, I'm still not sure why there's reluctance to use the existing mechanism of deprecation for *common*, recurrent typos. (Again, I'm not suggesting a full listing of every possible typo regardless of their tendency to show up organically, only those that do keep popping up, as seems to be the case with "seperate").

waldyrious:
@tordans ping :) I'd still like clarification regarding my previous comment, in case you're able to provide a response.

matkoniecz:
And if you have reviewed OSM data and found bunch of tags that can be safely replaced: feel free to open thread at OSM community forum proposing bot edit and ping me there

Preferably list would be formatted like https://codeberg.org/matkoniecz/OpenStreetMap_cleanup_scripts/src/branch/master/recurrent_bot_edits/unified_fixing_typos_in_keys_of_lesser_interest_for_me.py#L71 or https://codeberg.org/matkoniecz/OpenStreetMap_cleanup_scripts/src/branch/master/recurrent_bot_edits/surface_fix_bad_values.py

waldyrious:
Thanks for taking the time to re-read the discussion and comment, @tordans. I apologize for insisting but it feels to me like we might be talking over each other... I keep hearing you (and @tyrasd) saying that we should not be "adding many small deprecations" which "might never trigger", that values like `seperate` are rare, that "the list of deprecated tags is not meant to include every possible typo imaginable".

I want to be clear: **I fully agree** with that position, which is why I tried to address those statements directly (quoting them explicitly to make it clear what I was responding to), but I realize that I only once mentioned what drove me to open this PR, and what I think is the strongest argument for this change (emphasised in the quote):

> I recently fixed a bunch of typos in the values for the `sidewalk:*` tags, and there was a lot of variety, but these two (`none` and `seperated`) were the only ones that **iD prominently suggested in the autocomplete box** when I started typing them. 

(I also think there might have been some unintentional confusion regarding [`seperate`](https://taginfo.openstreetmap.org/search?q=seperate#values) vs. [`seperated`](https://taginfo.openstreetmap.org/search?q=seperated#values) — the latter indeed having a lot fewer results, especially as values of `sidewalk:*` tags; **the PR diff, however, added only the former, actually frequent value, `seperate`.**)

So when you say

> it still sounds to me you are looking for a feature to prevent tagging mistakes. That is not what the deprecations system is used for or designed for.

...allow me to specify: What I am looking for is implementing a system that **prevents iD from suggesting**, and thus **propagating**, tagging mistakes that are common and recurring (and recur independently from iD's suggestion of them).

That last point in parenthesis is important: we could argue that once these typos are in the deprecation list, that they would no longer become prevalent following a cleanup, and therefore the deprecation list would indeed be listing residual, rare values. However, as I mentioned above, "seperate" is a [well-known common misspelling](https://en.wiktionary.org/wiki/seperate) of "separate", and not just in general but in OSM as well: we can see that e.g. for [`sidewalk=seperate`](https://taghistory.raifer.tech/?#***/sidewalk/seperate) [`sidewalk:right=seperate`](https://taghistory.raifer.tech/?#***/sidewalk%3Aright/seperate), and [`sidewalk:left=seperate`](https://taghistory.raifer.tech/?#***/sidewalk%3Aleft/seperate), the value has been substantially lowered (and even zeroed) several times, but it keeps coming back. That's why I believe it is valid to keep the entry permanently in the deprecations list.

waldyrious:
As a _separate_ comment, just because I didn't want to make the previous one even longer: I must agree that my inclusion of `sidewalks` in this PR was ill-advised, as @tyrasd pointed out. Not only it has a [small distribution of values](https://taginfo.openstreetmap.org/keys/sidewalks#values), but it had [never been very frequent](https://taginfo.openstreetmap.org/keys/sidewalks#values) (at least not in the same scale as `sidewalk`/`sidewalk:*`=`seperate`), and indeed it's not suggested by iD even when typing it out almost fully:

image

So to sum up, my proposal would be to reinstate the deprecations from this PR (`sidewalk`/`sidewalk:*`=`seperate` and `sidewalk`/`sidewalk:*`=`none`), but keep `sidewalks` out, for the reasons you have mentioned: we don't want an exhaustive list of all possible (or even actually occurring with low frequency) typos and tagging mistakes.

tordans:
> What I am looking for is putting measures in place that **prevent iD from suggesting**, and thus **propagating**, tagging mistakes that are common and recurring (and recur independently from iD's suggestion of them).

I created a new PR https://github.com/openstreetmap/id-tagging-schema/pull/1507 this morning that will address this issue. Please test the preview there. (This PR took a long while due to the discussion that is linked over there.)
Once we have this merged, there is little reason for users to use the "Tags" section. Instead the fields UI will guide users to use the right tags. I disabled auto suggestions for this field so only the three valid options are given.

I think this will improve the situation considerably. 

The "Tags" section of iD will still allow to add wrong keys and values, but that is by design as this is intended to be used by users who know what they are doing.

waldyrious:
Oh, cool! I didn't know that was so close to being implemented! :heart:

> I disabled auto suggestions for this field so only the three valid options are given.

> The "Tags" section of iD will still allow to add wrong keys and values, but that is by design as this is intended to be used by users who know what they are doing.

Am I understanding correctly that this will prevent auto-suggesting typos in the dedicated sidewalks field editor, but will continue to show them in the tags UI?

If so, honestly I would argue that the reinstatement of these deprecations is still warranted, because even experienced editors are subject to misclicking or selecting a subtly erroneous value by distraction. I know I nearly did more than once, precisely because I would not expect iD to bubble up incorrect values, and I don't think it should be my responsibility as an editor to look out for improper suggestions by the editor, especially when there is already a mechanism that can prevent that from happening.

It would be one thing if the tag merely "allowed" introduction of wrong keys and values, as you say; but in the case above, it would be effectively _propagating_ them. :frowning_face:

matkoniecz:
> seperate

bot edit for these was reviewed and is now rerunning and will rerun from time to time

see https://github.com/openstreetmap/id-tagging-schema/issues/1110#issuecomment-2879660926 for details

PR #1284, opened 2024-07-13
Preview github action: allow manual rerun

tordans:
With this change, it should be possible to rerun the workflow manually to re-create the preview once it is removed due to being too old.

### DIFF
--- a/.github/workflows/build-preview.yml
+++ b/.github/workflows/build-preview.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
 
 permissions:
   contents: read


.github/workflows/build-preview.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
```
k-yle, line 7: This CI script relies on `github.event.number`, which will be `undefined` if the script is triggered using `workflow_dispatch`. So I think we'll need to configure [an input](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs):

```suggestion
  workflow_dispatch:
    inputs:
      pr_number:
        required: true
        type: number
```

This will also require changes to other parts of the script:
```diff
- echo ${{github.event.number}} > ./pr_number
+ echo ${{github.event.inputs.pr_number || github.event.number}} > ./pr_number
```

After that, I think we'll also have to change this line:

https://github.com/tordans/id-tagging-schema/blob/3cdd38dfad5026563382a2d000623bf38e7abc7f/.github/workflows/deploy-preview.yml#L12

I'm less familiar with this syntax so I don't have exact suggestion

tyrasd: If I see this correctly, we can just omit the `github.event.workflow_run.event == 'pull_request' &&` part in the `deploy-preview.yml` workflow, as it only runs when the `"Build Preview"` workflow is _completed_ (and that one is only triggered by pull requests and with this PR also by manual triggering it), which is precisely when we want the deploy step to happen.

k-yle:
for reference: my attempt is at https://github.com/openstreetmap/id-tagging-schema/commit/f955b26875d854e68e7acdeff13ab19849ac0125. This resolves the limitation from https://github.com/openstreetmap/id-tagging-schema/pull/1284#discussion_r1677088743, but there was some other problem which could not be bypassed. But I've since forgotten what the issue was 🤦

matkoniecz:
my plan is to tackle it experimentally for now: setup a different repository, put this code there on the main branch and make some test PRs to check is it actually working

done at https://github.com/matkoniecz/test_manual_rerun

matkoniecz:
https://github.com/matkoniecz/test_manual_rerun/pull/8 is now runnning build

retriggered it manually from actions tab

.github/workflows/build-preview.yml
matkoniecz: @tordans 

Are you interested in acting on these suggestions?

I am considering taking over this PR and trying to implement it myself.

tordans: @matkoniecz Feel free to take over, I have too many open things ATM to prioritize this. - FYI: I had good experience with LLM to iterate on GitHub Action topics; they ca be very tedious to research and test...

matkoniecz: yeah, rubber ducking with LLM was helpful already

its claims and advise were wrong but listing why lead me to the problem

matkoniecz: > If I see this correctly, we can just omit the github.event.workflow_run.event == 'pull_request' && part in the deploy-preview.yml workflow, as it only runs when the "Build Preview" workflow is completed (and that one is only triggered by pull requests and with this PR also by manual triggering it), which is precisely when we want the deploy step to happen.

done in https://github.com/openstreetmap/id-tagging-schema/pull/2583 as - as I understand - it makes sense in general

PR #1291, opened 2024-07-15
Update strings cross-referenced from seamark/beacon_isolated_danger/shape

kjonosm:


tyrasd:
@kjonosm please add this change to the existing PR #1014

Issue #1271, opened 2024-06-23
Deprecate `cycleway=opposite` family


tordans:
### Deprecated OSM Tag(s)

cycleway=opposite

### Replacement OSM Tag(s)

None?

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Proposal:Deprecate_cycleway%3Dopposite_family

### Further Information

The proposal to deprecate the `cycleway=opposite` family is accepted (or as good as, given that right now the state of page was not change but all votes are in favor).

So lets talk about what we can do from the id-tagging-schema point of view to deprecate the tagging…

Following the examples https://wiki.openstreetmap.org/wiki/Proposal:Deprecate_cycleway%3Dopposite_family#Tagging_and_Examples we should be able to add these deprecations:

_(Note: It is possible to add deprecation rules with multiple tags like in [this example](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/deprecated.json#L78-L81).)_

```json
  {
     "old": {"oneway": "yes", "cycleway": "opposite"},
     "replace": {"oneway": "yes", "oneway:bicycle": "no"}
  },
  {
     "old": {"oneway": "yes", "cycleway": "opposite_lane"},
     "replace": {"oneway": "yes", "oneway:bicycle": "no", "cycleway:right": "no", "cycleway:left": "lane", "cycleway:left:oneway": "-1"}
  },
  {
     "old": {"oneway": "yes", "cycleway": "opposite_share_busway"},
     "replace": {"oneway": "yes", "oneway:bus": "no", "oneway:bicycle": "no", "cycleway:right": "no", "cycleway:left": "share_busway", "cycleway:left:oneway": "-1"}
  },
```

But looking at the usage numbers in https://wiki.openstreetmap.org/wiki/Proposal:Deprecate_cycleway%3Dopposite_family#Current_usage_in_the_OSM_database we really should only include the first two plus rather deprecations for `cycleway:left=opposite` and `cycleway:left=opposite_lane`. What would those two look like @SupaplexOSM?

tordans:
Update
* https://github.com/openstreetmap/id-tagging-schema/pull/1295
* https://github.com/openstreetmap/id-tagging-schema/pull/1297

PR #1295, opened 2024-07-25
Add deprecation rules for `cycleway=opposite` #Part1

tordans:
This solves the main issues described in https://github.com/openstreetmap/id-tagging-schema/issues/1271 to put the [recently approved](https://wiki.openstreetmap.org/wiki/Proposal:Deprecate_cycleway%3Dopposite_family) deprecation of the `cycleway=opposite` tagging into our deprecation rules.

1. It updates the custom translations for the two opposite option to reflect the deprection.
2. It adds the most common and simplest deprecation to to automatically updated
   This is the one with the most usage (46k + 4k) ATM ([Usage stats](https://wiki.openstreetmap.org/wiki/Proposal:Deprecate_cycleway%3Dopposite_family#Current_usage_in_the_OSM_database))

I suggest we merge this separate of other possible deprecations. I will create a separate issue for those.

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -2031,5 +2031,13 @@
   {
     "old": {"industrial": "brickworks"},
     "replace": {"industrial": "brickyard"}
+  },
+  {
+     "old": {"cycleway": "opposite"},
+     "replace": {"oneway": "yes", "oneway:bicycle": "no", "cycleway:both": "no"}
+  },
+  {
+     "old": {"cycleway:left": "opposite"},
+     "replace": {"oneway": "yes", "oneway:bicycle": "no", "cycleway:left": "no"}
   }
 ]
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -2034,10 +2034,10 @@
   },
   {
      "old": {"cycleway": "opposite"},
-     "replace": {"oneway": "yes", "oneway:bicycle": "no", "cycleway:both": "no"}
+     "replace": {"oneway:bicycle": "no", "cycleway:both": "no"}
   },
   {
      "old": {"cycleway:left": "opposite"},
-     "replace": {"oneway": "yes", "oneway:bicycle": "no", "cycleway:left": "no"}
+     "replace": {"oneway:bicycle": "no", "cycleway:left": "no"}
   }
 ]
--- a/data/fields/cycleway.json
+++ b/data/fields/cycleway.json
@@ -36,17 +36,17 @@
                 "title": "Bike Lane Shared With Bus",
                 "description": "A bike lane shared with a bus lane"
             },
-            "opposite_lane": {
-                "title": "Opposite Bike Lane",
-                "description": "A bike lane that travels in the opposite direction of traffic"
-            },
-            "opposite": {
-                "title": "Contraflow Bike Lane",
-                "description": "A bike lane that travels in both directions on a one-way street"
-            },
             "separate": {
                 "title": "Cycleway Mapped Separately",
                 "description": "Indicates that cycleway was mapped as a separate geometry"
+            },
+            "opposite_lane": {
+                "title": "(Deprecated) Opposite Bike Lane",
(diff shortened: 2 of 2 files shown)


data/deprecated.json
```diff
@@ -2031,5 +2031,13 @@
...
     "old": {"industrial": "brickworks"},
     "replace": {"industrial": "brickyard"}
+  },
+  {
+     "old": {"cycleway": "opposite"},
+     "replace": {"oneway": "yes", "oneway:bicycle": "no", "cycleway:both": "no"}
+  },
+  {
+     "old": {"cycleway:left": "opposite"},
+     "replace": {"oneway": "yes", "oneway:bicycle": "no", "cycleway:left": "no"}
```
tordans, line 2041: Oh, wait, does this need to be `cycleway:right`? I will double check.

tordans:
## Testing

Query https://overpass-turbo.eu/s/1OIw

### test 1
https://pr-1295--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w114097468&locale=en&map=19.01/52.48624/13.42682
image
image


==> All good


### test2
https://pr-1295--ideditor-presets-preview.netlify.app/id/dist/#background=Berlin-2023&disable_features=boundaries&id=w5216567&locale=en&map=19.27/52.54829/13.19451

https://www.mapillary.com/app/user/tordans?lat=52.548182299999&lng=13.194176100002&z=17&mapStyle=Mapillary+light&pKey=785476499277461&focus=photo&x=0.492695052003444&y=0.5125723764267668&zoom=0

image
image

==> all good; would be cycleway:both=no in this case, but that can be updated manually

### test3

https://pr-1295--ideditor-presets-preview.netlify.app/id/dist/#background=Brandenburg-DOP20c&disable_features=boundaries&id=w10612191&locale=en&map=20.00/52.39148/13.09470&photo_overlay=mapillary
image

> [!INFO]
> This one is more interesting, I will update this issue later today

data/deprecated.json
k-yle, line 1: I was confused to see many cases of `cycleway:right=opposite` in countries that drive on the right. But regardless, _I think_ the deprecation for `:both`, `:left`, and `:right` is safe.... Is there a reason why `:right` is not included, besides that it's less common?

k-yle (approved):

data/deprecated.json
tordans: > Is there a reason why :right is not included, besides that it's less common?

[Its ~500](https://taginfo.openstreetmap.org/search?q=cycleway%3Aright%3Dopposite) usage so below my ~1k threshold that I tend to look at when adding deprecation rules. We said in our guidelines that we don't want to add them for cases that are low, but we have no specific number for that.

tordans:
Thanks for the review. I will merge this now.

data/deprecated.json
```diff
@@ -2031,5 +2031,13 @@
   {
     "old": {"industrial": "brickworks"},
     "replace": {"industrial": "brickyard"}
+  },
+  {
+     "old": {"cycleway": "opposite"},
+     "replace": {"oneway:bicycle": "no", "cycleway:both": "no"}
+  },
+  {
+     "old": {"cycleway:left": "opposite"},
```
tyrasd, line 2040: maybe for consistency there could also be a rule for [`cycleway:right=opposite`](https://taginfo.openstreetmap.org/tags/cycleway%3Aright=opposite#overview); I'm thinking of countries with left hand default traffic where this tag would have made sense?

tordans: To me the ~500 times that is used does not merit adding it here. I think those need to be cleaned up with a MapRoulette Challenge or something similar.

But if you prefer, we can add it, of course. Ideally we should try to update the guidelines with something along the lines "low usage is OK when is "a system" of deprecations being added" (wording TODO).

tyrasd: Ah, true. It also seems like much of the [data](https://overpass-turbo.eu/s/1Ynp) is not even in left hand traffic countries. So, a deprecation rule might be even less useful for those. Let's keep it as is. :+1:

PR #1297, opened 2024-07-26
Add deprecation rules for `cycleway=opposite_lane` #Part2

tordans:
This is based on top of https://github.com/openstreetmap/id-tagging-schema/pull/1295 and https://github.com/openstreetmap/id-tagging-schema/pull/1295 should be merged first.

---

I split off the `opposite_lane` because they are a bit more complex.
We are talking about 5.136 + 7.158 cases that this rule would apply to.

---

The replacement is following the examples from the proposal https://wiki.openstreetmap.org/wiki/Proposal:Deprecate_cycleway%3Dopposite_family#Tagging_and_Examples an I checked them with @SupaplexOSM.

---

One thing that makes this more complex is, that countries with Left-hand traffic, those deprecations might not be right.
However, it looks like those are only 18 cases for UK https://taginfo.geofabrik.de/europe:britain-and-ireland/search?q=cycleway%3Aleft%3Dopposite#tags because the more common tag would be https://taginfo.geofabrik.de/europe:britain-and-ireland/search?q=cycleway%3Aright%3Dopposite#tags there (also only ~200 cases).

---

I think it would be fine to merge this.
However I also think, it would be fine to leave it and handle those cases with a MapRoulette Project instead.

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -2031,5 +2031,13 @@
   {
     "old": {"industrial": "brickworks"},
     "replace": {"industrial": "brickyard"}
+  },
+  {
+     "old": {"cycleway": "opposite"},
+     "replace": {"oneway": "yes", "oneway:bicycle": "no", "cycleway:both": "no"}
+  },
+  {
+     "old": {"cycleway:left": "opposite"},
+     "replace": {"oneway": "yes", "oneway:bicycle": "no", "cycleway:left": "no"}
   }
 ]
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -2034,10 +2034,10 @@
   },
   {
      "old": {"cycleway": "opposite"},
-     "replace": {"oneway": "yes", "oneway:bicycle": "no", "cycleway:both": "no"}
+     "replace": {"oneway:bicycle": "no", "cycleway:both": "no"}
   },
   {
      "old": {"cycleway:left": "opposite"},
-     "replace": {"oneway": "yes", "oneway:bicycle": "no", "cycleway:left": "no"}
+     "replace": {"oneway:bicycle": "no", "cycleway:left": "no"}
   }
 ]
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -2039,5 +2039,13 @@
   {
      "old": {"cycleway:left": "opposite"},
      "replace": {"oneway:bicycle": "no", "cycleway:left": "no"}
+  },
+  {
+     "old": {"cycleway": "opposite_lane"},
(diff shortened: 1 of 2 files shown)


data/deprecated.json
```diff
@@ -2031,5 +2031,21 @@
...
+     "old": {"cycleway": "opposite"},
+     "replace": {"oneway:bicycle": "no", "cycleway:both": "no"}
+  },
+  {
+     "old": {"cycleway:left": "opposite"},
+     "replace": {"oneway:bicycle": "no", "cycleway:left": "no"}
+  },
+  {
+     "old": {"cycleway": "opposite_lane"},
+     "replace": {"oneway:bicycle": "no", "cycleway:left": "lane", "cycleway:left:oneway": "-1"}
```
k-yle, line 2045: I think this rule is wrong in countries that drive on the left (which affects [67 ways](https://overpass-turbo.eu/s/1OWv))

It's also wrong if the road has `oneway=-1` (which affects [2 ways](https://overpass-turbo.eu/s/1OWw))

So I think these 69 examples need to be manually fixed before we can introduce the rule

...or we could make a fairly simple change to iD, so that deprecation rules can be limited to certain countries or driving sides. (we already have this data from https://github.com/rapideditor/country-coder)

tordans: Thanks for confirming the issue on left-hand traffic!

I will wait for https://github.com/openstreetmap/id-tagging-schema/pull/1295 to be merged before thinking more about  this, though. Just not to invest too much time – given our longer backlog. :)

matkoniecz: #1295 is now merged

matkoniecz: > So I think these 69 examples need to be manually fixed before we can introduce the rule

the problem is that new ones can appear

I would say that such changes, if accepted by community, can be easily mass-executed in a bot edit
And if not accepted by community they should not be done

(and technical limitations are not a problem here: if someone obtains community approval I can run the bot edit itself)

tordans:
> I plan to close it on account of offering bad migration in countries with left-hand drive 

I agree, we did not find a way to resolve this issue so cleaning up the old data is something that has to happen with MapRoulette or by other means.

> AND being a bot edit that should be first approved by community.

I strongly disagree and I hope that we find a place where we can align our communication on this soon. Not every QA suggestion is a bot edit. Framing it as such is harmful to the progress we can make to maintain quality data. I applaud your work on true bot edit cleanups but I wished you would look at other means of data cleanup with more nuance and a more open mindset. Based on my work with real world cases for this specific case of cycleway=opposite* it does not allow for bot edits; it needs a human in the loop process like MapRoulette and the Tag-Update-Validations that iD provides because the cases need to be reviewed.

PR #1298, opened 2024-07-26
Add preset for office=translator

deevroman:
https://wiki.openstreetmap.org/wiki/Tag:office=translator

### DIFF
--- a/data/presets/office/translator.json
+++ b/data/presets/office/translator.json
@@ -1,0 +1,17 @@
+{
+    "icon": "fas-language",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "office": "translator"
+    },
+    "terms": [
+        "translator",
+        "translation",
+        "translation office",
+        "translation agency"
+    ],
+    "name": "Translation Office"
+}


Zverik:
I endorse this. Saw dozens of translation offices in Tallinn, but had to use `office=company` for those because didn't know of this tag.

kjonosm:
image

tordans:
A few research notes:

**Icon:** 
- We don't use the icon, yet https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema+language&type=code
- => I think it a good fit here

**Fields language**
- https://wiki.openstreetmap.org/wiki/Tag:office=translator mentions the `language:*=*` tags which are "in use"  but only ~15k times https://wiki.openstreetmap.org/wiki/Key:language:*#Statistics
- However, we have a language https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/language_multi.json field which we use on the language school preset https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/language_school.json#L10
- => I think we should test this field and add it if it works like defined in the `office=translator` docs

**Fields general**
- This will pull all fields from https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/office.json, right?
- => Lets check if those all fit and look at a few examples

**Testing**
Ideally lets add the testing notes as described in https://github.com/openstreetmap/id-tagging-schema/pull/1239/files#diff-18813c86948efc57e661623d7ba48ff94325c9b5421ec9177f724922dd553a35R37-R70 to have a nice overview of things…

deevroman:
> Fields language

It seems to me that we should not focus on this key yet, since it is difficult to implement it due to the suffix of the language. At the same time, it was used only 26 times: https://overpass-turbo.eu/s/1OSO

> => Lets check if those all fit and look at a few examples



It seems to look normal

Search example:


https://github.com/user-attachments/assets/c1581c18-3e5f-4239-b38c-5cdea6b8d6b9

tordans (approved):
Looks good, thanks.

Example https://pr-1298--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=n11376034657&locale=en&map=20.00/50.47281/17.32693

PR #1299, opened 2024-07-26
Add preset for entrance=shop

deevroman:
https://wiki.openstreetmap.org/wiki/Tag%3Aentrance%3Dshop

### DIFF
--- a/data/presets/entrance/shop.json
+++ b/data/presets/entrance/shop.json
@@ -1,0 +1,13 @@
+{
+    "icon": "maki-entrance-alt1",
+    "geometry": [
+        "vertex"
+    ],
+    "tags": {
+        "entrance": "shop"
+    },
+    "terms": [
+        "door"
+    ],
+    "name": "Shop Entrance"
+}


Zverik (approved):
Looks good, tag used 33k times

kjonosm:
image

tordans:
What are the search terms you would use to add this preset?
I wonder if we are on the right track by adding separate presets for all of those entrance values.
([There are a few already…](https://github.com/openstreetmap/id-tagging-schema/tree/main/data/presets/entrance))

My usage scenario is, to search for "entrance" in iD – but never more concrete like "main entrance". I then select the general entrance and specify the type if appropriate.

Give this usage, it would be better to make all the specific presets unsearchable and add search terms to the general entrance preset. (Something similar was talked about for crossings in https://github.com/openstreetmap/id-tagging-schema/pull/1044.)

Does this make sense? And would you be up to expand this PR to include this refactoring?


---

An argument for this preset is, that there is a dedicated Wiki page https://wiki.openstreetmap.org/wiki/Tag%3Aentrance%3Dshop and [Wikidata Item](https://wiki.openstreetmap.org/wiki/Item:Q22169). However, usage is quite lower (for now) than the other values:
> image
[Source](https://taginfo.openstreetmap.org/keys/entrance#values)

Dimitar5555:
> My usage scenario is, to search for "entrance" in iD – but never more concrete like "main entrance". I then select the general entrance and specify the type if appropriate.
> 
> Give this usage, it would be better to make all the specific presets unsearchable and add search terms to the general entrance preset. (Something similar was talked about for crossings in #1044.)
> 
> Does this make sense? And would you be up to expand this PR to include this refactoring?

The point of having them as searchable presets is to reduce the amount of clicks needed to apply them. Try to add the entrances of a dozen long panel buildings using your method and then using the specific preset.

deevroman:
> What are the search terms you would use to add this preset?

shop :) Note that the preset only works for vertex

>  I wonder if we are on the right track by adding separate presets for all of those entrance values.

I don't think that's the problem. The most common tag `entrance=yes ' is now in the first place



The only problem that bothers me when adding more presets for inputs is the same icons

tordans (approved):

tordans:
All right, thanks for the discussion. I think we should merge this and see if it bothers us to have too many entrance presets. We can always rework the system later…

PR #1302, opened 2024-07-28
added memorial=ghost_bike

mnalis:
as documented on https://wiki.openstreetmap.org/wiki/Tag:memorial=ghost_bike and requested in e.g. https://github.com/Zverik/every_door/issues/764

### DIFF
--- a/data/fields/memorial.json
+++ b/data/fields/memorial.json
@@ -7,6 +7,7 @@
             "blue_plaque": "Blue Plaque",
             "bust": "Bust",
             "cross": "Memorial Cross",
+            "ghost_bike": "Ghost Bike",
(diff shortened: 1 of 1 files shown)


tordans:
I think we can merge this.
The previous cut of in tags was right before gost bike in https://taginfo.openstreetmap.org/keys/memorial#values

Example: https://pr-1302--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=n9635107500&locale=en&map=20.00/52.50019/13.44397
Overpass: https://overpass-turbo.eu/s/1ORe

tordans (approved):

PR #1308, opened 2024-08-08
Revert "Add deprecation rules for typos in sidewalk-related tags"

tyrasd:
Reverts openstreetmap/id-tagging-schema#1278

For reasoning, see https://github.com/openstreetmap/id-tagging-schema/pull/1278#issuecomment-2276024097 and https://github.com/openstreetmap/id-tagging-schema/pull/1278#issuecomment-2276031945

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -687,6 +687,10 @@
     "replace": {"sidewalk": "both"}
   },
   {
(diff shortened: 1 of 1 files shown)


tordans (approved):
I agree with low usage we should not add them here but use other means to cleanup the data.

PR #1309, opened 2024-08-09
Guidelines: Clarify rules for deprecations

tordans:
Clarify when a deprecation should be considered based on tag volumen.

This is a follow up to https://github.com/openstreetmap/id-tagging-schema/pull/1278#issuecomment-2276024097 to have a reference for the next time this comes up.


tyrasd (approved):
:+1: It's good to have this info included. FYI: I've tweaked the wording slightly to indicate that "high" might depend on the use case, and potentially other factors as well.

PR #1320, opened 2024-08-22
Add field for `irrigation=pivot`

arch0345:

Adds a [Center-Pivot Irrigation](https://en.wikipedia.org/wiki/Center-pivot_irrigation) field for the Farmland preset. These types of fields are pretty common in the US (and by the looks of things seem to exist around the world)
image

image

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Tag:irrigation=pivot

**Relevant tag usage stats:**
> [image](https://taginfo.openstreetmap.org/tags/irrigation=pivot)
image

### DIFF
--- a/data/fields/irrigation_pivot.json
+++ b/data/fields/irrigation_pivot.json
@@ -1,0 +1,16 @@
+{
+    "key": "irrigation",
+    "type": "defaultCheck",
+    "label": "Center-Pivot Irrigation",
+    "terms": [
+        "central pivot irrigation",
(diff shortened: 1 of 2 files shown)


tordans:
@arch0345 this looks good to me. Could you add the testing docs to this PR so we document and double check the less obvious cases, like the wikidata item (info-I) things. Happy to merge then.

tordans (approved):

PR #1324, opened 2024-08-25
Add `building_area_yes` field to Funeral Service Hall preset

### LINKED ISSUE #1313: Adding amenity=funeral_hall from search menu removes building=funeral_hall
Adding amenity=funeral_hall from search menu removes tag building=funeral_hall.
1. original tag

2. choosing amenity=funeral_hall from menu

3. tags after adding amenity=funeral_hall

arch0345:

This will prevent building tags from being removed when selecting this preset on a building

### Related issues

Closes #1313

### DIFF
--- a/data/presets/amenity/funeral_hall.json
+++ b/data/presets/amenity/funeral_hall.json
@@ -3,6 +3,7 @@
     "fields": [
         "name",
         "operator",
+        "building_area_yes",
         "address"
     ],
     "moreFields": [


tordans (approved):
https://github.com/openstreetmap/id-tagging-schema/issues/1313#issuecomment-2323362533 says its all good now and I also see that the reported issue does not happen anymore. 

I think we can merge this.

PR #1326, opened 2024-08-28
Adjust parking suggestions: promote street_side to reflect usage

### LINKED ISSUE #1325: Parking lane and street side should be inverted
According to the [wiki](https://wiki.openstreetmap.org/wiki/Tag:parking=lane), you could map the whole city with `parking=lane`, but I'm sure that in most cases, when a mapper wants to add parking, it's probably `parking=street_side`. This is confirmed by [taginfo](https://taginfo.openstreetmap.org/keys/parking#values).

The current drop-down menu looks like this:
img
It's even worse in French because the words are basically synonyms, I can't even tell if the translations are inverted. (I'm French btw)
This has led me and I think many other mappers to favor "lane" instead of "street_side".

Also, not sure if this is intended, but when you select "Street-Side", the Type field disappears.

Nekzuris:

The usage of Street-Side parking has greatly increased, but it was still ranked very low in the suggestions.

### Related issues
Closes #1325

### Links and data

**Relevant OSM Wiki links:**
https://wiki.openstreetmap.org/wiki/Key:parking

**Relevant tag usage stats:**
https://taginfo.openstreetmap.org/keys/parking#values
https://taghistory.raifer.tech/#***/parking/lane&***/parking/street_side

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

     [Learn more…](https://github.com/tordans/id-tagging-schema/blob/main/CONTRIBUTING.md#info-i)
 -->

### Wording

- [ ] American English
```

</details>


Nekzuris:
## Test-Documentation
Before & After
ba

maro-21:
It should be on the second position. The order of the list should be I think the same as in:
https://taginfo.openstreetmap.org/keys/parking#values

tordans (approved):

tordans:
I am happing with merging the change. I think street side and roadside lane should be next to eachother so people research which to pick


https://wiki.openstreetmap.org/wiki/Tag:parking%3Dlane


---

Another improvement would be to add more help text right in iD, like with the strings description field.
More on that in https://github.com/ideditor/schema-builder/pull/152/files and https://github.com/openstreetmap/id-tagging-schema/blob/e25f91e8c0590024cfa6e23fe2a425594f0985b5/data/fields/marker.json#L7-L10 is an example. Especially for `lane` and `street_side` those inline text would help some users, I think. They are pretty hard to discover ATM because they are not visible by default.

---

A nice follow up would be to add icons to the most important options, like we have for crossings.

It might be possible to reuse some from StreetComplete, but the need to be simplified I think – see https://github.com/zlant/parking-lanes/issues/69

The schema is https://github.com/ideditor/schema-builder?tab=readme-ov-file#icons

PR #1329, opened 2024-09-01
Add preset aeroway=heliport

SteveLz:
Wiki: https://wiki.openstreetmap.org/wiki/Tag:aeroway%3Dheliport

taginfo: https://taginfo.openstreetmap.org/tags/aeroway%3Dheliport

### DIFF
--- a/data/presets/aeroway/heliport.json
+++ b/data/presets/aeroway/heliport.json
@@ -1,0 +1,28 @@
+{
+    "icon": "maki-heliport",
+    "fields": [
+        "name",
+        "iata",
+        "icao",
+        "address",
+        "operator",
+        "operator/type",
+        "{@templates/internet_access}"
+    ],
+    "moreFields": [
+        "{@templates/contact}",
+        "{@templates/internet_access}",
+    ],
+    "geometry": [
+        "area",
+        "point"
+    ],
+    "tags": {
+        "aeroway": "heliport"
+    },
+    "terms": [
+        "aeroway",
+        "helicopter"
+    ],
+    "name": "Heliport"
+}
--- a/data/presets/aeroway/heliport.json
+++ b/data/presets/aeroway/heliport.json
@@ -11,7 +11,7 @@
     ],
     "moreFields": [
         "{@templates/contact}",
-        "{@templates/internet_access}",
+        "{@templates/internet_access}"
     ],
     "geometry": [
         "area",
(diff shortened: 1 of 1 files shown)


tordans:
Could someone look into adding a test case for this https://github.com/openstreetmap/id-tagging-schema/blob/main/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L24-L70

novolife:
@tordans  I want to test this PR but the link seems failed, should I open a new PR then test or this PR can be test after some actions?

tordans:
@novolife any change triggers the preview build. I just moved the name so now the preview works again.

novolife:
@tordans Could you please check the test results?

tordans:
@novolife I am trying to learn more about this area of OSM.

So we have [`aeroway=helipad`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/aeroway/helipad.json) which is the actual "H+Circle". Used 66k times.
* https://wiki.openstreetmap.org/wiki/Tag:aeroway%3Dhelipad

And now we want do add the `heliport` which is the area around this `helipad`, right?
But not all `helipad` have a `heliport` it could also be a [`aerodrome`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/aeroway/aerodrome.json) or other.

I am very unsure if our users will understand this distinction based on the search results in the presets.

One thing that would make this easier is to understand IMO, if we only allowed geometry=area for `heliport`. The usage is ~500 for area and ~100 for point ([source](https://wiki.openstreetmap.org/wiki/Tag:aeroway%3Dheliport)). I did not spot check the point data in the wild but I think it is better to prevent miss-tagging than handling those 100 cases…

---

TBH I ran into this issue myself and I debugged a while why internet access did not show as field but now I understand I looked at the heliPAD AREA and not the heliport area…

---

Out of curiosity: Why is https://pr-1329--ideditor-presets-preview.netlify.app/id/dist/#locale=en&disable_features=boundaries&map=21.60/51.46941/-0.17863&background=Bing&id=w108957278 an helipad when there are landing zones for planes? Our icon suggest that this is for helicopter only?

matkoniecz:
Wait, is heliport supposed to include `aeroway=parking_position` objects, `aeroway=helipad` objects or both?

matkoniecz:
Maybe heliport can have helicopter + some extra stuff in icon to signify that it is about wider facility?

PR #1335, opened 2024-09-04
Include more countries for Footwear Decontamination Station preset

arch0345:

This will prevent iD from erroneously flagging points tagged with `man_made=footwear decontamination` as invalid.
image

These stations are pretty common at trailheads in the PNW
image

I also added 'boot brush' as a term


data/presets/man_made/footwear_decontamination-AU-NZ.json
k-yle, line 1: this file will need to be renamed as well, to remove the `-AU-NZ` suffix

k-yle (commented):
Regarding the iD warning - this seems like a bug that should be fixed, regardless of whether this PR is accepted. Right now, it's not allowed as a point nor an area 😆




The parent preset (`man_made=*`) is supposed to allow any geometry:

https://github.com/openstreetmap/id-tagging-schema/blob/6d863935200bbe1150377be01ba1f03c0aa82848/data/presets/man_made/_yes.json#L8-L13

data/presets/man_made/footwear_decontamination-AU-NZ.json
tordans: I went ahead and did this.

matkoniecz:
As I understand problem that this PR tried to fix was resolved by https://github.com/openstreetmap/iD/pull/10459 - right?

In such case reason for its existence seems gone.

arch0345:
Yeah I think it'll still be worth removing the regional restriction for this preset. It would be quite useful to have when mapping hiking areas. I've also encountered some in Canada and Korea a few years back, not sure how prevalent these are in other parts of the world though.

k-yle:
> maybe instead make PR adding those areas?

if we go with this option, I suggest also adding `SG`, `ID`, and `HK` where I've seen them too

data/presets/man_made/footwear_decontamination.json
FloEdelmann, line 1: Not sure if renaming this file should be done in two steps like [newly](https://github.com/openstreetmap/id-tagging-schema/commit/b22bc01788bb3238dd81ea492ebe39cafca59689) described here?

https://github.com/openstreetmap/id-tagging-schema/blob/main/GUIDELINES.md#file-name-conventions-and-directory-structure

data/presets/man_made/footwear_decontamination.json
```diff
@@ -18,7 +18,8 @@
         "shoe cleaning",
         "boot cleaning",
         "contamination station",
-        "decontamination station"
+        "decontamination station",
+        "boot brush"
```
FloEdelmann: Please keep the search terms ordered alphabetically, see https://github.com/openstreetmap/id-tagging-schema/blob/main/GUIDELINES.md#2-design-the-preset

matkoniecz:
note

(and I am neither maintainer nor PR author so I cannot resolve them, as I have no write permission on the branch)

data/presets/man_made/footwear_decontamination.json
k-yle: in this case, NSI does not use the preset, so the rename won't be as chaotic as https://github.com/openstreetmap/id/issues/11527 , but it will still create extra work for translators unfortunately.

so we probably don't need a placeholder preset for this case?

tordans: I agree, kyle. And our process is still to change the filenames, so I think we should continue following this process.

matkoniecz:
is this PR abandoned?

arch0345:
All conflicts should be resolved now

matkoniecz (commented):
ideally https://github.com/openstreetmap/id-tagging-schema/issues/2767 would be resolved first to not lose translation info...

PR #1348, opened 2024-09-27
deploy-preview: Change link to paris

tordans:
This change the deploy preview link by adding `?id=n17807753` which is Paris so users don't start at the work which in my experience is never where I do mapping.
My idea is, that it is quicker and easier to review PRs when I don't have to find a node or way to see the changes, first.

We can pick any location…

**[Example](https://pr-1344--ideditor-presets-preview.netlify.app/id/dist/?id=n17807753#locale=en**

Another change is to make the click target bigger than the short "here".

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -78,7 +78,7 @@ jobs:
                 owner: context.repo.owner,
                 repo: context.repo.repo,
                 issue_number: pullRequestNumber,
-                body: `${start} You can preview the tagging presets of this pull request [here](https://pr-${pullRequestNumber}--ideditor-presets-preview.netlify.app/id/dist/#locale=en).`
+                body: `${start} You can preview the tagging presets of this pull request [in this deploy preview](https://pr-${pullRequestNumber}--ideditor-presets-preview.netlify.app/id/dist/?id=n17807753#locale=en).`
               });
             } else {
               console.log(`Preview URL comment already added to PR #${pullRequestNumber}`);


.github/workflows/deploy-preview.yml
```diff
@@ -78,7 +78,7 @@ jobs:
                 owner: context.repo.owner,
                 repo: context.repo.repo,
                 issue_number: pullRequestNumber,
-                body: `${start} You can preview the tagging presets of this pull request [here](https://pr-${pullRequestNumber}--ideditor-presets-preview.netlify.app/id/dist/#locale=en).`
+                body: `${start} You can preview the tagging presets of this pull request [in this deploy preview](https://pr-${pullRequestNumber}--ideditor-presets-preview.netlify.app/id/dist/?id=n17807753#locale=en).`
```
tyrasd, line 81: The standalone version of iD does only look at the hash part of the url:

```suggestion
                body: `${start} You can preview the tagging presets of this pull request [in this deploy preview](https://pr-${pullRequestNumber}--ideditor-presets-preview.netlify.app/id/dist/#id=n17807753&locale=en).`
```

But for the  above mentioned alternatives, using the `map` parameter would be more appropriate:

```suggestion
                body: `${start} You can preview the tagging presets of this pull request [in this deploy preview](https://pr-${pullRequestNumber}--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=17.00/14.65485/121.06466).`
```

tyrasd (changes requested):
Hmm… I was originally thinking that starting from a global view is slightly better, because one cannot know which map features are interesting for a particular PR… So, one essentially has to search a matching test area anyway. Starting in an urban area would help for features that are common in cities, though :thinking: so, why not.

Regarding which location to choose: Maybe we can use London, as a nod to  it being the birthplace of OSM? Or set it to the next host city for the SotM (Manila), and update it once a year… 

Also, I'd not use the `id` parameter to set the starting location, but instead the `map` parameter (see below).

tordans:
Superseeded by https://github.com/openstreetmap/id-tagging-schema/pull/1478

matkoniecz:
> Maybe we can use London, as a nod to it being the birthplace of OSM? Or set it to the next host city for the SotM (Manila), and update it once a year…

I would wait a bit with updating it to Bogota, even if SOTM venue is announced - it is nice to have reminder about nice SOTM in Paris for now

PR #1353, opened 2024-10-06
Add field self_checkout to shop preset

lightonflux:

This MR adds the established tag self_checkout to the shop preset. 

Self checkouts are helpful for many people when shopping. E.g. when not speaking the local language or having social anxiety.

The tag is well-defined, and the adoption is accelerating. Adding it to the tagging schema would help mappers when using EveryDoor and similar mobile mapping apps (related issue: https://github.com/Zverik/every_door/issues/774). 

Self checkout counters are already established in the US and growing rapidly in many countries like [Germany](https://www.self-checkout-initiative.de/markterhebung-2023/), and according to projections the number of them will grow rapidly in the coming years. 

### Links and data

Wiki: https://wiki.openstreetmap.org/wiki/Key:self_checkout

Tag info: https://taginfo.openstreetmap.org/keys/self_checkout 

tag info graph of self_checkout

### DIFF
--- a/data/fields/self_checkout.json
+++ b/data/fields/self_checkout.json
@@ -1,0 +1,12 @@
+{
+    "key": "self_checkout",
+    "type": "radio",
+    "strings": {
+        "options": {
+            "yes": "Yes",
+            "only": "Only",
+            "no": "No"
+        }
+    },
+    "label": "Self Checkout"
+}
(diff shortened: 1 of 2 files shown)


Dimitar5555:
The addition of `self_checkout` is definitely a good idea, but it shouldn't appear for every `shop`. Limiting it to `shop=supermarket` for now seems reasonable, given it's usage.

Data as of 2024-12-30:
image

matkoniecz:
I will close/reopen to trigger preview build

PR #1366, opened 2024-10-23
Use American English spelling for Amphitheater preset

arch0345:

Currently, the `theatre:type=amphi` preset is using the British spelling for its label (amphitheatre), however according to the [contributing documentation](https://github.com/openstreetmap/id-tagging-schema/blob/main/CONTRIBUTING.md), the JSON files in this repository should be in American English. This PR changes the label of this preset to "Amphitheater"

tldr 🇺🇸 🇺🇸 🇺🇸 🦅💥 🦅💥

### Links and data

> An amphitheatre ([U.S. English](https://en.wikipedia.org/wiki/American_English): amphitheater) is an open-air venue used for entertainment, performances, and sports.

https://en.wikipedia.org/wiki/Amphitheatre

### DIFF
--- a/data/presets/amenity/theatre/type/amphi.json
+++ b/data/presets/amenity/theatre/type/amphi.json
@@ -20,5 +20,5 @@
         "amenity": "theatre",
         "theatre:type": "amphi"
     },
-    "name": "Amphitheatre"
+    "name": "Amphitheater"
 }


tordans:
Could you / someone also have a look at the translation system if the GB translation is in place?

1ec5:
Yes, [the British English translation](https://app.transifex.com/openstreetmap/id-editor/translate/#en_GB/presets/153998337) is in place, though it’ll need to be reconfirmed anyways after Transifex syncs with this change. The string has not yet been translated into other regional English locales.

tordans (approved):

tordans:
Sounds good, merged and over to you for the Transifex update :).

1ec5:
I updated all the English localizations except English (Philippines). Not sure of the correct form there, so I’ll leave it to that localization team’s translators.

PR #1375, opened 2024-11-02
fix link

matkoniecz:
fix link in docs

note that even with this fix it is still dead link, there is no expected content there


tordans (approved):
Thanks for the fix

tordans:
Btw, this is the section that this links to … https://github.com/openstreetmap/id-tagging-schema/pull/1257/files

PR #1378, opened 2024-11-03
Docs: Change default label to `needs-triage`

tordans:
The `enhancement` label does really add any details to the issue. We have better labels like `new-preset` and `first-good-issue` and `considering`. Instead I am using it like a triage flag. To this PR makes this more explicit by renaming it to `needs-triage` which is something many GitHub projects use.

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug.md
+++ b/.github/ISSUE_TEMPLATE/1-bug.md
@@ -2,7 +2,7 @@
 name: Bug report
 about: Create a report to help us improve
 title: ''
-labels: bug
+labels: bug,needs-triage
 assignees: ''
 
 ---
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/1-bug.md
```diff
@@ -2,7 +2,7 @@
 name: Bug report
 about: Create a report to help us improve
 title: ''
-labels: bug
+labels: bug,needs-triage
```
k-yle, line 5: should we also remove https://github.com/openstreetmap/id-tagging-schema/labels/bug until someone has assesed if it's genuinely a bug?

k-yle (approved):
I think this is a useful change which the main iD repo would also benefit from

tyrasd (approved):
:+1:  Agree that this would be a much more useful label for new issues compared to the current "enhancement". I think we can leave the `bug` label for bug reports, as that makes it more visible that the reporter thinks this is a bug (if that's not the case, the label can be corrected during the triage phase) and might help prioritizing incoming issues.

PR #1383, opened 2024-11-05
Add strolling to sac_scale.json

felagund:
Following an approved proposal ( ​https://wiki.openstreetmap.org/wiki/Proposal:Add_strolling_to_sac_scale_and_some_further_refinements ), sac_scale got a new value: strolling. Adding it to the schema. it has not T1-T6 value.

### DIFF
--- a/data/fields/sac_scale.json
+++ b/data/fields/sac_scale.json
@@ -5,6 +5,7 @@
     "placeholder": "Mountain Hiking, Alpine Hiking...",
     "strings": {
         "options": {
+            "strolling": "Strolling",
             "hiking": "T1: Hiking",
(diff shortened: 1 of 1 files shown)


tordans:
I thinks this is good to merge.

There is a wikidata item for sac_scale https://wiki.openstreetmap.org/w/index.php?search=sac_scale&title=Special:Search&profile=advanced&fulltext=1&ns120=1&searchToken=a4s4iji2g2a7tfm7ey0mynax9 but none for the values. But we already have the custom text for the value, so that is something and the wiki page https://wiki.openstreetmap.org/wiki/Key:sac_scale has a good section on which to choose.

PR #1385, opened 2024-11-13
Cycle_barrier fields

Zverik:
Adds commonly used fields for [barrier=cycle_barrier](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dcycle_barrier). Adding after a request from the Danish Cyclists' Association, with the goal of having those in Every Door eventually.

Right now the `cycle_barrier` key has no extra fields except `access`.

**Relevant tag usage stats:**

* `barrier=cycle_barrier`: 132k
* `cycle_barrier=*`: 32.5k, of which 31.3k account for `double`, `single`, and `triple`
* `cycle_barrier:installation=*`: 22k
* `maxspeed:physical=*`: 10k
* `spacing`, `opening`, and `overlap`: 500-600 each.

### DIFF
--- a/data/fields/cycle_barrier.json
+++ b/data/fields/cycle_barrier.json
@@ -1,0 +1,14 @@
+{
+    "key": "cycle_barrier",
+    "type": "combo",
+    "label": "Barrier Type",
+    "strings": {
+        "options": {
+            "single": "Single",
+            "double": "Double",
+            "triple": "Triple",
+            "diagonal": "Diagonal",
+            "tilted": "Tilted"
+        }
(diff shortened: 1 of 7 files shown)


Zverik:
id_cycle_barrier_fields

Huh, this preview feature is fun!

Zverik:
@k-yle okaaaaay, you are correct, this is almost the full duplicate. Except this one does not require a schema change. Maybe we could do this in two steps, first using the 1-out-of-3 approach used here, and then changing to lists when the schema is changed?

PR #921, opened 2023-06-13
Add support for `building=house` + `house=*`

arch0345:
Adds:
* a field for [`house=*`](https://wiki.openstreetmap.org/wiki/Key:house) to the House preset
* unsearchable presets for
  * [`house=detached`](https://wiki.openstreetmap.org/wiki/Tag:house=detached)
  * [`house=semi-detached`](https://wiki.openstreetmap.org/wiki/Tag:house=semi-detached)
  * [`house=terrace`](https://wiki.openstreetmap.org/wiki/Tag:house=terrace)
* preset for [`house=terraced`](https://wiki.openstreetmap.org/wiki/Tag:house=terraced)

Also modified label for the `building=terrace` preset to disambiguate it with the `house=terraced` preset

PR #1412, opened 2024-12-21
Add strings for `house=` field and disable Taginfo suggestions

Dimitar5555:

Currently the `house` field doesn't have any strings and it shows any value with 100+ uses. This creates a lot of variants for tagging the same thing.

This PR introduces strings for the documented values of `house=` tag and disables Taginfo's suggestions for the field. 

P.S. It doesn't affect the tag editor, I'm not sure if we can influence it.

### Related issues

None

### Links and data

**Relevant OSM Wiki links:**
https://wiki.openstreetmap.org/wiki/Key:house

**Relevant tag usage stats:**
https://taginfo.openstreetmap.org/keys/house#values

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/data/fields/house.json
+++ b/data/fields/house.json
@@ -26,7 +26,7 @@
             },
             "terrace": {
                 "title": "Row of Terraced Houses",
-                "description": "A row of houses that are attached to each other in a line"
+                "description": "A row of houses that are attached to each other in a line (mapping each house as a separate area is preferred)"
             },
             "terraced": {
                 "title": "Single Terraced House",
--- a/data/fields/house.json
+++ b/data/fields/house.json
@@ -1,5 +1,38 @@
 {
     "key": "house",
     "type": "combo",
-    "label": "House Type"
+    "label": "House Type",
+    "strings": {
+        "options": {
+            "bungalow": {
+                "title": "Bungalow",
+                "description": "A small house with a single storey"
+            },
+            "detached": {
+                "title": "Detached House",
+                "description": "A house that is not connected to any other houses"
+            },
+            "link-detached": {
+                "title": "Link-Detached House",
+                "description": "A house that is connected to another house by a garage or other non-living area"
+            },
+            "maisonette": {
+                "title": "Maisonette",
+                "description": "A residential dwelling, split over two floors with private external entrances rather than a communal entrance"
+            },
+            "semi-detached": {
+                "title": "Semi-Detached House",
+                "description": "A house that is attached to another house on one side"
+            },
+            "terrace": {
+                "title": "Row of Terraced Houses",
+                "description": "A row of houses that are attached to each other in a line"
+            },
+            "terraced": {
+                "title": "Terraced House",
+                "description": "Individual house in a row of terraced houses"
+            }
+        }
+    },
+    "autoSuggestions": false
 }
--- a/data/fields/house.json
+++ b/data/fields/house.json
@@ -25,12 +25,12 @@
                 "description": "A house that is attached to another house on one side"
             },
             "terrace": {
-                "title": "Row of Terraced Houses",
+                "title": "Row of Townhouses",
                 "description": "A row of houses that are attached to each other in a line (mapping each house as a separate area is preferred)"
             },
             "terraced": {
-                "title": "Single Terraced House",
-                "description": "Individual house in a row of terraced houses"
+                "title": "Single Townhouse",
+                "description": "Individual house in a row of townhouses"
             }
         }
     },
--- a/data/fields/house.json
+++ b/data/fields/house.json
@@ -29,7 +29,7 @@
                 "description": "A row of houses that are attached to each other in a line"
             },
             "terraced": {
-                "title": "Terraced House",
+                "title": "Single Terraced House",
                 "description": "Individual house in a row of terraced houses"
             }
         }
--- a/data/presets/building/house.json
+++ b/data/presets/building/house.json
@@ -4,8 +4,8 @@
         "area"
     ],
     "fields": [
-        "{building}",
-        "house"
+        "house",
+        "{building}"
     ],
     "tags": {
         "building": "house"
--- a/data/presets/building/house/_detached.json
+++ b/data/presets/building/house/_detached.json
@@ -3,6 +3,10 @@
     "geometry": [
         "area"
     ],
+    "fields": [
+        "house",
+        "{building}"
+    ],
     "tags": {
         "building": "house",
         "house": "detached"
--- a/data/presets/building/house/_semi-detached.json
+++ b/data/presets/building/house/_semi-detached.json
@@ -3,6 +3,10 @@
     "geometry": [
         "area"
     ],
+    "fields": [
+        "house",
+        "{building}"
+    ],
     "tags": {
         "building": "house",
         "house": "semi-detached"
--- a/data/presets/building/house/_terrace.json
+++ b/data/presets/building/house/_terrace.json
@@ -3,6 +3,10 @@
     "geometry": [
         "area"
     ],
+    "fields": [
+        "house",
+        "{building}"
+    ],
     "tags": {
         "building": "house",
         "house": "terrace"
--- a/data/presets/building/house/terraced.json
+++ b/data/presets/building/house/terraced.json
@@ -3,6 +3,10 @@
     "geometry": [
         "area"
     ],
+    "fields": [
+        "house",
+        "{building}"
+    ],
     "tags": {
         "building": "house",
         "house": "terraced"


data/fields/house.json
```diff
@@ -1,5 +1,16 @@
 {
     "key": "house",
     "type": "combo",
-    "label": "House Type"
+    "label": "House Type",
+    "strings": {
+        "options": {
+            "bungalow": "Bungalow",
+            "detached": "Detached House",
+            "link-detached": "Link-Detached House",
```
Dimitar5555, line 9: `house=link-detached` has only ~750 uses and less than 20 in the `building` key. There isn't a preset for it in iD. It is documented though, so it might be a good idea to create a preset.

Dimitar5555:
(The screenshot tool doesn't want to capture the dropdown menu)

tordans:
I looked into houses in https://github.com/openstreetmap/id-tagging-schema/issues/1155 when I tried to understand the different wording in SC vs iD vs Wiki but never finished it.

tordans:
@Dimitar5555 I think this is a good addition. But we need to make sure we are not missing something relevant and make the situation worse for some cases.

Could you double check that we have all values in this list that StreetComplete uses?

And we should also check document the taginfo most used values so we support any that are used often, like maybe 5k+?

tordans:
Btw, a great follow up to this would be to create small icons for those values like we have for crossings. Those are supported via `icons` https://github.com/ideditor/schema-builder?tab=readme-ov-file#icons

---

And once https://github.com/openstreetmap/iD/issues/10613 is done and we have a better display of the value `description`, we should IMO add a description for those as well.

Dimitar5555:
> Could you double check that we have all values in this list that StreetComplete uses?

Note that SC is using the `building=house/detached/semi-detached/...` while this field is shown on elements with `building=house + house=(empty value)`.

SC seems to be missing an answer for terraced houses.

https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/osm/building/BuildingType.kt#L9-L13

SC is treating `building=terraced` as deprecated. The documented replacement in SC is `terrace` but this is not always correct. `terraced` is for the individual houses while `terrace` is for the whole row of houses.

https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/osm/building/BuildingType.kt#L172-L174

P.S. Whether SC should be using `building=*_house` or `building=house + house=*` is a matter of discussion there. 

---

> And we should also check document the taginfo most used values so we support any that are used often, like maybe 5k+?

:white_check_mark:  

image

---

> Btw, a great follow up to this would be to create small icons for those values like we have for crossings. Those are supported via `icons` [ideditor/schema-builder#icons](https://github.com/ideditor/schema-builder?tab=readme-ov-file#icons)

It is a good suggestion in principle but I'm not sure how they could look. `detached` is easy but `semi-detached` and `terraced` could be represented with the same icon (which might be confusing). `terraced` could have 3 houses next to each other but we will have to change the icon for `terrace` as well.

---

> And once [openstreetmap/iD#10613](https://github.com/openstreetmap/iD/issues/10613) is done and we have a better display of the value `description`, we should IMO add a description for those as well.

I will add it in the coming days (or maybe next year :smile:).

tordans:
> Note that SC is using the `building=house/detached/semi-detached/...` while this field is shown on elements with `building=house + house=(empty value)`.

Ah, sorry, I did not fully realize, what we are actually looking at here …

So this list is only used for the `building=house` preset at https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/building/house.json#L8

(Side note: There is no other usage of this field https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema+%5C%22house%5C%22+path%3A%2F%5Edata%5C%2Fpresets%5C%2F%2F&type=code )

FYI: We have three unsearchable presets for specific values of this list in https://github.com/openstreetmap/id-tagging-schema/tree/main/data/presets/building/house which where all added in https://github.com/openstreetmap/id-tagging-schema/pull/921  – those unsearchable presets don't have "`fields`" themselves so they will reuse the parent `fields` which in will include the `house=*` field that we are editing here. This is important because it will allow to change the type of house even after one of those three was selected back to something else.

TBH I am not sure if we actually want those separate, unsearchable presets. The only benefit I can see is that https://github.com/openstreetmap/id-tagging-schema/pull/921/files#diff-cb5181126156c42a37b82bff9e25d5bf13d6c2b4331b6980a24262ceaf8b2c06R2 has a different icon now.




---

> P.S. Whether SC should be using `building=*_house` or `building=house + house=*` is a matter of discussion there.

The question for this repo that I see from this is: How can we make those sub-types discoverable when searching or browsing for houses in iD. Right now only https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/building/detached.json will be findable but https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/building/house/_detached.json will be hidden from the search and browser section.

Again, this is something we don't have to solve now, but we might want to create a follow up issue for this to see if there is a solution for it. Right now it looks like conflicting tagging practices that we cannot really resolve(?)

tordans:
## Test-Documentation

Find test cases: https://overpass-turbo.eu/s/1Wan

- 🟢  https://pr-1412--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w1271824908&locale=en&map=20.00/52.20455/13.58388
  - 🟡 one thing we should improve here is to move the "type of house" to the top of the list by changing the order of https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/building/house.json#L7-L8
   > image

- 🔴 https://pr-1412--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w1219479104&locale=en&map=20.00/52.30051/13.78133
   - I consider this a blocking issue. Right now those unsearchable sub presets remove the "type of house" field from the list of fields. I remember from crossing refactoring that this is some kind auf automatism that I don't think is smart. It means that once I select a "type of house" the UI changes and the only way back is to cmd+z or to start again or edit the raw tags. Instead the "type of house" should always be present at the top to allow to change the preset easily.

- same issue https://pr-1412--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w1324673770&locale=en&map=20.00/52.30858/13.50386
- same issue https://pr-1412--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w1328793108&locale=en&map=20.00/52.40519/13.53411

### Search

Just to document the "two tagging standards" issue described above: This find the `building=*` but now the `house=*` [that as mapped (manually)](https://pr-1412--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w1328793108&locale=en&map=20.00/52.40519/13.53411)
> image



### Info-`i`



### Wording

- [ ] American English

Dimitar5555:
> * 🟡 one thing we should improve here is to move the "type of house" to the top of the list by changing the order of [`main`/data/presets/building/house.json#L7-L8](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/building/house.json#L7-L8)

Done.

> 🔴 [pr-1412--ideditor-presets-preview.netlify.app/id/dist#background=Bing&disable_features=boundaries&id=w1219479104&locale=en&map=20.00/52.30051/13.78133](https://pr-1412--ideditor-presets-preview.netlify.app/id/dist/#background=Bing&disable_features=boundaries&id=w1219479104&locale=en&map=20.00/52.30051/13.78133)
> 
>     * I consider this a blocking issue. Right now those unsearchable sub presets remove the "type of house" field from the list of fields. I remember from crossing refactoring that this is some kind auf automatism that I don't think is smart. It means that once I select a "type of house" the UI changes and the only way back is to cmd+z or to start again or edit the raw tags. Instead the "type of house" should always be present at the top to allow to change the preset easily.

Fixed.
image

> 🟡 this line [#1412 (files)](https://github.com/openstreetmap/id-tagging-schema/pull/1412/files#diff-5fbab217bab0453164ec163e40f00955f65ebb1900518439ae0066393f5c2bceR12) is written different than the others. Should that be "Row of Terraced Houses"?

Updated accordingly.

---

> `residential` => MISSING

I skipped this value on purpose. It isn't clear to me what it means. Is it `building=residential`, `building:use=residential` or something else? It is also undocumented despite the large usage.

Dimitar5555:
@tordans I believe that all of your comments have been addressed, apart from adding description texts.

> And once [openstreetmap/iD#10613](https://github.com/openstreetmap/iD/issues/10613) is done and we have a better display of the value `description`, we should IMO add a description for those as well.

matkoniecz:
> Closes-and-reopen to retrigger the preview.

no idea why it has not happened

will try again

matkoniecz:
tried #1694 as workaround

but now I see that https://github.com/openstreetmap/id-tagging-schema/actions/runs/17338221058/job/49227935375?pr=1412 also has woken up

matkoniecz:
from testing it seems to work as expected codewise and in my testing, seems to handle https://github.com/openstreetmap/id-tagging-schema/issues/1678 mess well as nonsense `house=residential` suggestion is now gone

(for descriptions: I am curious how case where house was built as detached but adjoined house was constructed is supposed to be tagged)

data/fields/house.json
```diff
@@ -1,5 +1,38 @@
...
+            "semi-detached": {
+                "title": "Semi-Detached House",
+                "description": "A house that is attached to another house on one side"
+            },
+            "terrace": {
+                "title": "Row of Terraced Houses",
+                "description": "A row of houses that are attached to each other in a line"
+            },
+            "terraced": {
+                "title": "Terraced House",
```
matkoniecz:
```suggestion
                "title": "Single Terraced House",
```

maybe that would be more clear that it applies to individual ones, not big blob?

data/fields/house.json
```diff
@@ -1,5 +1,38 @@
...
+                "title": "Maisonette",
+                "description": "A residential dwelling, split over two floors with private external entrances rather than a communal entrance"
+            },
+            "semi-detached": {
+                "title": "Semi-Detached House",
+                "description": "A house that is attached to another house on one side"
+            },
+            "terrace": {
+                "title": "Row of Terraced Houses",
+                "description": "A row of houses that are attached to each other in a line"
```
matkoniecz:
```suggestion
                "description": "A row of houses that are attached to each other in a line (mapping each house as a separate area is preferred)"
```

maybe this hint would be useful?

Dimitar5555: It seems fine to me. We are giving a lot of mapping/tagging advice for various fields already.

matkoniecz: If you think it is improvement - can you apply this suggestion?

Or are you not against but also not supporting it?

data/fields/house.json
k-yle: Should we also change the preset name to match the name here?

matkoniecz: I would support that.

Dimitar5555: The current names for the presets are "Townhouse" / "Row of Townhouses" which as far as i can tell is American English. "Single Terraced House" / "Terrace House" are likely the British terms which I've used in the fields. Maybe we should swap them since iD uses American English usually?

matkoniecz: And UK terms can go to UK dialect translation.

Dimitar5555: Updated accordingly.

tordans (approved):

tordans:
Given the long discussion and wait for this and good feedback I think we can merge this now. Thanks for the help!

PR #1422, opened 2025-01-09
add building=annexe

andrewharvey:

`building=annexe` is currently in use and documented on the wiki.

Ideally we would have a tagging preset for this tag to make it easier for users to add, and to translate into local terms.

### Links and data

**Relevant OSM Wiki links:** https://wiki.openstreetmap.org/wiki/Tag:building=annexe

**Relevant tag usage stats:** https://taginfo.openstreetmap.org/tags/building%3Dannexe

### DIFF
--- a/data/presets/building/annexe.json
+++ b/data/presets/building/annexe.json
@@ -10,7 +10,8 @@
         "annexe",
         "adu",
         "granny",
-        "in-law unit"
+        "in-law unit",
+        "secondary suite"
     ],
     "matchScore": 0.5,
     "name": "Accessory Dwelling Unit"
--- a/data/presets/building/annexe.json
+++ b/data/presets/building/annexe.json
@@ -1,0 +1,16 @@
+{
+    "icon": "maki-home",
+    "geometry": [
+        "area"
+    ],
+    "tags": {
+        "building": "annexe"
+    },
+    "terms": [
+        "annexe",
+        "adu",
+        "granny"
+    ],
+    "matchScore": 0.5,
+    "name": "Annexe House"
+}
--- a/data/presets/building/annexe.json
+++ b/data/presets/building/annexe.json
@@ -9,8 +9,9 @@
     "terms": [
         "annexe",
         "adu",
-        "granny"
+        "granny",
+        "in-law unit"
     ],
     "matchScore": 0.5,
-    "name": "Annexe House"
+    "name": "Accessory Dwelling Unit"
 }


andrewharvey:
> 🍱 You can preview the tagging presets of this pull request [here](https://pr-1422--ideditor-presets-preview.netlify.app/id/dist/#locale=en).

Looks good!

data/presets/building/annexe.json
```diff
@@ -0,0 +1,16 @@
...
+    "icon": "maki-home",
+    "geometry": [
+        "area"
+    ],
+    "tags": {
+        "building": "annexe"
+    },
+    "terms": [
+        "annexe",
+        "adu",
```
1ec5, line 11: Other terms include “in-law unit” and “[secondary suite](https://en.wikipedia.org/wiki/Secondary_suite)”.

data/presets/building/annexe.json
```diff
@@ -0,0 +1,16 @@
...
+        "granny"
+    ],
+    "matchScore": 0.5,
+    "name": "Annexe House"
```
1ec5, line 15: The `name` needs to be in American English, but “annexe” is completely unknown to American English speakers. In California, the formal term is “Accessory Dwelling Unit” (hence ADU), but this might be unknown in other parts of the state, compared to something more crass like “in-law unit” or “granny suite”. I suggest asking folks in the #language channel of OSMUS Slack for second opinions.

1ec5:
By the way, thanks for posting this – it’s a frequently asked tagging question in Slack, but I can never remember the tag.

data/presets/building/annexe.json
andrewharvey: thanks, added.
Ideally these terms would be translatable too.

data/presets/building/annexe.json
andrewharvey: I'll use "Accessory Dwelling Unit" for now. Translators can choose the best term for their region.

tordans:
As a review note: Did someone check the Info-i (Wikidata item)? https://github.com/openstreetmap/id-tagging-schema/blob/main/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L51-L62

andrewharvey:
>> The best term depends on the type of ADU:
> For a detached ADU, Nebenwohnung fits better.
> For an attached or basement-style ADU, Einliegerwohnung is the most widely recognized and appropriate term in German-speaking regions.

Interesting. In Australian English, an isolated living quarters within a detached house would be "self contained accommodation" ie. a basement, lower lever or upper level which you access from a separate entry door to the main living quarters. Sometimes these are converted or added after to rent out while you're still living in the main residence. Typically this wouldn't be mapped at all in OSM though as it's more to do with an internal residential building configuration.

The "detached ADU", in Australian English is "Granny Flat" and it's a detached building not connected or part of the main residence, but on the same property/parcel/lot. It's distinguished from the main building by being smaller/secondary. You could map these as building=detached, but they are a bit different and it's good to capture this via the building=annexe tag.

> Are we sure the wiki is correct in associating building=annexe so strongly with ADUs, and that usage is consistent?

Yeah, maybe not. It may be best to hold off on this PR then until the tagging is clarified.

From Google "an annex is a self-contained living space", so it sounds like it might cover anything from self contained living inside an existing building (eg. the lower/upper floor or basement, or even just a room blocked off from inside access and with it's own dedicated external door), or a building added on adjacent to the existing building with a shared wall, or a completely detached building.

Perhaps then we need something like `building:part=annexe` for an addon to a building to create a self contained space and  `building=detached_annexe` where it's completely detached and for residential use, eg. ADU/Granny Flat.

matkoniecz:
I am confused why `building=detached` is not preferred tagging here

https://wiki.openstreetmap.org/wiki/Tag:building=annexe has

> Distinction from building=detached
> A detached house typically has an official address where the annexe might not.

so if legal change happens (it gets address assigned by legal authority) its `building` type changes?

andrewharvey:
> I am confused why building=detached is not preferred tagging here, possibly with extra tags

`detached=main`, `detached=secondary`. At first that sounds okay, however I don't think it's ideal, as then it's ambiguous if `building=detached` is main or secondary, and therefore you would expect every `building=detached` to also tag a `detached=*` value, given the vast majority would be `detached=main` this would be undesirable. Further based on my examples below, a secondary dwelling may be detached from the main building or attached to it or within in, so it's not correct to say they are all detached.

----

Stepping back I tried to look at this again from a fresh perspective.

In Australia we have a concept of a "granny flat", however in official documents this is also referred as a "secondary dwelling", which from everything I've read is the same concept as the ADU/Accessory Dwelling Unit term used in the US. This also matches the current OSM Wiki documentation of `building=annexe` "A small self contained apartment on the same property as the main residential unit."

A granny flat is a self-contained secondary dwelling that can be attached to, within, or separate from the main house. 

Breaking that down, you have the the functional requirements:

- **self-contained** ie. a living space that has it's own facilities like a toilet/shower/kitchen that aren't shared facilities.
- **secondary** not the main dwelling on the property (the main property would be called the Freestanding House), and by property I mean the land parcel/cadastral parcel/lot/title.

and the physical properties:

- attached to the main house,
- within the main house, or
- separate from the main house

attached to or within may be hard to tell, for example a basement or attic secondary dwelling would be considered within, but a building build on the side of the house, sharing a wall would be attached, but other times it can be hard to decide.

Example of separate from the main house:

image

[Example of either attached to or within the main house](https://www.domain.com.au/12-campbell-avenue-normanhurst-nsw-2076-2019759123):

image

Tagging those within or attached to the main house is difficult and it's hard to survey, so I'm mostly just thinking about the "separate from the main house" scenario here. That said `building:part=` as I suggested previously could work in this example.

andrewharvey:
We also have https://wiki.openstreetmap.org/wiki/Key:house which applies to `building=house`. Under that tagging scheme I would consider then `building=annexe` (secondary house) + `annexe=detached` (for a free standing annexe/secondary dwelling/adu/granny flat) to be appropriate.

jarek:
Cross-posting from [the wiki talk page](https://wiki.openstreetmap.org/wiki/Talk:Tag:building%3Dannexe):

Thank you for looking for tagging for ADUs. These have been popping up in my area and I wondered how to tag it; I ended up using the local term (building=house + house=laneway_house) but my usages are so few that this would be easily retagged to an accepted worldwide tag.

I share concerns raised by Kovposch on the wiki talk page:

*    Per history chart, most usages of this tag are from 2017 when there was a major import/retagging, yet the wiki was only created in 2024
*    There are examples of usage that do not correspond to how it was defined in wiki in 2024, including https://www.openstreetmap.org/way/435684867/history apparently tagged as part of the 2017 retagging - it is very clearly not an ADU in American/Australian/Canadian sense
*    As another example, https://www.openstreetmap.org/way/660755738/history was retagged to annexe in 2021, but its previous tag indicated it is a service building, not residential (the name "budynek gospodarczy" includes something like a shed, storage, workshop or similar)

Personally I am in agreement with [matkoniecz above](https://github.com/openstreetmap/id-tagging-schema/pull/1422#issuecomment-2626878270) that tagging by purpose should be adequate, so building=detached or building=house, and then it could be narrowed down with house=* if desired. We could probably do worse than house=adu or house=accessory_dwelling or house=secondary_dwelling or something or maybe even house=annexe_dwelling if you want to go British (*are* there ADUs in Britain?). But in the end sometimes it might not be possible to tell if a residential building is an ADU or not, so tagging that does not depend on being able to make this distinction while allowing the distinction where possible would be good -- building=* + house=* is one way to do that.

By the way, some ADUs in Vancouver do have house numbers, [n8400781795](https://osm.org/node/8400781795) for example. (House numbers in Vancouver usually aren't consecutive, so the missing numbers can be assigned to newly constructed laneway houses. I do not know if they get direct mail delivery to the laneway house - I would guess not.)

A comment above mentioned the issue of "self-contained" residential units "within the main house" but I believe these are outside the scope of OSM since they're within one building and it's not clear it's possible to reliably distinguish them even from street-side survey. If anything, a tag like [building:flats](https://wiki.openstreetmap.org/wiki/Key:building:flats)=* with the number of units in the building appears more easily verifiable (government records, or visible number of mailboxes, or number of electric meters or gas meters in regions where those are usually visible from the street).

May I also suggest that going through with a proposal or a discussion on the forum may have helped to uncover these issues _before_ making a request to add a preset to iD?

andrewharvey:
> May I also suggest that going through with a proposal or a discussion on the forum may have helped to uncover these issues before making a request to add a preset to iD?

I agree, at the time I made this PR I thought it was pretty clear cut that this tag was for ADUs, but it's clear now from the discussion there are issues. A proposal or forum discussion is a good idea first. I'd he happy to contribute or help out.

> Personally I am in agreement with https://github.com/openstreetmap/id-tagging-schema/pull/1422#issuecomment-2626878270 that tagging by purpose should be adequate, so building=detached or building=house, and then it could be narrowed down with house=* if desired. We could probably do worse than house=adu or house=accessory_dwelling or house=secondary_dwelling or something or maybe even house=annexe_dwelling if you want to go British (are there ADUs in Britain?).

Under the `building=house` scheme ADU's would meet the definition of `house=detached` being freestanding residential buildings usualyl housing a single family. You could then do `detached=secondary` but as I mentioned this then means you need a corresponding `detached=primary` so you can determine if it's primary/secondary or just unknown. This is also important for apps like StreetComplete which need a tag for each answer. I'm hesitant to go with this since there's still not that many ADUs compared to primary residences and do we really want to have to use 3 tags just to tag a bog standard house?

Therefore I would suggest if using the `building=house` schema then go with `house=auxiliary` or something like that, or if not using the `building=house` schema then just `building=auxiliary`, or similar.

> I ended up using the local term (building=house + house=laneway_house) but my usages are so few that this would be easily retagged to an accepted worldwide tag.

To be honest I'm not fussed, every region has their own name "annex, laneway house, ADU, granny flat, secondary dwelling..." each region can translate the preset however they like.

So long as "laneway" is not how the tag is defined since not all ADUs will have laneway access, in Australia some do but most don't. It would be nice to tag which street the building is access by, but that applies to all residential buildings and not just limited to ADUs. We have `addr:street` but sometimes the postal address street is not where the building is accessed by. We have `service=driveway` but not all properties will have vehicle access, tagging `highway=footway` leading to the front door could work, but seems perhaps a tad too invasive. A separate issue to this discussion.

andrewharvey:
I'll close this PR, it does seem the tagging issue is not resolved yet. After wider tagging discussion we can try again to add a preset.

Issue #362, opened 2022-01-22
sync NSI tags


tyrasd:
The following tags are used in the [NSI](https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/amenity/parcel_locker.json), but don't exist here. For best user experience, we should consider adding these as presets, so that iD can show the best description for the "brands" in NSI using those tags.

```
Warning - no iD source preset found for brands/amenity/training
Warning - no iD source preset found for brands/club/scout
Warning - no iD source preset found for brands/office/bail_bond_agent
Warning - no iD source preset found for brands/office/union
Warning - no iD source preset found for brands/shop/nuts
Warning - no iD source preset found for operators/boundary/aboriginal_lands
Warning - no iD source preset found for operators/boundary/national_park
Warning - no iD source preset found for operators/boundary/protected_area
```

matkoniecz:
> Warning - no iD source preset found for

Where such logs can be found?

bhousel:
> Where such logs can be found?

This is just output when someone runs `npm run dist` to regenerate the presets.

As of today the current list looks like this:

```
🏗   Building data...
Warning - no iD source preset found for brands/amenity/food_sharing
Warning - no iD source preset found for brands/amenity/training
Warning - no iD source preset found for brands/club/scout
Warning - no iD source preset found for brands/office/bail_bond_agent
Warning - no iD source preset found for brands/shop/gold_buyer
Warning - no iD source preset found for brands/shop/plant_hire
Warning - no iD source preset found for brands/shop/truck_repair
Warning - no iD source preset found for operators/boundary/aboriginal_lands
Warning - no iD source preset found for operators/boundary/national_park
Warning - no iD source preset found for operators/boundary/protected_area
Warning - no iD source preset found for operators/office/visa
Warning - no iD source preset found for operators/route/tracks
```

matkoniecz:
https://wiki.openstreetmap.org/wiki/Tag:shop%3Dtruck_repair - seems OK?

https://wiki.openstreetmap.org/wiki/Tag:shop%3Dplant_hire - oh, a bit confusing name at least for me. I expected place mass-hiring people to work at factory. Though it may be just me failing at English.

Carnildo:
"Plant hire" is the British term for it.  In American English, it would be "equipment rental", and I'm sure that Indian English has a third, totally different, name for it.

matkoniecz:
[...]

PR #1430, opened 2025-01-30
add shop-bbq preset

andrewharvey:

Add a preset for [`shop=bbq`](https://wiki.openstreetmap.org/wiki/Tag:shop=bbq) a shop selling bbq's and bbq equipment.

**Relevant tag usage stats:**

currently limited use

https://pr-1430--ideditor-presets-preview.netlify.app/id/dist/#background=NSW_LPI_Imagery&disable_features=boundaries&id=n12543268487&locale=en&map=20.96/-33.79695/151.18588

image

### Search

image

image

image

image

### Info-`i`

image

### Wording

- [x] American English

unsure, would you call this a "Grill Store"?

We already have "Barbecue Restaurant" and "Barbecue/Grill" so based on that I've used "Barbecue Store" but happy to take input for Americans.

Apparently "Barbecue" is mostly used globally and "Barbeque" in AU/NZ so I've gone with "Barbecue" for the US name and listed both as terms. Then translators can take it from there.

### DIFF
--- a/data/presets/shop/bbq.json
+++ b/data/presets/shop/bbq.json
@@ -11,6 +11,7 @@
         "area"
     ],
     "terms": [
+        "barbecue",
         "barbeque",
         "bbq",
         "grill"
@@ -18,5 +19,5 @@
     "tags": {
         "shop": "bbq"
     },
-    "name": "BBQ Shop"
+    "name": "Barbecue Shop"
 }
--- a/data/presets/shop/bbq.json
+++ b/data/presets/shop/bbq.json
@@ -1,0 +1,22 @@
+{
+    "icon": "maki-bbq",
+    "fields": [
+        "{shop}"
+    ],
+    "moreFields": [
+        "{shop}"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "bbq",
+        "barbeque",
+        "grill"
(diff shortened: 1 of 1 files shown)


tyrasd:
> currently limited use

75 is actually really low, well below the "threshold" of the other shop presets (IIRC, the shop presets with the fewest occurrences are around the 500 mark).  :thinking:

tordans:
One thing that would speak to adding it is that it is use in NSI https://github.com/osmlab/name-suggestion-index/pull/10416/files
I kind of think we should try to add all the presents that NSI references (which is #362) … but then again that is not really a reason since users will never see the preset given that NSI will overwrite most occurrences(?)

---

We could put this on hold and look at the numbers again in a year…

tyrasd:
Good point with keeping in sync with NSI.

Also, I took a second look at the OSM data, and it seems like many BBQ/Grill stores seem to be currently wrongly tagged, for example as [`shop=outdoor`](https://www.openstreetmap.org/node/2684745796) or [`shop=kitchen`](https://www.openstreetmap.org/node/3729208281), [`shop=yes`](https://www.openstreetmap.org/node/7523744198), etc. – yet another argument to make an exception for the inclusion of this preset

tyrasd (approved):

PR #1462, opened 2025-02-25
Use couch icon for upholsterer

westnordost:
This icon is also used for furniture shops. It makes sense to use it for upholsterers, too, because they are in the furniture business, too. Particularly, couches etc.

Currently, a hammer and a wrench is used, which doesn't really fit that much and that icon is already used for many other presets.

Mitigates #41

### DIFF
--- a/data/presets/craft/upholsterer.json
+++ b/data/presets/craft/upholsterer.json
@@ -1,5 +1,5 @@
 {
-    "icon": "temaki-tools",
+    "icon": "fas-couch",
     "geometry": [
         "point",
         "area"


1ec5:
Would a vacuum be more intuitive, similar to the existing icon for `craft=cleaning`? I’ve mapped some upholsterers that specialize in car interiors or even the interiors of private airplanes. Granted, couches can occur in some private airplanes…

westnordost:
Vacuum? Like, vacuum cleaner? Why? Upholsterers don't clean, they craft (seats etc)

1ec5:
Hm, yeah, I was thinking of reupholstering as a kind of refreshing, but maybe that’s too much of a stretch.

tordans:
Testing ==> Looks good

> image

> image

> image



https://pr-1462--ideditor-presets-preview.netlify.app/id/dist/#locale=en&disable_features=boundaries&map=20.00/-24.31848/29.33802&background=Bing&id=n5901713076

tordans (approved):

PR #1472, opened 2025-03-04
Support education tag per approved proposal

quincylvania:
This PR models the `education` key like `healthcare` per the approved tagging proposal: https://wiki.openstreetmap.org/wiki/Proposal:Education_tag

No new presets are added. Existing education presets are moved under the `education` directory for convenience.


data/presets/education.json
```diff
@@ -0,0 +1,39 @@
...
+        "opening_hours",
+        "payment_multi",
+        "wheelchair"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "education": "*"
```
tordans, line 29: I don't think this `*` works. Looking at https://pr-1472--ideditor-presets-preview.netlify.app/id/dist/#locale=en&disable_features=boundaries&map=20.00/51.43931/-0.10160&background=Bing&id=w922125174 
- remove the `amenity` tag => this becomes a building preset
- change `education=school` to `education=*` => only now it becomes this "Education Facility" tag

tordans: Ah, I see, this is not supposed to work, but meant to be a template, right? So all fields, moreFields are specified here and the other inherit it from here.

The new approach we are using for this are templates. There are more docs on this at https://github.com/openstreetmap/id-tagging-schema/tree/main/data/presets/%40templates

🟡 Personally I would prefer moving this into a template because it clarifies what it is and does not cause the confusion I had earlier.

data/fields/education.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "education",
+    "type": "typeCombo",
+    "label": "Type"
+}
```
tordans, line 18: 🟡 Lets add named keys for those values that have their own preset.
It will give the dropdown a push to signal which values are official.
We should keep the full value list, though, I think.
image

tordans:
## Test-Documentation for `"education": "*"`
- Code https://github.com/openstreetmap/id-tagging-schema/pull/1472/files#diff-fdef83a194b63f4e41f1b0e4ef4c39ce53e2dcf2c90e75d72a6538b46fd85f28
- Cases https://overpass-turbo.eu/s/27um

Result: 🟡  see inline comment. I assume this is a template and I think we should model the files differently for clarity.

## Test-Documentation for `"education": "yes"`
- Code https://github.com/openstreetmap/id-tagging-schema/pull/1472/files#diff-9afc952db865ad02c02aacfc4cbe4659de746a7b553ca12e8c427b22ce67979c
- Cases https://overpass-turbo.eu/s/27ur

There is a lot of noise for education=yes with information board, `route=foot` and similar.

Found one that fits the bill https://pr-1472--ideditor-presets-preview.netlify.app/id/dist/#locale=en&disable_features=boundaries&map=20.00/51.48661/12.05291&background=Bing&id=n1930139866

Result: 
- 🟢 preset
- 🔴 info-i: https://wiki.openstreetmap.org/wiki/Item:Q3235 is not a good description after the proposal was approved
   > image
- 🟡  wiki: lets add the yes case to https://wiki.openstreetmap.org/wiki/Key:education


## Test-Documentation for `"amenity": "college"`
- Code https://github.com/openstreetmap/id-tagging-schema/pull/1472/files#diff-5f53514c73c6665efb1496e4eb4991690fe08b4d332782874e4512662839eb48
- Cases https://overpass-turbo.eu/s/27ut

Test case https://pr-1472--ideditor-presets-preview.netlify.app/id/dist/#locale=en&disable_features=boundaries&map=20.00/48.28871/6.96602&background=Bing&id=w44116127

Result:
- 🟢  the preset looks good (ignoring the broken `building=yes`)
- 🟢  the tag updating looks good
   > image
- 🟢  info-i looks good

We do have an issue with the `building=yes` which does not trigger the right preset https://github.com/openstreetmap/id-tagging-schema/blame/main/data/presets/building/college.json#L16 which requires `building": "college`. I will comment on that separately.


----

Follow up issues:

- It would be great if we move from the wiki data items and wiki pages from amenity to education. This requires to set up those wiki pages, first. Maybe moving the content from amenity to education and having amenity redirect to education? Maybe in about a year…

- Do we want to handle the case where only `education` is tagged without `amenity`? If so, we would need some deprecation rules or separate, unsearchable presets to handle this.

tordans:
🟡  One more thing to clarify:
What about the "College Building" preset? Right now that does not have the `education` tag. Should those building presets get the tag as well? What did the proposal say about this?
https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/building/college.json#L13-L15

data/fields/education.json
quincylvania: Good call, I've added the value labels. I agree we should keep custom values.

data/presets/education.json
quincylvania: Oh I see. This is supposed to be a generic preset like for `healthcare=*` (I just copied that file). The problem was the `matchScore` was lower than that of the building preset. I updated it.

quincylvania:
> What about the "College Building" preset? Right now that does not have the education tag. Should those building presets get the tag as well? What did the proposal say about this?


There is no direct connection between the two tags. It's the same as how `building=retail` is tagged separately from `shop=`.

tordans:
> > What about the "College Building" preset? Right now that does not have the education tag. Should those building presets get the tag as well? What did the proposal say about this?
> 
> There is no direct connection between the two tags. It's the same as how `building=retail` is tagged separately from `shop=`.

I see. Just to understand this better: My understanding of the proposal is – in the spirit of https://openstreetmap.us/our-work/layercake/ - to be able to select all education data via `education`. (Right?) As such, why would the building be excluded from that? Especially in cases when it also has the `amenity=college` tag as well as the `building=college` tag – see https://github.com/openstreetmap/id-tagging-schema/issues/1632 for more on that.
I did not check this in the data, but I assumed that the school area is a second level mapping step where the school building (or node) is likely the first level. So we should be missing out on data, if we don't add the tag here as well…

quincylvania:
> My understanding of the proposal is – in the spirit of https://openstreetmap.us/our-work/layercake/ - to be able to select all education data via education

Almost but not quite. `education` is not a categorical tag intended to group together everything education-related. It's a general tag for a discrete educational institution (i.e. school), where individual values are specific types of schools. This is like how `shop` is a general tag for retail stores but does apply to other features related to the store.

A single school may have multiple buildings, but should only have one `education=school` feature (probably a node at the center of an area surrounding all the buildings). Sometimes in OSM a tag like `amenity=college` is tagged on a `building=college`, and in this case `education=college` should be added, but only because of the `amenity` tag.

The proposal has this line:

> E.g. tagging [education](https://wiki.openstreetmap.org/wiki/Key:education)=[yes](https://wiki.openstreetmap.org/w/index.php?title=Tag:education%3Dyes&action=edit&redlink=1) on a school parking lot or [education](https://wiki.openstreetmap.org/wiki/Key:education)=playground on a school playground is discouraged.

Tagging every building with `education` would be like tagging every parking lot or every playground.

data/presets/education.json
tordans: Thanks, I did not have this patter in mind. Resolved now.

data/fields/education.json
tordans: This works now, thanks 
image

tordans (approved):

tordans:
@quincylvania I am going to merge this now. The technical feedback has been adressed.

Could you have a look at the wiki part that I mentioned in https://github.com/openstreetmap/id-tagging-schema/pull/1472#issuecomment-3051220863 I find those PRs are a great nudge to improve the wiki docs and they are a core part of the preset-experience, so having them in a good shape is usually something that I think we should do before mergin. Just don't want to postpone this further just because…

On the building topic: Thanks for the reply. Based on this there is nothing to do ATM. Will answer separately.

tordans:
@quincylvania I tried looking into the building case more. You write
> A single school may have multiple buildings, but should only have one education=school feature (probably a node at the center of an area surrounding all the buildings). Sometimes in OSM a tag like amenity=college is tagged on a building=college, and in this case education=college should be added, but only because of the amenity tag.

This does mean, we have a whole in our preset definitions, doesn't it?
Right now what the presets "allow" is
- (a) an area or node as "school grounds" (which will now have `education`)
- (b) a building "school building" with only `building=school`

I understand your comment that we need a third option, when there is no area (or node) of "school grounds" but the "school building" acts as the only data source.
- (c) a building "school building ("standalone")" with `building=* + amenity=school + education=school`

Users should then either use A+B or only C.

In https://github.com/openstreetmap/id-tagging-schema/issues/1632 I looked into buildings that have a different tag than `building=school` but still `amenity|education=school` which don't trigger the (B) case. Many of those could be (C) cases.

And I tried to get some feeling for how common this mapping practise (A+B vs C) is. This is a messy ChatGPT chat on this to get a Postpass query going https://chatgpt.com/share/6887145c-cb30-8010-a5c8-e1864696bed5 (only look at the end). The query only looked at Stuttgart Region (random…) and found that 155.894 cases are more like C where only 4.037 are like A+B. At least I hope I understand the results right.

Those numbers suggest to me, that the goal of having `education` as a tag for discrete educational institution is a long way off even with the changes in this PR being merged.

Am I on the right track here?

quincylvania:
@tordans Thanks for working through these things!

The prevalence of tagging facilities on buildings vs. separate mapping as features varies by location and by mapper. I think in Europe it is much more common to combine the tags than in North America. iD sort of tries to split the difference and will add `building=yes` to certain presets. Users can then select the actual building type from the dropdown field. I don't think it's necessarily great to introduce a bunch of similar-sounding presets for different combinations.

I do think the current "School Grounds" preset is a little wonky and could possibly be improved. I didn't attempt to resolve this stuff in this PR, the `education` tag just maps directly onto existing amenity tags.

PR #1473, opened 2025-03-04
Add Waterway Access Point preset and add fields to Slipway preset

quincylvania:
These are all well-established tags.

- https://wiki.openstreetmap.org/wiki/Tag:waterway%3Daccess_point
- https://wiki.openstreetmap.org/wiki/Key:canoe
- https://wiki.openstreetmap.org/wiki/Key:boat
- https://wiki.openstreetmap.org/wiki/Key:motorboat
- https://wiki.openstreetmap.org/wiki/Key:camping
- https://wiki.openstreetmap.org/wiki/Key:swimming
- https://wiki.openstreetmap.org/wiki/Key:trailer
- https://wiki.openstreetmap.org/wiki/Key:snowmobile


data/presets/leisure/slipway.json
```diff
@@ -2,16 +2,41 @@
...
+        "intermittent",
         "lit",
+        "motorboat",
         "opening_hours",
+        "picnic_table",
+        "reservation",
         "seamark/type",
+        "seasonal",
+        "shower",
+        "snowmobile",
```
matkoniecz, line 34: how snowmobiles are relevant on slipways?

data/presets/leisure/slipway.json
```diff
@@ -2,16 +2,41 @@
...
+        "group_only",
+        "intermittent",
         "lit",
+        "motorboat",
         "opening_hours",
+        "picnic_table",
+        "reservation",
         "seamark/type",
+        "seasonal",
+        "shower",
```
matkoniecz, line 33: why `shower` should be show on slipways?

data/presets/leisure/slipway.json
quincylvania: In places where surface water routinely freezes over, it is common to use slipways as an access point for snowmobiles to go out onto the ice. The tag indicates if this is possible and allowed.

data/presets/leisure/slipway.json
quincylvania: Public showers are sometimes available at boat launches where swimming is popular. People will take the boat out, swim in a lake, and then shower upon returning to shore. Shower access is also important for people on overnight canoe trips who want to know what facilities are available at a given launch point.

data/presets/waterway/access_point.json
```diff
@@ -0,0 +1,27 @@
+{
```
matkoniecz, line 1: probably icon distinct from one used for unrecognised tags would make sense, but I have no idea for anything

data/presets/leisure/slipway.json
```diff
@@ -2,16 +2,41 @@
...
+        "operator",
         "surface",
         "access_simple",
+        "canoe",
+        "trailer",
         "fee",
         "payment_multi_fee",
-        "charge_fee"
+        "charge_fee",
+        "tidal"
```
matkoniecz: it is a bit weird to see it inland? But not sure should it be relegated to moreFields

matkoniecz (approved):
cautious approval as I am not an expert on topic or tagging here

data/presets/waterway/access_point.json
tyrasd: maybe the coastline icon would fit?

image

data/presets/waterway/access_point.json
```diff
@@ -0,0 +1,27 @@
...
+        "paddleboard",
+        "portage",
+        "put-in",
+        "rowboat",
+        "take-out"
+    ],
+    "tags": {
+        "waterway": "access_point"
+    },
+    "name": "Waterway Access Point"
```
tyrasd, line 27: Is there no better name for such a location? :see_no_evil:

tyrasd (commented):
I would reuse the options from the `access_simple` field in all of the boat-specific access fields (and add `"autoSuggestions": false`). That way, the options are translated and recognizable. Or was there a specific reason to have them as generic combo fields that even allows free user input?

data/presets/leisure/slipway.json
quincylvania: I agree that by simply looking at a map, `tidal` does seem weird to show inland, but in practice data consumers can't really guess which areas are tidal or not. Some rivers (like the Hudson here in New York) remain tidal hundreds of kilometers inland from the open ocean. While some streams are non-tidal just meters from the beach. It would take a lot of extra work for an app to confidently say a given slipway definitely isn't tidal, so ideally we'd want 100% coverage on the `tidal` tag. 

This wouldn't really matter except that tides can heavily affect the safety of logistics of boating. Some access points can only be used at high or low tides, and some tidal rivers can only be paddled in the direction of the tidal flow. An app can alert users to these types of dangers if it knows that a slipway is tidal.

data/presets/waterway/access_point.json
quincylvania: In English, not really, but paddlers would know what you're talking about. People might say "Canoe Launch" but this implies a wide range of infrastructure. And not every waterway access point is a canoe launch.

data/presets/waterway/access_point.json
quincylvania: Coastline is okay. A better icon would be a path entering the water. I can make something when I have time.

tyrasd (approved):

data/fields/canoe.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "canoe",
+    "type": "combo",
+    "label": "Canoe Access"
```
k-yle, line 4: a bit late, but [the defintion of this tag](http://osm.wiki/Key:canoe) says:

> `canoe=*` is the [access](https://wiki.openstreetmap.org/wiki/Access) tag for canoes, kayaks, and similar paddlecraft.

Would it make sense to rename this label to _"Paddlecraft Access"_ instead of just  _"Canoe Access"_ ? 

I realise that paddling in the inland US waterways is rather different to my own experience, so perhaps it's fine as-is? (given that these are `en-US` labels)

or we ignore this topic and revisit it after https://github.com/openstreetmap/iD/pull/12011 ?

quincylvania: > Would it make sense to rename this label to "Paddlecraft Access" instead of just "Canoe Access" ?

This would probably be better, though "paddlecraft" isn't really a common word and could be confused with pedal boats or paddle wheel ships. I usually say "Canoe & Kayak" in this context but that still leaves out rafts and rowboats.

There has actually been an uptick in `kayak` and `sea_kayak` tagging lately, perhaps under the false assumption that `canoe` doesn't cover these.

matkoniecz: maybe new, less misleading tag can be invented?

PR #1489, opened 2025-03-24
Fix broken "Address" documentation link

Vectorial1024:

See https://github.com/openstreetmap/iD/issues/10905 , especially https://github.com/openstreetmap/iD/issues/10905#issuecomment-2747373701

### Related issues

None.

### Links and data

N/A

### DIFF
--- a/data/fields/address.json
+++ b/data/fields/address.json
@@ -1,6 +1,6 @@
 {
     "type": "address",
-    "key": "addr",
+    "key": "addr:*",
     "keys": [
         "addr:block_number",
         "addr:city",


Vectorial1024:
This should somehow fix https://github.com/openstreetmap/iD/issues/10905

data/fields/address.json
```diff
@@ -1,6 +1,6 @@
 {
     "type": "address",
-    "key": "addr",
+    "key": "addr:*",
```
k-yle, line 3: image

while this appears to work, it will break [other logic that uses `field.key`](https://github.com/openstreetmap/iD/blob/4a1f27b1f6d6a322bd4563ac41aa7f8beb8edd07/modules/ui/fields/address.js#L113). I think it would be safer to use [the dedicated `reference` attribute](https://github.com/ideditor/schema-builder#reference):

```suggestion
    "key": "addr",
    "reference": { "key": "addr:*" },
```

k-yle (dismissed):

Vectorial1024:
@k-yle I have fixed the formatting problems found by Prettier. May you re-run the workflows?

tyrasd (approved):

PR #1490, opened 2025-03-24
Add bench:type field

Zverik:
I have been mapping benches for years, and did not have any idea how to map swing benches. Until somebody in the chat hinted at [bench:type=swing](https://wiki.openstreetmap.org/wiki/Key:bench:type). And I was like, why this is not in the presets?!

This PR adds a bench type field to the `amenity/bench` preset. No predefined options for this combo, like in other `*:type` fields: everything should come from Taginfo, and the values are pretty self-explanatory.

The wiki page mentions a controversy with the `*:type` tags. I believe there is not the case with benches: the `bench=*` tag almost universally is used to mark a presence of a bench (like `shelter=*` and `atm=*`), so we don't have any other options. Also `bench:type` tag has been used on ~3k benches, which is a lot, considering it's not in any presets.

### DIFF
--- a/data/fields/bench/type.json
+++ b/data/fields/bench/type.json
@@ -1,0 +1,5 @@
+{
+    "key": "bench:type",
+    "type": "combo",
+    "label": "Bench Type"
+}
--- a/data/presets/amenity/bench.json
+++ b/data/presets/amenity/bench.json
@@ -4,7 +4,8 @@
         "backrest",
         "armrest",
         "material",
-        "seats"
+        "seats",
+        "bench/type"
     ],
     "moreFields": [
         "colour",


Kovoschiz:
In 2015, 96% of it was mass added. The `delta: 530` is still 16% of the 3293 instances now. https://taghistory.raifer.tech/?#***/bench:type/
`bench=` has eg 393+231+41+29+21+13+9+7+5+5+4+4+4+4+3+2+2=777 instances about types https://taginfo.openstreetmap.org/keys/bench#values

Zverik:
I have read about the mass addition, but still. `bench:type` has been documented on the wiki since 2020, unlike `bench=*`, so that alone makes this tag preferrable. And from my surveying experience I know I need some key to map those attributes.

Kovoschiz:
`bench=` is still more numerous on loungers, so that's inconclusive. There's also confusion compared to `amenity=lounger` https://wiki.openstreetmap.org/wiki/Talk:Key:bench:type#";ounger"_type
A greater controversy is whether `=stand_up` are `=bench` at all. That's another 25% of `bench:type=`

tordans:
I agree the tagging schema can be approved in this area and we should…
But I think the status quo is still too chaotic to add the change currently proposed here.

If we add the type as a field, users will be very motivated to fill it out. For the majority of benches, that will not add any additional information so it is useless added data and costs time. But even worse, I don't see a real "that is a normal bench" value in https://wiki.openstreetmap.org/wiki/Key:bench:type#Common_values … so what would users pick for a regular bench.

We could look for a way to work around this by building the UI in a way that shows the default as "regular bench" and provides a nil value. But I don't think the `combo` type can do that (can it?).

Another thing that I think we need to resolve is, that we have https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/lounger.json as a preset for lounger. In this case, iD would recommend two ways to map the same thing… The data for `lounger` is still hugely in favor of the separate tagging https://taghistory.raifer.tech/?#***/amenity/lounger&***/bench%3Atype/lounger — And given the difference in usage, I think that will not change.

What we could do to improve this, is…
- …create separate presets for the most important additional types. But given that all those variations have less than 1k usage and some are pretty new, I am not sure we want to do this. Also, it would require to know what you are looking for when you search for the preset; which is not ideal
   - https://taghistory.raifer.tech/?#***/bench%3Atype/stand_up&***/bench%3Atype/swing
- … use a checkbox UI to say "this is a swing" and maybe an additional "this is for standing" which then applies the tag (see schema repo for docs)
- … use a combo dropdown but only provide curated value options

tordans (changes requested):
I agree the tagging schema can be approved in this area and we should…
But I think the status quo is still too chaotic to add the change currently proposed here.

If we add the type as a field, users will be very motivated to fill it out. For the majority of benches, that will not add any additional information so it is useless added data and costs time. But even worse, I don't see a real "that is a normal bench" value in https://wiki.openstreetmap.org/wiki/Key:bench:type#Common_values … so what would users pick for a regular bench.

We could look for a way to work around this by building the UI in a way that shows the default as "regular bench" and provides a nil value. But I don't think the `combo` type can do that (can it?).

Another thing that I think we need to resolve is, that we have https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/lounger.json as a preset for lounger. In this case, iD would recommend two ways to map the same thing… The data for `lounger` is still hugely in favor of the separate tagging https://taghistory.raifer.tech/?#***/amenity/lounger&***/bench%3Atype/lounger — And given the difference in usage, I think that will not change.

What we could do to improve this, is…
- …create separate presets for the most important additional types. But given that all those variations have less than 1k usage and some are pretty new, I am not sure we want to do this. Also, it would require to know what you are looking for when you search for the preset; which is not ideal
   - https://taghistory.raifer.tech/?#***/bench%3Atype/stand_up&***/bench%3Atype/swing
- … use a checkbox UI to say "this is a swing" and maybe an additional "this is for standing" which then applies the tag (see schema repo for docs)
- … use a combo dropdown but only provide curated value options

matkoniecz:
also, for https://wiki.openstreetmap.org/wiki/File:Swinging_Bench_in_Towson_Manor_Park.jpg I would tag it as `playground=swing`

I was not aware that it could be tagged as a bench

matkoniecz:
closing per above

PR #1491, opened 2025-03-24
Change the placeholder to a country code that no other country uses.

youssefelzedy:

After a misunderstanding about the `+31` code, which represents a phone number for the country of Holden, @tyrasd  suggested changing this code to any other code that doesn't belong to any country. He then suggested the code `+876`. After searching [Wikipedia](https://en.wikipedia.org/wiki/Area_codes_876_and_658), it turned out to be for the country of North America.

After some thought, he realized that after choosing a random number, it's possible that one day another country will come and take it, and in that case, we wouldn't have done anything about it. So, he suggested changing this from a number to a symbol like +XX, so that it doesn't represent any country. Thus, we would have solved the problem once and for all, and we wouldn't need to change it again.

### Related issues

[#10917](https://github.com/openstreetmap/iD/pull/10917)

### Links and data

https://en.wikipedia.org/wiki/Area_codes_876_and_658

### DIFF
--- a/data/fields/phone.json
+++ b/data/fields/phone.json
@@ -6,7 +6,7 @@
     ],
     "type": "tel",
     "label": "Telephone",
-    "placeholder": "+00 42 345 6459",
+    "placeholder": "+000 0000 0000 0000",
     "terms": [
         "phone number"
     ]
--- a/data/fields/phone.json
+++ b/data/fields/phone.json
@@ -6,7 +6,7 @@
     ],
     "type": "tel",
     "label": "Telephone",
-    "placeholder": "+XX 42 345 6459",
+    "placeholder": "+00 42 345 6459",
     "terms": [
         "phone number"
     ]
--- a/data/fields/phone.json
+++ b/data/fields/phone.json
@@ -6,7 +6,7 @@
     ],
     "type": "tel",
     "label": "Telephone",
-    "placeholder": "+31 42 123 4567",
+    "placeholder": "+XX 42 345 6459",
     "terms": [
         "phone number"
     ]


Vectorial1024:
Alternatively, some sort of zero-value might just work. Like, "+000 0000 0000 0000". Then, it communicates there should be some sort of country phone prefix, followed by the local format of the phone number.

I was not there when the previous PR was drafted, but indeed, the problem was somehow selecting specifically +31 (the Netherlands) instead of admitting "the schema/system doesn't know". Let's see how others feel about whether "XX" or "00" should be used.

youssefelzedy:
@Vectorial1024 It is a perfect idea to use `+00`. I found that there is no country using `+00` as a country code.
What about you @tyrasd ?

Here is the list of all the code countries:
https://en.wikipedia.org/wiki/List_of_telephone_country_codes

data/fields/phone.json
```diff
@@ -6,7 +6,7 @@
     ],
     "type": "tel",
     "label": "Telephone",
-    "placeholder": "+31 42 123 4567",
+    "placeholder": "+00 42 345 6459",
```
tordans:
```suggestion
    "placeholder": "+000 0000 0000 0000",
```

tordans (approved):

tordans:
Lets go with @Vectorial1024's [suggestion](https://github.com/openstreetmap/id-tagging-schema/pull/1491#issuecomment-2753139579) and merge this as `+000 0000 0000 0000`

1ec5:
> He then suggested the code +876. After searching [Wikipedia](https://en.wikipedia.org/wiki/Area_codes_876_and_658), it turned out to be for the country of North America.

By the way, +876 is an unused country code. The list you linked to says that it’s reserved for future maritime service. You might have misunderstood the entry for the 876 area code used by Jamaica. This is part of a separate dialing scheme in North America that doesn’t use country codes. Although it corresponds to a country, it isn’t a country code and wouldn’t appear in this part of the phone number.

matkoniecz:
https://github.com/openstreetmap/id-tagging-schema/issues/2283 has reraised this issue - and I admit that I am not entirely sure why this PR was made and what issue it is fixing

> After a misunderstanding about the +31 code, which represents a phone number for the country of Holden, @tyrasd suggested changing this code to any other code that doesn't belong to any country. He then suggested the code +876. After searching [Wikipedia](https://en.wikipedia.org/wiki/Area_codes_876_and_658), it turned out to be for the country of North America.

links https://github.com/openstreetmap/iD/pull/10917 and links back to https://github.com/openstreetmap/iD/issues/10915

And problem claimed in https://github.com/openstreetmap/iD/issues/10915 was not actually fixed here. Is this change an improvement? If yes, why?

(either it should be rolled back or applied to mobile and fax, no idea what is better - see https://github.com/openstreetmap/id-tagging-schema/issues/2283 )

Vectorial1024:
Guy was probably vibe-coding. The texts feel context-free.

@matkoniecz If my memory serves me right, back then I was reviewing how Hong Kong phone hints were being handled, and then the guy randomly came in and said "wait the Egypt phone hints are also wrong" and made the 10915 issue. There are actually two layers:

- iD does not know about Egypt phone hint (https://github.com/openstreetmap/iD/pull/10916)
- Default phone hint should not be Dutch (we are here)

Ultimately, 10917 is just a weird PR; who would think it's a good idea?

This PR is still an improvement. Hints should be neutral, which I then proposed using `+000`.

youssefelzedy:
> Guy was probably vibe-coding. The texts feel context-free.

You should use respectful language when we talk in a public repo. You were also part of the original discussion back then.

The story is: @Vectorial1024 opened #10899 (`Hong Kong/Macau are wrongly getting Chinese phone/fax hint text
`). I checked Egypt if it has the same issue or not. And I found it, so I opened #10916 to solve this issue (by adding it manually). 

After that, I found that the default phone placeholder was using the Netherlands country code (`+31`), so I suggested changing it to a random number in #10917 instead of being a default. @tyrasd suggested to me just `open a PR on the tagging repository and replace the [+31] placeholder with a country code that does not exist (maybe something like +876 or something else that's fun).` 

Yeah, when I was searching, I thought the `+876` was already for a country, but @1ec5 explained that was incorrect.. anyway after that, @Vectorial1024, @tordans, and I agreed the best use case was using only zeros in the default number to be a neutral solution.

The issue could have been explained without resorting to personal remarks. Making fun of someone or belittling them is not constructive and has no place in a public repository.

Vectorial1024:
Ultimately, we still achieve something constructive.

Issue #1480, opened 2025-03-11
Recycling centre and container have incorrect descriptions


Oivo35:
Recycling centre has description of [recycling_type](https://wiki.openstreetmap.org/wiki/Key%3Arecycling_type)
Correct data item is [recycling_type=centre](https://wiki.openstreetmap.org/wiki/Item:Q21274)

![Image](https://github.com/user-attachments/assets/c3cf0dec-4480-416e-a894-faa3a48e580d)

Recycling bin has description of [amenity=recycling](https://wiki.openstreetmap.org/wiki/Tag:amenity=recycling)
Correct data item is [recycling_type=container](https://wiki.openstreetmap.org/wiki/Item:Q21244)

![Image](https://github.com/user-attachments/assets/3454e5bb-8147-4963-85dc-4d9d9b072dd4)

SanaliSLokuge:
In response to your request, I created a new preset for Recycling Centre inside data\presets\amenity\recycling. Now the containers and recycling centres are seperated with seperate presets for each.

Looking forward to your feedback.

PR #1496, opened 2025-03-29
fix: Added recycling centre preset separately from container.

SanaliSLokuge:
Added a Separate Preset for Recycling Centre inside the centre folder in data\presets\amenity\recycling.

Fixed the issue where recycling center and container had incorrect descriptions:
Recycling center did not have a preset for it instead recycling folder only had a preset for containers. Created a separate preset for Recycling centre and added necessary details.

### Related issues
Recycling centre and container have incorrect descriptions #1480

### Links and data

**Relevant OSM Wiki links:**
[Link 1](https://wiki.openstreetmap.org/wiki/Key%3Arecycling_type)
[Link 2](https://wiki.openstreetmap.org/wiki/Item:Q21274)
[Link 3](https://wiki.openstreetmap.org/wiki/Tag:recycling_type%3Dcentre)

</details>

### DIFF
--- a/data/presets/amenity/recycling/centre/centre.json
+++ b/data/presets/amenity/recycling/centre/centre.json
@@ -1,0 +1,32 @@
+{
+
+    "icon": "maki-recycling",
+    "fields": [
+        "{amenity/recycling_centre}",
+        "operator",
+        "opening_hours",
+        "collection_times"
+    ],
+    "moreFields": [
+        "{amenity/recycling_centre}",    
+        "website",     
+        "landuse",
+        "fee"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "recycling centre",
+        "waste processing",
+        "recycling"
+    ],
+    "tags": {
+        "amenity": "recycling",
+        "recycling_type": "centre"
+    },
+    "name": "Recycling Centre",
+    "searchable": true
+    
+}


Oivo35:
I would keep the current fields of the preset.
    "fields": [
        "name",
        "operator",
        "operator/type",
        "address",
        "building",
        "opening_hours",
        "recycling_accepts"
    ],
    "moreFields": [
        "{@templates/contact}",
        "charge_fee",
        "fee",
        "payment_multi_fee",
        "wheelchair"

The [collection_times](https://wiki.openstreetmap.org/wiki/Key:collection_times) is used at containers or post boxes and means collection frequency. However, containers and post boxes are mostly accessible 24/7 and this means opening_hours=24/7. Therefore, the recycling center should also have `opening_hours` instead.

Oivo35:
@Dimitar5555 
Mayby the name should also be changed from `Recycling Center` (American English) to `Recycling Centre` (British English).

Dimitar5555:
> @Dimitar5555 Mayby the name should also be changed from `Recycling Center` (American English) to `Recycling Centre` (British English).

iD's presets use American English by default. There is an `en-UK` locale where British English is used.

Dimitar5555:
@SanaliSLokuge There are presets for both containers and centers. The issue is caused by their reference tags which I've corrected in #1497.

PR #1500, opened 2025-04-01
Created preset for acequias

swaree:

I'm currently editing on and off the historical irrigation system of my area, which is predominantly based on irrigation canals, i.e., acequias (Spanish _acequias_, Catalan _séquies_). With the current Transifex translation, acequias use `waterway=ditch`, which is incorrect per the OSM wiki, which reserves this tag only for unuseful water, referring that sense to `waterway=canal`. I expect medium usage of this preset, and with the descriptive name, it would now be unmistakable. I also edited the translation for _cuneta_, which is more appropiate

I am aware the term "acequia" is very local, and is only used in the Southern US (including other Spanish- and Catalan-speaking areas). However, this won't obscure the OSM data descriptions or the Canal preset.

### Links and data

**Relevant OSM Wiki links:**
- [`waterway=ditch`](https://wiki.openstreetmap.org/wiki/Tag:waterway%3Dditch)
- [`waterway=canal`](https://wiki.openstreetmap.org/wiki/Tag:waterway%3Dcanal)

Thanks, cheers

### DIFF
--- a/data/presets/waterway/canal/acequia.json
+++ b/data/presets/waterway/canal/acequia.json
@@ -1,0 +1,15 @@
+{
+    "icon": "iD-waterway-ditch",
+    "fields": [
+        "name",
+        "intermittent",
+    ],
+    "geometry": [
+        "line"
+    ],
+    "tags": {
+        "waterway": "canal",
+        "usage": "irrigation"
+    },
+    "name": "Acequia"
+}
--- a/data/preset_categories/waterway.json
+++ b/data/preset_categories/waterway.json
@@ -6,6 +6,7 @@
         "waterway/drain",
         "waterway/river",
         "waterway/canal",
+        "waterway/canal/acequia",
         "waterway/ditch",
         "natural/water/stream",
         "natural/water/river",


data/presets/waterway/canal/acequia.json
```diff
@@ -0,0 +1,15 @@
...
+        "intermittent",
+    ],
+    "geometry": [
+        "line"
+    ],
+    "tags": {
+        "waterway": "canal",
+        "usage": "irrigation"
+    },
+    "name": "Acequia"
```
matkoniecz, line 14: this is surely not an English name?

if should not be added to English presets

matkoniecz: Displaying all `waterway=canal usage=irrigation` as "Acequia" will be simply wrong

the same goes for translation work

matkoniecz (commented):
I am not saying that it should not be solved but this is not a proper way to do this

data/presets/waterway/canal/acequia.json
swaree: > this is surely not an English name?
> 
> if should not be added to English presets

Hi, I based it on the name it usually takes in English, which indeed comes from Spanish: https://en.wikipedia.org/wiki/Acequia 

Presets in other languages aren't unheard of though: [`historic=stolperstein`](https://wiki.openstreetmap.org/wiki/Tag:memorial%3Dstolperstein)

matkoniecz: > An acequia (Spanish: [[aˈθekja]](https://en.wikipedia.org/wiki/Help:IPA/Spanish)) or séquia (Catalan: [[ˈsekiə, -a]](https://en.wikipedia.org/wiki/Help:IPA/Catalan), also known as síquia [[ˈsikiə, -a]](https://en.wikipedia.org/wiki/Help:IPA/Catalan)) is a community-operated [watercourse](https://en.wikipedia.org/wiki/Watercourse) used in [Spain](https://en.wikipedia.org/wiki/Spain) and former [Spanish colonies in the Americas](https://en.wikipedia.org/wiki/Spanish_colonization_of_America) for [irrigation](https://en.wikipedia.org/wiki/Irrigation). Acequias are found in parts of Spain, the [Andes](https://en.wikipedia.org/wiki/Andes), northern [Mexico](https://en.wikipedia.org/wiki/Mexico), and the modern-day [American Southwest](https://en.wikipedia.org/wiki/American_Southwest) (northern [New Mexico](https://en.wikipedia.org/wiki/New_Mexico) and southern [Colorado](https://en.wikipedia.org/wiki/Colorado)).

so for example it should not be used for `waterway=canal usage=irrigation` in Germany

> Presets in other languages aren't unheard of though: [memorial=stolperstein](https://wiki.openstreetmap.org/wiki/Tag:memorial%3Dstolperstein)

note that all memorial=stolperstein are stolpersteins

is every single `waterway=canal usage=irrigation` actually a [Acequia](https://en.wikipedia.org/wiki/Acequia)?

tyrasd (changes requested):
I'm agreeing with what @matkoniecz said above.

To summarize: While every Acequia is an irrigation canal, this is not true the other way around: not even in Spain, every `waterway=canal + usage=irrigation` is an Acequia. We would need at least one more tag to make this explicit (e.g. something like `canal=acequia`, but that would need to be discussed with the community first!).

What about the following approach, as a first step: change the proposed preset to a generic preset for any Irrigation Canal (with `acequia` as a search term). What do you think?

matkoniecz:
@swaree Is "acequia" simply used for all irrigation canals in some region? Or is there something distinguishing acequias from other irrigation canals?

Is it word that should be added to Spanish translation or would it be used by English-language speaker?

matkoniecz:
closing, as important questions blocking merge seem to be not answered

PR #1507, opened 2025-04-04
Add field `sidewalk` on `highway` presets

tordans:
Following the long discussion in https://github.com/openstreetmap/id-tagging-schema/pull/454 this PR add the field sidewal.

Supported values: `yes|no|separate`
Autocomplete is disabled to prevent miss tagging – improving the situation described in https://github.com/openstreetmap/id-tagging-schema/pull/1278

Supported key: `sidewalk|sidewalk:both|sidewalk:left|sidewalk:right`
It will use `sidewalk:both` when user change date or add new data based on the evaluation in https://github.com/openstreetmap/id-tagging-schema/pull/454#issuecomment-2731787841.

The third tagging practice of using the values `left, right, both` will be supported once https://github.com/openstreetmap/iD/issues/10839 is implemented.

All those parts have already been discussed in the linked PR.

The main part that is new here is, which presets show the field.
I added the field to all presents, that have the `"cycleway"` field today. Just like the cycleway field the sidewalk field is placed in the `moreFields`. We could do more data crunching and analysis to find out if we would want to move the field to the `fields` section in some cases but I suggest to do this later to keep this PR simple.


tordans:
## Test-Documentation

… for https://pr-1507--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=19.43/14.65478/121.06335&disable_features=boundaries&background=Bing&id=w4095389

- Looks good. But ich changed the label to "Sidewalk**s**" after taking this screenshot 
   > image

- The "old" tagging does not look ideal. But given that we have a plan to improve this and given that it does give a strong signal to users to click and update the field (which will improve the data) it think that is OK
   > image

### Search

All good:

> image



### Info-`i`

Info-i looks good
> image

- That links to https://wiki.openstreetmap.org/wiki/Key:sidewalk which I think should be updated (see https://github.com/openstreetmap/id-tagging-schema/pull/454#issuecomment-2777839092) but that is separate from this PR
- It links to https://wiki.openstreetmap.org/wiki/Item:Q699 which is fine


### Wording

- [x] American English

Dimitar5555:
The order of the items in the placeholder and the order of the dropdown values doesn't match.

image

---

`none` and *None* may look a bit confusing when they are on top of each other (a mapper might ask "Why is one blue and the other black?"). It might be better to use 'No' for 'sidewalk*=no` label. 

image

---

Clicking on `=none` shows `None` with dark grey background (which is likely due to the search ability of the field. It may seem like it is the selected option.

image

waldyrious:
Nice! Looking forward to see this finally be in iD 😃

I tested the preview a bit, and noticed that when there's `sidewalk=both`, the two fields show the value "both" (which I guess is inevitable; at least, I can't think of an alternative). But when I change one of them to e.g. "No", the result is `sidewalk:left=no`+`sidewalk:right=both`, with doesn't seem like the right choice. If it's already changing `sidewalk` to `sidewalk:right`, it might as well include a change from `both` to `yes`, which not just prevents the extra step of modifying the right sidebar by hand, but actually prevents **inserting** (😱) an erroneous tag value.

tordans:
@waldyrious let me refer this to …

> The third tagging practice of using the values `left`, `right`, `both` will be supported once https://github.com/openstreetmap/iD/issues/10839 is implemented.

Which is where we should discuss this to not derail the conversation if how and when to merge this PR. You fill find [a Draft PR](https://github.com/openstreetmap/iD/pull/10935) attached over there which mentions this case. Let me also point out that a spirit of incremental optimization will help us get at least something going and faster.

waldyrious:
Got it, thanks for the pointers :) I'll follow that thread.

> Let me also point out that a spirit of incremental optimization will help us get at least something going and faster.

Absolutely! To be clear, I only commented here because you asked for feedback in #1278. I definitely agree that this PR is an improvement over the current state of things and should be integrated regardless of the issue I pointed out.

gy-mate:
I wanted to try out this new field but unfortunately the [Netlify preview](https://pr-1507--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=19.43/14.65478/121.06335&disable_features=boundaries&background=Bing&id=w4095389) isn't available anymore :(

data/fields/sidewalk.json
```diff
@@ -0,0 +1,34 @@
...
+                "title": "Yes",
+                "description": "There is a sidewalk"
+            },
+            "no": {
+                "title": "No",
+                "description": "No sidewalk"
+            },
+            "separate": {
+                "title": "Sidewalk Mapped Separately",
+                "description": "The sidewalk are mapped using separate ways"
```
gy-mate:
```suggestion
                "description": "The sidewalk is mapped using separate ways"
```

A minor grammar suggestion.

tordans:
> I wanted to try out this new field but unfortunately the [Netlify preview](https://pr-1507--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=19.43/14.65478/121.06335&disable_features=boundaries&background=Bing&id=w4095389) isn't available anymore :(

Your wording change just now triggered a new build. It will, however, not include the changes that are part of https://github.com/openstreetmap/iD/pull/10935, yet

gy-mate:
> Your wording change just now triggered a new build.

Great—thanks for the heads-up! :)

Are `highway=service` roads left out from this PR on purpose? If yes, why?

gy-mate:
Thank you so much! :)

data/fields/sidewalk.json
```diff
@@ -0,0 +1,34 @@
...
+    "keys": [
+        "sidewalk:left",
+        "sidewalk:right"
+    ],
+    "reference": {
+        "key": "sidewalk"
+    },
+    "type": "directionalCombo",
+    "label": "Sidewalks",
+    "placeholder": "Yes, No, Mapped using separate ways",
```
gy-mate:
```suggestion
    "placeholder": "Yes, No, Mapped Separately",
```

In order to...

1. match the `title` (and capitalization) of the option in question
2. fit in the text box.

data/fields/sidewalk.json
```diff
@@ -0,0 +1,34 @@
...
+            "yes": {
+                "title": "Yes",
+                "description": "There is a sidewalk"
+            },
+            "no": {
+                "title": "No",
+                "description": "No sidewalk"
+            },
+            "separate": {
+                "title": "Sidewalk Mapped Separately",
```
gy-mate:
```suggestion
                "title": "Mapped Separately",
```

The word _Sidewalk_ is already included in the `label`.

gy-mate:
Thank you so much! :) Just a few additional minor suggestions:

gy-mate:
Thanks a lot, LGTM!

data/presets/highway/primary.json
```diff
@@ -15,6 +15,7 @@
         "charge_toll",
         "covered_no",
         "cycleway",
+        "sidewalk",
```
matkoniecz, line 18: and say [tertiary](https://github.com/openstreetmap/id-tagging-schema/blob/a0aa625ea273ed97029277bbaf9a1e2b0cd5b03f/data/presets/highway/tertiary.json) is inheriting from `primary` fields

tordans: Right, we only need to change this one place; it follows the same pattern that `cycleway` has.

matkoniecz (approved):

tordans:
> What we would need here is a way to define "equivalent" tagging variants, e.g. `sidewalk=left` is the same `sidewalk:left=yes` + `sidewalk:right=no`. For now we could hard-code that for the `sidewalk` tag in iD & co, but it would be good to think whether we should include something like that in the [schema-builder](https://github.com/ideditor/schema-builder/).

This is what https://github.com/openstreetmap/iD/pull/10935 would add, right? (I might be missunderstanding it…)

matkoniecz:
https://github.com/openstreetmap/iD/commit/92667076b7cddea9fdafd93b45f13dba8cf1479e just appeared in iD

would closing/reopening PR to trigger rebuild would deploy preview with it included? Or is it taking latest released iD?

matkoniecz:
> left/right and way direction is deeply nonobvious but I have no great ideas how to better handle it
> maybe have a visual GUI like StreetComplete has? but it is not exactly so intuitive
> maybe show sidewalk sides on map if user edits these fields? but I have no idea how to code it

I opened https://github.com/openstreetmap/iD/issues/11950

PR #1517, opened 2025-04-15
Add Fish Farm / Hatchery preset and `aquaculture=*` field

arch0345:

This PR adds a preset for [fish hatcheries](https://en.wikipedia.org/wiki/Fish_hatchery), which are prevalent around the world. I'm hoping that this preset will help prevent these from being mistagged, such as [this one](https://www.openstreetmap.org/way/369021245). As there doesn't seem to be a tag to differentiate between fish farms/hatcheries, I've named the preset 'Fish Farm / Hatchery'

I also replaced the `produce` field in the Aquaculture preset with a new `aquaculture=*` field given that only 8% of features tagged with `landuse=aquaculture` use the former compared to 28% for the latter.

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Tag%3Aaquaculture%3Dfish
- https://wiki.openstreetmap.org/wiki/Key:aquaculture

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/tags/landuse=aquaculture#combinations

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/data/presets/landuse/aquaculture.json
+++ b/data/presets/landuse/aquaculture.json
@@ -3,7 +3,7 @@
     "fields": [
         "name",
         "operator",
-        "produce"
+        "aquaculture"
     ],
     "moreFields": [
         "{@templates/contact}",
@@ -16,7 +16,6 @@
         "landuse": "aquaculture"
     },
     "terms": [
-        "fish farm",
         "marine farm",
         "crustacean",
         "algae",
--- a/data/presets/landuse/aquaculture.json
+++ b/data/presets/landuse/aquaculture.json
@@ -3,13 +3,14 @@
     "fields": [
         "name",
         "operator",
-        "aquaculture"
+        "produce"
     ],
     "moreFields": [
         "{@templates/contact}",
         "address"
     ],
     "geometry": [
+        "point",
         "area"
     ],
     "tags": {
--- a/data/presets/landuse/aquaculture/fish.json
+++ b/data/presets/landuse/aquaculture/fish.json
@@ -1,29 +1,28 @@
 {
     "icon": "fas-fish",
     "fields": [
-        "name",
-        "operator"
+        "{landuse/aquaculture}"
     ],
     "moreFields": [
         "{landuse/aquaculture}"
     ],
     "geometry": [
+        "point",
         "area"
     ],
     "tags": {
-        "aquaculture": "fish"
-    },
-    "addTags": {
         "landuse": "aquaculture",
-        "aquaculture": "fish"
-    },
-    "reference": {
-        "key": "aquaculture",
-        "value": "fish"
+        "produce": "fish"
     },
     "terms": [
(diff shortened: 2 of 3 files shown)


kjonosm:
Searching for the preset works fine, except one would expect the term `aquaculture` to also be in the list of aliases.

image
image
image

One would expect to read the description from `landuse=aquaculture` or [`aquaculture=fish`](https://wiki.openstreetmap.org/wiki/Tag:aquaculture%3Dfish) but not from the key `aquaculture=*` as currently shown. Additionally one would also expect to see the field for `aquaculture=*` to be able to switch to another value when necessary.

image

Kovoschiz:
Those `aquaculture=` are mostly mass added https://taghistory.raifer.tech/#***/aquaculture/&***/aquaculture/fish&***/aquaculture/shrimp
Author didn't respond to comments about `produce=` at that time https://wiki.openstreetmap.org/wiki/Proposal_talk:Aquaculture#Produce
This blocks `aquaculture=` for other undetermined use. There's a need to distinguish whether it is in the sea, ponds, or water tanks. https://en.wikipedia.org/wiki/Mariculture#Types

kjonosm:
According to the Wiki `landuse=aquaculture` is also allowed on nodes.

data/presets/landuse/aquaculture.json
```diff
@@ -10,13 +10,13 @@
         "address"
     ],
     "geometry": [
+        "point",
```
tordans, line 13:
```suggestion
```

Do we really want to promote this, given that right now there are 500 point data vs. 30k area data entries and this really should be an area?

Ideally, we would have a way to apply the concept of "unsearchable presets" to only some geometry types. But as it stands, we would have to create a second preset unsearchable for points. (However 500 would be too low for that).

data/presets/landuse/aquaculture/fish.json
```diff
@@ -0,0 +1,28 @@
+{
+    "icon": "fas-fish",
+    "fields": [
+        "{landuse/aquaculture}"
+    ],
+    "moreFields": [
+        "{landuse/aquaculture}"
+    ],
+    "geometry": [
+        "point",
```
tordans, line 10: See above.

tordans (commented):
I see it was merged just now. So you can ignore my comments :-)

data/presets/landuse/aquaculture.json
matkoniecz: I consider it worth promoting, given how often people use fake geometries rather than points

matkoniecz: though ideally it would be consistent across presets and I am not deeply opposed to decision going one way or another

PR #1534, opened 2025-04-29
Add description to Removable Bollard string to make it clear the values includes bollards which are removable only with a key

andrewharvey:
Without this it's unclear to mappers if this value is for removable by anyone who wants to go past or if it can also be for those removable bollards but which require a key or other token.

This is confirmed by the wiki description at https://wiki.openstreetmap.org/wiki/Key:bollard

### DIFF
--- a/data/fields/bollard.json
+++ b/data/fields/bollard.json
@@ -7,7 +7,7 @@
             "fixed": "Fixed Bollard",
             "flexible": "Flexible Bollard",
             "foldable": "Foldable Bollard",
-            "removable": "Removable Bollard (with or without a key)",
+            "removable": { "title": "Removable Bollard", "description: Removable with or without a key" },
             "rising": "Rising Bollard"
         }
     }
--- a/data/fields/bollard.json
+++ b/data/fields/bollard.json
@@ -1,17 +1,17 @@
 {
-  "key": "bollard",
-  "type": "combo",
-  "label": "Type",
-  "strings": {
-    "options": {
-      "fixed": "Fixed Bollard",
-      "flexible": "Flexible Bollard",
-      "foldable": "Foldable Bollard",
-      "removable": {
-        "title": "Removable Bollard",
-        "description": "Removable with or without a key"
-      },
-      "rising": "Rising Bollard"
+    "key": "bollard",
+    "type": "combo",
+    "label": "Type",
+    "strings": {
+        "options": {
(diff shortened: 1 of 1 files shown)


andrewharvey:
> My take is, to move this into the description and get an improved UI for those going …

That's a good point. Let's go with that then, so we can close this PR.

> https://wiki.openstreetmap.org/wiki/Item:Q23021

I'm not sure about this now.

"The bollard can be removed with our without tools."

With the right tools or enough force, anything can be removable. I think the intent is that it should be for bollards designed to be removed without force or destruction of the bollard, so that someone, either the public (no key needed) or the authorised person (those with a key) can easily remove it when needed then re-lock/place it.

data/fields/bollard.json
```diff
@@ -7,7 +7,7 @@
             "fixed": "Fixed Bollard",
             "flexible": "Flexible Bollard",
             "foldable": "Foldable Bollard",
-            "removable": "Removable Bollard",
+            "removable": "Removable Bollard (with or without a key)",
```
tordans, line 10:
```suggestion
            "removable": { "title": "Removable Bollard", "description: Removable with or without a key" },
```

tordans:
@andrewharvey I have no opinion on the wording. I just copied what I found on https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dbollard#Types which I now see is very different to https://wiki.openstreetmap.org/wiki/Key:bollard

Feel free to change all that!

data/fields/bollard.json
matkoniecz: > That's a good point. Let's go with that then, so we can close this PR.

would you accept this specific change and retest is PR working?

matkoniecz: or should this PR be closed as an abandoned?

andrewharvey: sorry not sure who these comments were directed to, but that seems fine to me, I didn't realise you could use title/description values like that.

matkoniecz:
Is this description supposed to appear somewhere? I cannot find it while testing.

BTW, probably "removable bollard" should be able to find at least basic bollard preset and it seems to fail now.

matkoniecz:
now visible

screen-2026-07-24-11-23-45

matkoniecz (approved):

matkoniecz:
Thanks for the PR @andrewharvey !

PR #1537, opened 2025-05-05
change `opening_hours`-like fields to `type=schedule` (iD will link oh evaluator)

k-yle:
see previous discussion at https://github.com/ideditor/schema-builder/pull/101

PR #997, opened 2023-09-07
Create unsearchable preset for associatedStreet

serhii-muchychka:
closes: https://github.com/openstreetmap/iD/issues/9639

PR #1538, opened 2025-05-05
add relation role definitions to 32 presets

### LINKED ISSUE #134: Presets should enumerate expected roles
A preset that applies to relations should be able to list the expected member roles, along with a translatable label and expected element types for each role.

Currently, id-tagging-schema has no data on roles, so iD’s relation member editor simply lists the most common roles in taginfo for a given relation type: openstreetmap/iD#3300. id-tagging-schema’s multipolygon preset could specify Inner Ring (`inner`) and Outer Ring (`outer`) roles, disallowing other roles which would be invalid: openstreetmap/iD#8268.

Some roles are counterintuitive; a translatable string would give us the opportunity to make them less misleading. For example, the English label for `admin_centre` could be “Capital”, so that mappers don’t incorrectly assign it to a capitol building.

If the preset could say how many members of a relation can have a given role, and what element types are eligible, then the editor can implement a validation role to that effect. For example, iD could warn if a boundary relation has two Focal Point (`label`) members or a restriction relation has a node as its From (`from`) member.

### LINKED ISSUE #1033: Suggest 'from', 'to' and 'via' only for public transport routes
**Describe the bug**
Currently, iD editor suggests `from`, `via` and `to` for all `type=route` + `route=*` relations. However, these keys are useful only for public transport routes and recreational routes where there are separate relations for A -> B and B -> A. For other route types there are only one relation for the route (with possible `forward` and `backward` roles of the members).

See:

- https://wiki.openstreetmap.org/wiki/Key:from
- https://wiki.openstreetmap.org/wiki/Key:via
- https://wiki.openstreetmap.org/wiki/Key:to

**To Reproduce**
1. Add a new relation.
2. Select road route.
3. See the preset.

**Expected behavior**
Bi-directional routes should not have `from`, `via` and `to` fields.

### LINKED ISSUE #2622: `label` and `admin_centre` roles have misleading labels
As of #2576, the Administrative Boundary preset’s `label` and `admin_centre` roles are labeled quite literally as “Label” and ”Admin Center” in English. These terms are as misleading as the raw role keywords and will become even more misleading in translation. I detailed the concerns in https://github.com/openstreetmap/schema-builder/issues/134#issue-2129505702, https://github.com/openstreetmap/id-tagging-schema/pull/1538#discussion_r3661318002, and https://github.com/openstreetmap/id-tagging-schema/pull/1538#discussion_r3661328686.

https://github.com/openstreetmap/id-tagging-schema/blob/7ba70f4abf3fc32f22d6e9a531f421f9c653540d/data/presets/type/boundary/administrative.json#L28-L29

We should assume the user will take these labels at face value based on their preconceptions about the terms, unless we choose alternative terms that give them pause. Here are some suggestions in order of preference:

* `label`: Focal Point, Representative Point, Equivalent Point, Conventional Center Point
* `admin_centre`: [Capital](https://www.wikidata.org/wiki/Property:P36), Capital / Seat

### LINKED ISSUE #9127: control empty role of boundaries
### Description

Hi,

currently, when a user creates a multipolygon and forgets to add a role to one or more members, iD shows a warning. This is a good thing if we read the wiki : https://wiki.openstreetmap.org/wiki/Relation:multipolygon = a member with no role is indicated as deprecated.

In the wiki, we find the same rule for boundaries : https://wiki.openstreetmap.org/wiki/Relation:boundary = a member with no role is also indicated as deprecated.

But ID shows no warning when a mapper forgets a role and sends to osm database a boundary relation with one or more members with an empty role.

Could you add a control on empty roles of boundary relations like the one for multipolygon relations ?

For your informations, there were more than 17,000 boundary relations with empty roles a few month ago. Yesterday, zero on a way. Today 5 relations with empty roles on a way = all have been created with ID.

Best regards

### Screenshots

_No response_

k-yle:
Closes https://github.com/ideditor/schema-builder/issues/134, Closes https://github.com/openstreetmap/iD/issues/9127, Closes #1033

~~Lets discuss the schema structure at https://github.com/ideditor/schema-builder/pull/174 and the keep this PR for bike-shedding the tagging and labels.~~

~~The CI will fail until https://github.com/ideditor/schema-builder/pull/174 is merged & released.~~


k-yle:
also `associatedStreet` - https://github.com/openstreetmap/id-tagging-schema/pull/997#issuecomment-3267669136

matkoniecz:
label changed per https://github.com/openstreetmap/id-tagging-schema/pull/1537#issuecomment-3269620962

data/presets/public_transport/stop_area.json
```diff
@@ -17,5 +17,67 @@
...
+                        "railway": "subway_entrance"
+                    }
+                ],
+                "geometry": [
+                    "vertex"
+                ]
+            },
+            {
+                "role": "",
+                "roleLabel": ""
```
matkoniecz: maybe some mentions of that this members from route would make sense


```suggestion
                "roleLabel": "part of route"
```

or would it be silly for some reason?

package.json
```diff
@@ -20,7 +20,7 @@
     "translations": "node scripts/translations.js"
   },
   "devDependencies": {
-    "@ideditor/schema-builder": "~6.5.1",
+    "@ideditor/schema-builder": "github:k-yle/schema-builder#kyle-deploy",
```
matkoniecz: this is temporary, to be changed on actual merge, right?

data/presets/type/route_master.json
```diff
@@ -19,5 +19,29 @@
...
+                "role": "",
+                "roleLabel": "",
+                "geometry": [
+                    "relation"
+                ],
+                "matchTags": [
+                    {
+                        "type": "route",
+                        "route": "$1",
+                        "ref": "$2"
```
matkoniecz: what about pathological cases where single Route Master has subvariants?

With main route being say 1 and having subrelations 1, 1, 1a, 1b?

matkoniecz: not sure is above considered as clearly valid or clearly invalid

package.json
k-yle: yes, this is just to make testing possible - more details in https://github.com/ideditor/schema-builder/pull/174#issuecomment-4004549758

data/presets/type/restriction.json
```diff
@@ -10,5 +10,49 @@
     "tags": {
         "type": "restriction"
     },
-    "name": "Restriction"
+    "name": "Restriction",
+    "relation": {
+        "id": "restriction",
+        "allowDuplicateMembers": true,
```
matkoniecz: we should have separate entry for u-turn allowing duplicate restrictions, and with them blocked for regular ones

at least I think so

data/presets/type/restriction.json
```diff
@@ -10,5 +10,49 @@
...
+            },
+            {
+                "role": "via",
+                "roleLabel": "Via",
+                "geometry": [
+                    "vertex",
+                    "line"
+                ],
+                "min": 1,
+                "max": 1
```
matkoniecz: multiple vias are possible

data/presets/type/route.json
```diff
@@ -17,5 +17,215 @@
...
+                "matchTags": [
+                    {
+                        "public_transport": "stop_position",
+                        "$1": "yes"
+                    }
+                ]
+            },
+            {
+                "role": "stop_on_demand_exit_only",
+                "roleLabel": "Stop (Exit Only) (only stops on request)",
```
matkoniecz:
```suggestion
                "roleLabel": "Stop (Exit Only, only stops on request)",
```

Though Inconsistent Capitalization Looks even worse now.

data/presets/type/route.json
```diff
@@ -17,5 +17,215 @@
...
+        "optionalTags": {
+            "route": "$1"
+        },
+        "allowDuplicateMembers": true,
+        "members": [
+            {
+                "role": "stop",
+                "roleLabel": "Stop",
+                "geometry": [
+                    "point",
```
matkoniecz: stop_position should be vertex within road, right?

data/presets/type/waterway.json
```diff
@@ -15,5 +15,52 @@
...
+                "role": "",
+                "roleLabel": "",
+                "matchTags": [
+                    {
+                        "waterway": "*"
+                    }
+                ],
+                "geometry": [
+                    "line",
+                    "area"
```
matkoniecz: why area would be valid here?

https://wiki.openstreetmap.org/wiki/Relation:waterway seems to claim otherwise

data/presets/area.json
```diff
@@ -12,5 +12,28 @@
         "polygon"
     ],
     "name": "Area",
-    "matchScore": 0.1
+    "matchScore": 0.1,
+    "relation": {
```
matkoniecz, line 16: it is mostly registering my confusion... But is this `area` entry representing both closed ways with area tags and multipolygons?

If yes, would adding this relations structure interfere with applying it to closed ways?

https://github.com/ideditor/schema-builder/pull/174/changes#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5 has `For relations,` so I guess nonrelations should ignore this structure?

data/presets/type/route.json
Dimitar5555:
```suggestion
                "roleLabel": "Exit only stop (stops on demand)",
```

"on demand" should be equivalent to "on request"

data/presets/type/enforcement.json
```diff
@@ -9,5 +9,74 @@
...
-    "name": "Enforcement"
+    "name": "Enforcement",
+    "relation": {
+        "id": "enforcement",
+        "allowDuplicateMembers": false,
+        "members": [
+            {
+                "role": "device",
+                "roleLabel": "Device",
+                "matchTags": [
```
matkoniecz, line 26: hmmmm

https://wiki.openstreetmap.org/wiki/Relation:enforcement gives example of red light camera placed on `highway=traffic_signals` without further tags

https://wiki.openstreetmap.org/wiki/Relation:enforcement#Example_5:_Distance_between_vehicles even wants to put it on tagless way

https://wiki.openstreetmap.org/wiki/Relation:enforcement#Example_7:_Height_check_before_tunnel also reccs tagsless device node

matkoniecz: https://www.openstreetmap.org/node/3524489413 is it valid?

What about https://www.openstreetmap.org/note/5192079 ?

data/presets/type/destination_sign.json
```diff
@@ -15,5 +15,55 @@
...
+    "name": "Destination Sign",
+    "relation": {
+        "id": "destination_sign",
+        "allowDuplicateMembers": true,
+        "members": [
+            {
+                "role": "from",
+                "roleLabel": "From",
+                "geometry": [
+                    "vertex",
```
matkoniecz, line 33: https://wiki.openstreetmap.org/wiki/Relation:destination_sign claims it may be also a node

(in such case can we express that line must be highway= but node may be tagless? )

data/presets/type/site/traffic_signals.json
```diff
@@ -0,0 +1,44 @@
+{
```
matkoniecz: this relation is barely documented, lease do not add it yet

see https://wiki.openstreetmap.org/wiki/Special:WhatLinksHere/Tag:site%3Dtraffic_signals

data/presets/type/site/climbing/crag.json
```diff
@@ -26,5 +26,23 @@
     "aliases": [
         "Climbing Cliff"
     ],
-    "name": "Climbing Crag"
+    "name": "Climbing Crag",
```
matkoniecz, line 29: is there anywhere documentation for it?

https://wiki.openstreetmap.org/w/index.php?title=Tag:climbing%3Dcrag has nothing

data/presets/type/route.json
```diff
@@ -17,5 +17,215 @@
...
+                    "area"
+                ],
+                "matchTags": [
+                    {
+                        "public_transport": "platform",
+                        "$1": "yes"
+                    }
+                ]
+            },
+            {
```
matkoniecz, line 30: wait, is entry of elements mattering?

if no, then it probably should 

if yes, then maybe put it higher before this exotic stop variants to make editing easier?

data/presets/type/waterway.json
```diff
@@ -15,5 +15,52 @@
...
+                "role": "main_stream",
+                "roleLabel": "Main Stream",
+                "matchTags": [
+                    {
+                        "waterway": "*"
+                    }
+                ],
+                "geometry": [
+                    "line",
+                    "area"
```
matkoniecz: why area is allowed?

data/presets/type/route_master.json
k-yle: no longer relevant, removed all this complexity, see https://github.com/ideditor/schema-builder/pull/174#discussion_r2911019599

data/presets/type/site/climbing/crag.json
k-yle: it's documented on [Climbing#Crags](https://osm.wiki/Climbing#Crags). I know nothing about this topic, I just added it because of https://github.com/openstreetmap/id-tagging-schema/pull/1538#issuecomment-2954774423

data/presets/public_transport/stop_area.json
k-yle: done for `route` relations. For other relations like `stop_area`, [any junk](https://osm.wiki/Tag:public_transport=stop_area#Members) can be added with `"role": ""`, so there's no appropriate label.

data/presets/type/restriction.json
k-yle: sure, done

data/presets/type/waterway.json
k-yle: i don't remember why i added this 😅, removed now

data/presets/type/destination_sign.json
k-yle: hmm, that's not really possible. I'll just remove `matchTags`

data/presets/type/waterway.json
k-yle: https://github.com/openstreetmap/id-tagging-schema/pull/1538#discussion_r2911266802

data/presets/type/enforcement.json
k-yle: i have no idea. I guess we should trust the wiki and remove `matchTags` ?

data/presets/area.json
k-yle: > But is this `area` entry representing both closed ways with area tags and multipolygons?

yes, it's used by both 

> If yes, would adding this relations structure interfere with applying it to closed ways?

no, the code would just ignore this property for ways. Agree that it's a bit weird, but there's no other way to model it

data/presets/type/route.json
k-yle: the problem with putting `Exit only ` first, is that the list looks more chaotic. 

To me, it seems a bit better when everything starts with `Stop ...` 

image

data/presets/type/enforcement.json
matkoniecz: I guess it depends how this rules should be treated
- general claims about which relations are valid
- as guidance for editors

Because if it is guidance than being a bit prescriptive makes sense, but if violation of this rules should indicate invalidness of relation then they should be more relaxed and more matching what OSM Wiki permits and got used

data/presets/type/route.json
matkoniecz: is this mockup? or is there way to test it in iD somehow?

k-yle: There's a lot of half-complete features in my fork of iD, which I will eventually make PRs for. Maybe I'll open some PRs tonight, but TLDR: no this can't be tested yet

matkoniecz: sounds like one more reason to process PRs you already opened :)

matkoniecz (changes requested):
in data/presets/type/route.json I would put "Part of the route" higher, so if selected from some dropdown it will be higher on the list, above exotic stop types

data/presets/type/route.json
```diff
@@ -17,5 +17,194 @@
     "tags": {
         "type": "route"
     },
-    "name": "Route"
+    "name": "Route",
+    "relation": {
```
matkoniecz, line 21: type=route is also for hiking routes etc

maybe it would be better to put that definition into one of specific public transport routes and reference it from there

data/presets/type/boundary.json
```diff
@@ -10,5 +10,69 @@
...
+                ],
+                "matchTags": [
+                    {
+                        "place": "*"
+                    }
+                ],
+                "max": 1
+            },
+            {
+                "role": "subarea",
```
matkoniecz: do we want to promote and support these?

IIRC these are fairly bad ideas but I may misremember

matkoniecz: @k-yle Do you know what is the community consensus on those?

k-yle: the wiki says [it's questionable](https://osm.wiki/Relation:boundary#Relation_members), and some communities [are removing it](https://c.osm.org/t/122865/16), so lets remove it.

data/presets/type/route.json
k-yle: good idea, done

data/presets/type/route.json
k-yle: yes, the order matters for the dropdown list of options, just like fields. 

I've moved `forward` and `backwards`  and ` ` to the top now.

data/presets/type/enforcement.json
k-yle: as a first step, i would like to use this data for [translated role labels](https://github.com/openstreetmap/id-tagging-schema/pull/1538#discussion_r2911352599) in the UI. Some sort of [relation validator ](https://github.com/openstreetmap/iD/commit/cd10ed531b852f79ae5820aeff35108c1f2d6799) can be considered later.

So we don't need to list every possible role, since people can still enter custom values.

data/presets/type/waterway.json
```diff
@@ -15,5 +15,52 @@
...
+                ],
+                "geometry": [
+                    "line"
+                ]
+            },
+            {
+                "role": "main_stream",
+                "matchTags": [
+                    {
+                        "waterway": "*"
```
matkoniecz, line 43: maybe listing of viable ones would be preferred?

currently it will happily include `waterway=dam`

matkoniecz: (the same for blank role)

matkoniecz:
conflicts after #2576, so parts of it should be removed (or next part send to cherry picking)...

though maybe waiting for iD to implement some support would be nice? It would allow live testing

data/presets/type/boundary.json
```diff
@@ -10,5 +10,58 @@
...
     },
-    "name": "Boundary"
+    "name": "Boundary",
+    "relation": {
+        "reference": "boundary",
+        "allowDuplicateMembers": false,
+        "role_labels": {
+            "inner": "Inner",
+            "outer": "Outer",
+            "label": "Label",
```
1ec5: This is often misunderstood as a labeling hint, like mapping for the renderer. Really unfortunate keyword naming decision. In geography, this is called a “node” or “focal point”. Not many laypeople know these words, though they do appear [in school curricula](https://education.nationalgeographic.org/resource/geography-standard-5/). Of the two terms, I suspect “focal point” would be the more intuitive. If we have to resort to a circumlocution, maybe something like “equivalent point representation”?

data/presets/type/boundary.json
```diff
@@ -10,5 +10,58 @@
...
+            "admin_centre": "Admin Center"
```
1ec5: This value was named after the geographic concept of an [administrative center](https://en.wikipedia.org/wiki/Administrative_centre) – in other words, a capital or seat of government. Unfortunately, many mappers mistake it to mean either “geographic **center** of the **admin**istrative boundary” or “**center** where the **admin**istration is headquartered” (that is, the capitol). I’d favor calling this “Capital / Seat” to avoid this confusion.

data/presets/type/boundary.json
```diff
@@ -10,5 +10,58 @@
...
         "type": "boundary"
     },
-    "name": "Boundary"
+    "name": "Boundary",
+    "relation": {
+        "reference": "boundary",
+        "allowDuplicateMembers": false,
+        "role_labels": {
+            "inner": "Inner",
+            "outer": "Outer",
```
1ec5: Originally, these were `enclave` and `exclave`, but they got [renamed](https://wiki.openstreetmap.org/wiki/Special:Diff/967662) for consistency with `multipolygon` relations. Maybe this would be an opportunity to bring back “Enclave” as a term that laypeople are familiar with. But `outer` isn’t always an exclave, so I don’t know if that would work.

data/presets/area.json
```diff
@@ -12,5 +12,30 @@
...
+        "reference": "multipolygon",
+        "allowDuplicateMembers": false,
+        "role_labels": {
+            "inner": "Inner",
+            "outer": "Outer"
```
1ec5: “Inner Ring” and “Outer Ring”?

data/presets/type/route.json
```diff
@@ -17,5 +17,34 @@
...
+        "reference": "route",
+        "allowDuplicateMembers": true,
+        "role_labels": {
+            "": "Part of the route",
```
1ec5: How about “Route Segment”? It should be title case anyhow.

data/presets/type/route.json
```diff
@@ -17,5 +17,34 @@
...
+            "forward": "Forwards",
+            "backward": "Backwards"
```
1ec5, line 27: This is going to be confusing unless the user understands way directions. iD’s `oneway=*` field abstracts this complexity by calling both one-way and providing a button to flip the way around. Ideally, iD would consider the route relation’s `direction=*` tag and override these labels to e.g. “Northbound Only” and “Southbound Only” as appropriate.

data/presets/type/route/bus.json
```diff
@@ -17,5 +17,196 @@
...
     },
-    "name": "Bus Route"
+    "name": "Bus Route",
+    "relation": {
+        "reference": "route",
+        "allowDuplicateMembers": true,
+        "role_labels": {
+            "stop": "Stop",
+            "stop_exit_only": "Stop (Exit Only)",
+            "stop_entry_only": "Stop (Entry Only)",
```
1ec5, line 27: Is it just me, or do terms like “Enter” and “Exit” feel weird on a bus route? These are words that only ever occur on the vehicle, rather than the stop. Here in the San Francisco Bay Area, all the relevant signs at stops would say “Pick-Up Only” for entry and “Drop-Off Only” for exit. For the rail presets that are currently based on this preset, “Boarding Only” and “Deboarding Only” (or “Alighting Only”) would sound slightly more natural.

data/presets/type/waterway.json
```diff
@@ -15,5 +15,52 @@
...
     },
-    "name": "Waterway"
+    "name": "Waterway",
+    "relation": {
+        "reference": "waterway",
+        "allowDuplicateMembers": false,
+        "role_labels": {
+            "": "",
+            "main_stream": "Main Stream",
+            "spring": "Spring"
```
1ec5, line 25: Is the wiki correct that this role is only for a `natural=spring`? Or do we also tag it on other kinds of headwaters?

data/presets/type/boundary.json
matkoniecz: maybe make an issue then for this? Or new PR changing it? that was merged in #2576 and will be released soon

I admit that I see no difference between these descriptions

data/presets/type/boundary.json
matkoniecz: maybe make an issue then for this? Or new PR changing it? that was merged in #2576 and will be released soon

I admit that I see no difference between these descriptions

data/presets/type/boundary.json
matkoniecz: maybe make an issue then for this? Or new PR changing it? that was merged in https://github.com/openstreetmap/id-tagging-schema/pull/2576 and will be released soon

data/presets/area.json
matkoniecz: maybe make an issue then for this? Or new PR changing it? that was merged in https://github.com/openstreetmap/id-tagging-schema/pull/2576 and will be released soon

BTW, my initial impression is that I like this change

@k-yle what you think about this one?

data/presets/type/route/bus.json
matkoniecz: Exit/Entry sounds more natural to me, but I am not a native speaker

data/presets/type/waterway.json
matkoniecz: I guess that natural=hot_spring may be possible?

data/presets/type/boundary.json
1ec5: Oh, are we not expecting to land this PR soon? If so, I can PR a more targeted change.

data/presets/type/boundary.json
1ec5: In other words, `label` is for a `place=city`/`town`/`village`/`hamlet` point representing the most important point in that place, where the place “begins” and spreads out from. `admin_centre` is the `place=city`/`town`/`village`/`Hamlet` where the government is headquartered.

Neither role is for a building, even if its official name ends with “Administrative Center”. In English, we often describe it as the difference between a capit**a**l and a capit**o**l.

(In theory, `label` would only occur on a boundary that corresponds to a human settlement, while `admin_centre` would only occur on a broader administrative area that doesn’t correspond to a human settlement. However, this is far from the case in practice, because of ongoing confusion and data consumers relying too heavily on `label`.)

data/presets/type/boundary.json
matkoniecz: This comment is on part already removed here,  so may be lost

Also not sure when this one will be merged - but it is possible that new version of id tagging scheme will be released,  iD changed to support relation data that will be in turn used to changes here

Or maybe this one will be considered as ready earlier

data/presets/type/route/bus.json
k-yle: i quite like _Pick-Up Only_ and _Drop-Off Only_. [in NZ](https://www.rnz.co.nz/news/education/272034/'keeping-it-simple'-while-avoiding-cliches), a[nd AU](https://archive.md/LYCHJ) the word _alighting_ is being phased out, since many people don't understand it. although singapore [uses it a lot](https://i0.wp.com/spjg.com/wp-content/uploads/2017/01/no-alighting.jpg?w=825&ssl=1).

data/presets/type/waterway.json
k-yle: `hot_spring` is also allowed (see line 57 below). i did not bother adding [`seamark:type=spring`](https://taginfo.osm.org/tags/seamark:type=spring) for obvious reasons.

i'm not aware of any other tags for a headwater, unless we're counting tags like `manhole=drain` and `outlet=*`

data/presets/type/route.json
k-yle: that would be quite a neat feature. even if we do that, we'll still need some sort of label for other data consumers.

is this even worse?

```suggestion
            "forward": "Route Segment (Forward)",
            "backward": "Route Segment (Backward)"
```

data/presets/type/route.json
k-yle: sounds good, done

data/presets/area.json
k-yle: sounds good to me. done in this PR

data/presets/type/route.json
1ec5: That would be an improvement. Maybe we could further clarify the association with the way‘s direction using something like “Same Direction” and “Opposite Direction”. But we do use the forward/backward terminology in the Direction field, so I guess it’s a broader issue to address at some point.

data/presets/type/waterway.json
1ec5: I’m assuming that some headwaters don’t have any tags (or any that iD would recognize, anyway). Sometimes it’s just where enough rainwater collects from the surrounding hillsides to form a stream. I’m unsure if it qualifies for the `spring` role; that’s what I was double-checking.

data/presets/type/boundary.json
1ec5: #2622

data/presets/type/boundary.json
1ec5: #2622

data/presets/type/boundary.json
1ec5: I don’t have a strong opinion on this. Consistency with `multipolygon` is important. I just wanted to start a discussion about it in case you had any ideas.

data/presets/type/route/bus.json
1ec5: Yeah, my local transit agencies use “alighting” in public communications, but only in context. In plain American English, “drop-off” is more common when describing a bus stop and “deboarding” is more common when describing a train or ferry stop.

matkoniecz:
now that v7.1.0 is released iD can start using first definitions merged into a separate PR

BTW, https://github.com/openstreetmap/id-tagging-schema/issues/2622 should be looked at

data/presets/type/waterway.json
matkoniecz: if something is not a spring it should presumably not get spring role

data/presets/type/boundary.json
1ec5: Looks like this branch now fixes this half of #2622. 👍

data/presets/type/route/ferry.json
```diff
@@ -17,6 +17,8 @@
         "key": "route",
         "value": "ferry"
     },
+    "name": "Ferry Route",
+    "relationCrossReference": "{type/route/bus}"
```
matkoniecz:
```suggestion
    "relationCrossReference": "{type/route/bus}",
```

sorry, tried to be helpful and messed up merge (stupid lack of tailing comma in json - I am tempted to propose moving to json5 for source code and only emit regular json)

scripts/lib/build.ts
```diff
@@ -451,7 +451,6 @@ function generatePresets(
     if (preset.relationCrossReference) {
       references.presets[id] ||= {};
       references.presets[id].relation = preset.relationCrossReference;
-      delete preset.relationCrossReference;
```
matkoniecz, line 454: is it intentional here?

data/presets/area.json
```diff
@@ -17,8 +17,8 @@
         "reference": "multipolygon",
         "allowDuplicateMembers": false,
         "role_labels": {
-            "inner": "Inner",
```
matkoniecz, line 20: that sounds annoyingly confusing when it is only part of ring an inner/outer is built of many segments...

@1ec5

scripts/lib/build.ts
k-yle: yeah, that's actually a bug. but we didn't notice it, because this code was written before the schema-builder repo merge

k-yle:
this can now be tested from https://pr-12768--ideditor.netlify.app/, see https://github.com/openstreetmap/iD/pull/12768 for details

it will not work for NSI presets until https://github.com/osmlab/name-suggestion-index/pull/12522 is released

data/presets/area.json
1ec5: We could make it even more specific, like “Inner Ring Segment” or “Inner Edge”.

iD’s validator rule has the suggestions “Set as inner” and “Set as outer”. I noticed some languages are translating these strings as more than just “inner”, because there’s no such thing as “an inner”.

data/presets/type/boundary.json
k-yle: i'm closing this thread since the discussion moved to https://github.com/openstreetmap/id-tagging-schema/pull/1538#discussion_r3748244497

data/presets/type/site/climbing/crag.json
k-yle: i'm trying to figure out which of the 87 comments are still outstanding. is there anything i need to do here?

data/presets/type/waterway.json
k-yle: can we assume that the wiki is correct and mark this discussion as resolved?

data/presets/area.json
k-yle: "Inner Edge" / "Outer Edge" seems nice imo. and it avoids the need for the word "Segment"

matkoniecz:
can you add https://github.com/openstreetmap/iD/issues/5329 to what will be closed by it?

data/presets/type/waterway.json
matkoniecz: resolving now, even if wiki is wrong this is fixable detail

(and in this case wiki is not suspect, and while wiki may be wrong in this case we have no good reason to suspect this)

if someone thinks wiki is wrong - open discussion and in 100% clear cases edit wiki (on your own responsibility)

matkoniecz (changes requested):
https://github.com/openstreetmap/id-tagging-schema/pull/1538#pullrequestreview-4756923880

data/presets/type/site/climbing/crag.json
matkoniecz: I posted review with link to one at least, I am also reviewing this again and PR in general

matkoniecz: I have problems with relation but it in presets already, so it is not something to be fixed in this PR

see https://wiki.openstreetmap.org/wiki/Talk:Climbing#%22You_can_add_other_objects_to_the_relation_(parking,_cliff,_peak,_shelter_etc.)%22

data/presets/area.json
```diff
@@ -17,8 +17,8 @@
...
-            "outer": "Outer"
+            "inner": "Inner Ring",
```
matkoniecz:
```suggestion
            "inner": "Inner Edge",
```

follow idea from review, idea by @1ec5

data/presets/area.json
```diff
@@ -17,8 +17,8 @@
...
+            "outer": "Outer Ring"
```
matkoniecz:
```suggestion
            "outer": "Outer Edge"
```

follow idea from review, idea by @1ec5

data/presets/type/waterway.json
matkoniecz: so valid values would be:

stream, ditch, river, drain, canal, flowline

maybe tidal_channel, wadi, derelict_canal

data/presets/type/route.json
matkoniecz: > is this even worse?

seems better, I would apply it and consider it resolved

data/presets/type/restriction.json
```diff
@@ -10,5 +10,50 @@
...
+        },
+        "members": [
+            {
+                "role": "from",
+                "geometry": [
+                    "line"
+                ],
+                "matchTags": [
+                    {
+                        "highway": "*"
```
matkoniecz, line 30: maybe list valid values so it does not get bus stops etc? Though list can be fairly long here...

data/presets/type/enforcement.json
```diff
@@ -9,5 +9,76 @@
...
+            },
+            {
+                "role": "to",
+                "geometry": [
+                    "vertex",
+                    "line"
+                ],
+                "matchTags": [
+                    {
+                        "highway": "*"
```
matkoniecz, line 72: maybe list valid values so it does not get bus stops etc? Though list can be fairly long here...

data/presets/type/boundary.json
```diff
@@ -10,5 +10,30 @@
...
     },
-    "name": "Boundary"
+    "name": "Boundary",
+    "relation": {
+        "reference": "boundary",
+        "allowDuplicateMembers": false,
+        "role_labels": {
+            "inner": "Inner Ring",
+            "outer": "Outer Ring"
+        },
```
matkoniecz, line 20:
```suggestion
            "inner": "Inner Edge",
            "outer": "Outer Edge"
```

also here?

data/presets/type/route/bus.json
```diff
@@ -17,5 +17,196 @@
...
+                "role": "platform_on_demand_entry_only",
+                "geometry": [
+                    "point",
+                    "vertex",
+                    "line",
+                    "area"
+                ],
+                "matchTags": [
+                    {
+                        "public_transport": "platform"
```
matkoniecz, line 221: is it intentional to to not list `highway=bus_stop` ?

data/presets/type/route/tram.json
```diff
@@ -17,5 +17,6 @@
         "key": "route",
         "value": "tram"
     },
-    "name": "Tram Route"
+    "name": "Tram Route",
+    "relationCrossReference": "{type/route/bus}"
```
matkoniecz, line 21: I understand that we want to avoid duplicating stuff, but it would be nice to limit it to actual tram stops...

Though that may limit things too much if these are not fully tagged yet?

Either way can be left probably for future

data/presets/type/boundary/administrative.json
```diff
@@ -23,10 +23,10 @@
         "reference": "boundary",
         "allowDuplicateMembers": false,
         "role_labels": {
-            "inner": "Inner",
-            "outer": "Outer",
+            "inner": "Inner Ring",
+            "outer": "Outer Ring",
```
matkoniecz:
```suggestion
            "inner": "Inner Edge",
            "outer": "Outer Edge",
```

matkoniecz: also here

data/presets/type/boundary/administrative.json
```diff
@@ -23,10 +23,10 @@
...
             "label": "Label",
-            "admin_centre": "Admin Center"
+            "admin_centre": "Capital / Seat"
```
matkoniecz, line 29: presumably seat being confusing is just me due to being nonnative speaker?

(I remember usage only in church context, as in "seat of a bishop")

matkoniecz:
> this can now be tested from https://pr-12768--ideditor.netlify.app/, see https://github.com/openstreetmap/iD/pull/12768 for details

though it only shows labels and is not yet doing any smart filtering

data/presets/type/waterway.json
k-yle: i would suggest keeping it flexible for now and making it stricter later. 

i can't even reply to threads eaily, there are so many comments it crashes github 😭

data/presets/type/route.json
k-yle: donedoneddone

data/presets/area.json
k-yle: done

data/presets/area.json
k-yle: duplicate of some other thread

data/presets/area.json
k-yle: also duplicate thread

data/presets/type/enforcement.json
k-yle: duplicate of https://github.com/openstreetmap/id-tagging-schema/pull/1538#discussion_r3967772930

data/presets/type/boundary.json
k-yle: closing duplicate threads

data/presets/type/boundary/administrative.json
k-yle: dup

data/presets/type/route.json
```diff
@@ -17,5 +17,34 @@
     "tags": {
         "type": "route"
     },
-    "name": "Route"
+    "name": "Route",
+    "relation": {
```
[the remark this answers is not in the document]
k-yle: maybe this is out of date, but we already fixed this.

data/presets/type/route/bus.json
k-yle: fixed

data/presets/type/route/bus.json
k-yle: done

k-yle:
sorry for the delayed reply. there are so many code reviews that i can't even open this PR from my personal laptop, github just crashes 😭

listing the last 3 unresolve threads:
- [do we add lots more hardcoding in this PR?](https://github.com/openstreetmap/id-tagging-schema/pull/1538#discussion_r3632318364) 
- [how to we label `admin_centre`?](https://github.com/openstreetmap/id-tagging-schema/pull/1538#discussion_r3845315800) 
- [how to we label `label`?](https://github.com/openstreetmap/id-tagging-schema/pull/1538#discussion_r3661318002)

data/presets/type/boundary/administrative.json
k-yle: this is an american term, needs an american english speaker to reply. maybe _"Capital / Administrative Seat"_ is better?

data/presets/type/waterway.json
matkoniecz: > i can't even reply to threads eaily, there are so many comments it crashes github 😭

maybe close this PR and open new one, linking back old one?

If this is handled for now (I will open issue when merging this PR) and similar for `highway=` will be presumable also skipped for now, I think I do not have any suggestions/requests from me left

data/presets/type/boundary/administrative.json
matkoniecz: for me it is, but I am not from USA

summoning @1ec5

matkoniecz:
heh, bot also seems to be overwhelmed by amount of comments

matkoniecz:
> https://github.com/openstreetmap/id-tagging-schema/pull/1538#discussion_r3632318364 ( do we add lots more hardcoding in this PR? )

this one is resolved - it was only idea, and it got rejected/postponed so it is now handled

> https://github.com/openstreetmap/id-tagging-schema/pull/1538#discussion_r3845315800  how to we label admin_centre?
> https://github.com/openstreetmap/id-tagging-schema/pull/1538#discussion_r3661318002 how to we label label?

this PR is not making it worse as it got merged already and is tracked by https://github.com/openstreetmap/id-tagging-schema/issues/2622

matkoniecz:
@k-yle can you take look at https://github.com/openstreetmap/id-tagging-schema/pull/2818 ?

If you think it is a good idea it could be nice to merge it before this PR is merged to see it in action (or merge this PR, merge that one and it also should trigger)

data/presets/type/restriction/only_u_turn.json
```diff
@@ -14,5 +14,6 @@
         "key": "restriction",
         "value": "only_u_turn"
     },
-    "name": "U-Turn Only"
+    "name": "U-Turn Only",
+    "relationCrossReference": "{type/restriction}"
```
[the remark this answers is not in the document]
matkoniecz: U-Turn Only sounds kind of insane and U-Turn Only leading into itself even weirder, but LLM sounds right here?

verified it and LLM convinced me

data/presets/type/waterway.json
```diff
@@ -15,5 +15,52 @@
...
+                    {
+                        "natural": "spring"
+                    },
+                    {
+                        "natural": "hot_spring"
+                    }
+                ],
+                "geometry": [
+                    "point"
+                ]
```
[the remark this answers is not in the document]
matkoniecz: welp, this makes sense

maybe even dropping point would make sense?

data/presets/public_transport/stop_area.json
```diff
@@ -17,5 +17,69 @@
...
+            {
+                "role": "subway_entrance",
+                "matchTags": [
+                    {
+                        "railway": "subway_entrance"
+                    }
+                ],
+                "geometry": [
+                    "vertex"
+                ]
```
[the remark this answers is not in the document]
matkoniecz: also seems right?

matkoniecz (changes requested):
LLM seems to found nonhallucinated bugs

(if this is right then after fixing those requesting another round would make sense)

data/presets/public_transport/stop_area.json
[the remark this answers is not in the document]
k-yle: imo the preset is wrong, `subway_entrance` must be a vertex, just like `entrance`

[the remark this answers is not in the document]
matkoniecz: lets offload it into separate issue

data/presets/type/restriction/only_u_turn.json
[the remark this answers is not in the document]
k-yle: good job 🤖, yeah this is correct

data/presets/type/waterway.json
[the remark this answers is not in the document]
k-yle: fixed

data/presets/public_transport/stop_area.json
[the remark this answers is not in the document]
matkoniecz: opened https://github.com/openstreetmap/id-tagging-schema/issues/2841

matkoniecz:
opened https://github.com/openstreetmap/id-tagging-schema/issues/2841 to offload discussion rather than block this PR

matkoniecz:
created https://github.com/openstreetmap/id-tagging-schema/issues/2842 to offload more discussion/analysis/planning

data/presets/type/route/bus.json
```diff
@@ -17,5 +17,214 @@
...
         "value": "bus"
     },
-    "name": "Bus Route"
+    "name": "Bus Route",
+    "relation": {
+        "reference": "route",
+        "allowDuplicateMembers": true,
+        "role_labels": {
+            "stop": "Stop",
+            "stop_exit_only": "Stop (Drop-Off Only)",
```
matkoniecz, line 26:
```suggestion
            "stop_exit_only": "Stop (drop-off only)",
```

or capitalise "Stop (only stops on request)" ?

matkoniecz: (few more Off and similar are later)

data/presets/type/route/bus.json
```diff
@@ -17,5 +17,214 @@
...
+            "stop_entry_only": "Stop (Pick-Up Only)",
+            "stop_on_demand": "Stop (only stops on request)",
+            "stop_on_demand_exit_only": "Stop (Drop-Off Only) (only stops on request)",
+            "stop_on_demand_entry_only": "Stop (Pick-Up Only) (only stops on request)",
```
matkoniecz, line 30: is this value documented anywhere on wiki? I do not see it at https://wiki.openstreetmap.org/wiki/Tag:route%3Dbus or https://wiki.openstreetmap.org/wiki/Buses#Bus_stops_and_stations or https://wiki.openstreetmap.org/wiki/Proposal:Public_Transport

matkoniecz: (the same may apply for other complex ones)

matkoniecz:
@k-yle 

I hate it but I spotted some cases where taking extra look may make sense. If this R explodes your computer feel free to close it and push the same code to new one

data/presets/type/route/bus.json
matkoniecz: or is it fine to have such mismatch?

data/presets/type/enforcement.json
```diff
@@ -9,5 +9,76 @@
...
         "type": "enforcement"
     },
-    "name": "Enforcement"
+    "name": "Enforcement",
+    "relation": {
+        "reference": "enforcement",
+        "allowDuplicateMembers": false,
+        "role_labels": {
+            "device": "Device",
+            "force": "Force",
```
matkoniecz, line 18:
```suggestion
            "device": "Measuring Device",
            "force": "Enforcement Device",
```

(?)

At least for me "Force" is not really clear

data/presets/type/restriction.json
matkoniecz: https://github.com/openstreetmap/id-tagging-schema/issues/2842

data/presets/type/waterway.json
matkoniecz: https://github.com/openstreetmap/id-tagging-schema/issues/2842

matkoniecz (changes requested):
can you take look at (even more) suggestions that somehow I had only now?

data/presets/type/enforcement.json
matkoniecz: @1ec5 what you think about these labels?

data/presets/type/route/bus.json
matkoniecz: @1ec5 what you think about capitalization of these labels?

matkoniecz:
@k-yle Can you merge in latest changes from master? I am confused why https://github.com/openstreetmap/id-tagging-schema/pull/2847 bug is not visible here - either I am misunderstanding code, or bug was introduced after this branch diverged (and we may have other ones like this)

data/presets/type/route/bus.json
1ec5: Better to put everything in title case or sentence case consistently. It looks weird if we suddenly lowercase everything inside parentheses. Title case is the norm for any UI element where we tend to have short strings.

data/presets/type/enforcement.json
1ec5: Good catch. These roles have always been particularly unintuitive, to the point of hampering adoption of the entire tagging scheme.

[The examples](https://wiki.openstreetmap.org/wiki/Relation:enforcement#Examples) suggest that `device` has multiple functions depending on the situation. It can be a traffic control device like a traffic light that the enforcement relation reinforces, or it can be an enforcement device like a speed camera or truck scale that enforces a signposted restriction. In the latter case, `force` is an additional traffic control device to reinforce _that_ device.

Maybe we need to split up the preset by `enforcement=*` so that each specialized preset can label these two roles differently and enforce (heh) different feature tags on them. Until we take that step, the best I can think of is “Regulatory Device” for `device` and “Reinforcement Device” for `force`.

matkoniecz: > “Regulatory Device” for device and “Reinforcement Device” for force.

This does not really explains much for me. Maybe dropping of adding schema here and leaving it for later would make sense?

matkoniecz:
@k-yle - re https://github.com/openstreetmap/id-tagging-schema/pull/1538#pullrequestreview-5154142024 - would making PR changing this PR and fixing this part would help?

k-yle:
i think i'll just abandon this and make some smaller PRs

data/presets/type/boundary/administrative.json
1ec5: The full term is “seat of government”, but mappers might take that too literally and tag the legislative chamber, or even an `amenity=bench` within it. 😛 The other term we use in the U.S. is “county seat”, but obviously that only applies to counties.

“Capital city” by itself would already be quite good, but for some reason we never call a county seat a capital, and even some state capitals are mere towns.

matkoniecz: @1ec5 copied reply to now active https://github.com/openstreetmap/id-tagging-schema/pull/2923#discussion_r4063311131

PR #1539, opened 2025-05-07
Exclude `stop` field for stop signs in Bulgaria, as all are `stop=minor`

Dimitar5555:

All stop signs in Bulgaria are `stop=minor`. The literal meaning of the stop sign according to local law is "Stop! Give way to the priority road" so "all ways stop" should be an impossible situation.

### Related issues

none

### Links and data

**Relevant OSM Wiki links:**
n/a

**Relevant tag usage stats:**
There were 14 uses in Bulgaria of which 11 tagged as `minor` (some of them being on residential crossroads, go figure which one is "minor"), 1 incorrectly tagged as `all` and 1 of each `forward` and `backward`. The highest usage was around 35 uses in Jan 2022.

image

https://taginfo.geofabrik.de/europe:bulgaria/keys/stop

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>


Dimitar5555:
From what I've seen stop signs are rarely mapped at the intersection node. Does this field make sense when the sign is not at the intersection?

> Are you sure stop signs should not be tagged `stop=minor` under any circumstance?

Legally there shouldn't be such cases. In practice officials don't always follow the law and they might create `stop=all` junction...

Dimitar5555:
Sidebar screenshot:
https://www.openstreetmap.org/node/7899835401
image

1ec5:
> From what I've seen stop signs are rarely mapped at the intersection node. Does this field make sense when the sign is not at the intersection?

[The documented tagging scheme](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dstop#All-way_stop) is counterintuitive: effectively, an all-way stop should be tagged as `stop=all` on the intersection node, but a two-way stop should be tagged as `stop=minor` on each stopline. Implicitly, a data consumer can infer that any intersection node is `stop=all` and any other node is `stop=minor`, unless otherwise specified. This appears to be how routers are treating the key, though renderers may be doing something different. (Regardless of the documentation, some mappers are mapping all stop signs at the stoplines for consistency.)

> Legally there shouldn't be such cases. In practice officials don't always follow the law and they might create `stop=all` junction...

That sounds similar to the situation with `highway=give_way` in some jurisdictions. What does it even mean for everyone to give way to everyone else? 😵‍💫

Dimitar5555:
> > All stop signs in Bulgaria are `stop=minor`.
> 
> Should be a separate `highway=stop` preset for Bulgaria (and maybe other countries?) that automatically adds `stop=minor`?

Probably no. Navigation software usually knows where it is driving and should take the national law into consideration. Just like it knows that `maxspeed:type=bg:urban` is 50 km/h.

tordans:
It sounds to me that at this point of research the PR does what it is supposed to do, right?
Does anyone see a reason to not merge it as is?

matkoniecz:
> The bottom line is that this PR will make it more difficult for mappers to add stop=all/minor or even know when it’s already present, so you should really be sure this is just hypothetical and not a realistic scenario:

on the other hand it will make harder for mappers to become confused when they get nonsense (from their perspective) field

I am inclined to accept idea behind this PR

---------------------------------------

note: stop= is rare there ( https://overpass-turbo.eu/s/2tFu ) and when mapped it is always `stop=minor`, see https://overpass-turbo.eu/s/2tFw

matkoniecz:

tried to use online tool, but Microsoft vibecoded something again and it fails silently with 422 error in logs

EDIT: now it worked without explanation

matkoniecz:
> @[tordans](https://github.com/tordans) tordans added[ waitfor-consensus ](https://github.com/openstreetmap/id-tagging-schema/issues?q=state%3Aopen%20label%3Awaitfor-consensus)preset labels [on May 23, 2025](https://github.com/openstreetmap/id-tagging-schema/pull/1539#event-17783015047)

@tordans do you think we need consensus among maintainers, mappers in Bulgaria or global community or some other group?

tordans:
> > @[tordans](https://github.com/tordans) tordans added[ waitfor-consensus ](https://github.com/openstreetmap/id-tagging-schema/issues?q=state%3Aopen%20label%3Awaitfor-consensus)preset labels [on May 23, 2025](https://github.com/openstreetmap/id-tagging-schema/pull/1539#event-17783015047)
> 
> 
> 
> @tordans do you think we need consensus among maintainers, mappers in Bulgaria or global community or some other group?

@matkoniecz no objections to merging this as is reading the discussion. We agreed a while back that country specific presets/fields are ok (unless complexity explodes) and from what I understand this change is OK to Good.

matkoniecz:
And AFAIK country-specific fields are much less harmful than country-specific presets.

matkoniecz (approved):

PR #1552, opened 2025-05-20
Add roof shape and roof height fields to building presets

trs998:

https://github.com/openstreetmap/iD/issues/1829 - roof type.  This adds roof shape and roof height, as this currently required manual entry of the tags in iD. Data would be more consistent with presets.

### Related issues

https://github.com/openstreetmap/iD/issues/1829 - opened back when tags were internal to iD

### Links and data

(https://wiki.openstreetmap.org/wiki/Key:roof:shape)
- …

**Relevant tag usage stats:**
> …
Roof shape is tagged 8.3 million times - https://taginfo.openstreetmap.org/keys/roof%3Ashape

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/data/fields/roof/height.json
+++ b/data/fields/roof/height.json
@@ -2,5 +2,5 @@
     "key": "roof:height",
     "minValue": 0,
     "type": "number",
-    "label": "Roof height from facade to top"
+    "label": "Roof height"
 }
--- a/data/fields/roof/height.json
+++ b/data/fields/roof/height.json
@@ -1,0 +1,6 @@
+{
+    "key": "roof:height",
+    "minValue": 0,
+    "type": "number",
+    "label": "Roof Height from facade to top"
+}
--- a/data/fields/roof/height.json
+++ b/data/fields/roof/height.json
@@ -2,5 +2,5 @@
     "key": "roof:height",
     "minValue": 0,
     "type": "number",
-    "label": "Roof height"
+    "label": "Roof Height"
 }
--- a/data/fields/roof/height.json
+++ b/data/fields/roof/height.json
@@ -2,5 +2,5 @@
     "key": "roof:height",
     "minValue": 0,
     "type": "number",
-    "label": "Roof Height from facade to top"
+    "label": "Roof height from facade to top"
 }
--- a/data/fields/roof/height.json
+++ b/data/fields/roof/height.json
@@ -2,5 +2,5 @@
     "key": "roof:height",
     "minValue": 0,
     "type": "number",
-    "label": "Roof Height"
+    "label": "Roof Height (Meters)"
 }
--- a/data/fields/roof/shape.json
+++ b/data/fields/roof/shape.json
@@ -1,7 +1,7 @@
 {
     "key": "roof:shape",
     "type": "combo",
-    "label": "Roof shape",
+    "label": "Roof Shape",
     "strings": {
         "options": {
             "flat": "Flat",
--- a/data/fields/roof/shape.json
+++ b/data/fields/roof/shape.json
@@ -15,7 +15,6 @@
             "cone": "Cone",
             "dome": "Dome",
             "onion": "Onion",
-            "saltbox": "Saltbox",
             "round": "Round"
         }
     },
--- a/data/fields/roof/shape.json
+++ b/data/fields/roof/shape.json
@@ -1,0 +1,30 @@
+{
+    "key": "roof:shape",
+    "type": "combo",
+    "label": "Shape of building (part) roof",
+    "strings": {
+        "options": {
+            "flat": "Flat",
+            "gabled": "Gabled",
+            "gabled_height_moved": "Gabled Height Moved",
+            "skillion": "Skillion",
+            "hipped": "Hipped",
+            "half-hipped": "Half-Hipped",
+            "side_hipped": "Side Hipped",
+            "side_half-hipped": "Side Half-Hipped",
+            "hipped-and-gabled": "Hipped and Gabled",
+            "mansard": "Mansard",
+            "gambrel": "Gambrel",
+            "pyramidal": "Pyramidal",
+            "crosspitched": "Crosspitched",
+            "sawtooth": "Sawtooth",
+            "butterfly": "Butterfly",
+            "cone": "Cone",
+            "dome": "Dome",
+            "onion": "Onion",
+            "round": "Round"
+        }
+    },
+    "autoSuggestions": false,
+    "customValues": false
+}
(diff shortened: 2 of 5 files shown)


tordanik:
As a 3D renderer developer, I would prefer if the list of `roof:shape` values was restricted to those which are also part of [Simple 3D Buildings](https://wiki.openstreetmap.org/wiki/Simple_3D_Buildings) . Some of the extra values (e.g. `gabled_height_moved`) have unclear semantics when combined with other tags such as `roof:direction`, and pretty much all of them they were added to the wiki without first forming a consensus with the wider community or with data consumers.

tordans:
> I would prefer if the list of roof:shape values was restricted to those which are also part of [Simple 3D Buildings](https://wiki.openstreetmap.org/wiki/Simple_3D_Buildings)

We have a couple of options here.
My suggestion would be to use…
- only translate some values via the `strings` property
- set     `"autoSuggestions": true,` to expose other values

iD will then mix both the translated and raw values which gives users a way to see which are recommended and better documented.

I would also suggest to improve the Wikidata items (see "Info-i" section in the Testing recommendations) for the values to either signal that they are well or baldy defined.

In general, we don't want to take a strong stance on tagging inside the Repo but rather ask to resolve those discussions in the Forum and via Proposals. It sounds like the tags need to be documented better and maybe even deprecated.

trs998:
I have revised the commit - I agree the odder roof shapes don't have clear 3d rendering and don't seem used much - so I have shortened the list into the one on simple 3d buildings - as per todans' suggestion.

data/fields/roof/height.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "roof:height",
+    "minValue": 0,
+    "type": "number",
+    "label": "Roof Height from facade to top"
```
tordans, line 5:
```suggestion
    "label": "Roof height from facade to top"
```

or even just … because the definition definition is something that the info-i can handle.
```suggestion
    "label": "Roof height"
```

What is the unit here? We should add this field to https://github.com/ideditor/schema-builder/pull/198 right?

data/fields/roof/shape.json
```diff
@@ -0,0 +1,24 @@
+{
+    "key": "roof:shape",
+    "type": "combo",
+    "label": "Shape of building (part) roof",
```
tordans, line 4:
```suggestion
    "label": "Roof shape",
```

Having those fields consistent "Roof …" is better IMO

data/presets/building.json
```diff
@@ -19,6 +19,8 @@
         "not/name",
         "operator",
         "roof/colour",
+        "roof/shape",
+        "roof/height",
```
tordans, line 23: If you want to be fancy about it, you could extract those three fields into a template to share them, see https://github.com/openstreetmap/id-tagging-schema/tree/main/data/presets/%40templates

That will make it easier to not forget one later…

tordans (changes requested):
Thanks for the PR. I left a few comments. Please share the testing documentation once you feel this is ready for the next review.

tordans:
One more thing, those roof shape options would be a great case for field option icons. Some docs on that at https://github.com/ideditor/schema-builder?tab=readme-ov-file#icons
It is used on the crossing markings ATM for example. We can do this in a follow up, though, because I image getting the right Icons going will take a bit.

trs998:
I agree icons would be good. I am concerned people may put the height of the total building in Roof Height rather than specifically the height which is just the roof part, but I think the label field is used for the english version of the field name and doesn't have hints in it. And the documentation seems good, with illustrative pictures.

When I tried the preview, it did not show the new field definitions on a building with roof:shape=something - but it now does - see below screenshots. 

image
image

Existing feature example:
https://pr-1552--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=20.25/52.62865/1.29159&disable_features=boundaries&background=Bing&id=w1332031670
image

data/fields/roof/height.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "roof:height",
+    "minValue": 0,
+    "type": "number",
+    "label": "Roof height from facade to top"
```
tordans, line 5: TBH I think the "from facade to top" makes it less clear. The facade is the whole side, right? So does this say from ground to to …?. How can "Roof height" be miss-read?

data/fields/roof/shape.json
```diff
@@ -0,0 +1,24 @@
...
+            "skillion": "Skillion",
+            "hipped": "Hipped",
+            "half-hipped": "Half-Hipped",
+            "mansard": "Mansard",
+            "gambrel": "Gambrel",
+            "pyramidal": "Pyramidal",
+            "cone": "Cone",
+            "dome": "Dome",
+            "onion": "Onion",
+            "saltbox": "Saltbox",
```
tordans:
```suggestion
```

Lets remove this, it is not listed on https://wiki.openstreetmap.org/wiki/Key:roof:shape
> Inconsistent definitions of "saltbox" by some editors/users (through displayed images).

data/fields/roof/shape.json
```diff
@@ -0,0 +1,23 @@
+{
+    "key": "roof:shape",
+    "type": "combo",
+    "label": "Roof shape",
```
tordans:
```suggestion
    "label": "Roof Shape",
```

data/fields/roof/height.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "roof:height",
+    "minValue": 0,
+    "type": "number",
+    "label": "Roof height"
```
tordans:
```suggestion
    "label": "Roof Height"
```

tordans (changes requested):
Labels always start with capital letter; will apply change right away https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema%20%22label%22%3A&type=code

data/fields/roof/height.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "roof:height",
+    "minValue": 0,
+    "type": "number",
+    "label": "Roof Height"
```
tordans:
```suggestion
    "label": "Roof Height (Meters)"
```

tordans (approved):
I will merge this now.

Ich change a view things
- Use "Label Name" (not "Label name")
- Add "Roof Height (Meters)" to indicate the default unit (same as "height")
  - Update the https://wiki.openstreetmap.org/wiki/Item:Q1839 on this as well
- Remove one value ("saltbox") that was disputed in the Wiki

PR #1560, opened 2025-05-24
add `heritage=*` field

novolife:

The number of `heritage=*` used is 288,995 and the number of combinations with `historic=*` is 111,080, which is a very substantial number, and I think that heritage should have a field and can be filled in easily in historic. 

Since it is likely that not all historics have a heritage level, it would be fine to put it in moreFields.

Also I added the name field to historic, taginfo shows that 45.36% of `historic` have `name`.

Additionally I added some terms to the field `historic` and `memorial` for quick selection based on taginfo.

### Links and data

https://wiki.openstreetmap.org/wiki/Key:heritage

**Relevant tag usage stats:**

111,080

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/data/fields/heritage.json
+++ b/data/fields/heritage.json
@@ -1,0 +1,6 @@
+{
+    "key": "heritage",
+    "type": "number",
+    "minValue": 1,
+    "label": "Heritage Registration Level"
+}
--- a/data/presets/historic.json
+++ b/data/presets/historic.json
@@ -1,10 +1,12 @@
 {
     "icon": "temaki-ruins",
     "fields": [
+        "name",
         "historic",
         "inscription"
     ],
     "moreFields": [
+        "heritage",
         "gnis/feature_id-US"
     ],
     "geometry": [


novolife:
## Test-Documentation

https://pr-1560--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=17.00/14.65485/121.06466

ss1

ss3


### Search

ss2

ss4


### Info-`i`

ss5

### Wording

- [x] American English

data/fields/heritage.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "heritage",
+    "type": "number",
+    "minValue": 1,
+    "label": "Heritage Registration Level"
```
1ec5, line 5: I wonder if this will make any sense to someone who’s mapping heritage buildings but hasn’t encountered `admin_level=*` yet (since boundaries are hidden by default). Some heritage registries categorize listed structures into what the user might confuse with “registration levels”. But this field is about the heritage organization, not the structure. Since the `admin_level=*` field is called “Admin Level”, could we call this field “Heritage Register Admin Level”?

data/fields/historic.json
```diff
@@ -4,28 +4,46 @@
...
             "charcoal_pile": "Charcoal Pile",
             "church": "Church",
             "city_gate": "City Gate",
             "citywalls": "Citywall",
+            "farm": "Historic Farm",
             "fort": "Historic Fort",
+            "grave": "Historic Grave",
+            "hollow_way": "Hollow Way",
+            "house": "Historic House",
+            "heritage": "Historic Heritage",
```
1ec5, line 27: This is a bit redundant. American English doesn’t have a universal word for `historic=heritage`, since it depends on the specific register. But I guess something generic would be “Historic Resource” or “Listed Property”.

data/fields/memorial.json
```diff
@@ -1,20 +1,35 @@
 {
     "key": "memorial",
-    "type": "combo",
+    "type": "typeCombo",
```
1ec5, line 3: This will result in `memorial=yes` if unset. Is that intended?

data/presets/historic.json
```diff
@@ -1,10 +1,12 @@
 {
     "icon": "temaki-ruins",
     "fields": [
+        "name",
         "historic",
         "inscription"
     ],
     "moreFields": [
+        "heritage",
```
1ec5, line 9: `heritage=*` is usually tagged on buildings, so it would be useful to add it to those presets too.

data/fields/heritage.json
1ec5, line 1: If we’re going to add `heritage=*`, we should add `heritage:operator=*` and `heritage:ref=*` at the same time. Otherwise, we’d be incentivizing users to add only `heritage=*`, which could be meaningless without more details.

data/fields/memorial.json
novolife: Just find out that `memory=yes` is meaningless, sorry about that.

data/fields/heritage.json
novolife: `heritage:ref` looks different from what the wiki requires for `ref`, so I'm not sure it should be added.

As for `heritage:operator`, most of the options on the wiki are abbreviations, and if introduced I'm worried that it will make people who don't know the corresponding institution fill in something that doesn't meet the standards in the wiki.

But in taginfo, `heritage:operator` contains the full name of many institutions, the tag may not already follow the wiki. And as many countries don't give tag specifications in the wiki, I'm not so sure that introducing this field would bring more chaos to already chaotic data.

data/fields/historic.json
novolife: Maybe "heritage" should be remove since the wiki says 

> Use the preferred feature heritage=* instead.

data/presets/historic.json
novolife: But only 0.03% of the buildings are tagged heritage, it might be better to use the preset `Historic Building` and then fill the heritage field.

data/fields/heritage.json
1ec5: Yeah, the original idea was that `heritage:operator=*` would only contain acronyms, to save typing, and the wiki would have a decoder ring explaining what each acronym means. This system has fallen apart to some extent because there are a lot more operators than the tagging scheme’s creators imagined. We could provide strings for some of these values, though it’s a bit of a rabbit hole, so I would support deferring it to a separate PR.

I wasn’t aware that the [`heritage=*`](https://wiki.openstreetmap.org/wiki/Key:heritage) documentation these days only recommends <code>ref:<var>operator</var>=*</code> and not `heritage:ref=*`. The latter is still quite common (41,228 occurrences as of writing), but I guess the idea is that <code>ref:<var>operator</var>=*</code> would distinguish between multiple programs that list the same structure. This syntax also makes use of the acronyms, not the freeform values of `heritage:operator=*`.

This syntax would be rather annoying to implement. We’d need to define a <code>ref:<var>operator</var>=*</code> field for each major program that depends on a particular `heritage:operator=*` value, scoped to a particular region. Again, another rabbit hole to avoid for now. I’m just concerned that, if we don’t require anything beyond `heritage=*`, mappers might start using the key subjectively, similar to `tourism=attraction`.

data/fields/historic.json
```diff
@@ -4,28 +4,45 @@
     "label": "Type",
     "strings": {
         "options": {
+            "aircraft": "Aircraft",
             "archaeological_site": "Archaeological Site",
+            "battlefield": "Historic Battlefield",
```
matkoniecz: vast majority of battlefields are not mappable and it would be misleading to add this value - can it be dropped?

data/fields/historic.json
```diff
@@ -4,28 +4,45 @@
...
             "bomb_crater": "Bomb Crater",
             "boundary_stone": "Boundary Stone",
+            "bridge": "Historic Bridge",
             "building": "Historic Building",
+            "bunker": "Historic Bunker",
+            "cannon": "Historic Cannon",
             "castle": "Castle",
+            "cemetery": "Cemetery",
```
matkoniecz: I see that that `historic=bridge` got "Historic Bridge" label and `historic=cemetery` got "Cemetery" label.

What is the logic behind this?

data/fields/memorial.json
```diff
@@ -4,17 +4,32 @@
...
+            "bench": "Bench",
             "blue_plaque": "Blue Plaque",
+            "building": "Building",
             "bust": "Bust",
+            "cannon": "Cannon",
+            "column": "Column",
             "cross": "Memorial Cross",
             "ghost_bike": "Ghost Bike",
+            "grave": "Grave",
+            "hazard_memorial": "Hazard Memorial",
```
matkoniecz: where it is documented on wiki? what `memorial=hazard_memorial` means?

data/fields/memorial.json
```diff
@@ -4,17 +4,32 @@
...
+            "milestone": "Milestone",
             "obelisk": "Obelisk",
             "plaque": "Commemorative Plaque",
+            "railway_car": "Railway Car",
             "sculpture": "Sculpture",
             "statue": "Statue",
             "stele": "Stele",
             "stolperstein": "Stolperstein",
             "stone": "Stone",
+            "tank": "Tank",
```
matkoniecz: what is the difference between `historic=tank` and `historic=memorial memorial=tank`

(note also that this new values are very orthogonal with each other, partially with root cause in that `memorial=` is a mess of orthogonal values at different level of specificity - but maybe especially problematic ones like `immortality_tower` should not be promoted?)

matkoniecz (changes requested):
this PR is trying to do multiple things at once

maybe splitting it into a bit more granular parts would make sense?

novolife:
> this PR is trying to do multiple things at once
> 
> maybe splitting it into a bit more granular parts would make sense?

@matkoniecz You're right, I should do one thing in one PR, so I've removed the strings.

data/fields/heritage.json
matkoniecz: I have not reviewed it thoroughly but it looks like support for this other tags is also needed :(

matkoniecz: looking at it again: sadly, schema exists in such way that supporting only `heritage=` will result in problems

but implementing viable support for even viable minimum will be annoying/complex :(

Are you interested in expanding to such, much more complex, PR?

I opened https://github.com/openstreetmap/id-tagging-schema/issues/2374 for now to not forget about it (was not caught by https://github.com/openstreetmap/id-tagging-schema/issues/1641 before as it has own dedicated keys )

matkoniecz:
see https://github.com/openstreetmap/id-tagging-schema/pull/1560#discussion_r3472039515

matkoniecz:
If you would be interested in this, much much more complex task, let me know and I will reopen it

or submit a new PR

Issue #31, opened 2020-11-10
Preset for sport=climbing


mfbehrens99:
Hi,

it would be very useful if there was a preset for climbing. It should mainly be for nodes

The wiki article can be found here: https://wiki.openstreetmap.org/wiki/Climbing

Best, Michael

tyrasd:
I think the primary tag is simply `climbing` in this case. For example `climbing=area`, `climbing=route`, `climbing=crag` or `climbing=bolder`.

mfbehrens99:
If anybody can create the raw structure and an example I can implement the rest of the tags and keys

zbycz:
Hi, I summarized all climbing features in a table. 

1) as you can see, crag and boulder have both two ways of tagging which are equally popular. What is the proper procedure to include both in iD? Perhaps create two presets with same name, where one is searchable (prefered) and second not? And second references first's name?

2) Any suggestions to the preset names? I don't like `Climbing route - bottom` much, but as I am not a native speaker I can't say. Perhaps "start" could be better?

3) Can we also create a preset for the freestanding "sport=climbing"? It would help people evolve it to something more specific. 


|     preset name    |  tags                                      |  | types from wiki | # |
| ----------------------- | ------------------------------------------ | --   | --------  | -- |
| Climbing gym  (done)          | leisure = sport_center<br>sport = climbing |  | node               |  4k
| Rock climbing crag      | natural = cliff<br>sport = climbing        |  | node,way          | 7k
|                         | climbing = crag                            |  | node,relation | 3k
| Rock climbing boulder   | natural = stone<br>sport = climbing        |  | node,area         | 1k
|                         | climbing = boulder                         |  | node,relation | 1k
| Rock climbing area      | climbing = area                            |  | node,area,relation | 0k
| Climbing route          | climbing = route                           |  | node,way          | 1k
| Climbing route - bottom | climbing = route_bottom                    |  | node               | 4k
| Climbing route - top    | climbing = route_top                       |  | node               | 0k
|                      |                                            |  |      |      |          |
|   . . . fallback category:                    |                                            |  |      |      |          |
| Climbing site           | sport = climbing                           |  | any | 17k

Full table [here](https://docs.google.com/spreadsheets/d/1UzJrSPcJzeBFxKPpLtnL8AqzgXd1Eh5W3Ufwn9K2jqo/edit?usp=sharing) -  with counts per type.

zbycz:
[...]

PR #1590, opened 2025-06-07
Add preset for `climbing=route_bottom`

zbycz:

Hi, here is my another take on adding climbing features to iD. This time I focused only on what is already prevalent in OSM.  Last year I proposed a PR (#1113) which stalled - perhaps it was too large, or perhaps I add too much of unused stuff from OSM wiki.

In this first PR, I would like to add only `climbing=route_bottom` node preset + its fields.

### Preset: Climbing Route Start (`climbing=route_bottom`)

 - This tag is used to denote the position of the start of the climbing route. It is usually the most important to have in GPS.
 - This node usually holds all the information about the upcoming route. See [picture](https://github.com/user-attachments/assets/7dd89c39-dbf7-48ad-b92f-a87f2f256ff2) for clarification. 
 - There is a possible evolution, where node can be replace by `way` with `climbing=route`, but it is not as usual in OSM DB yet.

#### Linked fields from the preset

I wrote a script which measured all the tag combinations - see results [here](https://gist.github.com/zbycz/05895772588052c8743ce6f32da7c1c1). I propose to add only fields which are used more often and are already established on the OSM wiki.

| count | tag | label | description
|--|--|--|--|
| 15582 | `name=*` |  |
| 12631 | `climbing:grade:uiaa=*` | Grade (UIAA) | Different classification scales - typically used in different countries
| 911 | `climbing:grade:french=*` | Grade (French)
| 757 |`climbing:grade:saxon=*` | Grade (Saxon)
| 145 |`climbing:grade:yds_class=*` | Grade (YDS) | [Yosemite Decimal System](https://en.wikipedia.org/wiki/Yosemite_Decimal_System) - not that prevalent, but important
| 8866 | `website=*` | |
| 2323 | `climbing:length=#` | Length (m) | Length of the upcoming route - useful to have long enough rope
| 1661 | `climbing:sport=yes` | Sport climbing | Type of climbing, where the route is secured by pre-drilled bolts ([wikipedia](https://en.wikipedia.org/wiki/Sport_climbing)) 
| 1007 | `climbing:trad=yes` | Traditional climbing | Type of climbing, where climber has to place their own protection ([wikipedia](https://en.wikipedia.org/wiki/Traditional_climbing))
| 723 | `climbing:bolts=#` | Number of Bolts | Number of the pre-drilled bolts in the upcoming route.
| 284 | `description=*` | |

### Related issues

- https://github.com/openstreetmap/id-tagging-schema/issues/31

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Climbing

**Relevant tag usage stats:**
- 15 582 nodes - https://taginfo.openstreetmap.org/tags/climbing=route_bottom

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>


data/fields/climbing/bolts.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "climbing:bolts",
+    "type": "number",
+    "minValue": 0,
+    "label": "Number of Bolts"
```
tordans: Please check if https://github.com/ideditor/schema-builder?tab=readme-ov-file#prerequisitetag should be applied here. From your table in the PR description it sounds like the bolts should only be an option when sport=yes, right?

```suggestion
    "key": "climbing:bolts",
    "type": "number",
    "minValue": 0,
    "label": "Number of Bolts",
    "prerequisiteTag": {
        "key": "climbing:sport",
        "value": "yes"
    }
```

data/fields/climbing/length.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "climbing:length",
+    "type": "number",
+    "minValue": 0,
+    "label": "Length (m)"
```
tordans:
```suggestion
    "key": "climbing:length",
    "type": "number",
    "minValue": 0,
    "label": "Length (Meters)"
```

We always write it like this https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema+%22%28Meters%29%22&type=code not the short hand https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema+%22%28m%29%22&type=code

---

Ping https://github.com/ideditor/schema-builder/pull/198 and https://github.com/ideditor/schema-builder/issues/15#issuecomment-2948386541

data/fields/climbing/sport.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "climbing:sport",
+    "type": "check",
```
tordans, line 3: What are those number values? https://taginfo.openstreetmap.org/keys/climbing%3Asport#values


Also, could you create a wiki page and check the wiki data item https://taginfo.openstreetmap.org/keys/climbing%3Asport#wiki

data/fields/climbing/trad.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "climbing:trad",
+    "type": "check",
```
tordans, line 3: What are those number values? https://taginfo.openstreetmap.org/keys/climbing%3Atrad#values

Also, could you create a wiki page and check the wiki data item https://taginfo.openstreetmap.org/keys/climbing%3Atrad#wiki

Both should mention the `:traditional` as a tagging mistake for this, I assume https://taginfo.openstreetmap.org/keys/climbing%3Atraditional

data/fields/climbing/bolts.json
tordans: Once that changed, we could make it a `fields` field so it becomes visible as soon as the sport checkbox is checked.

data/fields/climbing/sport.json
```diff
@@ -0,0 +1,5 @@
...
+    "label": "Sport climbing"
```
tordans: Do users really understand that this is "with bolts" where the other is "bring your own"? Can we make this part of the label without making it wrong in some cases…?

Eg. https://de.wikipedia.org/wiki/Sportklettern says "bouldern" is also sport climbing?

data/presets/climbing/route_bottom.json
```diff
@@ -0,0 +1,33 @@
...
+        "climbing/grade/french",
+        "climbing/grade/saxon",
+        "climbing/grade/yds_class",
+        "website",
+        "climbing/length",
+        "climbing/sport",
+        "climbing/trad",
+        "climbing/bolts",
+        "description"
+    ],
```
tordans:
```suggestion
    "fields": [
        "name",
        "climbing/sport",
        "climbing/trad",
        "climbing/bolts",
    ],
    "moreFields": [
        "climbing/grade/uiaa",
        "climbing/grade/french",
        "climbing/grade/saxon",
        "climbing/grade/yds_class",
        "website",
        "climbing/length",
        "description"
    ],
```

Does every route have a name? Having the name field that prominent will result in users adding it as a free form field. What is the recommended name for a route?

Is there some way to promote a source in this schema? Like "From the signs at the wall" or "From the book XY"?

I think some fields should be `fields`. Aren't sport and trad fields we always want to be filled? (With bolts being conditional, see above)

data/fields/climbing/grade/french.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "climbing:grade:french",
+    "type": "text",
```
tordans: I wonder if we should try to improve the data validation aspects for those grade fields.
`type:text` is the most open field which provides no validation.
However, the values are well defined, eg. https://taginfo.openstreetmap.org/keys/climbing%3Agrade%3Afrench#values

What are our options?
* We could make it an [`combo` field](https://github.com/ideditor/schema-builder?tab=readme-ov-file#combodropdown-fields) with `strings` and either list all or rely on taginfo to fill the missing ([see `autosuggest`](https://github.com/ideditor/schema-builder?tab=readme-ov-file#autosuggestions))
* If there where a good website that we can link to, we could make it an `identifier` field and use [the `pattern`](https://github.com/ideditor/schema-builder?tab=readme-ov-file#pattern) to validate the input via Regex. The website should be something like a generic page that handles the input as search param, maybe something that shows the value in different grades?

This applies to all 3 grade fields.

I think we should at least do the `combo` so users are guided to well structures data.

data/presets/climbing/route_bottom.json
tordans: About the `name`: Looking at a few examples https://overpass-turbo.eu/s/25II the data looks good ATM. That might change, however, once we promote it to less experienced users.

data/presets/climbing/route_bottom.json
```diff
@@ -0,0 +1,33 @@
...
+    "terms": [
+        "rock climbing route"
```
tordans:
```suggestion
        "rock climbing route",
        "sport climbing",
        "route bottom",
        "climbing point",
        "route start",
        "climbing area"
```

Maybe some more…

tordans:
> I added the wikibase entry, but it doesn't show. Instead it links to "key:climbing" (https://wiki.openstreetmap.org/wiki/Item:Q3036) How to link this new one https://wiki.openstreetmap.org/wiki/Item:Q23038 ?

You have to add a description in the data item table, eg  https://wiki.openstreetmap.org/wiki/Item:Q8221

> image

zbycz:
@tordans - thanks for a comprehensive review. I agree that this is a good oportunity to update the wiki. I will work on it this week. Once it is visible on wiki, i will respond to the comments with links.

ad link to Q230238) i dont understand your comment - this item has the label specified, but iD still ignores the dataitem altogether. It seems to only fetch a wiki page, not the item. Maybe it gets fixed after i create the route_bottom page? or do i have to link it to the item?

image

data/fields/climbing/bolts.json
musoke: Some routes have a few bolts but require other protection for other sections of the climb.  These are not sport routes.

tordans: I see. We could still use the prerequisiteTags to expose the field right away once a user chooses the sport route. Any existing bolts values would still be visible, because whenever something is there, we show the field. (I just clarified this in https://github.com/ideditor/schema-builder/commit/89fa4644bce9c74329b376a5d1ccfae4c37c99ee).
But, we would loose the "Add field" Dropdown option to kind of promote the field on the initial version of this preset.

I still think this is a good change. We would promote the bolts quite a bit more this way. We still show existing data. Only in some cases, the field is promoted less … but in general a lot more than it is now.

Thoughts?

PS: https://github.com/openstreetmap/id-tagging-schema/blob/8b93d184e1b4bd071229717fef4e3f3e4d6f2289/data/fields/overlap.json#L7 is a good way to test this right now on https://ideditor.netlify.app/

musoke: Sport implies bolts in every scenario I can think of, so I think that makes sense.

data/fields/climbing/sport.json
musoke: Users will understand.  Bouldering occasionally gets lumped under sport climbing for political reasons related to "sport climbing" being an Olympic sport. There may be some linguistic differences too, but most Germans I know also make a distinction between the two.

data/presets/climbing/route_bottom.json
musoke: Most routes have a name.  Rarely there is a sign.  More often books, online guides, word of mouth.

data/fields/climbing/grade/french.json
musoke: Grades are very diverse.   French grades might benefit from a combo, but other grades get more complicated.

data/fields/climbing/length.json
zbycz: Thanks, fixed.

data/fields/climbing/bolts.json
zbycz: That sounds good. Added and pushed the changes.

data/fields/climbing/sport.json
zbycz: I created the pages with explanation. In short - the numbers are on different features (crags,areas):
- https://wiki.openstreetmap.org/wiki/Key:climbing:sport
- https://wiki.openstreetmap.org/wiki/Key:climbing:trad

data/fields/climbing/trad.json
zbycz: I created the pages with explanation. In short - the numbers are on different features (crags,areas):
- https://wiki.openstreetmap.org/wiki/Key:climbing:sport
- https://wiki.openstreetmap.org/wiki/Key:climbing:trad

Good point with the tagging mistake - I added it to the wiki page "Key:climbing:trad".

data/fields/climbing/sport.json
zbycz: I explored this topic little bit more, and usually the hypernym Rock climbing, consists of Sport climbing (with bolts+rope), Bouldering (shorter, without bolts), etc. The [wikipedia states it clearly](https://en.wikipedia.org/wiki/Sport_climbing#:~:text=While%20competition%20climbing%20consists%20of%20three%20distinct%20rock%20climbing%20disciplines%20of%20lead%20climbing%20(the%20bolted%20sport%20climbing%20element)%2C%20bouldering%20(no%20bolts%20needed)%2C%20and%20speed%20climbing%20(also%20not%20bolted%2C%20but%20instead%20top%20roped)%2C%20it%20is%20sometimes%20confusingly%20referred%20to%20as%20%22sport%20climbing%22.).

It seems the Olympics really added some confusion here.

To make it clearer, I added a bracket: `Sport climbing (bolted climbing)`  – specifically because we don't have the bouldering option (yet), so begginers could try to fit it wrongly. What do you think?

data/presets/climbing/route_bottom.json
zbycz: Yes, usually routes have names - I have never heard about a route without name. But when is someone mapping in the field – they could probably see there is a bolted route, but the name is not usually written there. It is maintained by the local climbing organization, usually appears in printed guides, and sometimes there is a signboard nearby with the names, difficulties and a map (topo).

I added the suggested fields to `fields` property.


New route looks like this, which is probably fine - the info buttons shows the additional clarification:
image

data/fields/climbing/grade/french.json
zbycz: I took the values from the [grading table on wiki](https://wiki.openstreetmap.org/wiki/Climbing#Grading), and made them into a `combo` field. I took only the basic ones (without slash), i think it is a good start for the users to select, and of course it allows users to enter their own value, if they prefer.

data/presets/climbing/route_bottom.json
zbycz: added those + "climbing crag"

zbycz:
> In general, it would be great if you could take this opportunity and check if the Wiki pages are beginner friendly already. I thing they need to be improved a bit.
> 
> When I looked at https://wiki.openstreetmap.org/wiki/Tag:sport%3Dclimbing I did not find any mention of the fields here. The fact that there is https://wiki.openstreetmap.org/wiki/Climbing is only visible if users actually read line by line. And only on https://wiki.openstreetmap.org/wiki/Climbing are there more details on what we are doing here.
> 
> Adding some of the fields here as redirects will help. But also adding some more links and callouts to the pages will help. The German page is also bigger than the english page which I always find weird for something that is international https://wiki.openstreetmap.org/wiki/DE:Tag:sport%3Dclimbing (and not scoped to one country in the presets here).

I changed this on the wiki + created correct wikitiems to all of these pages:
- added page for https://wiki.openstreetmap.org/wiki/Tag:climbing%3Droute_bottom - listed all common values and some more from the Climbing page. 
- all keys/tags from the route_bottom page are created:
  - https://wiki.openstreetmap.org/wiki/Key:climbing:length
  - https://wiki.openstreetmap.org/wiki/Key:climbing:bolts
  - https://wiki.openstreetmap.org/wiki/Key:climbing:sport
  - https://wiki.openstreetmap.org/wiki/Key:climbing:trad
  - https://wiki.openstreetmap.org/wiki/Key:climbing:grade:uiaa
  - https://wiki.openstreetmap.org/wiki/Key:climbing:grade:french
  - https://wiki.openstreetmap.org/wiki/Key:climbing:grade:saxon
  - https://wiki.openstreetmap.org/wiki/Key:climbing:grade:yds_class
- i removed the misleading table with `leisure=sport_center` on the https://wiki.openstreetmap.org/wiki/Tag:sport%3Dclimbing page, and added what tags is used for differentiating among the climbing features. The details are left on the large Climbing page. In german the sport=climbing page basically just contains what is in english on the Climbing page. I don't see an issue with that. 
- I added a category https://wiki.openstreetmap.org/wiki/Category:Climbing

data/fields/climbing/grade/french.json
```diff
@@ -0,0 +1,45 @@
+{
+    "key": "climbing:grade:french",
+    "type": "combo",
+    "label": "Grade (French)",
+    "placeholder": "5c",
+    "options": [
```
tordans, line 9:
```suggestion
    "caseSensitive": false,
    "customValues": true,
    "autoSuggestions": false,
    "options": [
```

data/fields/climbing/grade/saxon.json
```diff
@@ -0,0 +1,36 @@
+{
+    "key": "climbing:grade:saxon",
+    "type": "combo",
+    "label": "Grade (Saxon)",
+    "placeholder": "VIIa",
+    "options": [
```
tordans, line 9:
```suggestion
    "caseSensitive": true,
    "customValues": true,
    "autoSuggestions": false,
    "options": [
```

data/fields/climbing/grade/uiaa.json
```diff
@@ -0,0 +1,48 @@
+{
+    "key": "climbing:grade:uiaa",
+    "type": "combo",
+    "label": "Grade (UIAA)",
+    "placeholder": "6-",
+    "options": [
```
tordans, line 9:
```suggestion
    "caseSensitive": false,
    "customValues": true,
    "autoSuggestions": false,
    "options": [
```

tordans (changes requested):
The combo field can be improved a bit by adding a few more flags; some are defaults but it makes it more explicit. More at https://github.com/ideditor/schema-builder?tab=readme-ov-file#autosuggestions

I will apply those right away to see them in action.

data/fields/climbing/grade/yds_class.json
```diff
@@ -0,0 +1,42 @@
+{
+    "key": "climbing:grade:yds_class",
+    "type": "combo",
+    "label": "Grade (YDS)",
+    "placeholder": "5.10a",
+    "options": [
```
tordans, line 9: (Missed one with my last change)

```suggestion
    "caseSensitive": false,
    "customValues": true,
    "autoSuggestions": false,
    "options": [
```

tordans (approved):
I think this is ready to be merged now.
I changed the combo fields a bit with my last two commits. The Grade (Saxon) requires the `caseSensitive=true` flag. And with explicit `autosuggest=off` the options are sorted just as they are specified in the file and also no network request is done.

To me, this looks good. Shall we merge?

PR #1591, opened 2025-06-08
Add climbing features

harahu:
Merge first: https://github.com/openstreetmap/id-tagging-schema/pull/1590
---

Climbing is an activity that is often heavily reliant on maps and guides. Such material often exists in digital format, but the standard is, unfortunately, that information on climbing routes and areas are hidden behind paywalls. I want to contribute to making it more common that (geographic) information about climbing routes is shared  in the open, and I think OSM is the obvious arena to do that.

The main barrier I currently see to sharing such information is that the beginner editor that newcomers to OSM reach for (iD) does not have any support for climbing related features. So I've decided I want to do something about that. I'll do that with this PR, as well as a subsequent PR to iD itself, if this PR goes though, where I'll improve rendering of climbing routes.

The `climbing` key today has ~26000 uses, so I think it deserves recognition in the editor.

### Links and data

**Relevant OSM Wiki links:**
- [Climbing](https://wiki.openstreetmap.org/wiki/Climbing)

**Relevant tag usage stats:**
https://taginfo.openstreetmap.org/keys/climbing#overview

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>


data/presets/climbing/area.json
```diff
@@ -0,0 +1,15 @@
+{
+    "icon": "temaki-climbing",
+    "fields": [
+        "name"
+    ],
+    "geometry": [
+        "relation",
+        "point"
```
tordans: https://wiki.openstreetmap.org/wiki/Climbing#Climbing_areas only mentions relations and https://taginfo.openstreetmap.org/tags/climbing=area#overview shows little usage on points (and ways).

I think we should only promote the relation here and use the #1590 for points. 

A cleanup via MapRoulette might be good as a parallel effort.

data/presets/climbing/crag.json
```diff
@@ -0,0 +1,15 @@
+{
+    "icon": "temaki-climbing",
+    "fields": [
+        "name"
+    ],
+    "geometry": [
+        "relation",
+        "point"
```
tordans: What about 1.6k ways? https://taginfo.openstreetmap.org/tags/climbing=crag#overview

data/presets/climbing/route.json
```diff
@@ -0,0 +1,15 @@
+{
+    "icon": "temaki-climbing",
+    "fields": [
+        "name"
+    ],
+    "geometry": [
+        "line",
+        "relation",
+        "point"
```
tordans: Should we try to migrate those points to #1590 nodes instead? Maybe via MapRoulette…

I wonder, how will a new(ish) user understand what to pick when searching in iD…

data/presets/climbing/route_bottom.json
```diff
@@ -0,0 +1,11 @@
+{
+    "icon": "temaki-climbing",
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "tags": {
+        "climbing": "route_bottom"
```
tordans: Lets remove this and first merge #1590 (as you already suggested)

data/presets/climbing/route_top.json
```diff
@@ -0,0 +1,11 @@
+{
+    "icon": "temaki-climbing",
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "tags": {
+        "climbing": "route_top"
```
tordans: Isn't the top always the end of the way? It does not make sense to have a top node without any way or relation between the start, right? And once there is a way, the top node does not add to the data… (except for the fact that it makes explicit that there is no `fixme=continue`) on the way.

Are there any additional fields that this preset would get?

I wonder if we should make it an unsearchable preset (renamte to `_foo.json` and add [`searchable:false`](https://github.com/ideditor/schema-builder?tab=readme-ov-file#searchable)), so it gets rendered in the UI and shows something when clicked… but it is not promoted via the search and category list.

tordans (changes requested):
Lets see if we can keep the momentum of #1590 and this going, this time… :-). Here are some initial thoughts. 

For the relation stuff – which I don't know a lot about – we should also look at https://github.com/openstreetmap/id-tagging-schema/pull/1538 to add some good validations and guidance to the UI (once that is merged…).

data/presets/climbing/area.json
harahu: I agree that areas should only be represented as relations. I changed this now.

data/presets/climbing/crag.json
harahu: Personally, I dislike the practice of tagging cliffs with `climbing=crag`. To me the concept of a crag is distinct from the cliff itself, and mixing the two feels like it breaks with the [one feature, one OSM element](https://wiki.openstreetmap.org/wiki/One_feature,_one_OSM_element) guideline.

However, this pattern seems so well established that I don't feel it's my role to argue against it at this point. I added "line" as a possible geometry.

data/presets/climbing/route_bottom.json
harahu: I've removed it now.

data/presets/climbing/route.json
harahu: For consistency I think that might be a good idea for now. I've dropped the "point" option for routes in this PR.

I think there's a case to be made for deprecating "route_bottom" in favor of just "route", but that will likely require a formal proposal, which won't be done anytime soon.

Whether you map a route as a point, line or relation comes down to the qualities of the route and the level of detail you are after. A climbing route with long horizontal travel (e.g. a ridge traverse) is in my opinion best mapped as a line, or a relation of lines. Whereas a short, vertical single-pitch route might only warrant a node.

data/presets/climbing/route_top.json
harahu: I'll be honest, I don't really see the purpose of this tag either. I added it to the PR mostly to respect the wiki, but I have no problem with making it unsearchable. I have done so now.

harahu:
@tordans Thank you for the (very useful) review! I've now made this PR completely independent of #1590, so the order in which they are merged doesn't matter anymore.

If you're happy with this PR in it's current state feel free to merge it. Alternatively, I am considering splitting this PR into four different PRs (one per feature), such that we can discuss and merge them all individually.

harahu:
@tordans: Closing this in favor of #1596, #1597 and #1598.

PR #1593, opened 2025-06-11
Add more detailed information on integration testing

harahu:

I had some trouble figuring out how to test schema changes locally on my machine. @tordans helpfully figured out how to do so and provided me with a solution that works.

We felt it would be appropriate to document this workflow, and this is my attempt at doing so.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,4 +71,11 @@ The following `npm` commands are used in this repository:
 * `npm run dist` – validates the source data and compiles output files for iD
 * `npm run translations` – fetches translations from transifex and compiles the translations files for iD
 
-If you have [set up](https://github.com/openstreetmap/iD#installation) your own local instance of the iD editor, you can [configure](https://github.com/openstreetmap/iD/blob/develop/API.md#environment-variables) it to use your local set of tagging presets by setting the `ID_PRESETS_CDN_URL` environment variable. If you do that, don't forget to run `npm run dist` to compile the tagging schema output, as iD will otherwise not see the changes you made.
+### Integration Testing With iD
+When making changes to the schema, you would usually want to visually inspect how your changes would affect the user experience in the iD editor. There are two ways to do this.
+
+You could create a PR to this repository, and we will automatically create an iD instance for you with your changes in effect.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -71,4 +71,11 @@ The following `npm` commands are used in this repository:
...
+### Integration Testing With iD
+When making changes to the schema, you would usually want to visually inspect how your changes would affect the user experience in the iD editor. There are two ways to do this.
+
+You could create a PR to this repository, and we will automatically create an iD instance for you with your changes in effect.
+
+Alternatively, if you have [set up](https://github.com/openstreetmap/iD#installation) your own local instance of the iD editor, you can [configure](https://github.com/openstreetmap/iD/blob/develop/API.md#environment-variables) it to use your local set of tagging presets by setting the `ID_PRESETS_CDN_URL` environment variable.
+* First build and serve the schema:`npm run build && npm run dist && npx serve -Cp 1234`. Remember that you need to run `npm run dist` again should you make further changes.
```
k-yle: minor suggestion to reduce ambiguity:
```suggestion
* First build and serve the schema: `npm run build && npm run dist && npx serve -Cp 1234`. Remember that you need to run this command again should you make further changes.
```

CONTRIBUTING.md
```diff
@@ -71,4 +71,11 @@ The following `npm` commands are used in this repository:
...
+* Now, in your iD repository, start an iD instance using your custom schema: `ID_PRESETS_CDN_URL=http://localhost:1234/ npm start`
```
k-yle: the syntax is slightly different for windows. we could either suggest a more universal (but slower) command:

```suggestion
* Now, in your iD repository, start an iD instance using your custom schema: `npx cross-env ID_PRESETS_CDN_URL=http://localhost:1234/ npm start`
```

or explictly state the windows equivalent:

```suggestion
* Now, in your iD repository, start an iD instance using your custom schema: 
 * on macOS & Linux: `npx cross-env ID_PRESETS_CDN_URL=http://localhost:1234/ npm start`
 * on Windows:  `set ID_PRESETS_CDN_URL=http://localhost:1234/ && npm start`
```

harahu: I'm more partial to your seconds proposal. But does the `npx cross-env` part serve a purpose on unix systems?

k-yle: it's not required on unix systems, the purpose of the [cross-env](https://npm.im/cross-env) npm package is just to create a cross-platform command for setting env variables. 

and the [npx](https://docs.npmjs.com/cli/v8/commands/npx) part of the command is a method to execute an npm package without first installing it.

either option would be a  from me

PR #1597, opened 2025-06-12
Add preset for `climbing=crag` point and relation

harahu:

This PR is part of a larger effort to add support for rock climbing related features to iD. In this PR we add a preset for a climbing "crag", which is understood to be a grouping of distinct climbing routes that are located at most a few minutes walk from each other, usually on the same cliff face. Climbing crags are currently in active use by data consumers, e.g. openclimbing.org.

Climbing crags can be equipped with additional fields, in addition to a name. Usually this would be tags representing summary statistics for the routes contained within the crag. I've consciously decided to leave such fields out of this PR, since in practice all this information can be computed from the member objects, if the crag is represented as a relation. If the crag is rather represented as a point or a cliff line, such additional tagging could be helpful, but I'd rather encourage mapping crags as relations of routes, since this increases the value of the data available in OSM.

I've set up a Wikibase entry and a corresponding article for climbing areas, which can be found here:
https://wiki.openstreetmap.org/wiki/Item:Q23047
https://wiki.openstreetmap.org/wiki/Tag:climbing%3Dcrag

### Related issues

- #31 

### Links and data

**Relevant OSM Wiki links:**
- [Climbing crags](https://wiki.openstreetmap.org/wiki/Climbing#Crags)

**Relevant tag usage stats:**
https://taginfo.openstreetmap.org/tags/climbing=crag
https://taghistory.raifer.tech/#***/climbing/crag

> 5868 current uses

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>


data/presets/climbing/crag.json
```diff
@@ -0,0 +1,24 @@
...
+    "icon": "temaki-climbing",
+    "fields": [
+        "name"
+    ],
+    "moreFields": [
+        "website"
+    ],
+    "geometry": [
+        "point",
+        "line"
```
tordans: I am not sure about the line. 
I [looked at one case ](https://pr-1597--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=20.00/47.21543/11.41928&disable_features=boundaries&background=geovekst-nib&id=w1078893396) which is more of an area than a line. But it is rendered as a line because the `area=yes` is missing (and this tagging is not recognized as area-like when the way is closed.

Did not digg deeper into other examples https://overpass-turbo.eu/s/2670 to see if most are areas and its harder to analyse with taginfo and such because of our data schema…

I assume that the wiki recommendation is to not use an area but relation instead? Does openclimbingmap support this area like tagging?

I wonder if we should habe a second `_crag_area.json` (unsearchable) that has "addTags: area=yes" and triggers on line+area "climbing=crag". That should make it that the defaults are relations and points. But when users view existing data they at least see a preset that guides them to add the area-tag.

tordans (changes requested):
This looks good. One thing about lines/areas that we should think about. 

As with #1598 I hope for input from others about the relation part of this PR.

data/presets/climbing/crag.json
harahu: A fairly common use case I've seen is tagging `natural=cliff` ways with `climbing=crag`. I strongly dislike this pattern. In my opinion crags should be mapped only as points or relations. Keeping the crag object separate from the cliff is in keeping with https://wiki.openstreetmap.org/wiki/One_feature,_one_OSM_element.

Mapping crags as areas sounds like a really bad idea, as the boundaries of a crag would be more or less unverifiable. Does it only include the parts of a cliff that has routes on it, or the whole cliff? Does it include the area below the cliff where climbers hang out? It's all difficult to determine.

I added the "line" option because it was common in practice. However, both "line" and "area" variants are undocumented on the wiki, so perhaps the best option is to just leave out the "line" option. We could always clean up the existing cases later on. Would you be OK with this approach?

tordans: Yes, let’s leave it out and look into a follow up PR or maybe MapRoulette challenge to transform them to relations.

harahu: Done :)

tordans (approved):

PR #1598, opened 2025-06-12
Add preset for `climbing=route` point, way and relation

harahu:

This PR is part of a larger effort to add support for rock climbing related features to iD. In this PR we add a preset for a climbing route. A climbing route is a is a path by which a climber reaches the top of a mountain, a rock face or an ice-covered obstacle. A climbing route can be mapped as a single line tagged `climbing=route`, or, a relation of lines, possibly including connecting paths. Climbing routes are currently in active use by data consumers, e.g. openclimbing.org.

Climbing routes can be equipped with additional fields, in addition to a name. Useful information could be the difficulty rating of the route (grade) and the type of protection used on the route, as well as the 3D length of the route. I've consciously decided to leave such fields out of this PR, since these are covered in #1590. This PR could be merged independently, and we can update the fields of the preset once that PR is merged.

I've set up a Wikibase entry and a corresponding article for climbing routes, which can be found here:
https://wiki.openstreetmap.org/wiki/Item:Q23048
https://wiki.openstreetmap.org/wiki/Tag:climbing%3Droute

### Related issues

- #31 

### Links and data

**Relevant OSM Wiki links:**
- [Climbing routes](https://wiki.openstreetmap.org/wiki/Climbing#Climbing_routes)

**Relevant tag usage stats:**
https://taginfo.openstreetmap.org/tags/climbing=route
https://taghistory.raifer.tech/#***/climbing/route

> 2057 current uses

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>


tordans:
A bit off topic, I wonder why the climbing route does not get the way style.
> image

This might happen automatically once we have https://github.com/openstreetmap/iD/issues/11133

data/presets/climbing/route.json
```diff
@@ -0,0 +1,29 @@
...
+    ],
+    "moreFields": [
+        "website"
+    ],
+    "geometry": [
+        "line",
+        "point"
+    ],
+    "tags": {
+        "climbing": "route"
```
tordans, line 22: Just double checking: We want to promote this additional tag and this is not controversial in the sub community?

tordans (approved):
This looks good to me. I will ask for some more feedback because I did not review relations before…

data/presets/climbing/route.json
harahu: @jvaclavik and/or @zbycz are the only ones I know of qualified to give a second opinion, but based on my conversations with them this is uncontroversial to them. At least if I understand correctly what you are referring to: it is a bit weird to have two tags for climbing routes (climbing=route and climbing=route_bottom). That is, however, how it is documented on the wiki and done in practice.

In order to resolve this, I am strongly considering making a proposal for deprecating the latter in favor of the first, but right now I think it is more important to just get editor support in place for the current de facto schema, and that means both tags.

If any strong objections appear, even after merging, I am more than happy to rework the schema as deemed appropriate by the sub community.

tordans: Just to clarify: I think my comment referenced the wrong line I was wondering about the Sport part because iD will now show „incomplete tagging“ warnings for many routes. 

I agree with the rest. A proposal to cleanup the double tagging would be great. But afterward.

harahu: Aha, I see what you mean. I ended up adding in `sport=climbing` because I noticed this was also done in #1590, and I wanted to remain consistent with that PR. A justification for it is described here: https://wiki.openstreetmap.org/wiki/Tag:climbing%3Droute_bottom

I guess this boils down to the question of whether `climbing` should be considered a root level key, or whether it is considered as a specifier underneath `sport=climbing`. In the case of `wetland=bog`, you would always use this tag in combination with `natural=wetland`, even though the first tag is more than sufficient to understand what the feature represents. To stay consistent with this pattern, it makes sense to add `sport=climbing` to all climbing related features, even if it might seem superfluous.

I'll admit that I don't have a strong preference, one way or the other here. I also don't know the history behind the pattern of explicitly representing a hierarchy, as described above. I am open to both including the tag and dropping it; whatever gets the PR merged faster.

For what it's worth, neither including the tag, nor leaving it out, seems to have an effect on how the route is rendered on [openclimbing.org](https://openclimbing.org), which is the most notable data consumer at the moment.

tordans: Thanks. Lets leave it in; this makes sense IMO.

zbycz: I think the main advantage for having `sport=climbing` mandatory is that it allows for gradual adoption by apps and renderers. Many of them are already showing `sport=*` features. 
And of course it also makes sense as a hierarchy as with the example with the `wetland`. I [added note to the wiki](https://wiki.openstreetmap.org/wiki/Tag:sport%3Dclimbing#Note_for_data_consumers_(developers)) for the data consumers, that they should show the `climbing=*` alone as well.

harahu:
> This looks good to me. I will ask for some more feedback because I did not review relations before…

I just re-based this PR and included the fields added in #1590 in the route presets, so I think this PR is now ready to go. If you prefer, I can split the relation part out into a separate PR.

harahu:
> > I just re-based this PR and included the fields added in #1590 in the route presets, so I think this PR is now ready to go.
> 
> TBH it feels weird to promote the fields for routes-way, route-relation and route-bottom which are all part of the same "object" ([🔗 ](https://pr-1598--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=17.50/68.16209/16.59931&disable_features=boundaries&background=geovekst-nib)). I guess the data users and consumers will have to keep an eye on double tagging and the resulting mismatch in data and come up with a proposal to clean this up later. Ideally only the route-bottom would hold the tags which is likely the lowest common denominator.

I agree that adding fields to both the` route_bottom`, and the way/relation would be unnecessary double tagging, and that this is a pitfall of the current schema. I don't think route_bottom is the lowest common denominator, though. In the case where a route is drawn as a way, having a `route_bottom` node isn't always appropriate (e.g. in the case of bidirectional ridge traverses), so one has to be able to add these fields to the way. In the case of multi-pitch routes represented as relations, it makes sense to add data fields to the individual pitches as well as the route as a whole, since these will be different. Individual pitches won't have `route_bottom` tags either.

I think the best option for cleanup here is to propose a change where `route_bottom` is deprecated in favor of
- a `route` node, in the case of mapping a route as a single node (still to be placed at the bottom of the route when possible)
- a field on the `route` way indicating whether the route/pitch is unidirectional (default) or bidirectional, in the case of routes drawn as ways.

That way we won't have potential for double tagging, and we also align tagging of directionality with how it's done for other features in OSM, e.g. highways.

tyrasd:
Sorry for jumping in a bit late here, but I noticed that we now have two presets with the same name "Climbing Route". One for the lines and one for the `type=route` relation. I also see that there are only a [handful](https://taginfo.openstreetmap.org/tags/route=climbing) of route relations with `route=climbing` and could not find it mentioned [on](https://wiki.openstreetmap.org/wiki/Key:type) [the](https://wiki.openstreetmap.org/wiki/Climbing) wiki… 

Intuitively, it would make sense to also have relations for climbing routes (that perhaps also include also the access paths to the route bottom, etc.), but those should be properly defined, discussed and documented fist, shouldn't they?

tordans:
@harahu this https://github.com/openstreetmap/id-tagging-schema/pull/1598#issuecomment-3118274242 is best answered by you. I did in fact not check what Martin mentioned but only the [route example from the preview](https://github.com/openstreetmap/id-tagging-schema/pull/1598#issuecomment-2969554274) which would then be one of those 5.

data/presets/type/route/climbing.json
```diff
@@ -0,0 +1,38 @@
...
+    "terms": [
+        "climb",
+        "climbing",
+        "rock climbing",
+        "rock climbing route",
+        "mountain climbing"
+    ],
+    "tags": {
+        "type": "route",
+        "route": "climbing"
```
matkoniecz, line 21: `route=climbing` has 9 uses worldwide, see https://taginfo.openstreetmap.org/tags/route=climbing

is it here intentionally?

matkoniecz: See https://github.com/openstreetmap/id-tagging-schema/issues/2152

PR #1506, opened 2025-04-04
Docs: Update wording "Data item", "OSM Wiki data item"

tordans:
Follow up to https://github.com/openstreetmap/id-tagging-schema/pull/1504
I don't think adding "Wikibase" as a term helps users understand where to find this data (in the OSM eco system).

The suggestions from https://github.com/openstreetmap/id-tagging-schema/issues/1479#issue-2908776914 where different.

I applied those suggestions but also added the Term "OSM Wiki" in some places to signal that we are talking about things that live in the context of the OSM wiki installation.

I updated the one new link to link to our own docs first (which then links to external docs).

---

For documentation purposes, those are the terms users find when they try to find the data items in our wiki…

- Search page: "Item"
   > image

- Wiki page: "Data item"
   > image

- URL of the item page: "/item"
   > image

- Docs: "Data item"
  > https://wiki.openstreetmap.org/wiki/Data_items

---

Closes https://github.com/openstreetmap/id-tagging-schema/issues/1479

PR #1600, opened 2025-06-13
Update CONTRIBUTING.md about wiki items and wiki links

zbycz:
Documentation was missing about how iD resolves the wiki items. I learned it from @1ec5 here https://github.com/openstreetmap/id-tagging-schema/pull/1590#issuecomment-2964737766 (thanks!)


tordans:
I would like to merge #1506 first and then review. 
(I need a review on that, first…)
In general I am in favor of improving this!

zbycz:
Cool, didn't know. There was another PR in progress! Great to see improvements coming 🙂

CONTRIBUTING.md
```diff
@@ -53,7 +53,10 @@ Icons from different sources (_icon sets_) can be used in the tagging schema. He
...
 
 iD and other tools provide users with a way to learn more about the main tag of a preset. It is important to provide good information in this information panel. Here are a few notes on how to do this:
-- Does your tag have a Wikibase entry? Click the small pencil icon next to the text to open the Wikibase item on the OSM wiki. Improve this wording if needed. If the Wikibase item is missing, [learn more about how to add it in "Current methods for creating new items"](https://wiki.openstreetmap.org/wiki/Data_items#Item_creation_process).
+- Does your tag have a Wikibase entry? Click the small pencil icon next to the text to open the Wikibase item on the OSM wiki. Improve this wording if needed.
+  - If the Wikibase item is missing, [learn more about how to add it in "Current methods for creating new items"](https://wiki.openstreetmap.org/wiki/Data_items#Item_creation_process).
+  - iD queries the `Tag:<key>=<value>` page or `Key:<key>` page and prefers the more specific variant. It then uses "wiki sitelinks" to get the data item from the page. These can be set [here](https://wiki.openstreetmap.org/wiki/Special:SetSiteLink//wiki), see the documentation in link above.
```
tordans, line 58: I think I never did this before. Lets double check first if this is actually needed for the lookup … or only for the Wiki-Relation part.

tordans: @harahu you added wiki data items for your last two PRs, did you add those SiteLinks described here? Or did it work with just the data item with the right values?

harahu: I added the SiteLinks as described in the guide, but I don't know whether that was on the critical path to getting the description and picture to show up.

zbycz: For me it was crtitical to show the description, though i am not sure about the image.

zbycz: [This comment](https://github.com/openstreetmap/id-tagging-schema/pull/1590#issuecomment-2964737766) helped me understand it is really needed and I also had to set the sitelink for all other wiki items i created.

k-yle: is this the only unresolved issue? I can confirm that [Special:SetSiteLink](https://osm.wiki/Special:SetSiteLink) is a required step.

and it's the hardest step to figure out imo, because the UI is broken[^1]

[^1]: the link appears in the top right corner of the screen [for a split second](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) while the page loads ([example](https://osm.wiki/Item:Q16020)). you have to be very quick to click it before it disappears

1ec5: The link to Special:SetSitelink should be in the Tools section, alongside the link to the permalink. Depending on your skin and appearance preferences, it might be tucked away in a menu somewhere. The flash is because of the gadget (installed by default) that puts the link in Tools.

1ec5: To be pedantic, iD queries the wiki for data items that are linked to these page titles. It doesn’t run a separate query for the article then the data item. MediaWiki (Wikibase) handles all that for the client.

But to answer the original question, yes, setting a sitelink is necessary. It’s a bit confusing because there’s also typically a [documentation wiki pages <small>(P31)</small>](https://wiki.openstreetmap.org/wiki/Property:P31) property listing all the different translations, but that only matters once the sitelink to the English version is established.

matkoniecz: @zbycz What you think about those comments?

zbycz: @matkoniecz - I take it, these two comments from December are merely answers. They confirm my comment about "sitelinks being needed", so my suggested documentation change is still correct. 👍 

Thanks for the ping, i didn't realize I should confirm it here.

CONTRIBUTING.md
```diff
@@ -55,7 +53,10 @@ Icons from different sources (_icon sets_) can be used in the tagging schema. He
...
 
 iD and other tools provide users with a way to learn more about the main tag of a preset. It is important to provide good information in this information panel. Here are a few notes on how to do this:
-- Does your tag have a OSM Wiki data item? Click the small pencil icon next to the text to open the data item on the OSM Wiki. Improve this wording if needed. If the data item is missing, [learn more about how to add it in "Current methods for creating new items"](https://wiki.openstreetmap.org/wiki/Data_items#Item_creation_process).
+- Does your tag have a OSM Wiki data item? Click the small pencil icon next to the text to open the data item on the OSM Wiki. Improve this wording if needed.
+  - If the data item is missing, [learn more about how to add it in "Current methods for creating new items"](https://wiki.openstreetmap.org/wiki/Data_items#Item_creation_process).
+  - iD queries the `Tag:<key>=<value>` page or `Key:<key>` page and prefers the more specific variant. It then uses "wiki sitelinks" to get the data item from the page. These can be set [here](https://wiki.openstreetmap.org/wiki/Special:SetSiteLink//wiki), see the documentation in link above.
```
matkoniecz, line 58:
```suggestion
  - iD queries the data item linked to `Tag:<key>=<value>` or `Key:<key>` page and prefers the more specific variant. This connection (called sitelink) must be set and it can be done [here](https://wiki.openstreetmap.org/wiki/Special:SetSiteLink//wiki), see the documentation in link above.
```

attempt to follow https://github.com/openstreetmap/id-tagging-schema/pull/1600/changes#r2637951895 suggestion to not introduce false claims

@1ec5 have I understood correctly what you suggested there?

matkoniecz: or have you not requested such change?

1ec5: Yes, that’s an improvement. I’d probably still mention “sitelink”. That’s the weird term that OSM Wikibase uses to tie the data item to the article that the mapper is more familiar with.

matkoniecz: I added `(called sitelink)` - is it in appropriate location?

matkoniecz: @1ec5 do you see need for any further changes?

1ec5: Looks good to me.

CONTRIBUTING.md
matkoniecz: @tordans @1ec5 do you think that changes are still needed here?

CONTRIBUTING.md
matkoniecz: @zbycz Can you take look at this? Do you agree that it is an improvement?

PR #1611, opened 2025-06-24
Add preset for `device_charging_station`

### LINKED ISSUE #1359: New tag: amenity=device_charging_station
### OSM Tag(s)

amenity=device_charging_station

As a Preset

### Label

Device charging station

### Aliases

_No response_

### Terms

Mobile Device Charging Station, Phone Charging Station

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddevice_charging_station

### Status of the Tag

Proposed

### Usage of the tag

905

### Replaces other Tag?

no

### Regional Tag?

_No response_

### Further Information

This clarifies the existing tag `amenity=charging_station` (https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station).

While the old tag did not specify the type of devices being charged, eventually the community has taken it to mean "electric vehicles" charging station. However, some places have started to offer mobile device charging spots. These charging spots may simply offer USB-A power sockets (users bring their own cables), or provide dedicated eg Lightning cables for users to charge their smart phone.

There is a great need to distinguish such small device chargers from the otherwise distinct and incompatible EV chargers.

Information about mobile device charging spots is quintessential in this age of smart mobile devices.

paulklie:
Adds the [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=[device_charging_station](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddevice_charging_station)  preset.

### Related issues

Closes #1359 

### Links and data

[amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=[device_charging_station](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddevice_charging_station)

**Relevant tag usage stats:**
[taginfo](https://taginfo.openstreetmap.org/tags/amenity=device_charging_station): 1084 uses

I seem to be unable to post screenshots here, hope this suffices. The Infobox works

https://imgur.com/a/seYQ89i
https://www.openstreetmap.org/node/9983467989

### Search
https://imgur.com/a/x57uggB

### Info-`i`
https://imgur.com/a/GeQBaRf

### Wording

- [x] American English

```

### DIFF
--- a/data/presets/amenity/device_charging_station.json
+++ b/data/presets/amenity/device_charging_station.json
@@ -6,13 +6,13 @@
         "fee",
         "payment_multi",
         "level",
-        "indoor",
         "lockable",
         "opening_hours"
     ],
     "moreFields": [
         "wheelchair",
-        "operator"
+        "operator",
+        "indoor"
     ],
     "geometry": [
         "point"
--- a/data/presets/amenity/device_charging_station.json
+++ b/data/presets/amenity/device_charging_station.json
@@ -1,0 +1,27 @@
+{
+    "icon": "temaki-electronic",
+    "fields": [
+        "operator",
+        "capacity",
+        "fee",
+        "indoor",
+        "opening_hours"
+    ],
+    "moreFields": [
+        "level",
+        "manufacturer",
+        "maxstay",
+        "network"
+    ],
+    "geometry": [
+        "point"
+    ],
+    "tags": {
+        "amenity": "device_charging_station"
+    },
+    "terms": [
+        "USB",
+        "Plug"
+    ],
+    "name": "Device Charging Station"
+}
--- a/data/presets/amenity/device_charging_station.json
+++ b/data/presets/amenity/device_charging_station.json
@@ -1,17 +1,18 @@
 {
     "icon": "temaki-electronic",
     "fields": [
-        "operator",
+        "access_simple",
         "capacity",
         "fee",
+        "payment_multi",
+        "level",
         "indoor",
+        "lockable",
         "opening_hours"
     ],
     "moreFields": [
-        "level",
-        "manufacturer",
-        "maxstay",
-        "network"
+        "wheelchair",
+        "operator"
     ],
     "geometry": [
         "point"
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -25,7 +25,7 @@ Follow these steps to test your PR yourself and make it a lot easier and faster 
 
 **This is how it works:**
 1. After you submit your PR, the system will create a preview and comment on your PR:
-   > 🍱 You can preview the tagging presets of this pull request here.
+   > 🍱 Your pull request preview is ready.
    If this is your first contribution to this project, the preview will not happen right away but requires a click from one of the project members. We will do this ASAP.
 
 2. Once the preview is ready, use it to test your changes.

### CODE REVIEW

File: data/presets/amenity/device_charging_station.json [hunk 1]
```diff
@@ -0,0 +1,28 @@
+{
+    "icon": "temaki-electronic",
+    "fields": [
+        "operator",
+        "capacity",
+        "fee",
+        "indoor",
```
tordans:
```suggestion
```

Are there any non indoor? This is not on the wiki.

data/presets/amenity/device_charging_station.json
```diff
@@ -0,0 +1,28 @@
+{
+    "icon": "temaki-electronic",
+    "fields": [
+        "operator",
+        "capacity",
+        "fee",
```
tordans, line 6: Wiki suggest `payment`, https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/payment_multi_fee.json

data/presets/amenity/device_charging_station.json
```diff
@@ -0,0 +1,28 @@
...
+        "indoor",
+        "opening_hours"
+    ],
+    "moreFields": [
+        "level",
+        "manufacturer",
```
tordans:
```suggestion
```

Not recommended by wiki; is this relevant?

data/presets/amenity/device_charging_station.json
```diff
@@ -0,0 +1,28 @@
...
+        "maxstay",
```
tordans:
```suggestion
```

Not recommended by wiki.

data/presets/amenity/device_charging_station.json
```diff
@@ -0,0 +1,28 @@
...
+        "network"
```
tordans:
```suggestion
```

Not recommended by wiki; is this relevant?

data/presets/amenity/device_charging_station.json
paulklie: There are five: https://overpass-turbo.eu/s/26Rx
Thus around 20% of those with the "indoor" tag.

paulklie: *poke*

paulklie:
@tordans *poke*

paulklie:
## Test-Documentation

https://www.openstreetmap.org/node/10313144876#map=19/50.329905/7.224757
image



### Search


image
image
image

### Info-`i`


image

### Wording

- [x] American English

tordans:
@matkoniecz I think this looks good. Could you give this a review? Will add the "ready for merge tag" and merge next "round".

There is the open question if we would rather want to remove the `indoor` field given that they all should be indoor (and exceptions can always be added via the tags UI).

tordans:
> > if we would rather want to remove the `indoor` field given that they all should be indoor
> 
> I don't understand why? 

Because we want to be careful which tags to promote via fields and in my mind an outdoor charging station is something very unexpected. We don't want fields that are answered with "no" all the time by users just because they are very visible. This is not even in the `moreFields` but suggested to be answered to every user of the preset.

>  I have seen two such oudoor charging stations, one directly connected to a solar panel, and one in a bus stop.

The bus stop would then be `covered=yes`, yes? And the solar panel maybe as well?
But we don't promoted covered as an alternative to indoor…

Are there really uncovered charing stations (in a relevant number that we want to handle those in a preset)?

This would be easier if there where example images with tags in https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddevice_charging_station for those cases… but the tag is not even mentioned there…

data/presets/amenity/device_charging_station.json
matkoniecz: Given that it is so rare (and sufficiently unexpected that I suspect tagging errors, I would ask mappers if I would not be on bus)

I would put it into moreFields so it is not shown by default

matkoniecz: Made some probing changeset comments in the end

tordans:
Given  https://github.com/openstreetmap/id-tagging-schema/pull/1611#discussion_r2325483404 I moved it to `moreFields` for now. We can always improve this later.

PR #1637, opened 2025-07-18
Add power terminal preset

### LINKED ISSUE #653: Add preset for `power=terminal`
### OSM Tag(s)

power=terminal

As a Preset

### Label

Power Terminal

### Aliases

_No response_

### Terms

Electric Terminal, Electrical Terminal

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag%3Apower%3Dterminal

### Status of the Tag

In Use

### Usage of the tag

39 728 

### Replaces other Tag?

no

### Regional Tag?

_No response_

### Further Information

_No response_

nlehuby:

Add a preset for power terminal

### Related issues

Fix #653 

### Links and data

**Relevant OSM Wiki links:**

- https://wiki.openstreetmap.org/wiki/Tag%3Apower%3Dterminal
- https://taginfo.openstreetmap.org/tags/power=terminal
- #529

### DIFF
--- a/data/presets/power/terminal.json
+++ b/data/presets/power/terminal.json
@@ -1,7 +1,8 @@
 {
     "icon": "temaki-power_device",
     "fields": [
-        "operator"
+        "operator",
+        "line_attachment"
     ],
     "moreFields": [
         "height",
--- a/data/presets/power/terminal.json
+++ b/data/presets/power/terminal.json
@@ -22,5 +22,5 @@
     "addTags": {
         "power": "terminal"
     },
-    "name": "Power Terminal"
+    "name": "Power line reaching building"
 }
--- a/data/presets/power/terminal.json
+++ b/data/presets/power/terminal.json
@@ -13,7 +13,8 @@
     ],
     "terms": [
         "electric terminal",
-        "electrical terminal"
+        "electrical terminal",
+        "service entrance"
     ],
     "tags": {
         "power": "terminal"
--- a/data/presets/power/terminal.json
+++ b/data/presets/power/terminal.json
@@ -1,8 +1,7 @@
 {
     "icon": "temaki-power_device",
     "fields": [
-        "operator",
-        "line_attachment"
+        "operator"
     ],
     "moreFields": [
         "height",
@@ -21,8 +20,7 @@
         "power": "terminal"
     },
     "addTags": {
-        "power": "terminal",
-        "line_attachment": "anchor"
+        "power": "terminal"
     },
     "name": "Power Terminal"
 }
--- a/data/presets/power/terminal.json
+++ b/data/presets/power/terminal.json
@@ -1,0 +1,26 @@
+{
+    "icon": "temaki-power_device",
+    "fields": [
+        "operator"
+    ],
+    "moreFields": [
+        "height",
+        "ref"
+    ],
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "terms": [
+        "electric terminal",
+        "electrical terminal"
+    ],
+    "tags": {
+        "power": "terminal"
+    },
+    "addTags": {
+        "power": "terminal",
+        "line_attachment": "anchor"
+    },
+    "name": "Power Terminal"
+}
--- a/data/presets/power/terminal.json
+++ b/data/presets/power/terminal.json
@@ -22,5 +22,5 @@
     "addTags": {
         "power": "terminal"
     },
-    "name": "Power line reaching building"
+    "name": "Power Line Reaching Building"
 }


data/presets/power/terminal.json
```diff
@@ -0,0 +1,26 @@
...
+        "height",
+        "ref"
+    ],
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "terms": [
+        "electric terminal",
+        "electrical terminal"
```
1ec5: Looks like the technical term in American English would be “service entrance”. (A service drop connects a distribution line to a service entrance.) I also wonder if we should rename the preset to something more descriptive, since a layperson unfamiliar with electrical terminology could mistake “power terminal” for a substation.

tordans: >  I also wonder if we should rename the preset to something more descriptive, since a layperson unfamiliar with electrical terminology could mistake “power terminal” for a substation.

+1 I would never look for the tagging that connect a house a power line with this. No better suggestion, though…

tordans (commented):
(See comments above to be reviewed)

tordans (changes requested):
(See comments above to be reviewed)

data/presets/power/terminal.json
flacombe: We should distinguish two independant concepts:
* [A power terminal](https://electropedia.org/iev/iev.nsf/display?openform&ievref=466-06-06) in general, meaning a line ends on a building (related to more recent line_management=termination). It covers both public grid infrastructure (lines enter a substation) and private customers supply (service drop).
* A [service drop](https://en.wikipedia.org/wiki/Service_drop), a particular kind of power terminal intended to supply a single customer with power from the grid. This one is currently covered by [a proposal under RFC](https://wiki.openstreetmap.org/wiki/Proposal:Service_supplies_proposal).

So not all terminals are service drops and additionally, power=terminal is intended for all terminals except ones on poles/towers/portals (particularly on buildings).  
That's why we consider line_management=termination for an even broader usage then.

flacombe:
Regarding the tagging to be used in combination with power=terminal, here is some options:
* line_attachment=anchor
* line_management=termination

A [termination](https://wiki.openstreetmap.org/wiki/Tag%3Aline_management%3Dtermination) (more specifically a power=terminal) is at the opposite from a [transition](https://wiki.openstreetmap.org/wiki/Tag:line_management%3Dtransition), so location:transition is incompatible with power=terminal.

I should add corresponding examples to the wiki to make it clearer

tordans:
> I should add corresponding examples to the wiki to make it clearer

That would be great. Something like tagging examples are always helpful to validate the presets.
What would also be helpful for me are suggestions on how we name those presets (plus icon) in a way that users that are not deep into this schema will understand what to pick.

data/presets/power/terminal.json
1ec5: > So not all terminals are service drops and additionally, power=terminal is intended for all terminals except ones on poles/towers/portals (particularly on buildings).

That’s fine, this is just a list of search terms.

flacombe:
I've improved several pages:
* power=terminal (description + examples)
* power=connection (mention line_management)
* line_management=termination (terminal example)

Common principle that should be reminded is power=* are specific concepts and line_management are more transversal and sometimes abstracted concepts. This allows us to adapt to a large diversity of situations.
I hope this helps to complete this PR that is still valid.

data/presets/power/terminal.json
```diff
@@ -0,0 +1,28 @@
...
+        "electric terminal",
+        "electrical terminal",
+        "service entrance"
+    ],
+    "tags": {
+        "power": "terminal"
+    },
+    "addTags": {
+        "power": "terminal",
+        "line_attachment": "anchor"
```
matkoniecz: Are you sure that `line_attachment=anchor` should be added to every single `power=terminal`?

If mass adding such tag is useful, it should be done as an automated bot edit. Though if it is really supposed to appear at every single one then I am dubious is there point in doing it.

Not via deprecation rule that wastes enormous amount of human time for no good reason.

Yes, I know that we have some rules like that in this repo already. I think that they are a horrible mistake and should be removed.

matkoniecz (changes requested):

data/presets/power/terminal.json
nlehuby: I'm not really sure I understand your comment :thinking: 

It's not a depreciation rule, it's an additional tag that is relevant (and recommended) to add to this object, and it seems to me that this corresponds perfectly to what addTags is for. 
image

I don't understand why you suggest not adding it.

matkoniecz: it it needs to be mass added to 64k objects (see https://overpass-turbo.eu/s/2hpJ ) then it should be done with a proper bot edit, rather than undiscussed distributed bot edit with people adding it via "upgrade" button in iD

(also, if it absolutely always applies then I do not get what is the point of adding it to every object)

nlehuby: Ok, I understand your point. I guess it's legitimate to discuss the relevance of a massive import or, conversely, the relevance of having this attribute at all if it's an implied default value for this type of object, but it seems to me that the presets should be based on the current wiki documentation and that the forum would be a more appropriate place for the debate than this issue.

Anyway, this addition is on the critical path to adding a quality control rule that has been expected for years, and the addition itself has already been the subject of several issues & even PR, so I would like to move forward without starting a new debate... So I am removing this field entirely for the moment.

matkoniecz (commented):
I tested it and it works.

But maybe more clear name exists?

Would "Power line termination" be correct and more clear?

data/presets/power/terminal.json
```diff
@@ -0,0 +1,26 @@
...
+        "electrical terminal",
+        "service entrance"
+    ],
+    "tags": {
+        "power": "terminal"
+    },
+    "addTags": {
+        "power": "terminal"
+    },
+    "name": "Power Terminal"
```
matkoniecz:
```suggestion
    "name": "Power line reaching building"
```

better name is welcome

@ZeLonewolf maybe you have better idea? Or can confirm that this one is not Engrish/Ponglish?

ZeLonewolf:
I don't understand what a power reaching building is, can someone post a picture?

flacombe:
They more relate to line_management=termination and proposed line_management=service_supply tags than power=terminal imho.

_Power line reaching building_ looks like the more accurate for power=terminal

matkoniecz:
not 100% sure are further changes needed - if optimal name is set already please ping me

matkoniecz:
As I proposed the current name in https://github.com/openstreetmap/id-tagging-schema/pull/1637#issuecomment-3878047065 and there were some protests that it is not optimal I want to ping @k-yle @tordans for an independent review of comments on that.

data/presets/power/terminal.json
```diff
@@ -0,0 +1,26 @@
...
+        "electrical terminal",
+        "service entrance"
+    ],
+    "tags": {
+        "power": "terminal"
+    },
+    "addTags": {
+        "power": "terminal"
+    },
+    "name": "Power line reaching building"
```
tyrasd: I'm also not sure if this is the best possible label for this preset (I'm also not a native speaker), but independently from that: the label should always use title case spelling, e.g. `"Power Line Reaching Building"`

nlehuby:
I've updated the label as suggested by tyrasd.

data/presets/power/terminal.json
```diff
@@ -0,0 +1,26 @@
...
+        "electrical terminal",
+        "service entrance"
+    ],
+    "tags": {
+        "power": "terminal"
+    },
+    "addTags": {
+        "power": "terminal"
+    },
+    "name": "Power Line Reaching Building"
```
1ec5, line 25: A `name` needs to be a noun. This is a gerund phrase, which is very awkward. Even more awkward is that it implies a line feature, but this can only be a point or vertex. I can only imagine what’ll happen when this gets translated into other languages.

I think the intent of this name is “Point Where Power Line Reaches Building”, which is very verbose. As a layperson, it still sounds equivalent to “Power Termination Point”, which was [rejected previously](https://github.com/openstreetmap/id-tagging-schema/pull/1637#issuecomment-3929021540). “[Building Power Anchor](https://github.com/openstreetmap/id-tagging-schema/pull/1637#issuecomment-3929129779)” is at least a better attempt at circumlocution. It sounds like a real term of art, even if it isn’t.

flacombe:
Hello everyone

The effort to find the most appropriate term is really appreciable and will certainly allow beginners to get it well.
In JOSM, which is intended for more expercienced mappers, we kept Power terminal and no one argue about it.
image

Should we give it a try and see if any comment rises there?

This to also allow https://github.com/openstreetmap/iD/pull/11200 to find its way and get merged too.
[Many quality issues](https://osmose.openstreetmap.fr/fr/issues/graph.png?item=1250&class=1) happen due to this long time we take here.

1ec5:
@matkoniecz pointed out in chat that “Power Terminal” could be confused for some other meaning of “terminal”. “Power Termination Point” would avoid this ambiguity. As a layperson, I don’t know why “terminal” would be fine but “termination” would be problematic. Maybe we could address that when we add support for `line_management=*`?

matkoniecz (commented):
Sorry for complaining about label I himself suggested but https://github.com/openstreetmap/id-tagging-schema/pull/1637#pullrequestreview-4639022626 makes a good point - maybe reconsider “Power Termination Point” or “Building Power Anchor” ?

https://github.com/openstreetmap/id-tagging-schema/pull/1637#issuecomment-3929021540 mentions `line_management=termination` but this will never be a preset, right?

maybe even absurdly verbose “Point Where Power Line Reaches Building” would be better?

Or maybe what we have now would be fine but I worry that it is more German/French and Polish than English. Or at least some sort of Engrish.

matkoniecz:
@nlehuby have you seen https://github.com/openstreetmap/id-tagging-schema/pull/1637#pullrequestreview-4855641834 ?

matkoniecz:
@nlehuby can you respond? I am considering taking this PR as base. making few adjustments and submitting it separately (original author would be credited)

nlehuby:
Hi, I am on vacation mostly afk, I won't have time to look at this again for a few weeks.

Le 12 septembre 2026 11:47:59 GMT+02:00, Mateusz Konieczny ***@***.***> a écrit :
>matkoniecz left a comment (openstreetmap/id-tagging-schema#1637)
>
>@nlehuby have you seen
>https://github.com/openstreetmap/id-tagging-schema/pull/1637#pullrequestreview-4855641834
>?
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/openstreetmap/id-tagging-schema/pull/1637#issuecomment-5645120765
>You are receiving this because you were mentioned.
>
>Message ID:
>***@***.***>

-- 
Envoyé de mon appareil Android avec Courriel K-9 Mail. Veuillez excuser ma brièveté.

matkoniecz:
No problem! Thanks for reply with update, I will try to process other PRs then (>100 just in iD tagging schema, more in iD repo)

PR #1648, opened 2025-07-30
Add area geometry support to runway preset

### LINKED ISSUE #11169: Render area:aeroway
### Description

Render aeroway mapped as area with [`area:aeroway`](https://taginfo.openstreetmap.org/keys/area:aeroway), mainly `area:aeroway=runway` for runways.
Render as area, identical to existing rendering when `aeroway=runway` is combined with `area=yes`.
(Old rendering should stay for the time being as the tag is still in transition, or might go into transition.)

See https://wiki.openstreetmap.org/wiki/Tag:aeroway%3Drunway for information

### Screenshots

Example https://www.openstreetmap.org/way/1410575668
Image

should render as
Image

`aeroway=runway` on its own should only render when mapped as way (centerline), of course.

youssefelzedy:
### Summary
Adds support for area geometry to the runway preset to enable mapping of runway surface areas in addition to centerlines.

### Changes
- Added `"area"` to supported geometry types in runway preset
- Added `addTags` and `removeTags` configuration for proper tag management
- Maintains backward compatibility with existing line-based runway mapping
- Confirmed runway preset is already discoverable through existing area category (aeroway is included in area.json members)

### Motivation
Addresses issue #11169 which requested support for `area:aeroway=runway` tagging to distinguish between:
- **Line geometry**: Runway centerline (`aeroway=runway`)
- **Area geometry**: Runway surface area (`area:aeroway=runway`)

### Files Modified
- `data/presets/aeroway/runway.json` - Added area geometry support and tag management
- `data/preset_categories/area.json` - Verified aeroway category inclusion (no changes needed)

### Implementation Notes
This schema change provides the foundation for geometry-specific tagging. The actual logic to apply `area:aeroway=runway` for area geometries (vs `aeroway=runway` for lines) will need to be implemented in iD editor's JavaScript code, specifically in the `setTags` method of `modules/presets/preset.js`.

### Related
Fixes openstreetmap/iD#11169


data/preset_defaults.json
```diff
@@ -9,6 +9,7 @@
         "amenity/place_of_worship",
         "amenity/cafe",
         "amenity/restaurant",
+        "aeroway/runway",
```
k-yle: the purpose of this file is to define the default presets which are shown to new users who have no "recently selected" presets:


image

So I think the changes to this file are unrelated to the goal of this PR?

data/presets/aeroway/runway.json
```diff
@@ -7,11 +7,19 @@
...
     ],
     "tags": {
         "aeroway": "runway"
     },
+    "addTags": {
+        "aeroway": "runway"
+    },
+    "removeTags": {
+        "aeroway": "*",
+        "area:aeroway": "*"
```
k-yle: a simpler solution would be to leave the existing preset untouched, and create a single new file called `runway_area.json`.

[`pedestrian_line.json`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/highway/pedestrian_line.json) and [`pedestrian_area.json`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/highway/pedestrian_area.json) is a good example

data/preset_categories/area.json
```diff
@@ -0,0 +1,10 @@
+{
```
matkoniecz: what is the expected result of adding this file?

why this new file is needed?

matkoniecz: @youssefelzedy ping?

youssefelzedy: area.json is not adding a new tag or changing map rendering by itself. It defines a new preset category in the iD tagging schema so area-based presets can be grouped together in the editor UI.

In this PR, the category is needed so the new area-based aeroway preset, especially aeroway=runway with area:aeroway=*, has a proper place in the schema. Without this file, the preset can still exist, but it would not be organized under an Area category in iD, which makes it harder for users to discover and use.

- iD shows an Area category in the preset picker.
- Area-based aeroway presets, including runway areas, are grouped there.
- This helps users find and apply the correct area representation for aeroway=runway while keeping the existing aeroway=runway centerline behavior unchanged.

Sorry for the late response!

matkoniecz:
> @[youssefelzedy](https://github.com/youssefelzedy) youssefelzedy requested a review from [matkoniecz](https://github.com/matkoniecz) [6 hours ago](https://github.com/openstreetmap/id-tagging-schema/pull/1648#event-24732948621)

https://github.com/openstreetmap/id-tagging-schema/pull/1648#pullrequestreview-3682179543 remains unprocessed for last three months - why this file got added?

matkoniecz:
And https://github.com/openstreetmap/id-tagging-schema/pull/1648#issuecomment-3149884209 waits since Aug 4, 2025

youssefelzedy:
> [#1648 (review)](https://github.com/openstreetmap/id-tagging-schema/pull/1648#pullrequestreview-3682179543) remains unprocessed for the last three months - why was this file got added?

@matkoniecz The functional part is runway_area.json, which adds the area runway preset. area.json only affects preset grouping in the UI. 

If you prefer keeping this PR strictly scoped to runway-area support, I can drop the category file and keep only the runway preset change.

If you meant the Kyle review

I added aeroway/runway to preset_defaults.json because I thought runway should be easier to discover after adding area support, but I agree that this is a separate default-presets UX change and not part of the core area-runway support.

data/preset_categories/area.json
matkoniecz: was answered in https://github.com/openstreetmap/id-tagging-schema/pull/1648#issuecomment-4300119429 - I missed it

matkoniecz (changes requested):
please remove new "Area" category - grouping random selection of subgroup of categories does not seem useful in any way

(note, I still need to check tagging situation with area tagging of runways - hopefully it is not controversial or confused)

data/presets/aeroway/runway_area.json
```diff
@@ -0,0 +1,27 @@
...
+        "ref_runway",
+        "surface",
+        "length",
+        "width"
+    ],
+    "geometry": [
+        "area"
+    ],
+    "tags": {
+        "aeroway": "runway"
```
k-yle, line 13: this is the wrong tag according to the linked issue. 
```suggestion
        "area:aeroway": "runway"
```

also `addTags` and `removeTags` is redundant

k-yle (changes requested):

matkoniecz (commented):
see above

k-yle:
i have no permission to [dismiss my previous review](https://docs.github.com/en/pull-requests/how-tos/review-pull-requests/dismissing-a-pull-request-review). but the comment was addressed, thanks!

image

matkoniecz (commented):
was addressed, dismissing my review

wait, it has not worked

data/presets/aeroway/runway_area.json
```diff
@@ -0,0 +1,20 @@
...
+    "geometry": [
+        "area"
+    ],
+    "tags": {
+        "area:aeroway": "runway"
+    },
+    "terms": [
+        "landing strip"
+    ],
+    "name": "Runway (area)",
```
matkoniecz, line 18:
```suggestion
    "name": "Runway (additional marking as an area)",
```
????

not entirely sure is it a good idea, but I really want to indicate it is an optional extra

youssefelzedy: It seems too long.. we can rename it to `area marking` or `additional marking area`?

and can you explain more about why you want to add this?

PR #1650, opened 2025-08-02
Added a preset for motorcycle parts, aliases for optician & greengrocer

NLBRT:

a. Adding `shop=motorcycle_parts`

There are a whole bunch of shops, all over places where motorcycles are predominantly used, such as India, Vietnam etc., 
where they exclusively sell parts for repairing and maintaining motorcycles.

b. Aliases for `shop=optician` and `shop=greengrocer`

Added aliases which are more commonly used in Indian English for better understanding.

### Related issues

N/A (Didn't see any yet, will update if I see one)

### Links and data

**Relevant OSM Wiki links:**

[`shop=motorcycle_parts`](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dmotorcycle_parts)

**Relevant tag usage stats:**

https://taginfo.openstreetmap.org/tags/shop=motorcycle_parts
https://taginfo.openstreetmap.in/tags/shop=motorcycle_parts

### DIFF
--- a/data/presets/shop/greengrocer.json
+++ b/data/presets/shop/greengrocer.json
@@ -22,6 +22,8 @@
     "name": "Greengrocer",
     "aliases": [
         "Produce Store",
-        "Produce Shop"
+        "Produce Shop",
+        "Vegetable Shop",
+        "Vegetable Store"
     ]
 }
--- a/data/presets/shop/motorcycle_parts.json
+++ b/data/presets/shop/motorcycle_parts.json
@@ -1,0 +1,26 @@
+{
+    "icon": "temaki-motorcycle_repair",
+    "fields": [
+        "{shop}",
+        "service/vehicle"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "auto",
+        "bike",
+        "part",
+        "motorcycle",
+        "motorbike",
+        "parts"
+    ],
+    "tags": {
+        "shop": "motorcycle_parts"
+    },
+    "name": "Motorcycle Parts Shop",
+    "aliases": [
+        "Motorcycle Parts Store"
+    ]
+}
--- a/data/presets/shop/optician.json
+++ b/data/presets/shop/optician.json
@@ -12,4 +12,7 @@
         "shop": "optician"
     },
     "name": "Optician"
+    "aliases": [
+        "Eyewear Store"
+    ]
 }


tordans:
Usage is not that hight. But if it works well with the other motorcycle presets in search and such I think we can merge this even below 1k usage to avoid misstagging. 
Please add the testing documentation, especially around the help Text and search.

data/presets/shop/motorcycle_parts.json
```diff
@@ -0,0 +1,28 @@
...
+        "parts"
+    ],
+    "tags": {
+        "shop": "motorcycle_parts"
+    },
+    "name": "Motorcycle Parts Store",
+    "aliases": [
+        "Motorcycle Parts Shop",
+        "Motorbike Parts Store",
+        "Bike Parts Store"
```
matkoniecz, line 26: "Bike Parts Store" - can it be interpreted as referring to bicycles rather motorcycles? Maybe move it to terms so mapper can see what was selected?

data/presets/shop/optician.json
```diff
@@ -11,5 +11,8 @@
     "tags": {
         "shop": "optician"
     },
-    "name": "Optician"
+    "name": "Optician",
+    "aliases": [
+        "Eyewear Store"
```
matkoniecz, line 16: actually, are sunglasses etc supposed to be tagged as shop=optician if shop sells only sunglasses and no vision-correcting glasses?

NLBRT: I don't think so. There's `shop=fashion_accessories` for that.

However, I added the alias based on my observation in my region that a lot of these shops, be it selling just vision-correcting glasses or with sunglasses and/or "zero-power glasses", calls themselves an "Eyewear Store(/Shop)".

matkoniecz: does search for Eyewear Store brings up also `shop=fashion_accessories` if these shops can be also described in this way?

data/presets/shop/motorcycle_parts.json
matkoniecz: @ZeLonewolf Do you think that "Bike Parts Store" can be understood to be about bicycles?

data/presets/shop/greengrocer.json
```diff
@@ -22,6 +22,8 @@
     "name": "Greengrocer",
     "aliases": [
         "Produce Store",
-        "Produce Shop"
+        "Produce Shop",
+        "Vegetable Shop",
```
matkoniecz, line 26: @ZeLonewolf is "Vegetable Shop" a term in US English?

(BTW, "produce shop" matches already)

matkoniecz: @1ec5 may I summon you?

data/presets/shop/motorcycle_parts.json
matkoniecz: @1ec5 may I summon you?

matkoniecz:
Did research pointing me quite firmly toward these terms not working in US English.

PR #1674, opened 2025-08-17
Allow mapping amenity=lounger as a line/area

TheComputerGuy96:
The OSM wiki mentions these as supported element types (and projects listed in Taginfo aren't against them either)

### DIFF
--- a/data/presets/amenity/lounger.json
+++ b/data/presets/amenity/lounger.json
@@ -14,7 +14,9 @@
         "operator"
     ],
     "geometry": [
-        "point"
+        "point",
+        "line",
+        "area"
     ],
     "tags": {
         "amenity": "lounger"


matkoniecz:
I am planning to close it as an abandoned in about 100 days if no reply will arrive

TheComputerGuy96:
> does OSM Wiki make sense?

I guess it does (considering weirder objects like bus route relations have been de facto for a while)

tordans:
> So do you think this lounger can be mapped as an area?

Personally, I would map this as a point lounger and a second point bench+backrest, both with direction.

---

But the more important thing is IMO: This repo provides a curated list of tagging recommendations to help users get going with common tagging tasks quickly. So it is not a goal to include all possible tagging options as a preset. Users can and should still use good old manual tagging for less common cases.

So personally, I don't see a reason to expand the preset to lines and ares at this point, given the low usage numbers. We can and should, of course, revisit this when numbers changed.

There is, however, a missing feature, IMO, that would soften the fact that we only showcase this preset for some geometries, which I just wrote about in https://github.com/ideditor/schema-builder/issues/244

PR #1684, opened 2025-08-23
Changed landuse=education preset name from Educational Campus to Educational Area

ilias52730:
Just like `landuse=residential` is a _residential area_, `landuse=religious` is a religious area and `landuse=retail` is a _retail area_, `landuse=education` is logically an _educational area_.

Apart from this logic, the [original proposal](https://wiki.openstreetmap.org/w/index.php?oldid=2261291) said it should be **mainly** be used for areas shared by 2 or more schools. See an example:
image

This tag _can_ also be used for single-school areas, even though it's not its main use.

In any case, this should logically be called an "Educational Area", as I think it's, broader, yes, but still clearer than Educational Campus.

### DIFF
--- a/data/presets/landuse/education.json
+++ b/data/presets/landuse/education.json
@@ -9,8 +9,8 @@
     ],
     "terms": [
         "campus",
-        "educational facilities",
         "educational campus",
+        "educational facilities",
         "school campus",
         "school center",
         "school grounds"


data/presets/landuse/education.json
```diff
@@ -18,5 +18,5 @@
         "landuse": "education"
     },
     "matchScore": 0.9,
-    "name": "Educational Campus"
+    "name": "Educational Area"
```
1ec5, line 22: Looks like we do have “campus” as one of the search terms, but at least “school campus” and “educational campus” should be there too for discoverability.

ilias52730: [Done](https://github.com/openstreetmap/id-tagging-schema/pull/1684/commits/10e6671dbf101264688143753bb5e3f8157012e4).

matkoniecz (approved):
though it run into some (harmlessly looking) edit conflict

PR #1696, opened 2025-09-02
Add fields to Huts/shelters

paulklie:

The fee, shower, toilets, fire place and drinking water tags are commonly used by shelter and wilderness huts.
### Links and data

**Relevant OSM Wiki links:**
[drinking_water](https://wiki.openstreetmap.org/wiki/Key:drinking_water)
[reservation](https://wiki.openstreetmap.org/wiki/Key:reservation)
[drinking_water](https://wiki.openstreetmap.org/wiki/Key:drinking_water)
[toilets](https://wiki.openstreetmap.org/wiki/Key:toilets)
[capacity](https://wiki.openstreetmap.org/wiki/Key:capacity)

**Relevant tag usage stats:**

*Lean-to*
6.61% (1 314) have fee with 540 fee=yes, even though the wiki suggests the opposite. fee=yes is mostly used in Denmark.
8.66% (1966) have fireplace set with 1722 fireplace = yes.
2.2% (438) have  "toilets" set
1.9% (358)  have "capacity" set
6.1% (1138) have   "reservation"
1.7% (249) have picnic_table
1.7% (225) have drinking_water

I also added the term "wind shelter" which is often used to describe lean-to's

*Wilderness_hut*
10.13 %(1590) have toilets set
606  have shower set.

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>
image

### DIFF
--- a/data/presets/amenity/shelter/lean_to.json
+++ b/data/presets/amenity/shelter/lean_to.json
@@ -3,7 +3,17 @@
     "fields": [
         "name",
         "operator",
-        "building_area_yes"
+        "building_area_yes",
+        "fee",
+        "fireplace",
+        "toilets",
+        "capacity"
+    ],
+    "moreFields": [
+        "drinking_water_available",
+        "picnic_table",
+        "reservation",
+        "shower"
     ],
     "geometry": [
         "point",
@@ -20,7 +30,8 @@
         "lodging",
         "overnight accommodations",
         "sleeping shelter",
-        "wilderness hut"
+        "wilderness hut",
+        "wind shelter"
     ],
     "name": "Lean-To"
 }
--- a/data/presets/tourism/wilderness_hut.json
+++ b/data/presets/tourism/wilderness_hut.json
@@ -6,6 +6,7 @@
         "building_area_yes",
         "access_simple",
         "fee",
+        "toilets",
         "payment_multi_fee",
         "charge_fee",
         "fireplace"
@@ -16,6 +17,7 @@
         "capacity",
         "drinking_water_available",
         "gnis/feature_id-US",
+        "shower",
         "kitchen",
         "mattress",
         "reservation",


data/presets/amenity/shelter/lean_to.json
```diff
@@ -3,7 +3,17 @@
...
-        "building_area_yes"
+        "building_area_yes",
+        "fee",
+        "fireplace",
+        "toilets",
+        "capacity"
+    ],
+    "moreFields": [
+        "drinking_water_available",
+        "picnic_table",
```
tordans, line 14: With tags like this I wonder, if we actually want to promote them or rely on separately mapped objects instead? The general rule would be "one feature one object", so those "has bench", "has toilets (outside)" on features like here and eg. more commonly the bus stop nodes tend to confuse me. Should I click "has picnic table" when the picnic table" is outside? Or only when there is a table inside? If the second, are our labels clear on that?
That is a more general issue we have with these attributes in OSM and the presets … however adding the fields will increase the issue, which is why I bring it up.

data/presets/amenity/shelter/lean_to.json
```diff
@@ -3,7 +3,17 @@
     "fields": [
         "name",
         "operator",
-        "building_area_yes"
+        "building_area_yes",
+        "fee",
+        "fireplace",
+        "toilets",
```
tordans, line 9: For both see below. Is this only for "fireplace inside" and "toilet inside" or "somewhere nearby"? Also, what does a toilet inside a (open) lean to look like; is that a thing?

data/presets/amenity/shelter/lean_to.json
```diff
@@ -3,7 +3,17 @@
...
+        "capacity"
+    ],
+    "moreFields": [
+        "drinking_water_available",
+        "picnic_table",
+        "reservation",
+        "shower"
```
tordans, line 16: Same here.

tordans (commented):
Could you add a few testing links for the features affected here and related features for comparison.

---

In general I wonder if now would be a good time to look into sharing the fields and tags between the different shelter presets. There is also https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/tourism/alpine_hut.json and maybe others…

---

I left some comment to clarify how we interpret some (new and existing) fields.

data/presets/amenity/shelter/lean_to.json
paulklie: I have never seen a toilet inside of a lean to. I understand this as "is there a toilet nearby" situation. Fireplaces are often inside.

data/presets/amenity/shelter/lean_to.json
paulklie: I think that adding this information to the shelter helps users filter/search better for shelters that fit their needs. And we are not really adding additional elements, but simply more information to existing ones. Also we kind of already do this, for example, with bathing place (https://github.com/openstreetmap/id-tagging-schema/blob/ff66003e33074b241b2ecaba240b5a128c1dce26/data/presets/leisure/bathing_place.json#L15-L18),

matkoniecz: maybe better description/labeling this fields would help to indicate that you can use both methods or one of them?

data/presets/amenity/shelter/lean_to.json
matkoniecz: I always understood it as "this facility is provided for XYZ feature"

for example `lit=yes` on bus stop does not mean that source of light is inside bus stop (or even existing specifically to lit bus stop)

for `toilets=yes` as long as there is toilet provided for object it would qualify

matkoniecz, line 9: I am wondering whether it would make more sense in `moreFields` and to push this tagging a bit less.

matkoniecz (changes requested):

data/presets/amenity/shelter/lean_to.json
tordans: > for example lit=yes on bus stop does not mean that source of light is inside bus stop (or even existing specifically to lit bus stop)
> 
> for toilets=yes as long as there is toilet provided for object it would qualify

I know this is not the places for this, but IMO we need to change this definition. As a data consumer, I have to be able to distinguish between "this is a thing" and "this has a thing somewhere but I want to tell you about it here".
Luckily we have a system for this, its the value `separate` which we use to say "`yes`, but somewhere else".
So if the bus stop does not have light, but lights are mapped separately and relevant here, this would be `lit=separate`. Same for `toilets=separate`.

How else would I be able to say "this are has two toilets", when I have data that is a separate toilet and an attribute on something…

matkoniecz: ouch, `=separate` redefinition is on my list of most disliked tagging ideas

problem here is suddenly making huge amount `=yes` wrong, and what worse now mapping previously not marked toilet requires fiddling with `=yes` values (admittedly already `=no` values need to be flipped).

matkoniecz: @paulklie what you think about `moreFields` ?

paulklie: only toilets?  
I think that would work, I would like to keep the toilets field in the wilderness hut though since toilets are more likely to be "inside" for these.

matkoniecz: > only toilets?

as in putting only toilets there or keeping only toilets in `fields`?

PR #1697, opened 2025-09-03
Fix: update payment field to correct Wikidata item (Q550)

srujanaelicherla:

The payment field was pointing to the wrong Wikidata item (Q22063, “payment card”) instead of the correct one (Q550, “payment”).
This fix updates the reference to ensure the schema correctly aligns with Wikidata.

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Key:payment

**Relevant tag usage stats:**
> Taginfo: https://taginfo.openstreetmap.org/keys/payment

<details>

### Info-`i`

- Verified that the `payment` field now points to Q550. 

</details>

### DIFF
--- a/data/fields/payment_multi.json
+++ b/data/fields/payment_multi.json
@@ -2,6 +2,7 @@
     "key": "payment:",
     "type": "multiCombo",
     "label": "Payment Types",
+    "wikidata": "Q550",
     "strings": {
         "options": {
             "account_cards": "Account Card",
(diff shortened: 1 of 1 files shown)


matkoniecz:
@tyrasd Can you approve workflow?

srujanaelicherla:
Hi @matkoniecz , just an update: I had added "wikidata": "Q550", but since the schema doesn’t allow wikidata in field definitions, I’ve removed it so validation passes. The underlying issue (wrong item Q22063 vs. Q550) likely needs to be fixed on the iD side instead of here. Could you confirm the right place for the fix?

Thank you!

matkoniecz:
I have no idea what is causing this bug and have not tried to investigate it

PR #1701, opened 2025-09-04
fields: reference payment_multi to 'payment:*' for correct wiki link

srujanaelicherla:
This PR adds

"reference": { "key": "payment:*" } to the payment_multi field.

This ensures that iD’s Tag Reference links correctly to the Key:payment wiki page instead of pointing to the unrelated Wikidata item for “payment card”.

Validation and build pass.
Only payment_multi.json was modified.

This addresses the root cause of [iD #11385](https://github.com/openstreetmap/iD/pull/11385?utm_source=chatgpt.com) without requiring code-level workarounds.

### DIFF
--- a/data/fields/payment_multi.json
+++ b/data/fields/payment_multi.json
@@ -2,6 +2,7 @@
     "key": "payment:",
     "type": "multiCombo",
     "label": "Payment Types",
+    "reference": { "key": "payment:*" },
     "strings": {
         "options": {
             "account_cards": "Account Card",


matkoniecz (approved):
there is complain about code formatting, but this PR is fixing mentioned according to my testing

thanks for tracking it down!

matkoniecz:
if merged as is then I can make followup PR running that formatting fix

srujanaelicherla:
@tordans @matkoniecz 
I’ve run Prettier and updated the formatting of payment_multi.json.
The file now follows the project’s code style conventions, and all checks should pass.

Thanks for the feedback and guidance!

tordans (approved):

tordans:
Thanks for [tracking this down](https://github.com/openstreetmap/iD/pull/11385#discussion_r2321432151) @k-yle and preparing the change @srujanaelicherla 

Ideally, this PR would also have fixed https://github.com/openstreetmap/id-tagging-schema/pull/1703 but I made a saw that too late and made a follow up.

PR #1702, opened 2025-09-05
Add wheelchair field to shop=health_food preset

### LINKED ISSUE #1262: Add wheelchair to shop=health_food
Add `wheelchair` to `shop=health_food`. I think this is a basic tag and every shop should have it.

srujanaelicherla:
Description, Motivation & Context

This PR adds the wheelchair field to the shop=health_food preset.
It ensures consistency with other shop presets (e.g., shop=bakery) and allows contributors to easily record accessibility information for health food shops.

Related issues
Closes #1262

Links and data

Relevant OSM Wiki links:
Key:wheelchair
Tag:shop=health_food

Relevant tag usage stats:
shop=health_food
wheelchair

### DIFF
--- a/data/fields/craft.json
+++ b/data/fields/craft.json
@@ -5,6 +5,7 @@
     "strings": {
         "options": {
             "agricultural_engines": "Agricultural Engines Mechanic",
+            "atelier": "Atelier",
             "basket_maker": "Basket Maker",
             "beekeeper": "Beekeeper",
             "blacksmith": "Blacksmith",
@@ -24,7 +25,6 @@
             "electronics_repair": "Electronics Repair",
             "electrician": "Electrician",
             "floorer": "Floorer",
-            "builder": "Builder",
             "gardener": "Gardener",
             "glaziery": "Glaziery",
             "grinding_mill": "Grinding Mill",
--- a/data/presets/shop/health_food.json
+++ b/data/presets/shop/health_food.json
@@ -19,7 +19,8 @@
         "shop": "health_food"
     },
     "moreFields": [
-        "fhrs/id-GB"
+        "fhrs/id-GB",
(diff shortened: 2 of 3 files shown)


data/fields/craft.json
```diff
@@ -5,6 +5,7 @@
     "strings": {
         "options": {
             "agricultural_engines": "Agricultural Engines Mechanic",
+            "atelier": "Atelier",
```
matkoniecz: I see unrelated changes

matkoniecz (changes requested):

data/fields/craft.json
srujanaelicherla: Thank you for pointing that out! I have made the necessary changes.

srujanaelicherla:
Hi @matkoniecz Thanks for checking! I confirmed that shop.json (the parent for shop presets) does not currently include the general wheelchair field—only toilets/wheelchair. That means shop=bakery doesn’t have it either.

So this PR is filling the gap for shop=health_food, and it could make sense to extend wheelchair to other shop presets like bakery in a follow-up.

data/presets/shop/health_food.json
```diff
@@ -19,7 +19,8 @@
         "shop": "health_food"
     },
     "moreFields": [
-        "fhrs/id-GB"
+        "fhrs/id-GB",
+        "wheelchair"
```
tyrasd:
```suggestion
        "{shop}",
        "fhrs/id-GB"
```

tyrasd (approved):
Changed the fix to inherit all of the `shop` preset's optional fields. This was a bug we introduced accidentally in #987. Note to self: There are a handful of similar presets which have the same issue (e.g. `shop=farm`, etc.) from that PR.

PR #1703, opened 2025-09-05
Fix reference `payment:*` for `payment_multi_fee` field

tordans:
Same as https://github.com/openstreetmap/id-tagging-schema/pull/1701

@matkoniecz could you review / approve this, then I can merge.

Test case:
1. Preview
2. Create node `leisure=pitch`
3. Add payment field from more fields
4. Should be right
   - Current/Wrong: image

### DIFF
--- a/data/fields/payment_multi_fee.json
+++ b/data/fields/payment_multi_fee.json
@@ -6,5 +6,8 @@
         "key": "fee",
         "valueNot": "no"
     },
-    "stringsCrossReference": "{payment_multi}"
+    "stringsCrossReference": "{payment_multi}",
+    "reference": {
+        "key": "payment:*"
+    }
 }


tordans:
## Test documentation

image

data/fields/payment_multi_fee.json
matkoniecz, line 1: Is there some reason why fixing payment_multi referenced here cannot work?

tordans: AFAIK only some properties can use the reference logic and they all have a paragraph on this in the docs https://github.com/ideditor/schema-builder but `reference` does not, so I assume it does not have this ability.

matkoniecz (approved):

tordans:
Fixed by https://github.com/openstreetmap/iD/pull/11402

PR #1705, opened 2025-09-05
added highway=hitchhiking preset

### LINKED ISSUE #1675: Add highway=hitchhiking
### OSM Tag(s)

highway=hitchhiking

As a Preset

### Label

Hitchhiking spot

### Aliases

hitching, thumbing

### Terms

_No response_

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:highway=hitchhiking

### Status of the Tag

In Use

### Usage of the tag

670

### Suggested Icon

fa-thumbs-up

### Replaces other Tag?

_No response_

### Regional Tag?

_No response_

### Further Information

Used to mark hitchhiking spots. Recently merged from multiple similar tags: https://wiki.openstreetmap.org/wiki/Mechanical_Edits/TillWenke/moving_hitchhiking_stops_to_%27highway%3Dhitchhiking%27_(mainly_in_Europe/Germany)

paulklie:

I have added the "shelter", "bench" and "bin" fields since these places seem similar to bus stops, and they seem to have decent usage.
I have also changed the name from the original proposal to "**designated** hitchhiking place" to better clarify that this is an official place.

### Related issues
Closes #1675 

### Links and data

**Relevant OSM Wiki links:**
[ hitchhiking ](https://wiki.openstreetmap.org/wiki/Tag:highway=hitchhiking)

**Relevant tag usage stats:**
670

**This is how it works:**

```

```

</details>

### DIFF
--- a/data/presets/highway/hitchhiking.json
+++ b/data/presets/highway/hitchhiking.json
@@ -2,6 +2,7 @@
     "fields": [
         "bench",
         "bin",
+        "informal",
         "shelter"
     ],
     "geometry": [
--- a/data/presets/highway/hitchhiking.json
+++ b/data/presets/highway/hitchhiking.json
@@ -16,14 +16,14 @@
     ],
     "locationSet": {
         "include": [
-            "fx",
-            "de",
+            "at",
             "ch",
-            "nl",
+            "de",
             "dk",
-            "au",
+            "fx",
             "it",
-            "li"
+            "li",
+            "nl"
         ]
     },
     "name": "Signed Hitchhiking Place",
--- a/data/presets/highway/hitchhiking.json
+++ b/data/presets/highway/hitchhiking.json
@@ -1,0 +1,18 @@
+{
+    "fields": [
+    "bench",
+    "bin"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "highway": "hitchhiking"
+    },
+    "terms": [
+        "hitchhiking"
+    ],
+    "name": "Designated Hitchhiking Place",
+    "icon": "fas-sign-hanging"
+}
--- a/data/presets/highway/hitchhiking.json
+++ b/data/presets/highway/hitchhiking.json
@@ -1,18 +1,19 @@
 {
     "fields": [
-    "bench",
-    "bin"
+        "bench",
+        "bin",
+        "shelter"
     ],
     "geometry": [
-        "point",
-        "area"
+        "point"
     ],
     "tags": {
         "highway": "hitchhiking"
     },
     "terms": [
-        "hitchhiking"
+        "hitchhiking",
+        "thumbing"
     ],
-    "name": "Designated Hitchhiking Place",
+    "name": "Signed Hitchhiking Place",
     "icon": "fas-sign-hanging"
 }
--- a/data/presets/highway/hitchhiking.json
+++ b/data/presets/highway/hitchhiking.json
@@ -15,5 +15,5 @@
         "thumbing"
     ],
     "name": "Signed Hitchhiking Place",
-    "icon": "fas-sign-hanging"
+    "icon": "far-thumbs-up"
 }
--- a/data/presets/highway/hitchhiking.json
+++ b/data/presets/highway/hitchhiking.json
@@ -14,6 +14,18 @@
         "hitchhiking",
         "thumbing"
     ],
+    "locationSet": {
+        "include": [
+            "fx",
+            "de",
+            "ch",
+            "nl",
+            "dk",
+            "au",
+            "it",
+            "li"
+        ]
+    },
     "name": "Signed Hitchhiking Place",
     "icon": "far-thumbs-up"
 }


paulklie:
## Test-Documentation

image



### Search

image
image

### Info-`i`

image

### Wording

- [x] American English

tordans:
> designated hitchhiking place

Maybe "Signed hitchhiking place" would be an option. The feature we map represents the sign and what it stands for and should be placed where the sign is.

But happy with the current name as well.

tordans:
@paulklie please have a look at the CI error

paulklie:
> > designated hitchhiking place
> 
> Maybe "Signed hitchhiking place" would be an option. The feature we map represents the sign and what it stands for and should be placed where the sign is.
> 
> But happy with the current name as well.

Good idea, I prefer signed to.
Btw, I was unable to use the thumb icon as originally planned as it is not in the repo.

matkoniecz:
@tordans can you approve workflow run?

data/presets/highway/hitchhiking.json
```diff
@@ -0,0 +1,19 @@
...
+    ],
+    "tags": {
+        "highway": "hitchhiking"
+    },
+    "terms": [
+        "hitchhiking",
+        "thumbing"
+    ],
+    "name": "Signed Hitchhiking Place",
+    "icon": "fas-sign-hanging"
```
1ec5: The icon would also appear without any caption in the map part of the editor, where users might mistake it for a different feature such as `office=estate_agent`, `information=sign`, or `advertising=sign`.

> Btw, I was unable to use the thumb icon as originally planned as it is not in the repo.

This repository doesn’t store any icons, but you can use any icon from [several icon repositories](https://github.com/ideditor/schema-builder/blob/main/ICONS.md#where-do-the-icons-come-from). FontAwesome’s [thumbs-up](https://fontawesome.com/icons/thumbs-up) looks a bit more intuitive, assuming that a raised thumb is associated with hitchhiking in enough cultures.

data/presets/highway/hitchhiking.json
```diff
@@ -0,0 +1,19 @@
...
+        "point"
+    ],
+    "tags": {
+        "highway": "hitchhiking"
+    },
+    "terms": [
+        "hitchhiking",
+        "thumbing"
+    ],
+    "name": "Signed Hitchhiking Place",
```
1ec5, line 30: > Maybe "Signed hitchhiking place" would be an option. The feature we map represents the sign and what it stands for and should be placed where the sign is.

Not to bikeshed the name too much, but to me “Hitchhiking” would already be descriptive enough. The help documentation button can take care of clarifying when or where to place the point, just as it does for other things like `amenity=waste_disposal` and `leisure=swimming_area` that can be rather informal. I figure that mappers probably won’t be tempted to add unofficial hitchhiking locations just because the preset exists, but if so, “designated” or maybe “official” would be how a native English-speaking layperson would make the distinction, not “signed”.

data/presets/highway/hitchhiking.json
paulklie: > This repository doesn’t store any icons

Well no, but iD does:

https://github.com/openstreetmap/iD/tree/develop/svg/fontawesome

And the thumb icon is not present there.

data/presets/highway/hitchhiking.json
paulklie: Not every editor has such a help / documentation button, and not every user is guaranteed to read and understand it. Also the wiki page is only available in two languages thus alot of users might not be able to read it anyway. Thus I really dont think "hitchhiking" alone would suffice.

matkoniecz: > I figure that mappers probably won’t be tempted to add unofficial hitchhiking locations just because the preset exists

I bet that without very strong hints it will be used for any good hitchhiking spot (and even with this hint it will be misused this way)

data/presets/highway/hitchhiking.json
```diff
@@ -0,0 +1,19 @@
...
+    "fields": [
+        "bench",
+        "bin",
+        "shelter"
+    ],
+    "geometry": [
+        "point"
+    ],
+    "tags": {
+        "highway": "hitchhiking"
```
matkoniecz, line 12: that really should have been `highway=signed_hitchhiking_place` to prevent degenerated use, but fixing it is out of scope for iD presets I guess

matkoniecz (approved):

data/presets/highway/hitchhiking.json
tordans: I don't remember how the previews work in this area. In general we can use all icons from the icon list. It might be, though, that the preview does not pick them up right away due to dependencies. Asked in the OSM US iD Chat for clarification.

data/presets/highway/hitchhiking.json
tordans: Yeah, that will have to be a future discussion :-).
Will "resolve" this here …

data/presets/highway/hitchhiking.json
tordans: I don't see any harm in having the name like this now and change it later. And I tend to go with the more explicit version, just in case. I see the point that the simpler name "Hitchiking (signed stop)" or something like this would also work. But I don't think that has to delay this PR.

data/presets/highway/hitchhiking.json
tordans: According to Kyle (https://osmus.slack.com/archives/CBK3JLUJU/p1757410815995009?thread_ts=1757392963.072509&cid=CBK3JLUJU) the previews should pick up any Icon from the [list of icon sources](https://github.com/ideditor/schema-builder/blob/main/ICONS.md#where-do-the-icons-come-from).

Could you test this with the thumbs icon again?

tyrasd (commented):
//edit2: this review is mostly obsolete, see below: https://github.com/openstreetmap/id-tagging-schema/pull/1705#issuecomment-3271199331

Looking at the osm data with this tag, I noticed that there are lots of places where the `highway=hitchhiking` is misused to tag locations that are actually [`amenity=car_pooling`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/car_pooling.json) spots.

I think that might be caused by the fact that the [wiki](https://wiki.openstreetmap.org/wiki/Tag:highway=hitchhiking) for the hitchhiking tag is using a photo and tagging example[^1] for something that is not actually a hitchhiking spot, but rather a car pooling location. :shrug: 

[^1]: The term _Mitfahrbank_ is referring to a location where you can join a shared ride, and not where you would hitchhike spontaneously.

Soo, unfortunately, I fear that because this tag is currently mostly used wrongly, adding a preset will lead to even more confusion. :unamused: 

//edit: PS: I think the main culprit for the misuse of the tag seems to be https://wiki.openstreetmap.org/wiki/DE:Mitfahrbank :cry:

tordans:
@tyrasd great catch. It looks to me that we need some cleanup of the concepts and wiki first…

https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcar_pooling has some good wording to separate the concepts

> Car pooling station or spot, where you meet to get in (or where you get dropped off) someone's car or to pick up (or drop off) someone.
>…
>It is different from Tag:amenity=car sharing, in which drivers can pick or drop off shared cars (in the manner of a short term rental).
>It is different from Tag:highway=hitchhiking, where the driver has no intention of ride-sharing until the moment the hitchhiking spot becomes visible to them and the ride is free.

However, the other wiki pages do not properly reference the other tagging and don't help in making a good decision on which to pick.


> The term Mitfahrbank is referring to a location where you can join a shared ride, and not where you would hitchhike spontaneously.

https://de.wikipedia.org/wiki/Mitfahrbank tells the story differently: 
> Durch das Platznehmen auf dieser Bank signalisieren die Wartenden, dass sie auf eine spontane, kostenlose Mitfahrgelegenheit im PKW zu einem bestimmten Ziel hoffen.

(which sounds to me like hitchhiking)

tillwenke:
Joining in as the author of highway=hitchhiking.

The benches (German Mitfahrbank) are indeed for hitchhiking not for car pooling as it is defined in the tag's wiki article.

I was also in close contact with the Mitfahrverband (German lobbying organization for ride sharing) who also work on the benches to next to classical ride sharing. They list such initiatives at https://mitfahrverband.org/mitfahrbaenke/ which read very similar to what is defined in the tag's wiki page.

My background is in actual hitchhiking (and hitchwiki) where one rarely comes across official hitchhiking spots (because there are not so many and often they are in rural areas). Nevertheless these benches and e.g. Rezo Pouce signs in France are what we would consider hitchhiking spots that are allowed to be mapped on OSM because they provide physical features that identify them as such.

tillwenke:
I agree that it might be that https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcar_pooling is underspecified as it is just stated here https://en.wikipedia.org/wiki/Carpool#Operation that contact is made prior to the ride.

And, yes, @tordans https://wiki.openstreetmap.org/wiki/Tag:highway=hitchhiking should reference [car pooling](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcar_pooling) [done].

tillwenke:
As part of the documentation effort, feel free to pick a more representative image for https://wiki.openstreetmap.org/wiki/Tag:highway%3Dhitchhiking from the images at the bottom :)

tillwenke:
> Looking at the osm data with this tag, I noticed that there are lots of places where the `highway=hitchhiking` is misused to tag locations that are actually [`amenity=car_pooling`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/car_pooling.json) spots.

@tyrasd maybe we can find another venue to discuss where the tag is misused. Maybe at https://wiki.openstreetmap.org/wiki/Talk:Tag:highway%3Dhitchhiking

tyrasd (approved):
The :+1: icon looks good for this preset IMO.

data/presets/highway/hitchhiking.json
1ec5: OK, I trust your judgment. To me, the focus on signs is understandable but a bit restrictive, since the sign is just a symptom of designation. In some areas where I map, this kind of thing would probably be just as likely to be marked on the ground rather than signposted, but I understand that the preset name isn’t to be taken literally.

For what it’s worth, iD and Go Map!! take the help documentation from [this data item](https://wiki.openstreetmap.org/wiki/Item:Q22966), which has descriptions in five languages. You can easily contribute a description in another language without having to create a full-fledged wiki page for the tag.

matkoniecz:
As far as I know that symbol is potentially problematic in areas where such POIs are not present, so it should not be a problem.

k-yle:
Currently 98% of tag usage is from Germany. Do "official" hitchhiking spots exist in any countries other than [these 4](https://commons.wikimedia.org/wiki/Category:Ridesharing_benches)? If this concept only exists in a few small number of countries, maybe a regional preset is better?

There are only 3 examples in the US, and they're all mistagged, because they're not actually signposted. 

Other overseas examples are from a [bot edit](https://osm.org/changeset/166119705), so no one checked if they're actually signposted before adding the tag.

I think restricting the countries for this preset would be good, otherwise OSM might end up like hitchmap.com which suggests many random unsignposted locations including some where [hitchhiking is illegal](https://hitchmap.com/#-36.852820512,174.7541871071)

1ec5:
It’s slowly dawning on me that `highway=hitchhiking` is a terrible misnomer. The approved definition is for sanctioned slugging locations or ridesharing benches, _not_ what’s normally known as hitchhiking.

matkoniecz:
@1ec5 

> It’s slowly dawning on me that highway=hitchhiking is a terrible misnomer. The approved definition is for sanctioned slugging locations or ridesharing benches, not what’s normally known as hitchhiking (i.e., hitchhiking=*).

why you think so?

paulklie:
> In the only English-speaking country we know of that has this kind of feature, the local authorities make an active effort to avoid the term hitchhiking. The signs always say something like “[Carpool Staging Area](https://wtop.com/local/2014/08/slug-lines-ride-sharing-relieves-dcs-dire-commute/)” instead (even where it’s colloquially known as slugging).

But wouldnt this be a [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=[car_pooling](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcar_pooling)?

tyrasd (approved):
From my side it is also fine to start this as a regionally limited version. More countries can be added over time.

data/presets/highway/hitchhiking.json
```diff
@@ -0,0 +1,31 @@
...
+        "hitchhiking",
+        "thumbing"
+    ],
+    "locationSet": {
+        "include": [
+            "at",
+            "ch",
+            "de",
+            "dk",
+            "fx",
```
1ec5, line 24: Did you mean `fr` for France? I don’t think `fx` is a valid ISO&nbsp;3166 country code.

data/presets/highway/hitchhiking.json
```diff
@@ -0,0 +1,31 @@
...
+            "it",
+            "li",
```
1ec5, line 26: No occurrences of `highway=hitchhiking` in Liechtenstein yet, but I suppose it’s OK to prospectively add presets for places where we’re sure we could have some occurrences in the future.

data/presets/highway/hitchhiking.json
paulklie: fx is Metropolitan France. It is supported by https://location-conflation.com/?

data/presets/highway/hitchhiking.json
paulklie: I mostly added it because the other German speaking countries (austria, germany switzerland) were included. Yes, no such nodes currently, but since Lichtenstein is culturally similar,  it seems more likely to have some added than other countries.

data/presets/highway/hitchhiking.json
1ec5: Ah, `fx` was deleted from ISO&nbsp;3166-2 in 1997 and replaced by the ISO&nbsp;3166-3 code `fxfr`. But as long as it works with location-conflation, I guess that’s good enough for this repo.

paulklie: Hmm yea, fxfr does not seem to be supported by location-conflation

1ec5:
To clarify, I was not suggesting to have the preset automatically add `informal=yes` – that would be pointless. I was suggesting to add the existing Informal field to the preset, so that someone who sees “Hitchhiking” and understands this word according to the plain English meaning would at least have an opportunity to clarify their intent rather than mislead data consumers. This would be consistent with the stated goal of increasing _reliable_ coverage of these facilities in OSM.

`informal=*` is a general attribute that isn’t limited to a particular tagging scheme. Granted, it’s currently rare on `highway=hitchhiking` features. I’m not wedded to this idea, but the alternatives don’t seem to be well-liked either:

* Do nothing and accept that some amount of `highway=hitchhiking` will implicitly contradict the approved tagging scheme because iD “pushed” this incorrect association
* Avoid the term “hitchhiking” in the English localization so that lay English speakers don’t use it for typical hitchhiking spots
* Pretend that these facilities don’t frequently occur in English-speaking regions, contrary to current usage

1ec5:
Only if we aren’t worried about `highway=hitchhiking` getting diluted by actual hitchhiking spots, which are implicitly informal. I also suggested two alternatives, in case that isn’t palatable.

paulklie:
> Apart from that this is ready fro merge IMO.

@tordans There seems to be no further objections. Do you think this could be merged now?

tillwenke:
This is up for almost a year, I am wondering what slowed it down.

tordans (approved):

tordans:
I merged this yesterday. Please have a look at https://github.com/openstreetmap/id-tagging-schema/pull/2298 which are follow up changes proposed by @1ec5 on this topic. I suggest to move the follow up discussion there.

matkoniecz:
> This is up for almost a year, I am wondering what slowed it down.

from my side: lack of clarity what is the correct tagging, is it actually used as documented, whether merging it will improve situation and lack of critical importance of this preset (it is not say `highway=motorway` being missing)

tillwenke:
can someone reproduce the icon not showing up on e.g. https://www.openstreetmap.org/edit#map=19/54.361922/10.080991

image

matkoniecz:
I see that https://github.com/openstreetmap/id-tagging-schema/releases/tag/v6.18.0 was released two days ago.

https://github.com/openstreetmap/iD/releases was NOT released

it means that iD lacks icons and is unable to display them

I have opened https://github.com/openstreetmap/iD/issues/12445 as it happened before and it should stop happening

PR #1708, opened 2025-09-08
added icon to bathing place

paulklie:
Currently leisure=bathing_place uses a drop icon:
image

The new icon:
image

looks quite similar to the "badplats" sign for Sweden, which is where most of these bathing places are:
image

and is much more fitting than the simple "drop" used before.
Example node: https://www.openstreetmap.org/node/2954848386#map=19/57.417123/12.958180

### DIFF
--- a/data/presets/leisure/bathing_place.json
+++ b/data/presets/leisure/bathing_place.json
@@ -7,7 +7,7 @@
         "Swimming Spot",
         "Wild Swimming Spot"
     ],
-    "icon": "maki-water",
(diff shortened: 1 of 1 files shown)


tordans:
Testing: 🟡 

I am not convinced this is an improvement.
In the map pin UI this is really hard to recognize as a bathing place.

https://pr-1708--ideditor-presets-preview.netlify.app/id/dist/#locale=en&node=2954848386&map=19/57.417123/12.958180

> image

PR #1714, opened 2025-09-09
create preset for 'Offshore Platform'

k-yle:

An ['Offshore Platform'](https://osm.wiki/Tag:man_made=offshore_platform) is structure built in the ocean, usually an [Oil Platform](https://en.wikipedia.org/wiki/Oil_platform) or [cew accommodation platform](https://en.wikipedia.org/wiki/Accommodation_platform) for offshore wind farms.

OSM has a `man_made` and `seamark` tag for this feature, so the preset accepts uses both, just like _Marina_ and other existing presets 

image

### Related issues

part of #683

### Links and data

**Relevant OSM Wiki links:**
- https://osm.wiki/Tag:man_made=offshore_platform
- https://osm.wiki/Tag:seamark:type=platform

**Relevant tag usage stats:**

- 5711<sup>[[1]](https://overpass-turbo.eu/s/2bjl)</sup>

<details>
<summary><h2>Test-Documentation (click to expand)</h2></summary>

[node 9228831984](https://www.openstreetmap.org/node/9228831984) (and [in OpenSeaMap](https://map.openseamap.org/?zoom=15.9&lon=173.45034&lat=-39.55433&layers=TFTFFFTFFFFFFFFFTFFFFF))

image

### Search

image

image
image

### Info-`i`

image

### Wording

- [x] American English

</details>

### DIFF
--- a/data/fields/seamark/platform/category.json
+++ b/data/fields/seamark/platform/category.json
@@ -1,0 +1,14 @@
+{
+    "key": "seamark:platform:category",
+    "type": "combo",
+    "label": "Category",
+    "strings": {
+        "options": {
+            "oil": "Oil Rig",
+            "production": "Production Platform",
+            "observation": "Observation/research platform",
+            "fpso": "FPSO: Floating production, storage and off-loading vessel",
+            "accommodation": "Accommodation Platform"
+        }
+    }
+}
--- a/data/fields/seamark/platform/category.json
+++ b/data/fields/seamark/platform/category.json
@@ -8,7 +8,7 @@
             "production": "Production Platform",
             "observation": "Observation/research platform",
             "fpso": "FPSO: Floating production, storage and off-loading vessel",
-            "accommodation": "Accommodation Platform"
+            "accommodation": "Accommodation Platform (Living Quarters)"
         }
     }
 }
--- a/data/fields/seamark/radio_station/mmsi.json
+++ b/data/fields/seamark/radio_station/mmsi.json
@@ -1,0 +1,7 @@
+{
+    "key": "seamark:radio_station:mmsi",
+    "type": "identifier",
+    "urlFormat": "https://www.vesselfinder.com/vessels/details/{value}",
+    "pattern": "^[0-9]{1,}$",
+    "label": "MMSI"
+}
--- a/data/presets/seamark/offshore_platform.json
+++ b/data/presets/seamark/offshore_platform.json
@@ -1,0 +1,33 @@
+{
+    "icon": "maki-harbor",
+    "fields": [
+        "name",
+        "seamark/platform/category",
+        "seamark/radio_station/mmsi"
+    ],
+    "geometry": [
+        "point",
+        "vertex",
+        "area"
+    ],
+    "tags": {
+        "seamark:type": "platform"
+    },
+    "addTags": {
+        "seamark:type": "platform",
+        "man_made": "offshore_platform"
+    },
+    "removeTags": {
+        "seamark:type": "*",
+        "man_made": "offshore_platform"
+    },
+    "terms": [
+        "bore",
+        "drilling rig",
+        "offshore platform",
+        "oil rig",
+        "petroleum well",
+        "platform"
+    ],
+    "name": "Offshore Platform"
+}
--- a/data/presets/man_made/petroleum_well.json
+++ b/data/presets/man_made/petroleum_well.json
@@ -14,12 +14,12 @@
         "vertex"
     ],
     "terms": [
+        "bore",
         "drilling rig",
         "oil derrick",
         "oil drill",
         "oil horse",
         "oil pump",
-        "oil rig",
         "petroleum well",
         "pumpjack"
     ],


data/fields/seamark/radio_station/mmsi.json
```diff
@@ -0,0 +1,7 @@
+{
+    "key": "seamark:radio_station:mmsi",
+    "type": "identifier",
+    "urlFormat": "https://www.vesselfinder.com/vessels/details/{value}",
```
k-yle, line 4: this is the same formatter URL that [wikidata uses](https://www.wikidata.org/wiki/Property:P587#P1630), so this matches [the url used by tag2link](https://github.com/JOSM/tag2link/blob/b7d171728bf0803f10c7d044264767b7dc1bbcc1/index.json#L2595-L2596)

data/presets/seamark/offshore_platform.json
```diff
@@ -0,0 +1,40 @@
+{
+    "icon": "temaki-oil_well",
```
matkoniecz: I am not an expert on this topic - but this icon looks very land-specific

I think that even generic icon nonspecific to anything would be better

matkoniecz: @k-yle Can you fix the icon? Or at least replace it with some generic one?

k-yle: sure, changed to the generic seamark icon (same as the other seamark presets).

Alternatively we could create a new icon, maybe a truss structure with water beneath it:


image


but this could be confused with a lighthouse

data/presets/seamark/offshore_platform.json
```diff
@@ -0,0 +1,40 @@
...
+    "geometry": [
+        "point",
+        "vertex",
+        "area"
+    ],
+    "tags": {
+        "seamark:type": "platform"
+    },
+    "addTags": {
+        "seamark:type": "platform",
```
matkoniecz, line 17: why we would add `seamark:type` duplicating  `man_made=offshore_platform`?

I know that for some unclear reason there is that project that refuses to use keys not prefixed with `seamark:`, but that is IMHO not a valid reason to add not needed tags

data/presets/seamark/offshore_platform.json
matkoniecz: thanks! I think it can be considered as resolved!

data/presets/seamark/offshore_platform.json
```diff
@@ -0,0 +1,40 @@
...
+        "offshore platform",
+        "oil derrick",
+        "oil drill",
+        "oil horse",
+        "oil pump",
+        "oil rig",
+        "oil rig",
+        "petroleum well",
+        "platform",
+        "pumpjack"
```
FloEdelmann: Duplicate terms can be removed:

```suggestion
        "bore",
        "drilling rig",
        "offshore platform",
        "oil derrick",
        "oil drill",
        "oil horse",
        "oil pump",
        "oil rig",
        "petroleum well",
        "platform",
        "pumpjack"
```

data/presets/seamark/offshore_platform.json
k-yle: i guess for the same reason that the railway community has duplicate tags like [`phone`](https://taginfo.osm.org/tags/railway=phone), [`radio`](https://taginfo.osm.org/tags/railway=radio), [`fuel`](https://taginfo.osm.org/tags/railway=fuel), [`waste_disposal`](https://taginfo.osm.org/tags/railway=waste_disposal), ...



my personal opinion is that we only use the `seamark` tag here, and not the `man_made` one. 
- since this is a quintessential example of a feature that is only of interest for `seamark`-related maps
- the seamark tag was always more popular until a recent mass update/import (see [taghistory](https://taghistory.raifer.tech/#***/man_made/offshore_platform&***/seamark%3Atype/platform))
- the `man_made` tag does not have any well-defined subtags

but to keep everyone happy i thought we could add both tags? like several existing presets

data/presets/seamark/offshore_platform.json
k-yle: thanks, fixed

data/presets/seamark/offshore_platform.json
matkoniecz: > i guess for the same reason that the railway community has duplicate tags like phone, radio, fuel, waste_disposal, ...

the difference is that seamark people consistenly create seamark: duplicates even if existing tagging exists

matkoniecz: and putting project name into key prefixes rubs me wrong way, every time I see it - it looks like spam to me

in the same way as StreetComplete tagging `streetcomplete:name=` instead of `name=` would be obnoxious

k-yle: > and putting project name into key prefixes rubs me wrong way, every time I see it - it looks like spam to me

is `seamark` a project name? I believe it's just [a generic english word](https://en.wikipedia.org/wiki/Sea_mark)

matkoniecz: Sorry for LLM-grade brain dysfunction

matkoniecz: I still deeply dislike `seamark:` mess but I guess that I am not God-Emperor of OSM tagging and I cannot pretend it does not exists solely because I dislike it.

data/presets/seamark/offshore_platform.json
```diff
@@ -0,0 +1,38 @@
...
+    "addTags": {
+        "seamark:type": "platform",
+        "man_made": "offshore_platform"
+    },
+    "removeTags": {
+        "seamark:type": "*",
+        "man_made": "offshore_platform"
+    },
+    "terms": [
+        "bore",
```
matkoniecz, line 25: is it valid also for regular oil well? Then it should be added also there, now it finds solely offshore platform

data/presets/seamark/offshore_platform.json
```diff
@@ -0,0 +1,38 @@
...
+        "drilling rig",
+        "offshore platform",
+        "oil derrick",
```
matkoniecz: can we do anything so it will put offshore platform lower?

`    "matchScore": 0.9` is ugly but maybe worth doing? But sounds like `!important` in CSS - band-aid that will make future worse for the same problem.

Image

what worse the same happens for petroleum well, pumpjack

data/presets/seamark/offshore_platform.json
```diff
@@ -0,0 +1,38 @@
...
+        "oil drill",
+        "oil horse",
+        "oil pump",
+        "oil rig",
+        "petroleum well",
+        "platform",
+        "pumpjack"
```
matkoniecz: is "pumpjack" even applicable at all to offshore oil rigs?

https://commons.wikimedia.org/wiki/Nodding_donkey_pumps

matkoniecz: (I have not looked into other terms)

data/presets/seamark/offshore_platform.json
k-yle: yes, done

data/presets/seamark/offshore_platform.json
k-yle: i removed some of them, to reduce the cases where `petroleum_well` is the second match

data/presets/seamark/offshore_platform.json
k-yle: yeah we could add `matchScore: 0.9`, I've improved the terms a bit, so this might not be required

data/fields/seamark/platform/category.json
```diff
@@ -0,0 +1,14 @@
...
+    "key": "seamark:platform:category",
+    "type": "combo",
+    "label": "Category",
+    "strings": {
+        "options": {
+            "oil": "Oil Rig",
+            "production": "Production Platform",
+            "observation": "Observation/research platform",
+            "fpso": "FPSO: Floating production, storage and off-loading vessel",
+            "accommodation": "Accommodation Platform"
```
matkoniecz:
```suggestion
            "accommodation": "Accommodation Platform (Living Quarters)"
```

or is "Accommodation" clear enough?

matkoniecz: @k-yle

k-yle: i don't really mind, seems reasonably. the wikipedia page is called "[Accommodation platform](https://en.wikipedia.org/wiki/Accommodation_platform)"

matkoniecz (approved):

PR #1715, opened 2025-09-09
create preset for 'Ship Berth' (`seamark:type=berth`)

k-yle:

A [Berth](https://en.wikipedia.org/wiki/Berth_(moorings)) is like [`amenity=parking_space`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dparking_space) for boats.

I called the preset _Ship Berth_ to avoid confusion for people who don't know what this word means.

### Related issues

part of #683

### Links and data

**Relevant OSM Wiki links:**
- https://osm.wiki/Tag:seamark:type=berth
- [example rende](https://github.com/user-attachments/assets/1515bd92-700c-4d7a-bd2b-c0fc9603b6d9)[ring in OpenSeaMap](https://github.com/user-attachments/assets/e18d1e97-10b6-47a5-a6d2-c85ffd00bfd5)

**Relevant tag usage stats:**
- [5827](https://taginfo.osm.org/tags/seamark:type=berth)

<details><summary><h2>Test-Documentation (Click to expand)</h2></summary>

[example](https://pr-1715--ideditor-presets-preview.netlify.app/id/dist/#background=LINZ_NZ_Aerial_Imagery&id=n9330938230)

image

### Search

image

### Info-`i`

see screenshot above

### Wording

- [x] American English

```

</details>

### DIFF
--- a/data/fields/seamark/name.json
+++ b/data/fields/seamark/name.json
@@ -1,0 +1,5 @@
+{
+    "key": "seamark:name",
+    "type": "text",
+    "label": "{ref}"
+}
--- a/data/fields/maxdraft.json
+++ b/data/fields/maxdraft.json
@@ -1,0 +1,6 @@
+{
+    "key": "maxdraft",
+    "type": "roadheight",
+    "label": "Maximum Draft",
+    "snake_case": false
+}
--- a/data/presets/seamark/berth.json
+++ b/data/presets/seamark/berth.json
@@ -1,0 +1,23 @@
+{
+    "icon": "temaki-boat_floating",
+    "fields": [
+        "seamark/name",
+        "maxweight",
+        "maxlength",
+        "maxdraft",
+        "maxstay"
+    ],
+    "geometry": [
+        "point",
+        "vertex",
+        "area"
+    ],
+    "tags": {
+        "seamark:type": "berth"
+    },
+    "name": "Ship Berth",
+    "terms": [
+        "berth",
+        "mooring"
+    ]
+}

### CODE REVIEW

**Comment by matkoniecz** [comment]
Is it supposed to be used also for marinas where small pleasure craft berths?

**Comment by matkoniecz** [comment]
> I think that description is reasonable. From my personal experience[1](https://github.com/openstreetmap/id-tagging-schema/pull/1715#user-content-fn-1-c0f2d6ae151faf564e5330d4ec2e5af4), berths always have a name or ref. Otherwise it's almost impossible to describe the location to someone

my point is that wiki claims (maybe wrongly!) that berth without assigned name/ref cannot be mapped as seamark:type=berth

either wiki should be changed (with community consultation?) or this PR should be changed

File: data/fields/seamark/name.json [hunk 1]
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "seamark:name",
```
matkoniecz, line 2: there are few problems here: using weird duplicate of `name` should  not be promoted, in my opinion

if current ship berth tagging schema requires it then for me it should block its support in editors

also, https://wiki.openstreetmap.org/wiki/Tag:seamark:type=berth claims that `seamark:berth:name` should be used (which manages to be an even weirder idea)

k-yle: lol, i agree that `seamark:berth:name` is a bit dumb.... 

I'm only aware of 3 apps that support seamark tags. OpenSeaMap, OsmAnd, ~~and {Organic,Co}Maps~~. All of them require `seamark:name` or `seamark:berth:name` instead of `ref`.

If you want iD to start encouraging a different tagging practice, those apps will need to be updated too

matkoniecz: If it is name then `name` would work - is it supported there?

If it is in fact reference then `seamark:berth:name` is even sillier.

> If you want iD to start encouraging a different tagging practice, those apps will need to be updated too

Maybe. In my opinion it should be prerequisite to supporting this tagging scheme anywhere (and while at it, maybe getting rid of `seamark:` prefix from key would be doable?)

Issue #1651, opened 2025-08-03
Preset for olive grove


tordans:
Hello, I found this preset very interesting and I can say it is a kind of landuse very widespread also in Italy, so I am trying to add a translation in Transiflex.
I would also ask for another preset, similar to this one, that in Italy we call "uliveto" (olive grove) and it's the combination of these tags:
landuse=orchard
trees=olive_trees

I don't know how to add a preset, could you please help me ? 
Thanks !

_Originally posted by @dp7x in https://github.com/openstreetmap/id-tagging-schema/issues/1218#issuecomment-3148603213_

tordans:
@dp7x there are docs in this repo in how the translations work. Do those help you?

A new preset is certainly possible and we can help. Can you share some links like wiki, usage numbers… that help as reference.

PR #1717, opened 2025-09-09
Create olive_grove.json preset for landuse

dp7x:

Please see here: https://github.com/openstreetmap/id-tagging-schema/issues/1651

### DIFF
--- a/data/presets/landuse/olive_grove.json
+++ b/data/presets/landuse/olive_grove.json
@@ -1,0 +1,24 @@
+{
+    "icon": "maki-park",
+    "fields": [
+        "name",
+        "operator"
+    ],
+    "moreFields": [
+        "{@templates/contact}",
+        "address",
+        "species/wikidata"
+    ],
+    "geometry": [
+        "area"
+    ],
+    "tags": {
+        "landuse": "orchard",
+		    "trees": "olive_trees"
+    },
+    "terms": [
+        "fruit",
+		    "olive"
+    ],
+    "name": "Olive grove"
+}


dp7x:
image

tordans:
Some notes for the review
- review the wiki documentation on this, when we [last talked, there was none](https://github.com/openstreetmap/id-tagging-schema/issues/1651#issuecomment-3149627804)
- compare the fields/moreFields to other similar presets 
- compare the fields/moreFields to common usage

dp7x:
Hi @tordans , thanks.
I created the wiki data item, I hope this is ok: https://wiki.openstreetmap.org/wiki/Item:Q23173
About the fields, I started from "landuse=orchard" ; I will add more when they come to my mind.

dp7x:
Because I think it would be useful for more accurate and informed mapping.
Otherwise, your reasoning would also apply, for example, to other presents like "meadow=meadow_orchard" or "orchard=meadow_orchard." 
Likewise, it wouldn't have been necessary to create "landuse=vineyard": we could have used "landuse=orchard" + "crop=grape."
I think a preset should simplify the mapping process by helping users find exactly what they're looking for, while maintaining accuracy.

matkoniecz:
why https://wiki.openstreetmap.org/wiki/Item:Q23173 is for landuse=olive_grove while https://github.com/openstreetmap/id-tagging-schema/pull/1717#issuecomment-3271261633 shows subtag for `landuse=orchard` ?

dp7x:
Hello @matkoniecz , there is a misunderstanding: I **don't** want to introduce a new landuse value.
Now I see my error: I tried to create the wiki data page but I did a big mistake in the tag value: sorry, i am new to this, I will try to correct.

So, at last, what I would like to obtain is exactly what you can see in the pull request preview and I showed in the picture of my [previous message](https://github.com/openstreetmap/id-tagging-schema/pull/1717#issuecomment-3271261633). I think it would be very useful for more than a reason and I see no problems with it, but if it can't be done please tell me and I will stop asking. Thanks for your support and understanding.

dp7x:
> is https://github.com/openstreetmap/id-tagging-schema/blob/main/CONTRIBUTING.md#info-i helpful?

Sorry @matkoniecz,  but no... it's trying to follow that link that I made the previous mess creating the new landuse value. I suppose that I should have to create a new item that combines both the tags I need, so that I can create a new short description for the preset, but I can't find a way on how to do it !

matkoniecz:
Can you then add reference ( https://github.com/ideditor/schema-builder?tab=readme-ov-file#reference ) property?

See say https://github.com/openstreetmap/id-tagging-schema/blob/d87f48d201a0cc49988fffeebaf4e026a841b596/data/fields/tee.json#L5

dp7x:
> Can you then add reference ( https://github.com/ideditor/schema-builder?tab=readme-ov-file#reference ) property?
> 
> See say
> 
> https://github.com/openstreetmap/id-tagging-schema/blob/d87f48d201a0cc49988fffeebaf4e026a841b596/data/fields/tee.json#L5

Ok, I committed the changes, please take a look when you can. 
Thanks

dp7x:
Hi @matkoniecz and @tordans , could you please review ?
Thanks

PR #1727, opened 2025-09-17
more specific railway crossing icons

k-yle:

Currently `highway=crossing` and `railway=crossing` use the same icon, which is confusing.

Also, the generic X icons look very similar.

We can improve this by using a new icon for each type of crossing:

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>image

 <td>image

</table>

### Related issues

_none_

### Links and data

_none_

image

_(I think the rest of this template is irrelevant for icon-only changes?)_

### DIFF
--- a/data/presets/railway/crossing.json
+++ b/data/presets/railway/crossing.json
@@ -1,5 +1,5 @@
 {
-    "icon": "temaki-pedestrian",
+    "icon": "temaki-crossing_rail_striped",
     "fields": [
         "crossing",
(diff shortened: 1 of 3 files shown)


k-yle:
do you think we should keep `X` and `×`, and only fix the pedestrian case?

k-yle:
okay, I removed the changes for road–rail and rail–rail crossings. 

only kept the path–rail changes, to fix this issue:  

image

PR #1729, opened 2025-09-17
fix preset with no icon (advertising=totem)

k-yle:

[`advertising=totem`](https://osm.wiki/Tag:advertising=totem) is one of the few remaining presest with no icon. Changed it to be the same as `advertising=*`:

image

### Related issues

none

### Links and data

see above

**Relevant tag usage stats:**

- [13,000](https://taginfo.openstreetmap.org/tags/advertising=totem)

see above

### Search

see above

### Info-`i`

unchanged 

### Wording

unchanged

### DIFF
--- a/data/presets/advertising/totem.json
+++ b/data/presets/advertising/totem.json
@@ -1,4 +1,5 @@
 {
+    "icon": "temaki-billboard",
     "fields": [
         "{advertising}"
     ],


k-yle:
> I would say that no icon was better

no icon looks like a tag-less node, so everytime i see this feature I waste time clicking on it to see if it's tagless:

image

I think we should aim to have an icon for every preset. Only [22 out of 1690](https://github.com/openstreetmap/id-tagging-schema/issues/41#issuecomment-3302595303) are missing icons. 

In this case `advertising=anything_except_totem` gets the _billboard_ icon, so I think it's reasonable if `totem` gets the same icon to match almost all other `advertising=*` features.

image

PR #1744, opened 2025-09-26
Add presets from Proposal:Power circuits routing

nlehuby:
This PR adds the modifications approved in the [Power circuits proposal](https://wiki.openstreetmap.org/wiki/Proposal:Power_circuits_routing):
* new preset for [power=circuit](https://wiki.openstreetmap.org/wiki/Tag:power%3Dcircuit)
* new preset for [power=line_section](https://wiki.openstreetmap.org/wiki/Tag:power%3Dline_section)
* new [topology](https://wiki.openstreetmap.org/wiki/Key:topology) field for power=circuit
* deprecations of type=route + route=power & power=branch in favor of power=circuit ([source](https://wiki.openstreetmap.org/wiki/Proposal:Power_circuits_routing#Tags_to_be_replaced))


data/fields/topology.json
```diff
@@ -0,0 +1,11 @@
+{
+    "key": "topology",
+    "type": "combo",
+    "label": "Topology",
+    "strings": {
+        "options": {
+            "linear": "linear",
+            "branched": "branched"
```
wolfy1339: Shouldn't these be 
```suggestion
            "linear": "Linear",
            "branched": "Branched"
```

wolfy1339:
Changes will be needed to https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/type/route/power.json

data/deprecated.json
```diff
@@ -1304,6 +1400,10 @@
     "old": {"postcode": "*"},
     "replace": {"addr:postcode": "$1"}
   },
+  {
+    "old": {"power": "branch"},
+    "replace": {"power": "circuit"}
```
matkoniecz: can it be always replaced this way?

if yes, why not do this as a bot edit rather than going through iD? Also, with 241 uses upgrading existing tagging (if that is useful and worth doing) will take comparable effort to doing PR review of that change https://taginfo.openstreetmap.org/tags/power=branch

matkoniecz: the same for other deprecation

note that these comments are not enabling bypassing https://wiki.openstreetmap.org/wiki/Automated_Edits_code_of_conduct

if you do automated edit you must follow these rules

matkoniecz: In case of no further activity on this PR it will be closed

nlehuby: According to [the proposal page](https://wiki.openstreetmap.org/wiki/Proposal:Power_circuits_routing#Tags_to_be_replaced): 
* `power=branch` is to replace by power=circuit and case by case required `power=line_section` relations
* `type=route` + `route=power` is to replace by `type=power` + `power=circuit`

For my part, I don't plan on making any large-scale, automatic changes. I think it’s up to the local communities interested in this topic to take on this issue, and some have already started doing so on their own perimeter. In particular, regarding the `power=branch` case, they seem to have been introduced mainly as part of an import project I know nothing about, so I certainly won’t be changing anything.

I simply suggest using the existing feature in the iD editor to notify contributors who might edit the object that this is no longer the correct way to label it, and to prevent new contributors from creating new objects with these tags. From what I understand, this is exactly what the feature in iD is meant to do, but if you don't think it's reasonable, I'll remove it.

matkoniecz: > but if you don't think it's reasonable, I'll remove it.

please, do it

in my view this is effectively doing "large-scale, automatic changes" as at least some users press "update tags" blindly

and if these tags can be mass-transformed then it should be done with bot edit, not some manual distributed one

As disclaimer: yes, in past this deprecation feature was used and intended to be used differently, and I am effectively trying to change it

matkoniecz: So sorry for any confusion here - and yes, situation partially changed over say last year.

nlehuby: ok, it's done.

data/fields/topology.json
```diff
@@ -0,0 +1,11 @@
+{
```
matkoniecz, line 1: please change file name (so also field code) to something like `power_line_topology`

you can have also other topologies and it will be more clear where context is not obvious

data/presets/power/line_section.json
```diff
@@ -0,0 +1,22 @@
...
+    ],
+    "tags": {
+        "type": "power",
+        "power": "line_section"
+    },
+    "reference": {
+        "key": "power",
+        "value": "line_section"
+    },
+    "name": "Section of Power Line"
```
matkoniecz, line 21: I am quite confused by this, also after reading docs

I started https://wiki.openstreetmap.org/wiki/Talk:Tag:power%3Dcircuit#What_is_the_difference_between_line_section_and_circuit%3F

matkoniecz (changes requested):
documentation of tags supported in iD should be understandable, current one refers to "tap point" without clarification what it is.

And what "tap point" seems to be obscure knowledge without obvious answers.

data/fields/topology.json
nlehuby: Isn't that a bit early? The only documented use so far is for power circuits (YAGNI principle)

matkoniecz:
Have taken another look at that and become confused again, see https://wiki.openstreetmap.org/wiki/Talk:Tag:power%3Dcircuit#Substations_or_not

matkoniecz:
@nlehuby are you interesting in clarifying this documentation? Or do you think it is actually clear right now?

nlehuby:
I've reread it, and it seems fairly clear to me.

The definition of the tap point seems fine to me:

> [Branched] circuits involve tapping points referring to particular towers or junctions. They connect several sections together often in a T or X pattern and relate to line_management=branch.

And there are photos available when you click on line_management=branch

I think the wiki page is not bad. We could maybe improve the wording of the preset a bit, but it seems acceptable to me as is.

matkoniecz:
> [Branched] circuits involve tapping points referring to particular towers or junctions. They connect several sections together often in a T or X pattern and relate to line_management=branch.

for me the problem is circular definition: circuits is defined as being between tapping points and tapping point is place (sometimes but not always power tower) where circuit is split

and if you need to go to other pages to have chance to understand basic part of definition I would say it is a fairly bad one

I will try to ask others what they think

data/presets/power/line_section.json
k-yle: i also became extremely confused trying to understand the difference between `line_section` and `circuit`. i don't know how a normal iD user would understand it,

linguistically, a _"Section of [a] Power Line"_ <sup>[relation]</sup> cannot contain a _"Power Line"_ <sup>[way]</sup>. It should be the other way around to make sense linguistically. Or this preset needs a different label...?

data/presets/type/route/power.json
```diff
@@ -1,25 +0,0 @@
...
-    ],
-    "tags": {
-        "type": "route",
-        "route": "power"
-    },
-    "reference": {
-        "key": "route",
-        "value": "power"
-    },
-    "name": "Power Route"
```
k-yle, line 24: [the wiki page](https://wiki.openstreetmap.org/wiki/Tag:route=power) says:
> _"The recommended replacement is: [`power`](https://wiki.openstreetmap.org/wiki/Key:power)`=`[`circuit`](https://wiki.openstreetmap.org/wiki/Tag:power%3Dcircuit)."_

does this mean that you can blindly upgrade `power=route` to `power=circuit`?
- if yes, there should probably be an an entry in `deprecated.json` for this upgrade
- if no, then the wiki page should be fixed.

data/presets/power/circuit.json
```diff
@@ -0,0 +1,24 @@
...
+    ],
+    "tags": {
+        "type": "power",
+        "power": "circuit"
+    },
+    "reference": {
+        "key": "power",
+        "value": "circuit"
+    },
+    "name": "Power Circuit"
```
k-yle:
```suggestion
    "name": "Power Circuit",
    "relation": { ... }
```

(if this PR is accepted despite the confusing documentation and labels) then it should define the relation roles. see #1538 for the syntax

data/presets/type/route/power.json
wolfy1339: There are some slight differences in the roles of relation members

nlehuby: I had initially made a change to deprecated.json, and I was asked not to.

see https://github.com/openstreetmap/id-tagging-schema/pull/1744/changes/BASE..f88815f43b822a46cd60c4f33797cd7dc62c398f#r2901658938

EDITED (fix wrong link to PR comment)

data/presets/power/circuit.json
nlehuby: I've updated the preset to add info about the members.

However, I didn't understand how to test it in iD :thinking:

matkoniecz: > I didn't understand how to test it in iD 🤔

it is not yet used by iD

nlehuby:
Looks like the `power=line_section` relation is confusing for many people. I suggest narrowing the scope and leaving it out for now.

I've deleted the preset I had originally added.

data/presets/type/route/power.json
matkoniecz: if there are some slight differences in the roles of relation members, then it cannot be simply be replacing (and ""The recommended replacement is: [power](https://wiki.openstreetmap.org/wiki/Key:power)=[circuit](https://wiki.openstreetmap.org/wiki/Tag:power%3Dcircuit)."" should be also clear what the differences are rather than suggesting blindly changing)

matkoniecz:
https://github.com/openstreetmap/id-tagging-schema/pull/1744#discussion_r3852819131 is unresolved

https://github.com/openstreetmap/id-tagging-schema/pull/1744#issuecomment-5318479394 is unresolved

documentation is still confusing (despite improvements, and thanks for them), and therefore this PR will be closed without merging

this PR can be reopened once documentation is comprehensible by a typical OSM mapper

BTW, I opened https://github.com/openstreetmap/id-tagging-schema/issues/2813 proposing removal of route=power from presets

------------------------------------------------------------------------------------------

It is disappointing/frustrating to end this way but iD tagging schema is intended to support tagging that is widely accepted and well-documented. Going through proposal process does not automatically indicate that documentation is clear and unambiguous.

In this specific case I have reread docs so many times, and I am still confused by it. I am pretty sure that it is not just me and https://wiki.openstreetmap.org/wiki/Tag:power%3Dcircuit is genuinely confusing.

matkoniecz:
pinging @tordans @tyrasd @k-yle to give them chance to override me (I am well-aware that this may be controversial as this in fact passed proposal process, has high usage numbers and got rejected based on quite subjective matter of documentation quality)

pinging also @1ec5 @andrewharvey 

Still, I am quite sure that relevant documentation is just incomprehensible - not sure is it fixable or is tagging schema flawed here on a deeper level. Or maybe I am just having early-onset dementia, who knows.

tyrasd:
Agree with @matkoniecz. A prerequisite to get this merged would be for the documentation on the osm wiki to be clearer, less technical, and generally understandable by regular mappers.

PR #1752, opened 2025-10-08
Add field specification for EVSE identifier

RoiEXLab:
The [EVSE](https://wiki.openstreetmap.org/wiki/Key:ref:EU:EVSE) identifier is currently used to uniquely identify charging ports on charging stations for electric vehicles. Even though wide adoption of the tag is not there yet (and it's sometimes used incorrectly) the [wiki page](https://wiki.openstreetmap.org/wiki/Key:ref:EU:EVSE) already documents the format pretty well. Often public charging infrastructure has more than one plug (usually 2, each with a unique id), so I changed the regex to allow semicolon-separated lists of ids (which seems to be a popular notation: https://taginfo.openstreetmap.org/keys/ref%3AEU%3AEVSE#values).

#### Caveats
The notation of the id is specified in a very lenient way, allowing many strings that represent the same id. The wiki page presents a canonical notation, but it could be confusing to users because on actual charging stations they might use spaces as separators, or use more than two separators when looking at them in person.

For `man_made=charge_point` and `amenity=charging_station` the `ref` key is always recommended. Since the `EVSE` identifier only applies to EU member states I added this new field for the preset, but the common `ref` still remains, even though it is kinda redundant in these places.

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Key:ref:EU:EVSE

**Relevant tag usage stats:**
- Currently 17037 total uses (according to https://taginfo.openstreetmap.org/)

Screenshot amenity=charging_station
[link](https://pr-1752--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=18.80/48.82406/2.32435&disable_features=boundaries&background=fr.ign.bdortho&id=n12415996660)

Screenshot man_made=charge_point
[link](https://pr-1752--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=19.84/48.81045/2.48160&disable_features=boundaries&background=fr.ign.bdortho&id=n13030359201)

### Search

Does not apply, only applicable to EU member states and there it is pre-selected for all relevant feature types.

### Info-`i`

Info-`i`

### Wording

- [x] American English

### DIFF
--- a/data/fields/ref/EU/EVSE-EU.json
+++ b/data/fields/ref/EU/EVSE-EU.json
@@ -1,7 +1,7 @@
 {
     "key": "ref:EU:EVSE",
     "type": "identifier",
-    "label": "Electric Vehicle Supply Equipment (EVSE) ID",
+    "label": "EVSE identifier (Electrical Vehicle Supply Equipment)",
     "placeholder": "DE*ABC*E12345ABCDEFGHIJKLMNOPQRSTUVWXYZ",
     "locationSet": {
         "include": [


data/fields/ref/EU/EVSE-EU.json
```diff
@@ -0,0 +1,16 @@
+{
+    "key": "ref:EU:EVSE",
```
matkoniecz: we are typically using lowercase in key values

is there an actual support in community for this format, and not just result of some mass retagging?

RoiEXLab: Doesn't look like it. The wikipage notation is very clearly not lowercase and the use seems to have gradually increased over time: https://taginfo.openstreetmap.org/keys/ref%3AEU%3AEVSE#chronology
And the all lowercase variant has 0 uses: https://taginfo.openstreetmap.org/keys/ref%253Aeu%253Aevse

matkoniecz: do you have any idea why it ended not as `ref:eu:evse` ?

FloEdelmann: The part after `ref:` is often uppercase, as it refers to the country (DE, SE, EU, ...) instead of the language (de, sv, ...).

RoiEXLab: @matkoniecz I did some number crunching (dataset not up-to-date so the numbers might be slightly outdated) and there are a lot of cases where the ref key is not strictly lowercase. As @FloEdelmann already pointed out in most cases it refers to a country (I only included the top 11, where the ratio of country vs non-country labels is a lot more balanced), and often it refers to some sort of imported identifier, for a city, a country, or in the case of EVSE the entire EU.

- `ref:GB:uprn` has 3,820,791 usages. ([Even has approved status](https://wiki.openstreetmap.org/wiki/Key:ref:GB:uprn))
- `ref:RS:kucni broj` has 2,456,459 usages.
- `ref:FR:FANTOIR` has 686,483 usages.
- `ref:LV:addr` has 573,647 usages.
- `ref:AGESIC` has 480,192 usages.
- `ref:BRN` has 406,324 usages.
- `ref:IFOPT` has 312,675 usages.
- `ref:UrbIS` has 295,303 usages.
- `ref:FR:SIRET` has 244,183 usages.
- `ref:RS:ulica` has 173,136 usages.
- `ref:MLIT PLATEAU` has 123,154 usages.

RoiEXLab:
Anything else I need to do to a) get the workflows approved and b) get this thing merged in the end?

data/fields/ref/EU/EVSE-EU.json
```diff
@@ -0,0 +1,18 @@
...
+    "locationSet": {
+        "include": [
+            "EU"
+        ]
+    },
+    "terms": [
+        "electrical vehicle supply equipment",
+        "evse",
+        "charging identifier",
+        "charger identifier"
```
FloEdelmann: Please sort these terms alphabetically:

```suggestion
        "charger identifier",
        "charging identifier",
        "electrical vehicle supply equipment",
        "evse"
```

data/fields/ref/EU/EVSE-EU.json
```diff
@@ -0,0 +1,18 @@
+{
+    "key": "ref:EU:EVSE",
+    "type": "identifier",
+    "label": "Electrical Vehicle Supply Equipment (EVSE)",
```
matkoniecz:
```suggestion
    "label": "EVSE code (Electrical Vehicle Supply Equipment)",

```

"Electrical Vehicle Supply Equipment (EVSE)" is quite confusing

maybe it can be clarified that it is some code/reference/id?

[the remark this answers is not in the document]
tyrasd: I think the correct spelling is actually _Electric Vehicle Supply Equipment_, not _Electrical …_.

Source, e.g. ISO
15118-2: https://downloads.regulations.gov/FHWA-2022-0008-0405/attachment_2.pdf

matkoniecz: also, is it Electric by any chance?

https://european-union.europa.eu/index_en search results use `electric vehicle supply equipment (EVSE)` rather electrical

I changed OSM Wiki page

matkoniecz: also other sources and pages seems to use "electric vehicle" not "electrical vehicle"

RoiEXLab:
@tyrasd Thank you for linking the related issue, I had no idea this was happening at the same time.

@matkoniecz I corrected the spelling, thanks for digging up an official source. I didn't quite follow your suggestion, but I added "ID" at the end which should hopefully make things clearer. (Will update the screenshots in the original post once the CI lets me, GH is so annoying with these limitations for first-time contributors, but oh well).

> Single charging point can have multiple sockets each with own code. Would it make sense to have some support for such multivalues?

I don't quite get what you mean by "multivalues". According to https://taginfo.openstreetmap.org/keys/ref%3AEU%3AEVSE#values it is already very common practice to simply separate multiple values using a semicolon. I don't think the tagging schema has a way to represent that unfortunately.

tyrasd (approved):

PR #1774, opened 2025-10-24
make the vscode 'iconify' extension work

k-yle:
following https://github.com/iconify/icon-sets/issues/206, we can now use [this vscode extension](https://marketplace.visualstudio.com/items?itemName=antfu.iconify) to preview icons while editing: 

image

this requires one minor change to the shared [settings.json](https://github.com/openstreetmap/id-tagging-schema/blob/main/.vscode/settings.json) file in this repo, for anyone who wants to use this extension

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -39,5 +39,12 @@
     ],
     "files.exclude": {
         "**/dist": true
-    }
+    },
+
+    // iD uses a custom prefix for these icon providers:
+    "iconify.customCollectionIdsMap": {
+        "fas": "fa-solid",
+        "far": "fa-regular",
+        "fab": "fa-brands",
+    },
 }


matkoniecz (approved):
it looks like special-cased support for specific software, but noninvasive one and we already have vscode-specific file so that ship has sailed anyway

not tested, but it looks like it was tested by PR author

.vscode/settings.json
```diff
@@ -39,5 +39,12 @@
     ],
     "files.exclude": {
         "**/dist": true
-    }
+    },
+
+    // iD uses a custom prefix for these icon providers:
```
tyrasd, line 44: I assume that vscode still reads this nonstandard json syntax just fine, but maybe we should stick to valid json? :wink:

k-yle:
yeah, I'm not sure. We could just close this, anyone who wants the extension can add this config to their own user-level settings.

For completeness, we _could_ alternatively pull the icons from iconify instead of our [copy-pasted folder of SVGs](https://github.com/openstreetmap/iD/tree/develop/svg/roentgen). This would help reduce [vendor lock-in](https://en.wikipedia.org/wiki/Vendor_lock-in) for fontawesome. but it would be a big breaking change for other consumers of this repo, and would need a proper discussion first

matkoniecz:
> yeah, I'm not sure. We could just close this

I would probably do this, missing support seems superior over unfixably broken one

PR #1776, opened 2025-10-24
add `hazard=*` tag for highways

aaei924:
By this field, iD editor allows people to map `hazard=*` key on highway features easily. `hazard=*` is for boundary and highway, but currently iD tagging schema supports `hazard` key only for boundary. This PR adds `hazard` key support also for highways. This single key has been divided into `hazard_boundary.json` and `hazard_highway.json` files, because the key has different values depending on features.

This tag has been approved by [proposal procedure](https://wiki.openstreetmap.org/wiki/Proposal:Hazard).

### Links and data

**Relevant OSM Wiki links:**
- [key:hazard](https://wiki.openstreetmap.org/wiki/Key:hazard): You can find Wiki page for each values in this page.

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/keys/hazard

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/data/fields/hazard_highway.json
+++ b/data/fields/hazard_highway.json
@@ -19,12 +19,12 @@
             "loose_gravel": "Loose Gravel",
             "low_flying_aircraft": "Low Flying Aircraft",
             "pedestrians": "Pedestrians",
-            "queues_likely": "Queues Likely",
+            "queues_likely": "Stopped Traffic",
             "school_zone": "School Zone",
             "side_winds": "Side Winds",
             "slippery": "Slippery",
             "turn": "Turn",
-            "turns": "Reverse Turn"
+            "turns": "Multiple Turns"
         }
     },
     "label": "Signed Hazard"


tordans:
> This single key has been divided into hazard_boundary.json and hazard_highway.json files, because the key has different values depending on features.

I don't see anything about `hazard_boundary` in the changes, though.

tordans:
@aaei924 the examples at https://wiki.openstreetmap.org/wiki/Proposal:Hazard#Examples all show a combination of tags … but different tags. Mostly `traffic_sign`, but also `boundary`, `landuse`. This PR does not handle any relation to other tags. Is the established practice to just use the `hazard` key on `highway=*` without addition?

Another thing I wonder is, that https://wiki.openstreetmap.org/wiki/Key:hazard#Traffic_hazards does not promote tagging the road way too much. Tagging the traffic sign itself is mentioned instead.

And TBH, I wonder why we would want to promote tagging the road way given that those signs are super hard to tag on a way: The have a start but usually not an "end" sign, so one has to come up with a length without being able to verify it. And once we split the road, the sign is copied which would suggest a new `traffic_sign` object that is not there. We have that same issue with `maxspeed` of course, but those have a better start and end definition or signage.

So personally I am not convinced we would want those tags to be promoted on ways. Rather having more `traffic_sign=* + hazard=*` nodes as a vertex or next to the road.

tordans:
On a technical note, there are some  highway classes missing like `trunk_link`. But lets first talk about the general idea.

1ec5:
At least in the Americas, warning signs of all sorts can come with “Next _X_ Miles” or “Next _X_ km” plaques. To the contrary, I couldn’t imagine tagging `hazard=curve` on just a vertex along the roadway, since the extent of the curve is plainly visible and the warning is usually tied to an advisory speed (`maxspeed:advisory=*`).

data/fields/hazard_highway.json
```diff
@@ -0,0 +1,31 @@
...
+    "type": "combo",
+    "strings": {
+        "options": {
+            "animal_crossing": "Animal Crossing",
+            "bump": "Bump",
+            "children": "Children",
+            "curve": "Curve",
+            "curves": "Multiple Curves",
+            "cyclists": "Cyclists",
+            "dangerous_junction": "Dangerous Junction",
```
matkoniecz, line 12: https://wiki.openstreetmap.org/wiki/Tag:hazard=dangerous_junction should be fixed first, right now it looks like permitting 100% subjective tagging

while https://wiki.openstreetmap.org/wiki/Key:hazard has ***signed*** hazards on roadways
```suggestion
            "dangerous_junction": "Junction Signed As Dangerous",
```

data/fields/hazard_highway.json
```diff
@@ -0,0 +1,31 @@
...
+            "pedestrians": "Pedestrians",
+            "queues_likely": "Queues Likely",
+            "school_zone": "School Zone",
+            "side_winds": "Side Winds",
+            "slippery": "Slippery",
+            "turn": "Turn",
+            "turns": "Reverse Turn"
+        }
+    },
+    "label": "Hazard"
```
tordans:
```suggestion
    "label": "Hazard (Signed)"
```

Or

```suggestion
    "label": "Signed Hazard"
```

Had the same thought as @matkoniecz 
But this goes for all those values, right?

data/fields/hazard_highway.json
aaei924: This point may be able to be resolved by changing label of 'hazard' tag as the suggestion below. It clarifies that all values under this tag are signed.

aaei924:
> @aaei924 the examples at https://wiki.openstreetmap.org/wiki/Proposal:Hazard#Examples all show a combination of tags … but different tags. Mostly `traffic_sign`, but also `boundary`, `landuse`. This PR does not handle any relation to other tags. Is the established practice to just use the `hazard` key on `highway=*` without addition?
> 
> Another thing I wonder is, that https://wiki.openstreetmap.org/wiki/Key:hazard#Traffic_hazards does not promote tagging the road way too much. Tagging the traffic sign itself is mentioned instead.
> 
> And TBH, I wonder why we would want to promote tagging the road way given that those signs are super hard to tag on a way: The have a start but usually not an "end" sign, so one has to come up with a length without being able to verify it. And once we split the road, the sign is copied which would suggest a new `traffic_sign` object that is not there. We have that same issue with `maxspeed` of course, but those have a better start and end definition or signage.

This tag on highway is seemed to have been intended to use with `traffic_sign=*` tags. 

Many hazard signs even don't have any explicit end sign, but in some cases, for example, most of `hazard=school_zone` signs in Korea have their end sign on the road. We had better add this tag on ways only in cases we can find out where they start and end. If we don't know where the end point is, just placing a node and applying `hazard` tag may be encouraged.

1ec5:
Even if a warning sign doesn’t come with an explicit distance, it often still has a natural start and end because the same sign is posted as you approach from either direction. (This only applies to two-way roads, of course.) I _think_ users would be able to figure that out themselves, but if not, the help text from the data item could suggest the right method.

data/fields/hazard_highway.json
```diff
@@ -0,0 +1,31 @@
...
+            "dip": "Dip",
+            "falling_rocks": "Falling Rocks",
+            "frost_heave": "Frost Heave",
+            "horse_riders": "Horse Riders",
+            "ice": "Ice",
+            "landslide": "Landslide",
+            "loose_gravel": "Loose Gravel",
+            "low_flying_aircraft": "Low Flying Aircraft",
+            "pedestrians": "Pedestrians",
+            "queues_likely": "Queues Likely",
```
1ec5: The default locale, American English, doesn’t normally use the word “queue”. It would be a “line”, but people don’t associate “line” with a queue of cars. A more effective label would be “Stopped Traffic”. (The signs in most of the U.S. say “Watch for Stopped Traffic”.) Other locales will be able to translate it based on “queue” as appropriate.

data/fields/hazard_highway.json
```diff
@@ -0,0 +1,31 @@
...
+            "school_zone": "School Zone",
+            "side_winds": "Side Winds",
+            "slippery": "Slippery",
+            "turn": "Turn",
+            "turns": "Reverse Turn"
```
1ec5: This is what the U.S. sign is technically called, but I’m not sure that laypeople would recognize this term, since it never appears on the sign as text. “Multiple Turns” might be easier to understand because it’d be consistent with “Multiple Curves”.

matkoniecz:
@tordans can you approve the run?

matkoniecz (changes requested):
note "and other tag pages should also be fixed" 

I checked random one - https://wiki.openstreetmap.org/wiki/Tag%3Ahazard%3Dchildren and it is also missing info that it is valid only for signed hazards

other value pages also should be fixed

aaei924:
> Is anyone planning to fix documentation?

I changed all documents of approved values at traffic hazards. It's good to proceed.

PR #1816, opened 2025-11-23
tweak toilet terms

matkoniecz:
* rename generic amenity=toilets preset to "Toilets" (was: "Restroom"), as restroom is typically only used for indoor toilets, but the tag can also be used for outdoor ones
* rename field for "toilets" tag to "Restroom" (for the opposite reason as above)
* move alias "Water Closet" to "Flush Toilets" preset

redoes large part of #1159

see #1159

### Related issues

closes #1159 (deletion of portable toilets file is fairly easy, no need to keep that broken PR)

### Links and data

**Relevant OSM Wiki links:**

none, I think

**Relevant tag usage stats:**

irrelevant

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/data/presets/amenity/toilets.json
+++ b/data/presets/amenity/toilets.json
@@ -32,17 +32,17 @@
         "privy",
         "public restroom",
         "public toilet",
+        "water closet",
         "w.c.",
         "wc"
     ],
     "tags": {
         "amenity": "toilets"
     },
-    "name": "Restroom",
+    "name": "Toilets",
     "aliases": [
-        "Toilets",
+        "Restroom",
         "Bathroom",
-        "Lavatory",
-        "Water Closet"
+        "Lavatory"
     ]
 }
--- a/data/fields/toilets.json
+++ b/data/fields/toilets.json
@@ -1,5 +1,16 @@
 {
     "key": "toilets",
     "type": "check",
-    "label": "Toilets"
+    "label": "Restroom",
+    "terms": [
+        "bathroom",
+        "latrine",
+        "lavatory",
+        "outhouse",
+        "privy",
+        "toilet",
+        "water closet",
+        "W.C.",
+        "WC"
+    ]
 }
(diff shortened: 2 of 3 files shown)


data/presets/amenity/toilets.json
```diff
@@ -32,17 +32,17 @@
...
         "wc"
     ],
     "tags": {
         "amenity": "toilets"
     },
-    "name": "Restroom",
+    "name": "Toilets",
     "aliases": [
-        "Toilets",
+        "Restroom",
```
1ec5, line 44: I’m not sure keeping “restroom” as an alias is adequate for the default American English localization. This preset was renamed in #419 because “toilet” in this dialect can only refer to the bathroom fixture, not the restroom as a whole. Calling the preset “toilets” could give mappers the impression that it’s for micromapping individual stalls within a restroom.

“Restroom” is not necessarily indoors: a rest area or gas station might have a standalone restroom accessible from the outside. At a park, it might only have enough shelter for privacy but not from the elements. As a rule of thumb, if it’s more substantial than a single toilet fixture, like if it has a sink, then it’s a restroom. If it’s just the toilet fixture with minimal housing around it, then it’s a toilet. Referring to anything else as a toilet is crass and impolite.

American English is not the only locale that lacks a unified term that covers both a restroom and an outhouse. Other languages also have a usual word for restrooms that literally calls it a kind of room. One possible way forward would be to make a more specific preset for a restroom, but I’m not sure what the tagging would be. In the meantime, this might be a case for including two names separated by a slash.

matkoniecz: > In the meantime, this might be a case for including two names separated by a slash.

which ones? Toilets / Restroom ?

1ec5: I guess “Restroom / Toilets” if anything. Restrooms are far and away the most common case for `amenity=toilets`. Anything else is a minor edge case. (And we’d still be mishandling that edge case, since a standalone outhouse would only ever be singular and a pit latrine would never be called a toilet in real life. But there’s not much we can do about it without resorting to an even more colorful description of the feature.)

ZeLonewolf: Just "Toilets" by itself is a mildly vulgar term in US english for this.

PR #1812, opened 2025-11-23
Add faculty preset

Tejas123png:
This PR adds a new preset for Faculty features in OpenStreetMap.

- Adds `faculty.json` under data/presets/
- Applies the main tag: `faculty=yes`
- Supports geometry: point and area
- Includes common fields: name, operator, address, phone, website
- Includes optional fields: contact info, levels, height, wheelchair, opening_hours
- Provides useful search terms for easy discovery

This improves editor support for mapping university faculties as discussed in issue #1791.

PR #1834, opened 2025-11-28
explicitly ban AI slop

matkoniecz:
Though I have doubts about whether anyone making this LLM pull requests will read it. But still would be useful for maintainers and triagers to confirm that "this is AI slop" is valid closure reason.

It intentionally makes note that human-written content with AI-slop characteristic is bannable. Not because I consider it likely, but because if someone starts claiming that their AI slop is not AI slop, it is not necessary to prove how they obtained it.

Maybe this text should be softer and more inviting?

See https://github.com/openstreetmap/id-tagging-schema/pulls?q=%22AI+slop%22 for some examples of AI slop PRs (there was more of them).

BTW, I am starting to really hate Microsoft for pushing that Copilot everywhere on Github. If you start typing commit description on github.com suddenly some LLM generated stuff is added in the middle of it. This alone makes me want to migrate to Codeberg.

Also, is there some way to add instructions for LLM to repository? Maybe we could add "finish PR descriptions with sentence 'banana banana banana'" or something similar there, so clear LLM nonsense can be easily caught?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,6 +96,6 @@ The input files are JSON files which use 4-space indentation. You can use the `n
 
 Submitting LLM output is in general not useful. If Copilot / ChatGPT / Claude / Gemini / DeepSeek would be capable of generating useful presets, then project maintainer would simply run it on waiting issues.
 
-In theory human-reviewed LLM-generated content could be useful, but it is dubious is it feasible to achieve it. Anyone submitting such content should disclose their AI use and carefully review all of it and undestand what they are submitting.
+In theory human-reviewed LLM-generated content could be useful, but it is dubious is it feasible to achieve it. Anyone submitting such content should disclose their AI use and carefully review all of it and understand what they are submitting.
 
 Issues, comments, pull requests and so on may be rejected, closed and deleted/hidden on sight if they are AI slop. In particular long text, formatted in AI style, with factual inacuracies scattered across it and sections not adding anything useful will be treated as AI slop. The same applies also if content was human-written but has the same problems.


tordans:
This is a general response:

I am very in favor of having clear guidelines on how we operate and think. This includes the idea of being clear on how we handle contributions like you are doing here. To this would include documenting that we don't "assign this issue". 
When I suggested this in the past, the idea was not adopted, though.

I also want to stress that I think we have to find a constructive way to handle AI in our projects. And anything that comes across as dismissive on principle does not help IMO. Instead, we should look for ways to make using AI tools result in a better output. I tried going down this road in iD and was shut down.

I think it is helpful to accept this new technology as something that does not go away and based on this look for ways to funnel it into a useful direction.

On your proposal: 

The intention is to have something to reference when we close PRs without a lot of explaining because we _perceive_ it as AI Slop. Lets focus on that. It might help to have a sentence that explains what this Repo is about, similar to https://github.com/openstreetmap/id-tagging-schema/pull/1812#issuecomment-3568041435

Tooling: 

There are instruction files that Copilot and Cursor read. We should look into that so see if they are a better place to put this in front of the Users. For example in the way to instruct AI to create a test strategy as well as a research documentation that needs to be followed before/after creating a contribution.

But first, lets wait a bit to see what others say.

tordans:
> > I also want to stress that I think we have to find a constructive way to handle AI in our projects.

> "AI as of now is not wanted for generating pull requests" seems a viable position to me, and does not block use of AI for analysis/detecting potential problems/debugging assistance.

That is the point I am trying to make. We do not care about which Editor – Vim, VS Code, Notepad2 or even the Github UI – a users uses to write code. In the same way we should not care if I create a preset by using Copilot. The point we have to make is, that the quality of what is submitted has to be high. And we should focus on describing what we mean by that and look for ways to enforce it (like the rules-files, PR template, Testing-Guidelines, …). 
Yes, we have an increase in low quality PRs. But by banning the tool we miss on the opportunity to improve our processes and take advantage of the (developing) opportunity.

matkoniecz:
wave of AI slop PRs went away, we can clearly close them anyway without specific AI rant rules, so probably spending time on regulations is not really optimal use of our time

this does not change that slop PRs will be closed

matkoniecz:
and if AI slop PRs will again be a problem it can be resurrected in this or different form

PR #1838, opened 2025-11-30
Add missing options to `playground`

cuatim:

This brings the playground devices up-to-date with the [wiki](https://wiki.openstreetmap.org/wiki/Key%3Aplayground).

### Links and data

**Relevant OSM Wiki links:**

https://wiki.openstreetmap.org/wiki/Key%3Aplayground

**Relevant tag usage stats:**

- https://taginfo.openstreetmap.org/tags/playground=agility_trail 183
- https://taginfo.openstreetmap.org/tags/playground=rotating_seesaw 36

Note: there are 10+ playground devices in `data/fields/playground.json` already that have even less objects tagged.

### DIFF
--- a/data/fields/playground.json
+++ b/data/fields/playground.json
@@ -6,6 +6,7 @@
         "options": {
             "activitypanel": "Activity Panel",
             "aerialrotator": "Aerial Rotator",
+            "agility_trail": "Agility Trail",
             "archimedes_screw": "Archimedes Screw",
             "balance": "Balancing Device (unspecified)",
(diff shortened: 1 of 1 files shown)


tordans:
@cuatim could you also look into adding OSM Wiki Data Items for those two entries?

Right now, we show the key docs: 

image


… Instead of specific docs: 
image
(^- which is a bad example because those docs are not helpful…)

There are docs here in the Repo on "Info-i" and how to add those entries to the Wiki

PR #1839, opened 2025-11-30
Add preset for `playground=pump` and align wording for `playground=water`

cuatim:

`playground=water` is the generic version of a water play device.

### Links and data

**Relevant OSM Wiki links:**

- https://wiki.openstreetmap.org/wiki/Key:playground

**Relevant tag usage stats:**

- https://taginfo.openstreetmap.org/tags/playground=water 1730
- https://taginfo.openstreetmap.org/tags/playground=pump 174

### DIFF
--- a/data/presets/playground/pump.json
+++ b/data/presets/playground/pump.json
@@ -1,0 +1,10 @@
+{
+    "icon": "temaki-water",
+    "geometry": [
+        "point"
+    ],
+    "tags": {
+        "playground": "pump"
+    },
+    "name": "Play Water Pump"
+}


tordans:
Thanks @cuatim for looking into this.

I agree we should rename the `playground=water` so it aligns better with https://github.com/cuatim/id-tagging-schema/blob/main/data/fields/playground.json#L80

However, I am not sure we actually need this separate preset or should add other presets. I started to look into this related to https://github.com/openstreetmap/id-tagging-schema/issues/960 twice already but never finished my work.

My take is: We should only add separate presets when they have something more to offer than the main preset.
- especially fields that only make sense for this equipment
- but also search terms

I don't see both in either of the presets.

So I wonder, if we should just remove the separate playground=water preset instead…

tordans (approved):
I will merge this as is but commented on https://github.com/openstreetmap/id-tagging-schema/issues/960#issuecomment-3619667560 to handle the generic presets different in the future. **Follow up PRs would be welcome.**


---

FYI and as possible follow up:

**Whenever we have specific presets, lets look into adding useful fields.**
In this case the usage is low, yet, so this is not so easy.

Below are some numbers on current usage. 
Possible follows ups could be to look int adding fields for `drinking_water` and `pump` with useful interaction/options.

```
% curl -g https://postpass.geofabrik.de/api/0.2/interpreter \
  --data-urlencode "options[geojson]=false" \
  --data-urlencode "data=
SELECT
  key,
  COUNT(*) AS usage_count
FROM postpass_pointlinepolygon,
LATERAL jsonb_object_keys(tags) AS key
WHERE tags->>'playground' = 'pump'
GROUP BY key
HAVING COUNT(*) > 5
ORDER BY usage_count DESC"

{
    "result": [
        {
            "key": "playground",
            "usage_count": 174
        },
        {
            "key": "pump",
            "usage_count": 39
        },
        {
            "key": "man_made",
            "usage_count": 32
        },
        {
            "key": "drinking_water",
            "usage_count": 20
        },
        {
            "key": "material",
            "usage_count": 9
        },
        {
            "key": "pump:status",
            "usage_count": 7
        },
        {
            "key": "access",
            "usage_count": 7
        },
        {
            "key": "check_date",
            "usage_count": 6
        },
        {
            "key": "mechanical_driver",
            "usage_count": 6
        }
    ],
    "metadata": {
        "generator": "Postpass API 0.2",
        "timestamp": "2025-12-06T06:36:50Z"
    }
}%
```

And more to the point

 ```
% curl -g https://postpass.geofabrik.de/api/0.2/interpreter \
  --data-urlencode "options[geojson]=false" \
  --data-urlencode "data=
SELECT
  kv.key,
  kv.value,
  COUNT(*) AS usage_count
FROM postpass_pointlinepolygon,
LATERAL jsonb_each_text(tags) AS kv(key, value)
WHERE tags->>'playground' = 'pump'
GROUP BY kv.key, kv.value
HAVING COUNT(*) > 10
ORDER BY usage_count DESC"

{
    "result": [
        {
            "key": "playground",
            "value": "pump",
            "usage_count": 174
        },
        {
            "key": "pump",
            "value": "manual",
            "usage_count": 39
        },
        {
            "key": "man_made",
            "value": "water_well",
            "usage_count": 26
        },
        {
            "key": "drinking_water",
            "value": "no",
            "usage_count": 17
        }
    ],
    "metadata": {
        "generator": "Postpass API 0.2",
        "timestamp": "2025-12-06T06:53:10Z"
    }
}%
```

PR #1843, opened 2025-12-02
Update Actions in CI and add Dependabot configuration

Harvester57:

Hi there ! :)

This pull request updates the project's continuous integration configuration by extending Dependabot's monitoring capabilities. The change ensures that GitHub Actions used within the repository are automatically checked for updates on a daily basis, same as the NPM current configuration.

This PR also update the current Actions used to their latest versions available, and use the recommended best practice to pin each version to their release SHA hash. Dependabot is able to update the hash of future releases and update the comment to reflect which semantic version it corresponds to.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,15 +13,11 @@ jobs:
   test:
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        node-version: ['20', '22', '24']
-
     steps:
     - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
       with:
-        node-version: ${{ matrix.node-version }}
+        node-version-file: '.nvmrc'
     - run: npm clean-install
     - run: npm run test


.github/workflows/build-preview.yml
```diff
@@ -12,13 +12,13 @@ jobs:
   build-preview:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
```
matkoniecz, line 15: it goes from v4 to (according to comment, not verified it yet) v6

are there breaking changes here

BTW, it would probably make sense to verify where this hash is pointing - or merge only `github/dependabot.yml` and let dependabot to open PRs.

.github/workflows/test.yml
```diff
@@ -15,12 +15,12 @@ jobs:
 
     strategy:
       matrix:
-        node-version: ['18']
+        node-version: ['20', '22', '24']
```
matkoniecz: what are consequences of trippling runs of tests? are we likely to hit some usage limits?

tyrasd: I'd say for the tagging repo it would sufficient to just test it on one node version, as the repo is really mostly data-heavy and testing the data itself instead of the code[^1]. So, adding more than one node version here just makes the tests run slower than necessary.

[^1]: the code for this repo is external at https://github.com/ideditor/schema-builder/

.github/workflows/build-preview.yml
Harvester57: Update to from v4.3.0 to v6.0.1 doesn't introduce any breaking change according to the Action changelog, it's only an internal update of the Action to use NodeJS 24.x.

Your last comment makes a lot of sense, although I'm not sure that with the current notation ('v4' without hash pinning), Dependabot would introduce the digest itself, it would probably migrate to 'v6' or 'v6.0.1'

.github/workflows/test.yml
Harvester57: The three tests are run in parallel, so I'm not sure it would increase the total duration of the pipeline. The logic behind this change was to test against the current LTS version (v24) along the maintenance LTS versions (v20 and v22), as the official NodeJS recommendation is the following:

> Production applications should only use Active LTS or Maintenance LTS releases.[^1]

I have no particular feeling about this, so sticking only to v24 is fine by me :)

Regarding the usage limits, they are documented [here](https://docs.github.com/en/actions/reference/limits) and [here (for the job concurrency limits, 20 for the free plans)](https://docs.github.com/en/actions/reference/limits#job-concurrency-limits-for-github-hosted-runners)

[^1]: https://nodejs.org/en/about/previous-releases#:~:text=Production%20applications%20should%20only%20use%20Active%20LTS%20or%20Maintenance%20LTS%20releases.

PR #1846, opened 2025-12-03
Drop unneeded deprecations (0 matches according to Taginfo)

Dimitar5555:
All removed deprecations have 0 potential matches according to Taginfo.

All changes which require >1 tags and drop/change >1 of their values are skipped.

<details>
<summary>Script used for this change</summary>

```js
import fs from 'fs';
const taginfo_url = 'https://taginfo.openstreetmap.org/api/4/';
const input = fs.readFileSync('input.json', 'utf-8');
const data = JSON.parse(input);

for(const item of data) {
    const { old, replace } = item;
    
    const old_tags = new Set(Object.entries(old).map(([k, v]) => `${k}=${v}`));
    const new_tags = new Set(Object.entries(replace || {}).map(([k, v]) => `${k}=${v}`));
    const diff = old_tags.difference(new_tags);
    if(diff.size > 1) {
        console.warn('Skipping item with multiple tag changes:', item);
        continue;
    }
    else if(diff.size === 0) {
        console.warn('No tag changes found for item:', item);
        continue;
    }
    const [key, value] = Array.from(diff)[0].split('=');
    let url;
    if(value != '*') {
        url = new URL(`${taginfo_url}tag/stats?key=${key}&value=${value}&filter=all`);
    }
    else {
        url = new URL(`${taginfo_url}key/stats?key=${key}&filter=all`);
    }

    console.log('Fetching stats for', url.toString());
    const stats = await fetch(url).then(res => res.json());
    item.count = stats.data.length > 0 ? stats.data[0].count : 0;
}

const filtered = data.filter(d => d.count > 0 || d.count === undefined)
.map(el => {
    delete el.count;
    return el;
});

function generate_json_el(element) {
    function tags_to_single_row(tags={}) {
        const result =  Object.entries(tags).map(([k, v]) => `"${k}": "${v}"`).join(', ');
        return `{${result}}`;
    }
    const rows = [];
    rows.push('  {');
    rows.push(`    "old": ${tags_to_single_row(element.old)}`);
    if(Object.keys(element.replace || {}).length > 0) {
        rows[rows.length - 1] += ',';
        rows.push(`    "replace": ${tags_to_single_row(element.replace)}`);
    }
    rows.push('  }');
    return rows.join('\n');
}

const filtered_text = `[
${filtered.map(el => generate_json_el(el)).join(',\n')}
]`
fs.writeFileSync('output_1+_instances.json', filtered_text, 'utf-8');

data.sort((a, b) => b.count - a.count);

fs.writeFileSync('output_sorted_by_count.json', JSON.stringify(data, null, 2), 'utf-8');
```

</details>

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1,13 +1,5 @@
 [
   {
-    "old": {"aerialway": "canopy"},
-    "replace": {"aerialway": "zip_line"}
-  },
-  {
-    "old": {"aeroway": "aerobridge"},
-    "replace": {"aeroway": "jet_bridge", "highway": "corridor"}
-  },
-  {
     "old": {"access": "customer"},
     "replace": {"access": "customers"}
   },
@@ -16,24 +8,8 @@
     "replace": {"access": "yes"}
   },
   {
-    "old": {"aerodrome": "military"},
-    "replace": {"military": "airfield"}
-  },
-  {
     "old": {"aerodrome:type": "military"},
     "replace": {"military": "airfield"}
-  },
-  {
-    "old": {"agrarian": "agrcultural_machinry"},
-    "replace": {"agrarian": "agricultural_machinery"}
-  },
-  {
-    "old": {"amenity": "advertising"},
-    "replace": {"advertising": "*"}
-  },
-  {
-    "old": {"amenity": "artwork"},
-    "replace": {"tourism": "artwork"}
   },
   {
     "old": {"amenity": "bail_bonds"},
@@ -48,40 +24,16 @@
     "replace": {"amenity": "dressing_room"}
   },
   {
-    "old": {"amenity": "citymap_post"},
-    "replace": {"tourism": "information"}
-  },
-  {
     "old": {"amenity": "club"},
     "replace": {"club": "*"}
-  },
-  {
-    "old": {"amenity": "community_center"},
-    "replace": {"amenity": "community_centre"}
-  },
-  {
-    "old": {"amenity": "dog_bin"},
-    "replace": {"amenity": "waste_basket", "waste": "dog_excrement"}
-  },
-  {
-    "old": {"amenity": "dog_waste_bin"},
-    "replace": {"amenity": "waste_basket", "waste": "dog_excrement"}
   },
   {
     "old": {"amenity": "embassy"},
     "replace": {"office": "diplomatic"}
   },
   {
-    "old": {"amenity": "ev_charging"},
-    "replace": {"amenity": "charging_station"}
-  },
-  {
     "old": {"amenity": "charging_station", "car": "*"},
     "replace": {"amenity": "charging_station", "motorcar": "$1"}
-  },
-  {
-    "old": {"amenity": "fire_hydrant"},
-    "replace": {"emergency": "fire_hydrant"}
   },
   {
     "old": {"amenity": "firepit"},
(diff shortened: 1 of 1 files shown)


data/deprecated.json
```diff
@@ -15,26 +7,10 @@
...
-  {
-    "old": {"aerodrome": "military"},
-    "replace": {"military": "airfield"}
-  },
   {
     "old": {"aerodrome:type": "military"},
     "replace": {"military": "airfield"}
   },
-  {
-    "old": {"agrarian": "agrcultural_machinry"},
```
matkoniecz, line 27: I see no reason to spend time on dropping such cases, this is typo that can plausibly come back

no benefit from dropping it, and we have plenty of typos in various tags - this can help against more creeping it

(adding all possible typo fixes, also ones with 0 current cases is not useful, but if present already I would not remove them)

Dimitar5555: > Deprecations are not for cleanup:
>
> Deprecation rules work such that the user sees a message with suggestions and can act only when editing the given element. This makes them well-suited for gradual, human-reviewed updates of taggings like crossings. **However, they are not suitable for cleaning up incorrect tagging from the database, especially for low-volume changes**.

From https://github.com/openstreetmap/id-tagging-schema/blob/f5a76958359104f9b16cc75e6d4390be6aeba5f4/GUIDELINES.md

> I see no reason to spend time on dropping such cases, this is typo that can plausibly come back

Those rules are so unlikely to be triggered that pigs will probably learn to fly first. In any case, most issues can be resolved in seconds by a bot or a mass edit.

For reference, this specific mistake had 100+ uses in 2019 and was added because it showed up in a dropdown. See `https://github.com/openstreetmap/iD/issues/7053`.

matkoniecz: > In any case, most issues can be resolved in seconds by a bot or a mass edit.

based on my own experience, it takes noticeable more time

data/deprecated.json
```diff
@@ -47,42 +23,18 @@
...
-    "old": {"amenity": "ev_charging"},
-    "replace": {"amenity": "charging_station"}
-  },
   {
     "old": {"amenity": "charging_station", "car": "*"},
     "replace": {"amenity": "charging_station", "motorcar": "$1"}
   },
-  {
-    "old": {"amenity": "fire_hydrant"},
-    "replace": {"emergency": "fire_hydrant"}
```
matkoniecz, line 84: and this one is also entirely plausible mistake

for me this kind of rules are actually proper use of iD deprecation list

Dimitar5555: There are quite a few allegedly plausible mistakes that are not in this list currently and don't exist in the database. Shall we have them added to `deprecated.json`? `*=y/ye/ys/Yes/YES/нет/да` to name a few.

matkoniecz: > that are not in this list currently and don't exist in the database

no, as commented already in https://github.com/openstreetmap/id-tagging-schema/pull/1846#discussion_r2590323305

data/deprecated.json
```diff
@@ -1488,18 +1132,6 @@
     "old": {"religion": "catholic"},
     "replace": {"religion": "christian", "denomination": "catholic"}
   },
-  {
-    "old": {"roof:color": "*"},
-    "replace": {"roof:colour": "$1"}
-  },
-  {
-    "old": {"roof:shape": "half_hipped"},
```
matkoniecz, line 1400: for example this is kind of likely mistake that happens over and over again and people keep fixing it over and over again (as it is blatantly obvious)

data/deprecated.json
```diff
@@ -1636,10 +1244,6 @@
     "old": {"shop": "telecommunications"},
     "replace": {"shop": "telecommunication"}
   },
-  {
-    "old": {"shop": "tickets"},
```
matkoniecz, line 1544: See https://wiki.openstreetmap.org/wiki/Special:WhatLinksHere/Tag:shop%3Dtickets

it is currently being eliminated by my bot edit, but please do not make it load bearing.

I am making no promise that it will be running in 6 months or 6 years or 16 years.

tyrasd:
I would agree with @matkoniecz that it is still worthwhile to keep some of these, e.g. `color -> colour` is a likely tagging mistake when using the raw tag editor. So, if we're at it, we should go through the list and to keep entries that are still likely to happen, even if the osm database does currently not have any instances.

Issue #1531, opened 2025-04-25
iD incorrectly tries to convert old to new parking scheme


wielandb:
I noticed a lot of streets in my vicinity started to pop up in the StreetComplete parking overlay as "no data", and after investigation, I found that the iD suggestions do incorrectly and/or incompletely convert the data, leading to information loss and corrupt data.

Examples:

https://www.openstreetmap.org/way/1039535438
https://www.openstreetmap.org/way/1053141214

Suggested action:

iD should not try to convert the old to the new street parking scheme. (Or there needs to be much more sturdy rulesets)

tordans:
This is probably from https://github.com/openstreetmap/id-tagging-schema/pull/1390. can you spot where the rules are wrong?

wielandb:
From what I could see so far (and citing from [my own ruleset](https://github.com/wielandb/maproulette-tagfixes/blob/main/challenges/parking_converter/tag_convert_description.txt) for my maproulette challenge):

- The parking position seems to get lost on some more occasions, but I'm not sure if it's consistent
  - `painted_area_only` (becomes `on_street` with additional tag `parking:<side>:markings=yes`)
     - Instead it gets converted to a generic "yes" and the parking position is lost
  - `half_on_kerb` should become `parking:<side>=half_on_kerb`
    - but instead just becomes yes and the parking position is lost
- A conversion should not happen if additional tags from the old parking scheme, especially conditions and restrictions are present and not also converted  

I will try to dig into the concrete rulesets and try to suss out a solution.

tordans:
TBH I wonder if we should just remove the rules again. The schema update is quite complex and it will be hard to build rules that are more right than wrong…
Other ways to update the data like your MR challenge are the better approach IMO.
Btw, did you test the new MR integration in Rapid? Unfortunately they don't support fix tag challenges, yet (see https://github.com/facebook/Rapid/issues/1340). But for regular challenges this integration is really useful.

wielandb:
I would be in favour of removing the rules too, if we cannot make them work properly. I will try my best to understand how the rules work.

> Btw, did you test the new MR integration in Rapid? Unfortunately they don't support fix tag challenges

I guess you mean cooperative challenges in general, not TagFix Challenges specifically. 
I have not dabbled in other cooperative challenges yet, and I did not know Rapid has a MR integration. 😳 But didn't iD also once considered implementing something like that? (that = beeing able to load .osc files into the editor)

tordans:
[...]

PR #1390, opened 2024-11-17
Deprecated parking

tiptoptom:
### Description, Motivation & Context

There was already a PR #918 about this. I'm willing to finish it now.

- regarding https://github.com/openstreetmap/id-tagging-schema/pull/918#pullrequestreview-1539142798 I added [parking:lane:<side>:parallel=on_street](https://github.com/openstreetmap/id-tagging-schema/commit/a6af08dcd5ce9a9ab6bff4f1aa80ba6cc59aa8cd)
- regarding https://github.com/openstreetmap/id-tagging-schema/pull/918#discussion_r1269383073 and following I [changed "lane" to "yes"](https://github.com/openstreetmap/id-tagging-schema/commit/b4b20ce5fab676ae4df4928da4d446a2d8ca103a)

### Related issues

Closes #879

### Links and data

**Relevant OSM Wiki links:**
- …

**Relevant tag usage stats:**
> …

### Checklist and Test-Documentation Template

<details><summary>Read on to get your PR merged faster…</summary>

Follow these steps to test your PR yourself and make it a lot easier and faster for maintainers to check and approve it.

**This is how it works:**
1. After you submit your PR, the system will create a preview and comment on your PR:
   > 🍱 You can preview the tagging presets of this pull request here.
   If this is your first contribution to this project, the preview will not happen right away but requires a click from one of the project members. We will do this ASAP.

2. Once the preview is ready, use it to test your changes.

3. Now copy the snippet below into a new comment and fill out the blanks.

4. Now your PR is ready to be reviewed.

```
## Test-Documentation

### Preview links & Sidebar Screenshots

<!-- Use the preview to find examples, select the feature in question and **copy this link here**.
     Find examples of nodes/areas. Find examples with a lot of tags or very few tags. – Whatever helps to test this thoroughly.
     Add relevant **screenshots** of the sidebar of those examples. -->

<!-- FYI: What we will check:
     - Is the [icon](https://github.com/ideditor/schema-builder/blob/main/ICONS.md) well chosen.
     - Are the fields well-structured and have good labels.
     - Do the dropdowns (etc.) work well and show helpful data. -->

### Search

<!-- **Test the search** of your preset and share relevant **screenshots** here.
     - Test the preset name as search terms.
     - Also test the preset terms and aliases as search terms (if present). -->

### Info-`i`

<!-- **Test the info-i** for your fields and preset and share relevant **screenshots** here.
     The info needs to help mappers understand the preset and when to use it.
[...]

PR #1849, opened 2025-12-06
Revert "Deprecated parking"

matkoniecz:
Reverts openstreetmap/id-tagging-schema#1390

see #1531


matkoniecz:
https://github.com/openstreetmap/id-tagging-schema/issues/1531#issuecomment-3619855339

PR #1850, opened 2025-12-08
Add `local_ref` to `public_transport=platform` & `=stop_position` nodes

gy-mate:

Lots of public transport stops have a signposted local reference letter/number (or their combination) in order to help wayfinding.

### Related issues

N/A

### Links and data

**Relevant OSM Wiki links:**

- https://wiki.openstreetmap.org/wiki/Key:local_ref
- https://wiki.openstreetmap.org/wiki/Tag:public_transport%3Dplatform
- https://wiki.openstreetmap.org/wiki/Tag:public_transport%3Dstop_position

**Relevant tag usage stats:**

> Used on 269 489 objects. 85.52% is in combination with the `public_transport` key.

https://taginfo.openstreetmap.org/keys/local_ref

https://taghistory.raifer.tech/#***/local_ref/

### DIFF
--- a/data/fields/local_ref.json
+++ b/data/fields/local_ref.json
@@ -1,0 +1,5 @@
+{
+    "key": "local_ref",
+    "type": "text",
+    "label": "Local Platform Number"
+}
--- a/data/fields/ref_stop_position.json
+++ b/data/fields/ref_stop_position.json
@@ -1,5 +1,5 @@
 {
     "key": "ref",
     "type": "text",
-    "label": "Stop Number"
+    "label": "Network-wide Reference Number"
 }
--- a/data/presets/public_transport/platform.json
+++ b/data/presets/public_transport/platform.json
@@ -1,7 +1,7 @@
 {
     "icon": "temaki-board_transit",
     "fields": [
-        "ref_platform",
+        "local_ref",
         "network",
         "operator",
         "vehicles",
@@ -15,6 +15,7 @@
         "layer",
         "level",
         "lit",
+        "ref_platform",
         "wheelchair"
     ],
     "geometry": [
--- a/data/presets/public_transport/platform.json
+++ b/data/presets/public_transport/platform.json
@@ -15,7 +15,6 @@
         "layer",
         "level",
         "lit",
-        "local_ref",
         "wheelchair"
     ],
     "geometry": [
--- a/data/presets/public_transport/platform.json
+++ b/data/presets/public_transport/platform.json
@@ -15,6 +15,7 @@
         "layer",
         "level",
         "lit",
+        "local_ref",
         "wheelchair"
     ],
     "geometry": [
--- a/data/presets/public_transport/platform.json
+++ b/data/presets/public_transport/platform.json
@@ -1,7 +1,7 @@
 {
     "icon": "temaki-board_transit",
     "fields": [
-        "local_ref",
+        "ref_platform",
         "network",
         "operator",
         "vehicles",
@@ -15,7 +15,6 @@
         "layer",
         "level",
         "lit",
-        "ref_platform",
         "wheelchair"
     ],
     "geometry": [
--- a/data/presets/public_transport/platform_point.json
+++ b/data/presets/public_transport/platform_point.json
@@ -15,6 +15,7 @@
         "gnis/feature_id-US",
         "level",
         "lit",
+        "local_ref",
         "tactile_paving",
         "wheelchair"
     ],
--- a/data/presets/public_transport/platform_point.json
+++ b/data/presets/public_transport/platform_point.json
@@ -2,7 +2,7 @@
     "icon": "temaki-sign_and_bench",
     "fields": [
         "name",
-        "ref_stop_position",
+        "local_ref",
         "network",
         "operator",
         "vehicles",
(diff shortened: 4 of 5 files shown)


matkoniecz (changes requested):
At least for me neither "Local Platform Number" nor "Letters and/or numbers identifying a public transport stop locally" makes meaning of this tag clear to me.

matkoniecz:
for me main problem mentioned in https://github.com/openstreetmap/id-tagging-schema/pull/1850#pullrequestreview-3601067060 remains

gy-mate:
I've updated https://wiki.openstreetmap.org/wiki/Item:Q1443 and https://wiki.openstreetmap.org/wiki/Key:local_ref accordingly. See the result at the end of https://github.com/openstreetmap/id-tagging-schema/pull/1850#issuecomment-3677745765.

@matkoniecz, does this look good to you?

matkoniecz (commented):
I am withdrawing my objection, as problem that was raised was resolved (at least in my opinion)

gy-mate:
@tordans Could you please merge this if it's fine by you? Thank you very much in advance! :)

andrewharvey:
This looks good to me. While I think the wiki pages could do with better documentation around the tags, that shouldn't stop them being used via a preset.

Around me I see `local_ref` used for bus stops within a transport interchange or larger bus station (comprising several stops) where each stop has a different stand letter, so we'd probably localise the preset to "Stand Number" (despite it being a letter).

eg. Stand A, Stand B etc. where `ref` is still being the official stop number, eg. [node/1263263493](https://www.openstreetmap.org/node/1263263493/)

image

[_Photo credit_](https://commons.wikimedia.org/wiki/File:Keolis_Downer_Northern_Beaches_3921.jpg)

For our train stations, the platform numbers, eg. 1, 2, 3 all go under `ref=*` despite them being local to the station.

k-yle:
+1 from me as well, at the interchanges where `local_ref` is used, our signage also makes the difference between the "local platform number" and the stop ID very clear:

Untitled

Untitled

tyrasd (changes requested):
I also found this very confusing. The wiki is not particularly [helpful](https://wiki.openstreetmap.org/w/index.php?title=Tag:public_transport%3Dplatform&oldid=2947106#Tags_used_in_combination):

> ref [is] used for numbers of platforms/platform edges in train stations or bus stations
> […]
> local_ref […] Track number/platform number

To me both tags' description read the exact same. _insert they are the same picture meme_[^1]

[^1]: I have to admit that in my region, the network wide platform ids are not posted on the ground, so I only ever encounter situations like `ref=2;3` for a railway platform or the occasional `ref=C` on a bus stop with multiple berths. 

One issue I see with `local_ref` in particular is that it originates from an import, was not included in the [public transport tagging](https://wiki.openstreetmap.org/w/index.php?oldid=1550843) proposal and was also never "formally" discussed/approved individually as far as I can see.

Ignoring that issue, we need to make this as little confusing as possible for mappers. For example:

* turn the priority of the fieds around: the important `local_ref` field should be always displayed for all platforms, and `ref` only optionally in case it is known.
* the label for the `local_ref` field should be easy to understand and easy to distinguish from the `ref` field. Currently, they are too similar IMO. Maybe the `ref` field label could have some indication that it should be network wide.…^2]
* the `local_ref` should also be included in `data/presets/public_transport/platform.json` shouldn't it?



[^2]: on the other hand, if the `ref` is network-specific: why is this not better mapped as `ref:<network>` instead?

matkoniecz:
will close/reopen to get preview up again

gy-mate:
@matkoniecz and @tyrasd, could you please review the updated PR? Many thanks in advance! :)

matkoniecz:
It is on my list. Sadly, list overall has 290 pull requests (to not count issues that are likely to produce more of pull requests when reviewed).

gy-mate:
Oh my...
I'll be patient! :) Thank you for your work!

matkoniecz (commented):
I am still trying to understand local_ref and has it got community consensus to be used and how it would be used

https://wiki.openstreetmap.org/wiki/Talk:Tag:railway%3Dplatform_edge#ref_vs_local_ref is discussion that worries me

slavish following of https://wiki.openstreetmap.org/wiki/Key:local_ref claims may result in putting actual visible reference into `local_ref` and putting internal GUID not labelled anywhere at all and not used by humans at all into `ref`

opened https://wiki.openstreetmap.org/wiki/Talk:Key:local_ref#%22operator-wide_or_network-wide_number_or_reference_code%22_should_be_qualified

Kovoschiz:
If there's any problem, it can be added to only `=bus_stop` first?

matkoniecz:
I am against admitting into presets key with serious unresolved issues. But I would be happy to help resolving them.

k-yle (approved):
per https://github.com/openstreetmap/id-tagging-schema/pull/1850#issuecomment-3876753607 , this seems like a much needed improvement. I don’t find this confusing at all, maybe it depends on where you live?

matkoniecz:
I agree with this one. The problem is that existing documentation can be easily (mis)intepreted to result in quite broken tagging.

matkoniecz:
> per https://github.com/openstreetmap/id-tagging-schema/pull/1850#issuecomment-3876753607 , this seems like a much needed improvement. I don’t find this confusing at all

well, in https://wiki.openstreetmap.org/wiki/Talk:Key:local_ref#%22operator-wide_or_network-wide_number_or_reference_code%22_should_be_qualified for example I got response that ***obviously*** `local_ref` is for what is signed (given as difference - so `ref` is for what is unsigned?)

Issue #835, opened 2023-03-23
Add preset for horse_walker


Hufkratzer:
### OSM Tag(s)

animal=horse_walker

### How would you like this tag to see supported?

As a Preset

### Label

Horse Walker

### Aliases

* [Hot Walker](https://en.wikipedia.org/wiki/Hot_walker)
* Horse Exerciser
* Mechanical Hot Walker

### Terms

horse riding, horse racing, horse training, equestrian

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:animal=horse_walker

### Status of the Tag

In Use

### Usage of the tag

2,196

### Replaces other Tag?

no

### Regional Tag?

no

### Possible icon
https://commons.wikimedia.org/wiki/File:Horse_(9585)_-_The_Noun_Project.svg
_(don't use the horse riding icon, no one rides in a horse walker)_

### Possible reason to add
As long as there is no preset for horse walkers available, users often mistakenly use the preset for riding centres (leisure=horse_riding) for them; most recent examples:
- https://www.openstreetmap.org/changeset/133983562
- https://www.openstreetmap.org/changeset/129900838

### Possible reasons NOT to add
* animal key is not ideal
* 48 % of all uses are in Ireland so far

Hufkratzer:
> https://fontawesome.com/v5/icons/horse?f=classic&s=solid could also work as an icon

I know, but this horse doesn't walk, it stands.

> also potentially of interest: there exists an abandoned proposal for this tag: https://wiki.openstreetmap.org/wiki/Proposed_features/animal%3Dhorse_walker

I know, I [made](https://wiki.openstreetmap.org/wiki/Tag:animal=horse_walker) the [tag page](https://wiki.openstreetmap.org/wiki/Tag:animal=horse_walker) like this proposal in 2018, and they are still quite similar.; I only removed barrier=fence as a refining tag recently, compare [talk page](https://wiki.openstreetmap.org/wiki/Talk:Tag:animal%3Dhorse_walker).

PR #1857, opened 2025-12-13
Add horse walker preset (`animal=horse_walker`)

Hufkratzer:

This adds a new preset for tag [animal=horse_walker](https://wiki.openstreetmap.org/wiki/Tag:animal%3Dhorse_walker)

It is important to have a preset for horse walkers to avoid mappers using other tags (e.g. `leisure=horse_riding`) for them instead.

I am not particularly happy with the key `animal` of this tag as a horse walker is not an animal. But we already have about 3.4k uses of this tag, so it is not so easy to change anymore. The tag was [proposed in 2013](https://wiki.openstreetmap.org/w/index.php?title=Proposal:Animal%3Dhorse_walker&oldid=857101).

I used the icon [temaki-amusement_park](https://rapideditor.github.io/temaki/docs/#icon-temaki-amusement_park) although it is not intended for horse walkers; but I think it looks more specific than [fas-horse](https://fontawesome.com/icons/horse?s=solid).

### Related issues

Closes [#835](https://github.com/openstreetmap/id-tagging-schema/issues/835)

### Links and data

Used about 3.4k times so far.

https://wiki.openstreetmap.org/wiki/Tag:animal=horse_walker

https://taginfo.openstreetmap.org/tags/animal=horse_walker

https://taghistory.raifer.tech/#***/animal/horse_walker

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/data/presets/animal/horse_walker.json
+++ b/data/presets/animal/horse_walker.json
@@ -2,10 +2,10 @@
     "icon": "temaki-amusement_park",
     "fields": [
         "building_area",
-        "covered",
-        "access_simple"
+        "covered"
     ],
     "moreFields": [
+        "access_simple",
         "operator",
         "lit",
         "indoor",


Hufkratzer:
## Test-Documentation

#### Horse walker with building tag -> ok

https://pr-1857--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=19.00/52.66601/-7.27228&disable_features=boundaries&background=MAPNIK&id=w177517745

horsewalkerdetails1

#### Horse walker without building tag -> ICON MISSING

https://pr-1857--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=20.00/54.11709/-6.28268&disable_features=boundaries&background=MAPNIK&id=w1167347375

hw-noicon

**WHY  IS  THERE  NO HORSE   WALKER   ICON  ? ? ?**

### Search



s1

search-horse-exercise

search-horse-training

search-horse






### Info-`i`



horsewalker-info

data/presets/animal/horse_walker.json
```diff
@@ -0,0 +1,28 @@
+{
+    "icon": "temaki-amusement_park",
+    "fields": [
+        "building_area",
+        "covered",
+        "access_simple"
```
tordans: Not sure about this one. In my mind, those features are always `private` or `customers` which does not really matter given that they are never(?) `public`/`permissive`. But if we promote this as a `fields` we suggest that we expect this to be filled in most of the time which I would not…

Should we rather move it into `moreFields` so only user that want to mark it with "open access" (for example) find it?

data/presets/animal/horse_walker.json
```diff
@@ -0,0 +1,28 @@
...
+    ],
+    "moreFields": [
+        "operator",
+        "lit",
+        "indoor",
```
tordans, line 11: I would use this in a case when this is not a building but indoor and mapped as separate point inside the building, yes?

data/presets/animal/horse_walker.json
```diff
@@ -0,0 +1,28 @@
+{
+    "icon": "temaki-amusement_park",
```
tordans, line 2: :-D I was wondering where that nice horse+ambrella icon came from

data/presets/animal/horse_walker.json
```diff
@@ -0,0 +1,28 @@
...
+        "lit",
+        "indoor",
+        "capacity"
+    ],
+    "geometry": [
+        "area",
+        "point"
+    ],
+    "tags": {
+        "animal": "horse_walker"
```
1ec5, line 19: > I am not particularly happy with the key animal of this tag as a horse walker is not an animal. But we already have about 3.4k uses of this tag, so it is not so easy to change anymore. The tag was [proposed in 2013](https://wiki.openstreetmap.org/w/index.php?title=Proposal:Animal%3Dhorse_walker&oldid=857101).

I was unaware that `animal=*` is being used thematically. I’ve [added](https://wiki.openstreetmap.org/wiki/Special:Diff/2932357) it to our [tagging hall of fame](https://wiki.openstreetmap.org/wiki/Homonymous_keys#Containment_versus_refinement). 😄

data/presets/animal/horse_walker.json
Hufkratzer: Good idea. I was just unsure where to put it. I moved it now, see:
access moved

data/presets/animal/horse_walker.json
Hufkratzer: There has just been a discussion (https://community.openstreetmap.org/t/should-amenity-be-tagged-on-a-building-way/138827) in the forum about whether amenity (horse_walker is also something like this) and building tags should always be used on separate objects. There are different opinions. It only seems to be necessary if otherwise it leads to confusion. In any case, it is very unusual for horse walkers so far. The only example I have found is: https://www.openstreetmap.org/node/8142210585. In fact, only 14 of a total of 3428 horse walkers are mapped as nodes. In contrast, 1805 horse walkers also have a building tag, which is 52%.

Only 3 of the mapped horse walkers have an operator tag (“Gestüt im Engels,” “Ziegelhof,” “Reitanlage Obernbergen”). I assume that this is both the operator of the building and the horse walker.

“lit” and “indoor” have not actually been used for horse walkers so far (only 2 x ‘lit’, but that was probably a mistagging, see [changeset discussion](https://www.openstreetmap.org/changeset/140287878). I had included these fields because they are also in the [preset for pitch](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/leisure/pitch.json). Shall I remove them here?

matkoniecz: Hmm.  On one hand they seem to make sense in this context,  on the other it starts to look like pushing new tagging by preset.

I am unsure

matkoniecz: lit and indoor in moreFields should be harmless as it is not presenting them when unfilled so it should not unduly push tagging, I think

Issue #1858, opened 2025-12-14
Don't suggest to add `education` tag to existing `amenity=school` features


tyrasd:
https://community.openstreetmap.org/t/why-is-it-ok-that-id-performs-undiscussed-semiautomatic-distributed-mass-edits/139011 discussed the potentially problematic inclusion of `education=school` in the `addTags` section of the School (and related) presets. Maybe it would be better to omit this for the time being until the tagging is more well established. Since https://github.com/openstreetmap/id-tagging-schema/commit/2fcb659c8511ddf8c652b25fa336ed50b3d2c94e we do have placeholder presets for `amenity=school` without `education` tag, so implementing this should be trivial.

PR #1863, opened 2025-12-20
(#1858) Remove auto-added education=school from school presets

Mayankaggarwal8055:
This PR removes the automatic addition of `education=school` from school-related presets #1858.

The `education=*` tagging scheme is still not well established, and automatically suggesting it can lead to unintended semi-automatic mass edits. The tag remains available for manual use.

### DIFF
--- a/data/presets/education/school.json
+++ b/data/presets/education/school.json
@@ -40,8 +40,7 @@
         "amenity": "school"
     },
     "addTags": {
-        "amenity": "school",
-        "education": "school"
+        "amenity": "school"
     },
     "reference": {
         "key": "amenity",


matkoniecz (changes requested):
https://github.com/openstreetmap/id-tagging-schema/blob/51443d45f6505e775ae12cc2776d2ef975e5bfe7/data/presets/education/school.json#L44

it seems to not work yet, you also need to change here (probably)

screen-2026-01-06-10-42-48

Mayankaggarwal8055:
Thanks for the review and for pointing this out.

I’ve removed the auto-added education=school tag from data/presets/education/school.json, so it no longer gets applied indirectly via the education preset. The school-related presets now only add amenity=school, and education=* remains available for mappers to add manually if they want to use it.

I also noticed that education=school still appears in presets.json, but since that file is generated, I didn’t modify it directly. If there’s a recommended way for contributors to regenerate or update that file as part of this change, I’m happy to follow your guidance.

PR #1864, opened 2025-12-23
Add Goods Aerialway Station preset (#1633)

### LINKED ISSUE #1633: Do not suggest `public_transport=station` for goods aerialway
Selecting the preset "Aerialway Station" adds aerialway=station and public_transport=station to the node.  However, the second tag is not appropriate for a goods aerialway because there is no public transportation involved.  The tag can be deleted, but then a "missing tag" warning is displayed.

Since public transportation aerialways are probably more common, at least in most locations anyway, it makes sense to add the tag automatically.  But to reduce the warning noise and avoid inappropriate tagging, I would suggest to not warn about the absence of or suggest adding a public_transport=station tag if the node is part of an aerialway=goods way.

jnakaigen:
Description, Motivation & Context
Currently, the "Aerialway Station" preset automatically adds public_transport=station. While correct for passenger lifts, this triggers a "missing tag" warning when mappers remove it for goods, freight, or agricultural lifts (common in the Alps and industrial areas).

This PR adds a dedicated "Goods Aerialway Station" preset that uses usage=freight and omits the public transport tag, resolving the validation noise reported in #1633.

Related issues
Closes #1633

Links and data
Relevant OSM Wiki links:

[Tag:aerialway=station](https://wiki.openstreetmap.org/wiki/Tag:aerialway=station)

[Key:usage](https://wiki.openstreetmap.org/wiki/Key:usage) (Specific to freight/industrial)

Relevant tag usage stats:

aerialway=station + usage=freight: [739 objects](https://www.google.com/search?q=https://taginfo.openstreetmap.org/tags/aerialway%3Dstation%2520usage%3Dfreight)

### DIFF
--- a/data/presets/aerialway/station_goods.json
+++ b/data/presets/aerialway/station_goods.json
@@ -24,6 +24,5 @@
         "freight",
         "goods",
         "mining"
-    ],
-    "matchScore": 2
+    ]
 }
--- a/data/presets/aerialway/station_goods.json
+++ b/data/presets/aerialway/station_goods.json
@@ -1,0 +1,28 @@
+{
+    "name": "Goods Aerialway Station",
+    "icon": "temaki-goods_lift",
+    "geometry": ["node", "vertex"],
+    "tags": {
+        "aerialway": "station",
+        "usage": "freight"
+    },
+    "addTags": {
+        "aerialway": "station",
+        "usage": "freight",
+        "access": "private",
+        "foot": "no"
+    },
+    "fields": [
+        "name",
+        "operator"
+    ],
+    "terms": [
+        "agriculture",
+        "cargo",
+        "farming",
+        "freight",
+        "goods",
+        "mining"
+    ],
+    "matchScore": 2
+}
--- a/data/presets/aerialway/station_goods.json
+++ b/data/presets/aerialway/station_goods.json
@@ -9,8 +9,7 @@
     "addTags": {
         "aerialway": "station",
         "usage": "freight",
-        "access": "private",
-        "foot": "no"
+        "access": "private"
     },
     "fields": [
         "name",
--- a/data/presets/aerialway/station_goods.json
+++ b/data/presets/aerialway/station_goods.json
@@ -7,9 +7,8 @@
         "usage": "freight"
     },
     "addTags": {
-        "aerialway": "station",
-        "usage": "freight",
-        "access": "private"
+        "access": "private",
+        "foot": "no"
     },
     "fields": [
         "name",
--- a/data/presets/aerialway/station_goods.json
+++ b/data/presets/aerialway/station_goods.json
@@ -1,7 +1,10 @@
 {
     "name": "Goods Aerialway Station",
     "icon": "temaki-goods_lift",
-    "geometry": ["node", "vertex"],
+    "geometry": [
+        "point",
+        "vertex"
+    ],
     "tags": {
         "aerialway": "station",
         "usage": "freight"
--- a/data/presets/aerialway/station_goods.json
+++ b/data/presets/aerialway/station_goods.json
@@ -9,10 +9,6 @@
         "aerialway": "station",
         "usage": "freight"
     },
-    "addTags": {
-        "access": "private",
-        "foot": "no"
-    },
     "fields": [
         "name",
         "operator"


data/presets/aerialway/station_goods.json
```diff
@@ -0,0 +1,28 @@
...
+    "geometry": ["node", "vertex"],
+    "tags": {
+        "aerialway": "station",
+        "usage": "freight"
+    },
+    "addTags": {
+        "aerialway": "station",
+        "usage": "freight",
+        "access": "private",
+        "foot": "no"
```
tordans:
```suggestion
        "access": "private"
```

The `foot` tag feels weird here, there `private` access should be enough.

jnakaigen: Thanks for the feedback, @tordans! That makes sense, I've applied the suggestion to remove the foot tag as access=private covers the intent. I'm also checking the build failure now to ensure everything is valid.

jnakaigen:
Hi @tordans, thanks for the guidance! I’ve updated the preset and added the requested documentation below.

Technical Changes
Strong Matching: I moved usage=freight into the tags object. This ensures the preset only matches existing nodes if they are explicitly marked as freight, preventing any "collision" with the standard passenger station preset.

Implicit Tags: I removed redundant tags from addTags. Since usage=freight is now in the tags block, iD will automatically apply it upon selection.

Safety Tags: I’ve included foot=no in addTags to ensure these stations aren't used for pedestrian routing.

Research & Wiki Verification
As requested, here are the links and community standards supporting this:

[Key:usage](https://wiki.openstreetmap.org/wiki/Key:usage): The Wiki states: "usage=freight is used for infrastructure used for transporting goods."

[Tag:aerialway=station](https://wiki.openstreetmap.org/wiki/Tag:aerialway%3Dstation): This is the correct node tag for all aerialway stops.

[Public Transport Schema](https://wiki.openstreetmap.org/wiki/Public_transport): Public transport tags are reserved for "transport services available to the general public." Goods/agricultural lifts (common in the Alps) do not meet this criteria, justifying the omission of public_transport=station.

Let me know if there’s anything else needed to get this merged!

data/presets/aerialway/station_goods.json
```diff
@@ -0,0 +1,29 @@
...
+    ],
+    "terms": [
+        "agriculture",
+        "cargo",
+        "farming",
+        "freight",
+        "goods",
+        "mining"
+    ],
+    "matchScore": 2
```
matkoniecz: is `matchScore` really needed?

jnakaigen: I added matchScore: 2 to ensure that when a node has aerialway=station and usage=freight, this specific preset would definitely take precedence over the generic 'Aerialway Station'.

However, since I've now moved usage=freight into the tags block (making it a more specific match than the generic preset), I agree that the higher score is likely redundant. I'll test it in the preview link provided by the bot and, if it matches correctly without the boost, I will remove it in the next commit!

matkoniecz:
I am trying to help with PRs/issues in this repo

Sadly there is quite large backlog and there is limit how much I can review as a hobby/volunteering :(

I will try to look at it, but I cannot promise any firm deadline.

data/presets/aerialway/station_goods.json
matkoniecz: 1) this suggestion is not applied

2) what about cases where it is in fact not with private access? Will this preset still work?
Will it try to force `access=private` ?

matkoniecz (changes requested):

data/presets/aerialway/station_goods.json
matkoniecz: @jnakaigen

ping?

matkoniecz:
@jnakaigen 

jnakaigensee https://github.com/openstreetmap/id-tagging-schema/pull/1864#pullrequestreview-3607134313 - review waits for processing since December 2025

Is this PR abandoned?

data/presets/aerialway/station_goods.json
```diff
@@ -0,0 +1,28 @@
...
+        "vertex"
+    ],
+    "tags": {
+        "aerialway": "station",
+        "usage": "freight"
+    },
+    "addTags": {
+        "access": "private",
+        "foot": "no"
+    },
```
matkoniecz:
```suggestion
```

in theory there may be public-use goods aerialway

matkoniecz:
taking over this PR as it was abandoned, so I cannot merge it myself

data/presets/aerialway/station_goods.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "Goods Aerialway Station",
```
matkoniecz, line 2:
```suggestion
    "name": "Cargo Aerialway Station",
```

matkoniecz: I think this is more legible label

@1ec5 what you think?

matkoniecz:
I see a major blocker, details at https://github.com/openstreetmap/id-tagging-schema/issues/1633#issuecomment-4824239998

data/presets/aerialway/station_goods.json
matkoniecz: Though "aerialway" is still osmism.

matkoniecz: "Cargo Lift Loading Area"? "cargo cablecar station"?

PR #1870, opened 2025-12-27
Add translatable labels for address components (#1788)

### LINKED ISSUE #1788: housenumber lacks a translatable placeholder string
I use id-tagging-schema to provide presets in the mobile app Go Map!!.

Most fields in the schema include a translatable name, such as "Building", "Levels", etc. which gets localized by all the hardworking translators, and I display these localizations in the app. For the "Address" field there aren't regular localizable names for things like "addr:street", but the placeholder text gets localized and works as a proxy for it. Unfortunately for several of the address fields (housenumber, conscriptionnumber) the placeholder text is just "123" which doesn't work as a name.

Ideally there would be some way to have a translatable name associated with those fields as well.

Maybe something similar to how there is a ""housenumber!jp" entry that has alternate placeholder text?

Image

bryceco:

See #1788: This may or may not be what @k-yle was suggesting there.

There are currently placeholder strings that provide translations of address components like "addr:city", but "addr:housenumber" uses placeholder text "123". This adds a new section to provide translations for such strings.

### Related issues

Closes #1788

### DIFF
--- a/data/fields/address.json
+++ b/data/fields/address.json
@@ -29,6 +29,10 @@
     ],
     "label": "Address",
     "strings": {
+        "labels": {
+            "conscriptionnumber": "Conscription Number",
+            "housenumber": "House Number"
+        },
         "placeholders": {
             "block_number": "Block Number",
             "block_number!jp": "Block No.",


bryceco:
Using "labels" is also the thing I figured would be most controversial. Opinions welcome!

bryceco:
Wow, that's actually a really nice feature!

But:
* iD doesn't make use of the changes added by this PR. This change is only useful for 3rd party users of id-tagging-schema.
* The changes are only visible after translators have translated the associated strings.

Based on Kyle's feedback I'm fairly confident about the PR. It seems unlikely for it to break things (since it adds data iD doesn't reference), and if there's a problem with it we can iterate without breaking things.

Edit: I'll add that I don't see any breakage or change using the link you provided.

tordans:
(FYI, will comment to ask some questions later; happy to merge afterwards.)

PR #1874, opened 2026-01-01
Expand translations for `tunnel` values

tordans:
I noticed that when picking a tunnel value in the dropdown we don't have a filter that shows only those values that apply to the elected feature – highway vs. waterway.

A quick way around this is, to add the "category" of tunnel to the translation string, like done here.

Wiki: https://wiki.openstreetmap.org/wiki/Key:tunnel

Example current UX:
image

### DIFF
--- a/data/fields/tunnel_combo.json
+++ b/data/fields/tunnel_combo.json
@@ -4,10 +4,10 @@
     "label": "Type",
     "strings": {
         "options": {
-            "avalanche_protector": "Avalanche Protector",
-            "building_passage": "Building Passage",
-            "culvert": "Culvert",
-            "flooded": "Flooded Tunnel"
+            "avalanche_protector": "Avalanche Protector (Highways)",
+            "building_passage": "Building Passage (Highways)",
+            "culvert": "Culvert (Waterways)",
+            "flooded": "Flooded Tunnel (Waterways)"
         }
     }
 }
--- a/data/fields/tunnel_combo.json
+++ b/data/fields/tunnel_combo.json
@@ -4,8 +4,8 @@
     "label": "Type",
     "strings": {
         "options": {
-            "avalanche_protector": "Avalanche Protector (Highways)",
-            "building_passage": "Building Passage (Highways)",
+            "avalanche_protector": "Avalanche Protector",
+            "building_passage": "Building Passage",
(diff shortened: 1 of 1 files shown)


tordans:
@1ec5 maybe you have an opinion on this (and can review the PR).

---

**Testing**

https://pr-1874--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=21.00/28.46490/-16.25873&disable_features=boundaries&background=fr.ign.bdortho&id=w78870461

image

matkoniecz:
"Building Passage (Highways)" will be quite confusing on `highway=footway`

matkoniecz (changes requested):
quite confusing on footway building passages

tordans:
> > Or just add "waterway" to the one and nothing to the other?
> 
> +1, this seems viable

I changed the PR just now. "Highway" tunnels are now only implicitly "grouped" because `(Waterways)` groups the other tunnels well.

A good minimal improvement, IMO.

matkoniecz (approved):

matkoniecz:
Though note that you can have actual path in actual culverts (especially under motorways and railways)

1ec5:
Do users actually tend to mark a roadway as a tunnel when it passes over a culvert? I’m not sure I see how the extraneous options actually lead to mistagging among English speakers, but maybe there are more opportunities for misunderstanding in translation?

matkoniecz:
I understood https://github.com/openstreetmap/id-tagging-schema/pull/1874#issuecomment-3964318476 as claim was that "culvert" is normal word that is clearly about waterways and therefore extra clarification is not helpful.

matkoniecz:
I quickly added https://wiki.openstreetmap.org/wiki/Tag:tunnel%3Dculvert#Not_only_water - intended as noncontroversial documentation of actual situation

tordans:
TBH, I think we need to move this to the Forum. For me, all those examples are just plain tunnels; just because they have round features, does not make them a pipe or pipe-like IMO. But then again, I don't really know much about this topic nor do I have the energy to dig into it. The fact that the Wiki was clearly focused on Water and we are IMO redefining this, is an indicator that this is not just a tagging schema topic but a question of definition.

matkoniecz:
I will close it per comments above then

Issue #106, opened 2021-01-07
Translate keys for recycling materials accepted


AntMadeira:
Hi there.
I would like to ask for multilingual support for the values in the amenity=recycling + recycling:* tag

BTW, how are the various values fetched? From the wiki, from taginfo or are they just inputted into iD?

Regards.

ireun:
Bump on this! Where are theses strings located on Transifex?

tyrasd:
These strings need to be defined first in the preset as possible "strings", similar to how it was done in the `diet` field:

https://github.com/openstreetmap/id-tagging-schema/blob/v3.1.0/data/fields/diet_multi.json#L17-L27

Afterwards they are automatically added to transifex for translation.

AntMadeira:
And how can that be done? 
Is there a tutorial or help file on how to do that? I wouldn't mind to add those strings if I knew how...


> These strings need to be defined first in the preset as possible "strings", similar to how it was done in the `diet` field:
> 
> https://github.com/openstreetmap/id-tagging-schema/blob/v3.1.0/data/fields/diet_multi.json#L17-L27
> 
> Afterwards they are automatically added to transifex for translation.

tyrasd:
This one is relatively easy: go to the field you want to add the strings for, which would be this one here: https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/recycling_accepts.json

Then click on the edit button (looks like a :pencil2: ), and add the strings similar to how it is done in the [example above](https://github.com/openstreetmap/id-tagging-schema/blob/v3.1.0/data/fields/diet_multi.json#L17-L27). You need to change the individual "options", of course: the so part before the `:` is what comes in the OSM tag key after the field's base part (i.e. after `recycling:`), and the part after the `:` is just a textual representation of the respective tag. For example, one line could read like this: `    "glass_bottles": "Glass Bottles",`. When you're done, you can save ("commit") your changes on the bottom of the page, which will then step-by-step guide you through the creation of a new pull request. That's basically it.[^1]


[^1]: If you are a bit more advanced: one additional step is to run the `npm run build` command locally after the field's strings are added, which would then allow transifex to automatically fetch the new translatable strings, but I would consider that as an optional step.

tyrasd:
this was done in #349, thanks @kjonosm!

AntMadeira:
@kjonosm can you please also add plastic, waste, cooking oil, books and electrical_items, which are very common?

[...]

PR #1876, opened 2026-01-05
add `recycling:drugs=yes`  recycling container type

mnalis:

Allows `recycling:drugs=yes` containers to be:
- easily be added by mappers (e.g. in StreetComplete Things Overlay), and 
- be translatable (so non-English speaking mappers can actually find it)

Just like [recycling:electrical_items=*](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/recycling/container/electrical_items.json), which is often present standalone (e.g. in supermarkets) so has its standalone preset, `recycling:drugs=*` is also mostly preset standalone in pharmacies. Thus, it makes sense to have a specific preset for it, especially for other data consumer then iD (like e.g. StreetComplete / SCEE in particular)

### Related issues

See StreetComplete issue https://github.com/streetcomplete/StreetComplete/issues/4802 for more background story (in short, without specific preset, people can only add unspecific recycling container in StreetComplete, and then leave an OSM Note stating that is a expired drugs container, depending on others to fix its type)

Also, for translation, it needs to be added as a explicit value (currently it seems to be auto-added in iD due to its taginfo-occurencies, but that cannot be translated then (see e.g. https://github.com/openstreetmap/id-tagging-schema/issues/106#issuecomment-756527761 for a similar issue)

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Key:recycling:drugs

**Relevant tag usage stats:**
> https://taghistory.raifer.tech/#***/recycling:drugs/

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/data/presets/amenity/recycling/container/expired_drugs.json
+++ b/data/presets/amenity/recycling/container/expired_drugs.json
@@ -1,0 +1,33 @@
+{
+    "icon": "maki-recycling",
+    "fields": [
+        "{amenity/recycling_container}"
+    ],
+    "moreFields": [
+        "{amenity/recycling_container}"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "drugs",
+        "expired drugs",
+        "medicine",
+        "medicament",
+        "pills",
+        "pharmaceutical",
+        "narcotic",
+        "prescription"
+    ],
+    "tags": {
+        "amenity": "recycling",
+        "recycling_type": "container",
+        "recycling:drugs": "yes"
+    },
+    "reference": {
+        "key": "recycling:drugs",
+        "value": "yes"
+    },
+    "name": "Expired Drugs Container"
+}
--- a/data/fields/recycling_accepts.json
+++ b/data/fields/recycling_accepts.json
@@ -20,7 +20,8 @@
             "scrap_metal": "Scrap Metal",
             "cooking_oil": "Cooking Oil",
             "engine_oil": "Engine Oil",
-            "food_waste": "Food Waste"
+            "food_waste": "Food Waste",
+            "drugs": "Expired Drugs"
         }
     }
 }


matkoniecz:
Are any drugs recycled anywhere at all?

data/presets/amenity/recycling/container/expired_drugs.json
```diff
@@ -0,0 +1,33 @@
...
+        "expired drugs",
+        "medicine",
+        "medicament",
+        "pills",
+        "pharmaceutical",
+        "narcotic",
+        "prescription"
+    ],
+    "tags": {
+        "amenity": "recycling",
```
matkoniecz, line 24: Why it is not `amenity=waste_basket` with `waste=` value?

(BTW, if I would mark it at all - then I would put it as a pharmacy property)

mnalis: > Why it is not `amenity=waste_basket` with `waste=` value?

Because [`recycling:drugs=yes`+`recycling_type=container` ](https://overpass-turbo.eu/s/2isn)is about order of magnitude more popular then [`amenity=waste_basket`+`waste=drugs` ](https://overpass-turbo.eu/s/2ism) :shrug: 

It is also how it is being specified in preset in [beautified-JOSM-preset](https://github.com/simonpoole/beautified-JOSM-preset/blob/5.95.0/master_preset.xml#L2755) (e.g. as used by Vespucci).

But if people feel using `amenity=waste_basket`+`waste=drugs` approach should be used instead, I could change it, sure.

> (BTW, if I would mark it at all - then I would put it as a pharmacy property)

You mean, as in adding `waste=drugs` on existing `amenity=pharmacy` elements? Thanks seems to be [extremely rare](https://overpass-turbo.eu/s/2isd)  tagging (2 uses worldwide so far)

mnalis:
> Are any drugs recycled anywhere at all?

Maybe? People seem to be mapping those separate-collection-containers as `recycling:*` on OSM in any case, and it seems unreasonable to require of them to know exact post-collection processing procedures in their municipality before being able to map those containers.

Note that (according to wikipedia at least) the term ["recycling"](https://en.wikipedia.org/wiki/Recycling) seems to include _"the recovery of energy from waste materials"_, so perhaps they _are_ recyclable in that sense. I really lack the knowledge to authoritatively make claim for or against. And (at last over here in Croatia) they are called _"drug recycling containers"_. 

<details> 

While I personally would like to know more _details_ about what and how _actually_ gets recycled (in my country at least), autoritativne information on that seems to be surprising hard to find.



 FWIW:

-  over here, they are called "_reciklažni spremnik za lijekove"_ (literal translation: _"drug recycling container"_),  just as the ones for e.g. `recycling:batteries=yes` are called _"battery recycling containers"_) 
- they are also being collected at `amenity=recycling` + `recycling_type=centre`
- it seems that `recycling:*` is being used for all those "disposable things" collected  separately from household waste, regardless of what may actually happens to them down the line (e.g. I have serious doubts that those alkaline household batteries are being recycled either here -- or asbestos, or old paints,  or waste oil, or solvents/other hazardous substances, or meat-containing food waste, etc -- even if they are all being tagged with `recycling:*` tags). 


It seems somewhat unreasonable to expect mappers to be intimate with actually implemented waste processing workflows in their municipality to be able to tag those _"recycling/waste/disposables collection containers"_ (e.g. differentiate between tagging as `recycling:asbestos/batteries/drugs=yes` and `waste=asbestos/batteries/drugs`)

</details>

andrewharvey:
I agree with @stragu. We tag based on how it is presented to the consumer, not what actually happens downstream of placing the material in the bin.

I'll add a note to the wiki as well to clarify that this tag is for medications, and not to be confused with medication blister pack recycling, eg https://www.pharmacycle.com.au/.

lucaceresoli:
+1 to add this, either as `waste=drugs` or as `recycling:drugs` or whatever.

I think `waste=drugs` is more correct, but definitely having none is bad.

In Italy drugs disposal boxes are very common, every small village has at least one, while I don't remember having seen any shoe disposal/recycling/whatever anywhere in my life.

PR #1920, opened 2026-01-29
Fix Roentgen icon naming mismatch for radio telescope

### LINKED ISSUE #1919: Wrong file name for telescope_radio.svg
I was updating my code in Go Map!! to use the new Roentgen icon repository and in doing so was getting warnings that the icon for data/presets/man_made/telescope/radio.json is missing. 

In https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/man_made/telescope/radio.json there is
 "icon": "roentgen-telescope-radio",
But the icon has name https://github.com/enzet/Roentgen/blob/main/icons/telescope_radio.svg
("-" instead of "_").

ak8abhinay:
Fixes #1919
This PR fixes a naming mismatch in the radio.json preset. The icon was incorrectly defined with a hyphen (roentgen-telescope-radio), which prevented the icon from loading in editors.

Changes :
Changed the icon ID from roentgen-telescope-radio to roentgen-telescope_radio.

I also verified all other Roentgen icon references in the presets using grep. It appears that roentgen-telescope-radio is the only one using a hyphen. All other multi-word Roentgen icons correctly use underscores.

### DIFF
--- a/data/presets/man_made/telescope/radio.json
+++ b/data/presets/man_made/telescope/radio.json
@@ -1,5 +1,5 @@
 {
-    "icon": "roentgen-telescope-radio",
+    "icon": "roentgen-telescope_radio",
     "geometry": [
         "point",
         "area"


tordans (approved):
Image

https://github.com/openstreetmap/id-tagging-schema/issues/1919#issuecomment-3830977585

Issue #1829, opened 2025-11-26
Replace querying taginfo by prefefined lists wherever possible


matkoniecz:
> IMHO the main thing to fix over anything else is to get rid of as many cases of iD relying on querying taginfo for tag values as possible.
>
> The system made sense for Mapbox as a quick fix to their competive issues at the time, but it has been and continues to be a constant source of problems and likely has more influence on actual tagging than any voting here.

@simonpoole at https://community.openstreetmap.org/t/id-tagging-schema-improvements-which-would-be-especially-worth-doing-which-should-be-avoided/138538/2

that would mean using https://github.com/ideditor/schema-builder?tab=readme-ov-file#strings

FloEdelmann:
Not sure where that querying is actually happening – but I guess it's not in the presets but somewhere in the iD code?

dieterdreist:
for the building tag we should not normalize to few values I think. Typical data consumers would just treat most values as yes and maybe special case a few hand selected values (e.g. roof, greenhouse, church, …) but if they are particularly interested in building types they will probably want to know about subtle differences, so going with taginfo seems a good approach for this tag.

matkoniecz:
Decrease the number of dropdowns #2049 is de facto (in addition to other effects) special case of doing this and blocking access to taginfo suggestions in the main GUI

PR #1940, opened 2026-02-12
Omit religion=none from suggested options of religion field for places of worship

### LINKED ISSUE #1939: don't suggest religious=none ?
> i know this is a 3 year old PR, but I'm a bit confused about `landuse=religious` + [`religious=none`](https://osm.wiki/Tag:religion=none):
>
> image
>
> I originally thought it was for religions that [don't perfectly meet the definition](https://en.wikipedia.org/wiki/Nontheistic_religion) of "religion", or for shared meditation spaces. But actually, the wiki [suggests](https://osm.wiki/Tag:religion=none) that [`religious=none`](https://osm.wiki/Tag:religion=none) is different to all the other values. 
> 
> So, should we be promoting it as a value here? (`landuse=religious` | `amenity=place_of_worship`) + [`religious=none`](https://osm.wiki/Tag:religion=none) is only used 167 times. The other 20,000 uses of [`religious=none`](https://osm.wiki/Tag:religion=none) are just to indicate that schools etc. are not religious

_Originally posted by @k-yle in https://github.com/openstreetmap/id-tagging-schema/issues/263#issuecomment-3713988747_

ak8abhinay:
This update follows the suggestion to stop promoting religion=none as a suggested value in the religion field.

Problem: Offering 'Non-religious' as a clickable option for 'Religious Area' or 'Place of Worship' presets created a logical contradiction.

Fix: By removing none from the options list in religion.json, the editor no longer 'offers' this value as a button.

Existing data (like the 20,000 schools using this tag) remains valid and will still display the translated 'Non-religious' label, and mappers can still manually enter the tag if needed."

Fixes #1939


ak8abhinay:
@matkoniecz,

Yes, I see now that deleting that will break the label for existing data. I will update the PR to restore the 'Non-religious' string while adding an explicit options array that excludes none. This keeps the label perfect for 20,000 schools but stops the button from being suggested for religious presets. Future mappers can still manually type the tag if needed, and the editor will recognize it through the restored string.

data/fields/religion.json
```diff
@@ -28,4 +51,4 @@
             "voodoo": "Voodoo"
         }
     }
-}
+}
```
matkoniecz:
```suggestion
}

```
restore trailing newline

data/fields/religion.json
```diff
@@ -16,7 +39,7 @@
             "jewish": "Jewish",
             "multifaith": "Multifaith",
             "muslim": "Muslim",
-            "none": "Nonreligious",
+            "none": "Non-religious",
```
matkoniecz: is adding hyphen deliberate?

ak8abhinay: I reverted the spelling to match the original

data/fields/religion.json
ak8abhinay: added the trailing newline

tyrasd (dismissed):
I think it is not ideal to not be able to select the `none` options for features like a school that also use the `religion` field, but would most often be tagged as nonreligious.

See https://github.com/openstreetmap/id-tagging-schema/issues/1939#issuecomment-3921386262 for an idea to better approach this.

matkoniecz (approved):
I guess it works, though you get anyway 'no' coming from taginfo

not sure is this list good enough to disable taginfo suggestions

matkoniecz:
https://github.com/openstreetmap/iD/commit/b71105a2b16ac3110fea0638792fda6f45e2817b should fix iD bug, so it has more noticeable effect once released

note: if our religion list is really complete, maybe we can and should disable showing values coming from taginfo, see https://github.com/ideditor/schema-builder?tab=readme-ov-file#autosuggestions and #1829

PR #1941, opened 2026-02-14
Add recycle to terms for recycling centre and container

simonpoole:

Currently searching for "recycle" doesn't return "recycling" related items as the top results, this is likely just an edge case and at least for vespucci tweaking the search algo will as a tendency make overall behaviour worse (see the linked issue).

### Related issues

Resolves https://github.com/simonpoole/beautified-JOSM-preset/issues/514

### DIFF
--- a/data/presets/amenity/recycling_centre.json
+++ b/data/presets/amenity/recycling_centre.json
@@ -26,6 +26,7 @@
         "dump",
         "garbage",
         "glass",
+        "recycle",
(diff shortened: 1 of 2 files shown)


k-yle (approved):
lgtm in principle, because other presets already define equivalent terms, like `charger/charging`, `change/changing`, `cleaner/cleaning`, `gamble/gambling`

PR #1948, opened 2026-02-20
Rename High Jump Pit to High Jump Area for mapping accuracy

### LINKED ISSUE #1947: "High Jump Pit" is a confusing name
On the wiki page for [athletics=high_jump](https://wiki.openstreetmap.org/wiki/Tag:athletics%3Dhigh_jump) the example image shows the entire high jump area mapped with this tag, not just the mat that you land on. It makes sense to map the entire area instead of just the mat because at many tracks, the mat is only there during track meets and is usually in storage somewhere else.
The preset however names athletics=high_jump as "High Jump Pit" which to me makes it seem like this is supposed to be used only on the area where the mat is. Because similarly named presets, for example "Pole Vault Pit" and "Long Jump Pit" are only used for the areas that you actually land on, because both of these also have presets for their runways. High Jump differs from these events because there isn't a strict runway like for pole vault and long jump, which is why there is no high jump runway preset.

The preset is here:
https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/leisure/pitch/athletics/high_jump.json

ak8abhinay:
Issue :
In the attached image, the current preset is labeled "High Jump Pit." This causes mappers to only trace the landing mat rather than the complete area.

image

Fix : Rename the preset to "High Jump Area".

Matches Wiki: The athletics=high_jump [Wiki page](https://wiki.openstreetmap.org/wiki/Tag:athletics%3Dhigh_jump) defines this as a field event facility. The image there show the entire paved semi circular path, not just the mat.

A "Pit" is portable equipment in this case, the "Area" is the permanent infrastructure.

Fixes #1947

### DIFF
--- a/data/presets/leisure/pitch/athletics/high_jump.json
+++ b/data/presets/leisure/pitch/athletics/high_jump.json
@@ -19,5 +19,5 @@
         "key": "athletics",
         "value": "high_jump"
     },
-    "name": "High Jump Pit"
+    "name": "High Jump Area"
 }


matkoniecz:
Makes sense to me. 

Is it looking good for basics speaker?

@ZeLonewolf @1ec5

PR #1950, opened 2026-02-21
Adding support for common waterway features (flowline, link) and paddling related waterway fields (open_water, rapids)

scarapella:
In order to facilitate waterway and canoe related mapping I propose to add two well established and commonly used waterway types:
* https://wiki.openstreetmap.org/wiki/Tag:waterway%3Dflowline  (REMOVED FROM THIS PR)
* https://wiki.openstreetmap.org/wiki/Tag:waterway%3Dlink

Furthermore, I propose including these safety related fields with hints
* https://wiki.openstreetmap.org/wiki/Key:rapids
* https://wiki.openstreetmap.org/wiki/Key:open_water
and updates to the hints for the access tag
* wiki.openstreetmap.org/wiki/Key:canoe

I've aligned the waterway=river, waterway=canal, waterway=stream, and waterway=tidal_channel to use the new fields.


matkoniecz:
I remember flowlines as controversial, see say https://community.openstreetmap.org/t/rfc-feature-proposal-flowlines/117361/38

Maybe we should wait for https://wiki.openstreetmap.org/wiki/Proposal:Flowlines to progress?

data/presets/waterway/tidal_channel.json
```diff
@@ -6,7 +6,10 @@
         "width",
         "salt",
         "tidal",
-        "intermittent"
+        "intermittent",
+        "open_water",
```
matkoniecz, line 10: why `open_water` is not in `moreFields` unlike in river preset?

data/presets/waterway/tidal_channel.json
```diff
@@ -6,7 +6,10 @@
...
+        "rapids",
```
matkoniecz, line 11: are rapids really likely or possible at all in tidal channels?

data/presets/waterway/tidal_channel.json
```diff
@@ -6,7 +6,10 @@
...
+        "canoe"
```
matkoniecz, line 12: I have feeling that these may be better as "moreFields" (shown when tagged, some help in adding it) rather than showing it prominently also when empty

(these applies to all cases where this PR adds fields to be always-displayed)

data/presets/waterway/tidal_channel.json
matkoniecz: from my (limited) understanding it sounds inherently contradictory

data/presets/waterway/link.json
```diff
@@ -0,0 +1,46 @@
...
+        "kayak",
+        "launch",
+        "paddleboard",
+        "portage",
+        "put-in",
+        "rowboat",
+        "take-out"
+    ],
+    "tags": {
+        "waterway": "link"
```
matkoniecz, line 43: I plan to check is it similarly controversial as `waterway=flowline`

hopefully not?

from chart at https://taginfo.openstreetmap.org/tags/waterway%3Dlink looks like a novel tagging idea, with relatively low use and without going through approval process (I consider neither to be blocker, but each requires more thorough check for controversies before getting iD-presets approval - especially if multiple are present)

data/presets/waterway/tidal_channel.json
scarapella: My rationale is that tidal_channels, by definition, tend to be closer to the ocean and therefore likely to be subject to open_water conditions on average compared to rivers.  However, I went back and forth on this. I could equally see that it should be consistent with rivers just for the sake of interface design.  

So I'm happy to take your guidance on this if you've an opinion one way or another.

matkoniecz: In general it is quite easy to throw in a new property into `moreFields` - as it will surface already mapped data and make a bit easier to edit for people who know it.

In opposition adding it to "fields" is a very big commitment, prominently changing interface, strongly encouraging all mappers to fill this info and risking ending with long overwhelming list.

In case of any doubt I would put things into "moreFields". Note that it can be easily changed by followup PR.

data/presets/waterway/tidal_channel.json
scarapella: I included the tag because rapids caused by the bi-directional movement of water caused by tides are unusual, but do exist. This can be in a minor way like the creek near my house which generates a very short rapid when the tide is going out due to the constriction of a causeway, or in a large named phenomenon such as:
* Tidal Rapids/Tidal Races: https://en.wikipedia.org/wiki/Tidal_race
* Reversing Falls including
** Saint John Reversing Falls: https://en.wikipedia.org/wiki/Reversing_Falls
** Pembrook Reversing Falls: https://en.wikipedia.org/wiki/Reversing_Falls_Park
** Sheepscot Reversing Falls: https://www.americanwhitewater.org/content/River/view/river-detail/3634/main

It is true however that these tidal features are quite rare and are almost non-existant in the current osm database. So, while I have a rationale, it's not one I feel super strongly about.

matkoniecz: then maybe move it into `moreFields`?

(I am quite happy that I qualified my surprise, I learned one thing from OSM it is to not declare things "impossible" too soon)

data/presets/waterway/tidal_channel.json
scarapella: My rationale on open_water and rapids are that they are at least as important as tidal from a boating routing safety perspective.  You wouldn't want to send someone down a river if it has rapids on it and not tell them, so incentivizing people to map it would make sense.

For canoe, my rationale is that on highways, access, foot, bicycle, horse, and motor_vehicle are always displayed by default as they are the most important access tags. For waterways, canoe is an important access tag so it should be a default. There are other tags we could imagine being default as well (e.g. boat).  But I was trying to make this PR (relatively) narrow to not try to boil the ocean.

As usual your additional thoughts are very welcome. :-)

matkoniecz: yes, but rapids on tidal channels are extremely rare - right?

note that showing it by default will result in all iD users being encouraged to fill it, and for tidal channels I bet that error rate will swamp actual rapids

data/presets/waterway/tidal_channel.json
scarapella: Yeah, moreFields works for me I'll make the update.

data/presets/waterway/tidal_channel.json
scarapella: Upon reflecting on your wise consul and the clear infrequency of the the phenomenon for sure moreFields is more than sufficient :-)  I will take care. 

Off topic, but an interesting (depending on your inclination) related human use of this phenomenon of bi-directional tidal flow at narrow openings are [tide mills](https://www.tidemillinstitute.org/definition-of-a-tide-mill/)

data/presets/waterway/tidal_channel.json
scarapella: Agreed, no need to encourage tagging the very rare. Might make ghosts appear.

data/presets/waterway/link.json
```diff
@@ -0,0 +1,46 @@
...
+    },
+    "name": "Link"
```
matkoniecz, line 45: this seems quite confusing name - though I have no idea for a better one right now

PR #1955, opened 2026-02-25
Stop deprecating landuse=basin (#1804)

### LINKED ISSUE #1804: Stop considering `landuse=basin` as a deprecated tag
`landuse=basin` is included in the list of deprecated tags (which triggers iD to suggest to update it to `natural=water + water=basin`), but the [OSM wiki](https://wiki.openstreetmap.org/wiki/Tag:landuse=basin) does not document this tag as deprecated. The deprecation was originally added in 
[`17087f42`](https://github.com/openstreetmap/iD/commit/17087f42c8eef77ab928de08684ffee39636bbbd#diff-e792bc72773891751cfd04c91afc7de0a55ed64ff1c6ce2d6faed5a12a5ea512R16) (for https://github.com/openstreetmap/iD/issues/6045, maybe https://github.com/openstreetmap/iD/issues/5591 is also related), without giving a reason why that particular tag should be considered deprecated.[^1] I assume it was perhaps added to the list of deprecated tags by mistake. //cc @quincylvania 

[^1]: I also checked the old version of the the wiki [at the time](https://wiki.openstreetmap.org/w/index.php?title=Tag:landuse%3Dbasin&oldid=1805061), but it did not document anything special about that tag even back then.

ashree2118:
This PR removes `landuse=basin` from the deprecated tags list (`data/deprecated.json`). As discussed by the maintainers in the issue thread, this tag was never officially deprecated by the OSM community, and the automatic upgrade prompt in iD was resulting in undiscussed, distributed bot-like mass edits.

To prevent confusion while maintaining a good user experience, I have also created an unsearchable preset for it (`data/presets/landuse/_basin.json`). This ensures the editor still properly recognizes and formats existing objects mapped with `landuse=basin` without promoting it in the search interface for new edits.

### Related issues
Closes #1804

### Links and data

**Relevant OSM Wiki links:**
- [Tag:landuse=basin](https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dbasin) (The legacy tag)
- [Tag:water=basin](https://wiki.openstreetmap.org/wiki/Tag:water%3Dbasin) (The modern preferred tagging)

**Relevant tag usage stats:**
> With over 80,000 active uses on Taginfo, this tag is still heavily present in the database, reinforcing that a mass deprecation/upgrade path should require a formal community proposal.
- [Taginfo: landuse=basin](https://taginfo.openstreetmap.org/tags/landuse=basin)

### DIFF
--- a/data/presets/landuse/_basin.json
+++ b/data/presets/landuse/_basin.json
@@ -1,10 +1,16 @@
 {
     "icon": "maki-water",
-    "name": "Basin",
+    "fields": [
+        "name",
+        "basin",
+        "intermittent_yes"
+    ],
     "geometry": [
         "area"
     ],
     "tags": {
         "landuse": "basin"
-    }
+    },
+    "name": "Basin",
+    "searchable": false
 }


matkoniecz (approved):
there is question whether `landuse=basin` should get blue coloring (as it effectively represents `natural=water` area, its key is misleading)

but this seems to be defined in iD

PR #1986, opened 2026-03-05
Add maxweightrating in France (in future may be tweaked to allow maxweightrating= outside France)

### LINKED ISSUE #1887: FR: incoming deprecation of `maxweight=*` in favor of `maxweightrating=*`; please add support for `maxweightrating=*`
### OSM Tag(s)

maxweightrating=*

As a Field

### Label

Gross combined weight rating

### Aliases

GCWR

### Terms

_No response_

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Key:maxweightrating

### Status of the Tag

In Use

### Usage of the tag

7,000 (many more with subtags)

### Suggested Icon

_No response_

### Replaces other Tag?

maxweight, but just for France

### Regional Tag?

global (used in other countries, although less often as in France)

### Further Information

Hello, there.

It has been recently noted by the French community that the maximum weight road signs from France did not mean limit on actual vehicle weight (`maxweight=*`) but max allowed weight (`maxweightrating=*`). After discussion, a progressive, nation-wide migration to `maxweightrating=*` [​has been decided by the community](https://forum.openstreetmap.fr/t/proposition-d-edition-automatisee-migration-de-maxweight-en-maxweightrating-pour-la-france/35439/71).

In February 2026, all objects tagged with `maxweight=*` will be nation-wide dual-tagged with `maxweightrating=*`, with a `note=*` warning of the incoming deprecation of `maxweight=*` for France, and that these `maxweight=*` tags should now be ignored and no longer updated. In July 2026, all these objects will have their `maxweight=*` tags removed, along with the `note=*`.

Validator rules for this deprecation will be asked for, or added, in due course. That being said, it seems that iD currently not supports `maxweightrating=*` as it supports `maxweight=*`; could you please add support for `maxweightrating=*`?

ashree2118:

added global support for the `maxweightrating=*` tag and clarified the existing `maxweight=*` label, as discussed in #1887.

1. Created a new global field `maxweightrating`
2. Added `maxweightrating` to the `moreFields` arrays for vehicle-applicable infrastructure (`bridge`, `ferry`, `residential`).
3. Updated the label for `maxweight` to clearly distinguish physical weight limits from legal weight ratings.

### Related issues

Closes #1887

### Links and data

**Relevant OSM Wiki links:**
(https://wiki.openstreetmap.org/wiki/Key:maxweightrating)

**Relevant tag usage stats:**
> …

**This is how it works:**

5. Now your PR is ready to be reviewed.

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>


data/fields/maxweight.json
```diff
@@ -1,6 +1,6 @@
 {
     "key": "maxweight",
     "type": "combo",
-    "label": "Weight Limit",
```
ZeLonewolf, line 4: What's wrong with weight limit?

ashree2118: This change is part of resolving issue itself. We are adding a new global field for `maxweightrating` (Limit on legal permitted weight / GVWR) and it was discussed in the issue thread that we should update the `maxweight` label so that both can be distinguished properly.



open to changing that if you think there is a better way to phrase this distinction in US English.

ZeLonewolf: Why not weight limit (tons) for maxweight?

ZeLonewolf: Yeah, I would use parantheticals here. Max weight (trucks) etc.  Max weight is the clearest label and we make understandability worse by removing it.

ashree2118: makes sense, but both these fields are technically measured in tons/lbs, so it might not help mappers tell them apart.

what do you think about using the parentheticals 'Weight Limit (GVW)'

matkoniecz (commented):
there is https://github.com/openstreetmap/id-tagging-schema/blob/e3bd511615b3a0699a22c29da4848cdf894051f1/data/fields/maxweight_bridge.json to make weight limit field more prominent on bridge sections

do we need adjusting it so France will get `maxweightrating` instead? Or maybe it was adjusted already and I missed it?

ashree2118:
## Test-Documentation

test confirming the new `maxweightrating` field is available in france instead of the `maxweight` label .

**1. Inside france:**
Screenshot 2026-03-07 020950


**2. Outside france:**
Screenshot 2026-03-07 021341




### Wording
- [x] American English

matkoniecz (changes requested):
more annoyances and complexity!

https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/aerialway/zip_line.json#L9 has `maxweight` which probably should get own field as GVW makes no sense whatsoever there

probably the same for 
https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/aeroway/helipad.json which is vehicle but GVWR/GVW is likely not applicable there

Not sure about https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/weighbridge.json - should it get `maxweightrating`?

data/presets/aeroway/helipad.json
```diff
@@ -5,7 +5,7 @@
         "ref",
         "operator",
         "surface",
-        "maxweight",
+        "maxweight_capacity",
```
matkoniecz: `maxweight_capacity` sounds a bit weird to me but I have no idea for the name that would be much better (`generic_maxweight`? `maxweight_generic`? this does not appear to be much better)

ashree2118: if `maxweight_capacity` sounds a bit niche, I suppose we can have it like `maxweight_general`

matkoniecz: lets wait whether anyone else seconds it, I am not a native speaker anyway

1ec5: This name is just for internal use, right? As long as nothing depends on it to match the key, we can technically call it anything. I agree that “capacity” is a bit confusing, since we have a `capacity=*` field on some of the same presets. “General” or “generic” should be fine.

data/fields/maxweightrating.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "maxweightrating",
+    "type": "combo",
+    "label": "Limit on legal permitted weight (GVWR)",
```
1ec5: We normally use Title Case for preset names. This is also a bit long, so the important “(GVWR)” suffix might get truncated if the user has resized the sidebar in iD to be too narrow. “Legal Weight Limit (GVWR)” would be shorter and eliminate the redundant “legal permitted”.

data/fields/maxweight_bridge.json
```diff
@@ -6,5 +6,10 @@
     "prerequisiteTag": {
         "key": "bridge",
         "valueNot": "no"
```
matkoniecz, line 8: hmmm, this gets put into `moreFields` not `fields`

I think it can and should be more prominent for bridges

but this change, if useful, should be done as a separate PR

matkoniecz (approved):
I run out of problems :)

This is significant change so I want give others some time to inspect it.

matkoniecz (changes requested):
I see one problem: when `maxweightrating` is tagged on road outside France it should be shown, but it seems not to be

it sounds like I misread how one of fields works, sorry

or one of `maxweightrating` fields should be added to additional places?

data/fields/maxweightrating.json
Bibi56: I'm a bit unsure about the wording as maxweightrating is the legal weight limit in France, but maxweight is the legal weight limit elsewhere. For the restriction I mean (not for the vehicle).

data/presets/highway/elevator.json
```diff
@@ -6,7 +6,8 @@
         "level_semi",
         "access_simple",
         "wheelchair",
-        "maxweight"
+        "maxweight",
+        "maxweightrating"
```
matkoniecz: we surely want here `maxweight_general`

data/presets/man_made/bridge.json
```diff
@@ -5,13 +5,13 @@
         "bridge_combo",
         "layer",
         "maxweight",
+        "maxweightrating_bridge",
```
matkoniecz: note that this will affect bridge outlines (man_made=bridge), not regular roads

data/fields/maxweight_bridge.json
matkoniecz: probably

data/presets/highway/busway.json
```diff
@@ -20,7 +20,6 @@
         "maxlength",
         "maxspeed/advisory",
         "maxweight_bridge",
-        "maxweightrating",
```
matkoniecz: on busways it made sense

matkoniecz: probably?

data/presets/highway/living_street.json
```diff
@@ -26,7 +26,6 @@
         "maxheight",
         "maxlength",
         "maxweight_bridge",
-        "maxweightrating",
```
matkoniecz: on living_street it makes sense to show it if tagged

data/presets/highway/motorway.json
```diff
@@ -26,7 +26,6 @@
         "maxspeed/advisory",
         "maxspeed/hgv",
         "maxweight_bridge",
-        "maxweightrating",
```
matkoniecz: definitely made sense on motorway etc, unless shown also from a different source

matkoniecz:
What about presets/man_made/bridge.json ?

It has `maxweight` field, and not sure what it means there. Is it supposed to be typical weight limit on bridge? But it can be different on different sections and it sounds like pointless duplicate of tags from roads. And in such case it should also get `maxweightrating`

Is it supposed to be design capacity? Maximum weight capacity? Both of limited verifiability, especially second one.

From looking at https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dbridge it is not listed as expected combination.

I guess it can be ignored here, I will open separate PR for handling it.

I opened #2009 - I suggest to ignore it altogether here.

matkoniecz:
I also created Why highway=footway has maxweight_bridge field? #2010

As either `maxweight` should be removed from there or `maxweighrating` added there, but it is a preexisting issue easy to change afterwards, so lets decide and change separately.

matkoniecz (changes requested):
```
rg GVW
data/fields/maxweight.json
4:    "label": "Weight Limit (GVW)",

data/fields/maxweightrating.json
4:    "label": "Legal Weight Limit (GVWR)"
```

why one has "Weight Limit" and another has "Legal Weight Limit"?

I know that it went through several iteration but probably either should both have Legal prefix or neither.

Or is it deliberate distinction? If deliberate - why?

(I assume that it is result of flip-flaping between million different labels and it got diverged at nth name change)

matkoniecz:
Based on iterations of label I think that dropping Legal in both or adding Legal to both would be fine.

If you cannot decide or have no preferences you can throw a dice :)

matkoniecz:
finally I will have time to finish review, sorry for waiting so long :/

EDIT: https://master.apis.dev.openstreetmap.org/changeset/640759


screen-2026-06-26-18-07-04

data/fields/maxweight_bridge.json
matkoniecz: https://github.com/openstreetmap/id-tagging-schema/issues/2385

matkoniecz: we already are not showing it prominently on bridges so nothing changes

matkoniecz:
Changed labels again based on https://community.openstreetmap.org/t/do-you-know-this-terms/144912/ and other feedback from US speakers.

Also based on me having no clue what this terms mean despite spending hours on this PR quite recently.

matkoniecz:
bridleway, footway, cycleway, path, informal path, pedestrian_line miss new presets while they have `maxweight_bridge`

this seems fine to me?

matkoniecz (approved):
I am also open to disabling `maxweightrating` outside France if it turns out to cause too many mistakes

matkoniecz:
Also, I have mixed feedback from US speakers whether "Current Weight Limit" or "Actual Weight Limit" would be better but "Current Weight Limit" is noticeably preferred

Overall @1ec5 ping

matkoniecz:
> maxweightrating globally

there was another discussion with @k-yle and decision for now is to actually restrict `maxweightrating` to France and start asking communities whether they want it enabled in their area

also, this PR is affected by https://github.com/openstreetmap/iD/issues/12491

matkoniecz:
testing fields: https://pr-1986--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=19.11/50.08588/19.95538&disable_features=boundaries and https://pr-1986--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=19.52/48.84300/2.58805&disable_features=boundaries&id=w4307534389 - both on dev server

data/fields/maxweight.json
```diff
@@ -1,6 +1,14 @@
...
-    "snake_case": false
+    "label": "Current Weight Limit (GVW)",
```
tyrasd: I think this label is not yet usable:
* IMO the abbreviation GVW is not well known, so it should either be spelled out, or removed
* It's not clear what "Current" is referring to: is it the current _weight limit_ (in contrast to a future weight limit), or is it the limit of the _current weight_ (of the vehicle)?
* current is also maybe not the best adjective, perhaps (as it has the temporal connotation, which could be confusing). The wiki uses _actual weight_, which is a bit more precise, I think. Other alternatives: _real weight_, _physical weight_?


```suggestion
    "label": "Limit of Actual Vehicle Weight",
```

or alternatively the way the wiki defines it:

```suggestion
    "label": "Permissible Maximum Actual Weight",
```


or (as a temporary solution for now):
```suggestion
    "label": "Weight Limit",
```

data/fields/maxweightrating.json
```diff
@@ -0,0 +1,11 @@
+{
+    "key": "maxweightrating",
+    "type": "combo",
+    "label": "Rated Weight Limit (GVWR)",
```
tyrasd:
```suggestion
    "label": "Limit of Rated Vehicle Weight",
```

or (as a temporary solution for now):
```suggestion
    "label": "Rated Weight Limit",
```

tyrasd (changes requested):
see below

matkoniecz:
Changed labels again based on further feedback from https://community.openstreetmap.org/t/do-you-know-this-terms/144912/ and other people

tyrasd (approved):
ok for now as an intermediate solution until we come up with a better label

1ec5:
Sorry for the belated reply; I was traveling the past several days.

> Also, I have mixed feedback from US speakers whether "Current Weight Limit" or "Actual Weight Limit" would be better but "Current Weight Limit" is noticeably preferred

I agree with @tyrasd that these changes were both worse than the alternatives. “Current” and “actual” are likely to be translated as the same thing in Spanish and other Romance languages. Even in English, the two words are meaningless in this context. Current what? Actual as opposed to what?

“Limit of Rated Vehicle Weight” is correct, though it sounds very stilted. A more idiomatic phrase would be “Vehicle Weight Rating Limit”, “Gross Vehicle Weight Rating”, or “Maximum Vehicle Weight Rating”.

By the way, this is a very technical distinction that only truckers are supposed to care about. In the few jurisdictions that have a mix of GVW and GVWR restrictions, the signs usually clarify the criterion explicitly via these abbreviations. So any attempt to make the label more self-explanatory via plain English will ironically make the field harder for a layperson to use when mapping from a sign or other source. We could at least add some more mainstream terms for the `maxweightrating=*` field.

matkoniecz:
> Current what? Actual as opposed to what?

my intention was to make it refer to weight of vehicle right now

>  In the few jurisdictions that have a mix of GVW and GVWR restrictions, the signs usually clarify the criterion explicitly via these abbreviations. 

I guess that Polish traffic sign design is being stupid again. Admittedly in way that is far less problematic to general public than  usual

> “Limit of Rated Vehicle Weight” is correct, though it sounds very stilted. A more idiomatic phrase would be “Vehicle Weight Rating Limit”, “Gross Vehicle Weight Rating”, or “Maximum Vehicle Weight Rating”.

I guess I will make a followup PR then.

Issue #1468, opened 2025-03-03
Power outlet (sockets)


2hu4u:
### OSM Tag(s)

socket:*=yes

### How would you like this tag to see supported?

As a Preset

### Label

Power Outlet - ([ISO 3166-2 code])

### Aliases

power point; power supply

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Key:socket:*

### Status of the Tag

In Use

### Usage of the tag

Over 15000

### Suggested Icon

See body text

### Regional Tag?

global but with different presets per region

### Rationale
I propose adding a preset for common types of power outlet sockets. This will encompass the family of keys relating to domestic power outlets: `socket:*`. Currently, there is a large barrier towards the uptake of mapping power outlets, because keys of format `socket:*` have cryptic and difficult-to-remember names such as "typeE", "as3112", "bs1363", "nema_5_15" amongst others. As a result, public power outlets are a useful amenity that are largely unmapped compared to similar amenities such as drinking water and EV charging stations because the tagging is difficult. I find that without this preset, I have to dig through a large list of international sockets (that include all the electric vehicle charging sockets) on the OSM Wiki every time.

### Tag usage and competing tags
There are a few similar (competing?) tags for domestic power outlets - the main two are `amenity=power_supply` (about 600 uses) and `power=outlet` (about 300 uses), with a more general `power_supply=yes` having about 36000 uses. Luckily, there is general concensus amongst all of these that the `socket:*` key group be used to specify the socket type. Therefore I don't think this will be problematic or favour any of the competing tags. Overall, the `socket:*` key group has about 15000 uses pertaining to domestic power outlets and close to 100K for electric vehicle charger sockets.

### Regional tagging example
I suggest regional presets use the ISO 3166-2 country code. In Australia, there should be a regional preset such as "Power Outlet (AU)" for a domestic single phase outlet specified by the key `socket:as3112`. This could be expanded as necessary per country, such as `socket:nema_5_15` for North America, `socket:bs1363` for UK/Hong Kong etc. 
In some cases, there may be an identical tagging structure for a preset called "Power Outlet (NZ)" which also uses `socket:as3112`. 
In some cases, there will be multiple outlet styles used in a country; for example "Power Outlet (AU - 3 phase)" may be a preset for `socket:as3123_5pin` in Australia. In China, there are 3 commonly used outlet types that would need consideration.

### Icon example
[...]

PR #1988, opened 2026-03-05
introduce type=measurement and migrate 48 fields to use it

### LINKED ISSUE #2889: Don't show auto-suggestions for value-fields with unit examples in placeholder
The examples at the top all have mm, cm, in, attached,
20210314T142622
But the choices below all lack them!

Seen on
`man_made=pipeline
`

k-yle:
Closes #2530, Closes #2889

This PR implements a new `measurement` field to replace `roadheight` and `roadspeed`, and support any unit of measurement. 

For prior discussion, see the 57 comments on https://github.com/ideditor/schema-builder/pull/198

See SCHEMA.md for explanation of how the schema works.

---

This code implementation is a bit complicated because:
- we want the json schema to validate that `dimension`, `usage`, and the keys of `units` are valid, using [`cldr-core`](https://npm.im/cldr-core)
- when fetching translations from transifex, we need to merge the translations with CLDR's translations, for the units that we use, using [`cldr-units-full`](https://npm.im/cldr-units-full).

The generated `dist/translations/*.json` files now contain a new property called `units`:

```diff
  // es.min.json
  {
    "presets": { … },
+   "units": {
+     "speed": {
+       "kilometer-per-hour": { "long": "kilómetros por hora",  "narrow": "km/h" },
+       "mile-per-hour": { "long": "millas por hora", "narrow": "mi/h" },
+       "knot": { "long": "nudos", "narrow": "kn" }
+     }
+   }
  }
```

The PR does not implement any logic to extract the default values per region, because [it's not yet clear where this data should be stored](https://github.com/openstreetmap/id-tagging-schema/issues/1468#issuecomment-2851009726).

---

My observations:
- [`charge`](https://osm.wiki/Key:charge) can't be supported, because it allows `{currency}` and `{currency}/{time}`
- [`fire_hydrant:diameter`](https://osm.wiki/Key:fire_hydrant:diameter) can't be supported, because it allows unitless values, as well as `m` and `in`
- [`incline`](https://osm.wiki/Key:incline), can't be supported, because it uses `%` which is dimensionless / unitless, it's not `dimension=angle`. Also, this tag frequently uses non-numeric values such as `up`/`down`/`yes`/`steep`
- [`railway:position`](https://osm.wiki/Key:railway:position) and [`railway:position:exact`](https://osm.wiki/Key:railway:position:exact) can't be supported, because they use `mi:12.3` instead of `12.3 mi` 😭  and that tag uses a "fake" unit called `pkm` in argentina.<sup>[[1]](https://github.com/openstreetmap/id-tagging-schema/pull/1988#discussion_r3044714544)</sup>


data/fields/ele.json
```diff
@@ -1,7 +1,22 @@
...
-    "label": "Elevation (Meters)",
+    "type": "measurement",
+    "measurement": {
+        "dimension": "length",
+        "usage": "default",
+        "units": {
+            "meter": [
+                ""
+            ],
+            "inch": [
```
matkoniecz: is specifying `ele` in inches making any sense at all?

data/fields/ele.json
```diff
@@ -1,7 +1,22 @@
...
+                "in"
+            ],
+            "foot": [
+                "'"
+            ]
+        }
+    },
+    "label": "Elevation",
```
matkoniecz, line 12: maybe keep `ele` in meters here to not mix different things?

Or mention it also in PR title that we are also doing `ele` demeterification?

data/fields/maxheight.json
```diff
@@ -1,6 +1,30 @@
...
+        "dimension": "length",
+        "usage": "vehicle",
+        "units": {
+            "meter": [
+                ""
+            ],
+            "centimeter": [
+                "cm"
+            ],
+            "millimeter": [
```
matkoniecz: mm? in which case `maxheight` will be in mm?

data/fields/maxspeed.json
```diff
@@ -1,6 +1,22 @@
...
+    "type": "measurement",
+    "measurement": {
+        "dimension": "speed",
+        "usage": "default",
+        "units": {
+            "kilometer-per-hour": [
+                ""
+            ],
+            "mile-per-hour": [
+                "mph"
```
matkoniecz: maybe it would make sense to contain it to areas where mph  are a thing?

data/fields/maxheight.json
k-yle: good point, removed `mm`, it's only used 3 times for `maxheight`. 

`cm` could probably be removed too, it's only used 30 times. [example](https://osm.org/way/35458500)

data/fields/ele.json
k-yle: oh yeah i forgot `ele` was the controversial one 😅 removed for now, to match the existing preset.

data/fields/ele.json
k-yle: https://github.com/openstreetmap/id-tagging-schema/pull/1988#discussion_r2910865866

data/fields/maxspeed.json
k-yle: currently iD offers both `km/h` and `mph` everywhere in the world, but the default option is chosen by [country-coder](https://github.com/rapideditor/country-coder).

So I think it would still be the responsibility of downstream apps to decide how they choose the default unit. 

For example, [some StreetComplete changesets](https://osm.org/changeset/179328330) contain [a BCP47 extension like `-u-mu-celsius`](https://www.unicode.org/reports/tr35/#MeasurementUnitPreferenceOverride), suggesting that SC already knows the users's unit preferences

data/fields/diameter.json
```diff
@@ -1,6 +1,30 @@
...
+        "usage": "default",
+        "units": {
+            "meter": [
+                "m"
+            ],
+            "centimeter": [
+                "cm"
+            ],
+            "millimeter": [
+                ""
```
matkoniecz: `mm` should still be supported, see https://github.com/ideditor/schema-builder/pull/198#pullrequestreview-3928128380

`diameter=120 mm` is still valid even if it is not preferable (and may be even preferable)

data/fields/ele.json
```diff
@@ -1,7 +1,16 @@
...
     "key": "ele",
-    "type": "number",
-    "label": "Elevation (Meters)",
+    "type": "measurement",
+    "measurement": {
+        "dimension": "length",
+        "usage": "default",
+        "units": {
+            "meter": [
+                ""
```
matkoniecz: `ele=2450 m` is also valid, right?

data/fields/ele_node.json
```diff
@@ -1,6 +1,21 @@
 {
     "key": "ele",
-    "type": "number",
+    "type": "measurement",
+    "measurement": {
+        "dimension": "length",
+        "usage": "default",
+        "units": {
+            "meter": [
+                ""
```
matkoniecz: `ele=2450 m` is also valid, right?

data/fields/min_height.json
1ec5, line 1: I’m surprised we aren’t already reusing this field for rollercoasters (“You must be yea tall to ride”). I think the `usage` would differ in that case.

data/fields/maxspeed.json
1ec5: This is a good argument for keeping `usage` in the schema. I don’t think it makes sense to keep bloating country-coder with redundant measurement properties by usage.

data/fields/maxheight.json
```diff
@@ -1,6 +1,27 @@
...
+        "dimension": "length",
+        "usage": "vehicle",
+        "units": {
+            "meter": [
+                ""
+            ],
+            "centimeter": [
+                "cm"
+            ],
+            "foot": [
```
1ec5: Doesn’t CLDR call the combined `8'10"` format `foot-and-inch`?

data/fields/ele.json
1ec5: Yeah, I’ll update #1216 to follow this pattern if this PR makes it in first.

data/fields/maxspeed.json
1ec5: > is gain from magic matches from usage is really worth specifying it over and over again and extra complexity?

In theory, `dimension` and `usage` mean that this PR shouldn’t need to repeat `"inch: ["in"]` in each preset that accepts inches.

If we omit `dimension` and `usage`, then id-tagging-schema will need to enumerate not only the allowable units but also the default unit by country. That would be much more verbose. The only exception is the combination of `length` and `vehicle`, because rapideditor/country-coder#30 implemented `roadheight` as a stopgap. It was tricky to implement. I don’t look forward to doing that a few more times over for the differences in https://github.com/ideditor/schema-builder/issues/15#issuecomment-1006823849.

data/fields/maxwidth.json
```diff
@@ -1,6 +1,30 @@
...
+            "foot": [
+                "'"
+            ],
+            "inch": [
+                "in"
+            ],
+            "yard": [
+                "yd"
+            ]
+        }
```
1ec5: These unit symbols are getting repeated all over. Can we centralize them in a new JSON file under data/? It shouldn’t ever be the case that some unit symbol is acceptable in one key but not another. (Some duration fields like `maxstay=*` use calendar units. They need a field type other than `measurement` but could potentially be powered by CLDR too.) We could go further and centralize most of these `units` specifications. The list of allowable units is generally determined by the dimension, not the individual key. `units` would only be necessary for overriding the usual list with some key-specific quirks, like `ele` only accepting meters.

1ec5 (commented):
Some more fields to migrate over:

https://github.com/openstreetmap/id-tagging-schema/blob/93113eccdddcedb502d951dc206c7add951b1d68/data/fields/distance.json#L2-L3 https://github.com/openstreetmap/id-tagging-schema/blob/93113eccdddcedb502d951dc206c7add951b1d68/data/fields/frequency.json#L2-L3 https://github.com/openstreetmap/id-tagging-schema/blob/93113eccdddcedb502d951dc206c7add951b1d68/data/fields/frequency_electrified.json#L2-L3 https://github.com/openstreetmap/id-tagging-schema/blob/93113eccdddcedb502d951dc206c7add951b1d68/data/fields/incline.json#L2-L3 https://github.com/openstreetmap/id-tagging-schema/blob/93113eccdddcedb502d951dc206c7add951b1d68/data/fields/gauge.json#L2-L3 https://github.com/openstreetmap/id-tagging-schema/blob/93113eccdddcedb502d951dc206c7add951b1d68/data/fields/voltage.json#L2-L3 https://github.com/openstreetmap/id-tagging-schema/blob/93113eccdddcedb502d951dc206c7add951b1d68/data/fields/voltage_electrified.json#L2-L3 https://github.com/openstreetmap/id-tagging-schema/blob/93113eccdddcedb502d951dc206c7add951b1d68/data/fields/maxweight.json#L2-L3 https://github.com/openstreetmap/id-tagging-schema/blob/93113eccdddcedb502d951dc206c7add951b1d68/data/fields/maxweight_bridge.json#L2-L3 https://github.com/openstreetmap/id-tagging-schema/blob/93113eccdddcedb502d951dc206c7add951b1d68/data/fields/maxaxleload_bridge.json#L2-L3

> (somehow I cannot rely inline)

(It’s because I responded to another comment as part of a larger review. You can click on the permalink for my comment to navigate to the original thread.)

data/fields/maxspeed.json
matkoniecz: My main worry is that relying on magic providing of "best" unit may end with ones not preferred for osm tagging, with this happening in way that will not be noticeable and may change as upstream package changes

Would there be at least way to see what is the outcome of that automatching?

1ec5: The unit preferences are enumerated in [this LDML file](https://github.com/unicode-org/cldr/blob/77319828794414b24554cea9043f56c881f629c8/common/supplemental/units.xml#L394-L618). CLDR accepts pull requests in case of any bugs.

matkoniecz: I am worrying less about actual bugs and more where CLDR reasonably lists X first while for OSM tagging Y should be used.

1ec5: This PR and ideditor/schema-builder#198 use `units` to enumerate the allowable units. Another possible interpretation is that they restrict the units from an external source such as CLDR to just the enumerated units. However, this PR does not address the default choice of unit when a user goes to tag something that hasn’t been tagged yet. It leaves the choice up to the client implementation. In order for the client implementation to choose correctly, it could:

<ol type="A">
<li>Use country-coder to find the preferred unit by coordinate. This library is not specifically about OSM tagging conventions and has generally <a href="https://github.com/rapideditor/country-coder/pull/132#issuecomment-2340744751">declined</a> to duplicate CLDR locale data. It made an exception for <code>roadheight</code> but probably would not attempt to replicate CLDR’s entire unit preference table. It’s just too difficult to manage in a single GeoJSON file.</li>
<li>Use country-coder to find the country code, then look up the country code in CLDR based on <code>dimension</code> and <code>usage</code>. On Android and iOS, this second step would use a simple method call built into the platform’s standard library. For now, a Web application needs to use a library provided by Unicode, but <a href="https://github.com/tc39/proposal-smart-unit-preferences/">a proposal to standardize <code>dimension</code> and <code>usage</code> lookups</a> is on the standards track behind <a href="https://github.com/tc39/proposal-amount/">another proposal for unit conversion</a>.</li>
<li>Use country-coder to find the country code, then look up the country code in a hard-coded table specific to that client. Every client would need to implement this lookup table separately.</li>
</ol>

Only a hard-coded heuristic could possibly account for an idiosyncratic choice by a local OSM community to tag in a unit that differs from real-world custom. Let’s say the Canadian community decides to tag `maxspeed=*` in kilometers per hour on railways, despite miles per hour being standard in the real world. In this case, option (B) would require an override, but it would be relatively rare and easier to maintain than a full implementation of (C).

Nothing is magical about this part of CLDR. In principle, id-tagging-schema could define a full lookup table for (C) so that clients wouldn’t have to maintain it themselves. If so, each field would still need to specify a `dimension` and `usage`, referring to a centralized unit-territory file under data/. Otherwise, these unit-territory preferences would bloat each field redundantly.

k-yle: There's no harm in adding a field for `ūsage`, but i want to clarify some things:

> https://github.com/openstreetmap/id-tagging-schema/pull/1988#discussion_r2918968363
> This is a good argument for keeping usage in the schema. I don’t think it makes sense to keep bloating country-coder with redundant measurement properties by usage.

The problem is, CLDR's `ūsage` data is very limited. So even if we add it, the regional defaults often wouldn’t make sense. For example:
1. [`dimension=length` + `ūsage=default` says](https://github.com/unicode-org/cldr/blob/834204882226ea025a7a61548f33689aaea4a840/common/supplemental/units.xml#L453-L512) that [001](https://r12a.github.io/app-subtags/index?lookup=001)
 uses kilometres to measure length. CLDR has no `ūsage` for the circumference of a tree, depth of the ocean, elevation, track-gauge of a railway, height of a building, etc. All of these examples would not use kilometers.
2. In some cases, OSM tags map to multiple CLDR `ūsage`s, For example, [minheight](https://osm.wiki/Key:minheight) is both `person-height` and `???` _(there is no `usage` value that makes sense for building height)_


> In theory, `dimension` and `usage` mean that this PR shouldn’t need to repeat `"inch: ["in"]` in each preset that accepts inches. 
> 
> https://github.com/openstreetmap/id-tagging-schema/pull/1988#discussion_r2921291057
> These unit symbols are getting repeated all over. Can we centralize them in a new JSON file under data/? It shouldn’t ever be the case that some unit symbol is acceptable in one key but not another.

Sure. But even if we define a mapping of CLDR units --> OSM units in a single global file, each field will still need to define:
1. the plausible units for that tag. For [maxheight](https://osm.wiki/Key:maxheight): `meter`, `centimeter`, `foot-and-inch`, `yard`. This is to exclude illogical values like `nanometer` and `light-year`.
2. the unit which does not require a suffix in the OSM tag. For [maxheight](https://osm.wiki/Key:maxheight): `meter`
3. some way to indicate whether the default value can be determined from CLDR data, to workaround the issue described above. For tags where CLDR's data is not sufficient, I guess data consumers have to use OSM's default unit and not support regional-specific defaults? :/

data/fields/maxheight.json
k-yle: ah, i missed that `-and-` is special-cased, just like `-per-`. fixed now.

data/fields/maxspeed.json
1ec5: I was under the impression that the `person-height` usage was intended to mean “things of a human scale”, the kind of thing that gets measured in meter-and-centimeter in many metric countries and feet and inches in some metric countries (but not vehicle dimensions that are only ever in meters). In other words, I think they only split out a new usage when needed for a regional distinction.

My suggestion of a central units file is based on the fact that we have [a single units article](https://wiki.openstreetmap.org/wiki/Map_features/Units) on the wiki. Some individual tagging schemes specify which units are acceptable or implied, but no tagging scheme comes with its own unique unit _symbol_. The individual field JSONs can continue to specify acceptable units that reference entries in the central units JSON.

data/fields/maxwidth.json
k-yle: done, unit suffixes are defined in a new file now.

If a unit is not defined in the global `unit.json` file, then it can't be used in any fields. for example we wouldn't include _lightyears_ or _imperial teaspoons_ since no OSM tag would ever want those values.

fields that have strict rules like `ele` can still further constrain the list of allowed units.

data/fields/min_height.json
k-yle: yeah, if we add this field to amusement park features in the future, we will need to split it into 2 JSON files. so no action required right now imo

data/fields/maxspeed.json
k-yle: i see, then I think https://github.com/openstreetmap/id-tagging-schema/pull/1988#discussion_r2999789091 (specifically https://github.com/ideditor/schema-builder/pull/198/commits/161112209ae9b18a6a548695209462cc51027bc4) addresses most of these issues.
 - units defined in a single global file.
 - each field must define a `usage`
 - each field can optionally further limit the allowed fields
 - each field can define the unit which is implied by default 

both PRs updated

data/fields/aerialway/duration.json
```diff
@@ -1,7 +1,15 @@
...
+        "dimension": "duration",
+        "usage": "default",
+        "units": [
+            "minute"
+        ],
+        "impliedUnit": "minute"
+    },
     "minValue": 0,
-    "label": "Duration (minutes)",
+    "label": "{duration}",
```
matkoniecz, line 13: I am not sure whether we want to remove hint

are we assuming that iD and other will add it back, somehow, using `impliedUnit`

matkoniecz: but then we have `maxspeed` where we have impliedUnit but it should be pushed far less strongly

data/fields/maxspeed.json
```diff
@@ -1,6 +1,11 @@
 {
     "key": "maxspeed",
-    "type": "roadspeed",
+    "type": "measurement",
+    "measurement": {
+        "dimension": "speed",
+        "usage": "default",
```
matkoniecz, line 6: maybe we should not need to state `default` and default to `default` as value?

data/fields/plant/output/electricity.json
```diff
@@ -1,8 +1,13 @@
 {
     "key": "plant:output:electricity",
-    "type": "typeCombo",
+    "type": "measurement",
+    "measurement": {
+        "dimension": "power",
+        "usage": "default"
```
matkoniecz, line 6: is there reason to not specify here and similar MW as preferred? Is `placeholder` enough?

I am not sure how it will work in iD in practice, I guess it can be tweaked later?

data/fields/water_tank/volume.json
```diff
@@ -1,6 +1,11 @@
 {
     "key": "water_tank:volume",
-    "type": "text",
-    "label": "Volume (Liters)",
+    "type": "measurement",
+    "measurement": {
+        "dimension": "volume",
+        "usage": "fluid",
```
matkoniecz: what is the benefit of specifying it here? (I am still confused by that entire `usage` thingy, and trying educate myself via https://cldr.unicode.org/ )

.vscode/settings.json
k-yle, line 1: i just learned that [`railway:position`](https://osm.wiki/Key:railway:position) and [`railway:position:exact`](https://osm.wiki/Key:railway:position:exact) use `mi:12.3` instead of `12.3 mi` 😭 

and that tag uses a "fake" unit called `pkm` in argentina. i guess we can't support that field

1ec5: Ah, post kilometers. But is there any other kind? I thought `railway:position=*` uses a different syntax in order to accommodate things like [milepost equations](https://en.wikipedia.org/wiki/Milepost_equation).

k-yle: if there is, that would ideally be documented on [the wiki page](https://osm.wiki/Key:railway:position). but i think it's clear that we can't support the syntax of this tag

data/fields/ele.json
```diff
@@ -1,7 +1,15 @@
 {
     "key": "ele",
-    "type": "number",
-    "label": "Elevation (Meters)",
```
matkoniecz, line 4: taking look at this again: what is the benefit of dropping unit from label here and elsewhere where field supports solely single label?

SCHEMA.md
```diff
@@ -671,6 +670,38 @@ Combo field types can accept key-label pairs in the `options` value of the `stri
 
 An optional property to reference to the icons of another field, indicated  by using that field's name contained in brackets, like `{field}`. This is for example useful when there are multiple variants of fields for the same tag, which should all use the same icons.
 
+##### `measurement`
+
+Used when `type = measurement`. Defines the unit of measurements that are supported by this field. For example:
```
matkoniecz:
```suggestion
Used when `type = measurement` defines the unit of measurements that are supported by this field. For example:
```

seems to flow better for me

disclaimer: not a native speaker

SCHEMA.md
```diff
@@ -671,6 +670,38 @@ Combo field types can accept key-label pairs in the `options` value of the `stri
...
+
+```json
+{
+  "key": "diameter",
+  "type": "measurement",
+  "measurement": {
+    // The dimension being measured. This constrains the permitted units.
+    // The ID id defined by CLDR.
```
matkoniecz: maybe link CLDR in description above?

at least I have not really understood what CLDR even is before

or is it case "obviously google for it if unclear"?

SCHEMA.md
```diff
@@ -671,6 +670,38 @@ Combo field types can accept key-label pairs in the `options` value of the `stri
...
+    // The corresponding 'usage' from CLDR.
+    "usage": "default",
+
+    // If the field only allows some units, you can list them here
+    // using CLDR's unit names. If not specified, then, all units from
+    // this dimension are allowed.
+    "units": ["meter", "centimeter", "foot-and-inch"],
+
+    // Some OSM tags have a default unit, which does not need to be explicitly included in the tag value.
+    // This field defines how to interpret a unit-less value:
```
matkoniecz:
```suggestion
    // This field defines how to interpret an implicit unit:
```

```suggestion
    // This field defines how to interpret case without explicit unit:
```

```suggestion
    // This field defines how to interpret case where unit is not specified:
```

whole fun is that it is not unitless.

SCHEMA.md
matkoniecz: "The ID id"?

SCHEMA.md
```diff
@@ -671,6 +670,38 @@ Combo field types can accept key-label pairs in the `options` value of the `stri
...
+}
+```
+
+To convert the unit IDs into the values used by OSM, see [§Units](#Units).
+
+Translations for the [`narrow` and `long` form](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat) of each unit are bundled into iD-tagging-schema's locale files.
+
 ### Deprecations
 
 Use `deprecated.json` ([Example](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/deprecated.json), [Schema](https://github.com/ideditor/schema-builder/blob/main/schemas/deprecated.json)) to specify tag deprecations.
```
matkoniecz, line 732: migration guide file also should get an entry, right?

matkoniecz:
> Closes https://github.com/ideditor/schema-builder/issues/15

Can you edit it here and in commit so it points to https://github.com/openstreetmap/id-tagging-schema/issues/2530 ?

(sorry for another round of nitpicking without promise or chance that it covers all remaining issues)

SCHEMA.md
matkoniecz: maybe include there (or in docs) some high-level overview how that thing works? Sadly, I am still not entirely sure and mostly scared by this

schemas/field.json
```diff
@@ -368,6 +367,40 @@
...
+                },
+                "usage": {
+                    "type": "string",
+                    "description": "The corresponding 'usage' from CLDR"
+                },
+                "units": {
+                    "type": "array",
+                    "minItems": 1,
+                    "uniqueItems": true,
+                    "description": "Optional, if only some units are allowed for this tag, then list the permitted units here, using the unit IDs from CLDR (see https://cdn.jsdelivr.net/npm/cldr-core/supplemental/unitPreferenceData.json).",
```
matkoniecz: hmm, always only some units will be allowed

maybe "if only some units of set remaining after filtering by usage and dimension..." wait, no, that is terrible

maybe "Optional, allows to further constrain allowed units" ?

schemas/field.json
```diff
@@ -368,6 +367,40 @@
...
+                    "items": {
+                        "type": "string"
+                    }
+                },
+                "impliedUnit": {
+                    "type": "string",
+                    "description": "Some OSM tags have a default unit, which does not need to be explicitly included in the tag value. This field defines how to interpret a unit-less value."
```
matkoniecz:
```suggestion
                    "description": "Some keys have a default unit, which does not need to be explicitly included in the tag value. This field defines how to interpret a unit-less value."
```

do we need to specify here that we think about OSM tags?

I think that we have no plans whatsoever to be expanding beyond OpenStreetMap with this project

also, keys would have default units, not tags, right?

schemas/field.json
```diff
@@ -393,4 +426,4 @@
             "enum": ["point", "vertex", "line", "area", "relation"]
         }
     }
-}
+}
```
matkoniecz, line 570: why remove final newline?

schemas/field.json
```diff
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "https://github.com/openstreetmap/id-tagging-schema/raw/main/schemas/field.json",
+    "$id": "https://cdn.jsdelivr.net/npm/@ideditor/schema-builder/schemas/field.json",
```
matkoniecz: is it some temporary change for CI to work?

SCHEMA.md
```diff
@@ -671,6 +670,38 @@ Combo field types can accept key-label pairs in the `options` value of the `stri
...
+  "type": "measurement",
+  "measurement": {
+    // The dimension being measured. This constrains the permitted units.
+    // The ID id defined by CLDR.
+    "dimension": "length",
+
+    // The corresponding 'usage' from CLDR.
+    "usage": "default",
+
+    // If the field only allows some units, you can list them here
```
matkoniecz, line 696: same as earlier in .json

SCHEMA.md
```diff
@@ -671,6 +670,38 @@ Combo field types can accept key-label pairs in the `options` value of the `stri
...
+    // using CLDR's unit names. If not specified, then, all units from
+    // this dimension are allowed.
+    "units": ["meter", "centimeter", "foot-and-inch"],
+
+    // Some OSM tags have a default unit, which does not need to be explicitly included in the tag value.
```
matkoniecz, line 701: same as earlier in .json

data/fields/aerialway/duration.json
k-yle: there's no need to repeat it in the label, since it's now shown in the UI:

image

data/fields/plant/output/electricity.json
k-yle: "preferred" units is out of scope for this PR. that's currently [country-coder](https://github.com/rapideditor/country-coder)'s job (for some dimensions). 

I suggest tackling that as a separate topic

data/fields/water_tank/volume.json
k-yle: we need `usage` so that iD can show sensible units in the dropdown.

for example: 
- we don't want `lightyears` to be an option for `maxheight=*`
- we don't want `tablespoons` to be an option for `water_tank:volume=*`

data/fields/ele.json
k-yle: I replied in https://github.com/openstreetmap/id-tagging-schema/pull/1988#discussion_r3645339250

schemas/field.json
k-yle: there's some formatting inconsistencies in this folder caused by the repo merge. fixed now

schemas/field.json
k-yle: no longer needed after the repo merge, removed

SCHEMA.md
k-yle: sorry, which earlier comment? there are 59 comments now 😆

k-yle:
I have commited the changes to dist/translations for `en`/`pl`/`de`/`vi`/`mi`, to illustrate the impact, and maybe make it easier to review

data/fields/water_tank/volume.json
1ec5: Haha, I’m going to have to find a sign somewhere that measures a bridge’s height in light years or a tank’s volume in tablespoons!

SCHEMA.md
matkoniecz: https://github.com/openstreetmap/id-tagging-schema/pull/1988#pullrequestreview-4765187268

data/fields/maxspeed.json
k-yle: i think it's fine to be explicit? otherwise it looks like an optional field that contributors can ignore

SCHEMA.md
k-yle: corresponds to https://github.com/openstreetmap/id-tagging-schema/pull/1988#discussion_r3638968548

k-yle:
(rebased just to get the new data into https://kyle.kiwi/iD, i will address code review comments later)

data/fields/generator/output/electricity.json
```diff
@@ -1,8 +1,13 @@
 {
     "key": "generator:output:electricity",
-    "type": "typeCombo",
+    "type": "measurement",
+    "measurement": {
+        "dimension": "power",
+        "usage": "default"
```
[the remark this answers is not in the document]
k-yle: this is a good point and it's quite annoying. here's [what it would look like](https://kyle.kiwi/iD):

image

"yes gigawatts" is a bit silly, not sure how we should handle this, since `yes` is only valid for some fields like `*:output:electricity`, but not `maxspeed`

image

data/fields/plant/output/electricity.json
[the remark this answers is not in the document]
k-yle: lets discuss in the other thread: https://github.com/openstreetmap/id-tagging-schema/pull/1988#discussion_r3852318430

PR #1990, opened 2026-03-05
Fix: enforce LF line endings to resolve prettier errors on Windows

Kayd-06:
### Description

When checking out this repository on a Windows machine (which typically defaults to `core.autocrlf=true`), Git checks out files with Carriage Return Line Feed (`CRLF`) line endings. However, `prettier` enforces Line Feed (`LF`) endings (via `.prettierrc`). This mismatch causes `npm run lint` to fail out-of-the-box on Windows with thousands of formatting warnings.

To prevent this error and guarantee consistent file endings across all operating systems, I added a `.gitattributes` file to enforce `LF` for text files.


matkoniecz:
if band aid for bad OSes is needed, maybe something as simple as

```
* text=auto
```

would be sufficient and work? What is the benefit of listing file types?

(idea taken from https://stackoverflow.com/questions/26075786/force-lf-line-endings-using-gitattributes-without-losing-automatic-text-binary - as I have no access to Windows I am unable to test it, but if I read https://git-scm.com/docs/gitattributes#_end_of_line_conversion correctly it should work well)

Kayd-06:
Yes, using just `* text=auto` is much cleaner and sufficient. I've tested it locally on Windows and `npm run lint` now passes successfully. I have pushed the updated `.gitattributes` file to the PR.

k-yle (commented):
`* text=auto` will [break binary files](https://www.drupal.org/project/drupal/issues/2333243) if git's autodetection fails. 

I'm not opposed, but if your laptop is misconfigured, it's usually better to fix your own settings. Otherwise you'll just experience the same issue in every single git repository that you edit.

If this is accepted, it should probably contain [explicit exceptions](https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings?platform=windows#example) for every binary file that we have committed (for example `*.png`)

PR #2009, opened 2026-03-08
Remove 'maxweight' field from man_made=bridge

matkoniecz:

It has maxweight field, and not sure what it means there. Is it supposed to be typical weight limit on bridge? But it can be different on different sections and it sounds like pointless duplicate of tags from roads. And in such case it should also get maxweightrating in/after #1986

Is it supposed to be design capacity? Maximum weight capacity? Both of limited verifiability, especially second one.
And encouraging to copy official stats seems dubious.

From looking at https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dbridge it is not listed as expected combination.

Removing as meaning, utility and usefulness is dubious and unclear.

### Related issues

Spotted as part of #1986

### Links and data

**Relevant OSM Wiki links:**

https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dbridge

**Relevant tag usage stats:**
> …

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/data/presets/man_made/bridge.json
+++ b/data/presets/man_made/bridge.json
@@ -4,7 +4,6 @@
         "name",
         "bridge_combo",
         "layer",
-        "maxweight",
         "ref_bridge_number"
     ],
     "moreFields": [


andrewharvey:
> It has maxweight field, and not sure what it means there. Is it supposed to be typical weight limit on bridge?

Signposted or legal weight limit on the bridge.

For example a ["Bridge Load Limit" sign](https://wiki.openstreetmap.org/wiki/Australian_Tagging_Guidelines/Road_Signage#R6-3)

image

> But it can be different on different sections and it sounds like pointless duplicate of tags from roads

It can and should still be tagged on the `highway=*` ways for routing, but it's useful to have it on the `man_made=bridge` too for simple queries like find all bridges with a load limit under n tonnes.

> And in such case it should also get maxweightrating in/after https://github.com/openstreetmap/id-tagging-schema/pull/1986

maxweight is based on the current actual weight of the vehicle, maxweightrating is based on the weight that the vehicle is rated to carry.

matkoniecz:
Duplicating road info onto bridge is both dubious duplication and not documented at wiki and not done reliably, so any such query is doomed anyway.

1ec5:
`man_made=bridge` itself is intentionally redundant to the `highway=*` `bridge=yes` ways, so keys like `name=*` and `ref=*` on it are naturally redundant to `bridge:ref=*` and `bridge:name=*`.

As for `maxweight=*`, I don’t normally see mappers in my region bothering to tag it on `man_made=bridge`, despite the field. This field could potentially cause a mapper unfamiliar with routers’ requirements to tag it on just the area but not the roadways. This has happened [over 6,700 times](https://qlever.dev/osm-planet/AfDKgJ). Fortunately, it’s pretty easy to catch using queries or validators.

In my experience, a prefabricated bridge usually comes with [a plaque](https://commons.wikimedia.org/wiki/File:Sojourner_Truth_bridge_placard.jpg) indicating the manufacturer and maximum capacity (vehicle load limit) in tons. This is a design limit that seems silly to tag as `maxweight=*` on the footway or cycleway that goes across it, so I do sometimes [tag it on the area](https://www.openstreetmap.org/way/1497539942). Prefabricated bridges are the norm for footways and cycleways and are becoming more common for roadways too.

If I’m not mistaken, a load limit is a total that applies to everything on the bridge at a given time, similar to `capacity=*`, It isn’t a limit on each individual vehicle. If you tag `maxweight=*` on the `highway=*` way and someone splits it, the tag technically becomes inaccurate. Tagging it on the area seems more prudent, but if we actually want to support load capacity, we should probably double-check with the community that `maxweight=*` is the right key for it.

1ec5:
So that’s the other thing… “load limit” is ambiguous. It can refer to either the total load the bridge has to support or the load that a single vehicle places on the bridge, or even just the load that a vehicle is carrying. For example, [some U.S. authorities](https://engineer.mcohio.org/departments/bridge-department/posted-bridge-load-limits/) use “bridge load limit” to refer to an individual vehicle weight that they signpost using the usual `maxweight`-related signs.

For a short bridge, it’s a distinction without a difference, but for a longer bridge, you’d need additional context before translating it to a tag. In the case of a prefabricated pedestrian bridge (or an elevator or escalator or amusement park ride), some mappers have been associating the posted design capacity with a key like `maxweight` out of common sense – what other weight would be posted? – but a more specific key could probably find support.

matkoniecz:
> It can refer to either the total load the bridge has to support or the load that a single vehicle places on the bridge, or even just the load that a vehicle is carrying.

ops, I assumed that it is always the first

second one would be just `maxweight` tagged on road, right?

1ec5:
Correct.

tyrasd (approved):

1ec5:
I started [a forum discussion](https://community.openstreetmap.org/t/load-limits-on-bridges-and-elevators/145077) about how to tag bridge load limits that are about the overall bridge rather than a traffic restriction.

matkoniecz:
If those would become documented at OSM Wiki and there would be consensus that such tagging makes sense, and anyone thinks then it should become supported in iD presets - please open an issue about that

PR #2012, opened 2026-03-08
Add `highchair` to more fields for `amenity=pub`

codeinabox:

Often pubs are family-friendly, and will offer highchairs to their customers with young children.


codeinabox:
> so I would tend toward "no" here, though for `moreFields`...

To clarify, do you mean this isn't significant enough to appear in `moreFields`, or that's where I should add it?

codeinabox:
In that case, I'll close this pull request for now. I've updated the wiki so hopefully more people will tag pubs that offer highchairs.

PR #2037, opened 2026-03-13
Use temaki-saddle icon for craft/saddler preset

### LINKED ISSUE #2036: Create dedicated icon for craft=saddler (saddle icon)
Extracted from #41 which has quite a lot of text and may be overwhelming. This issue seems fairly doable by new people, and no OSM-specific knowledge is actually needed. Therefore I will mark it as "good first issue". Having one or maybe two/three of those more specific icon creation requests, where there is a viable idea for a new icon, seems to make sense to me.

See https://github.com/ideditor/schema-builder/blob/main/ICONS.md#icons for icon info.

https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/craft/saddler.json may get as icon something like saddle (or something else! I am not great at icon design

See https://github.com/openstreetmap/id-tagging-schema/pull/1964 for a very similar case where things went well

similarly as there following steps are needed

- if icon is available in one of https://github.com/ideditor/schema-builder/blob/main/ICONS.md#where-do-the-icons-come-from - then just open PR changing icon code

or

- create/find freely licensed icon
- submit it to one of icon repositories mentioned at https://github.com/ideditor/schema-builder/blob/main/ICONS.md#icons (follow all requirements of that project)
- wait for it to be merged, new version released
- wait for dependabot to emit new PR in iD repository
- wait for merge in iD
- create here PR changing icon

It is a bit fiddly as seen above but as can be seen in #1964 it is not so bad with @rapideditor/temaki used as an icon repository.

----------------------------------

See https://github.com/openstreetmap/id-tagging-schema/issues/41#issuecomment-3491973553 for more cases where new icon would be welcome!

ak8abhinay:
Updated the icon for the Saddler preset from `temaki-tools` to `temaki-saddle`.

`temaki-tools` (hammer) is too generic and not representative of saddlery work.
Searched across all icon libraries, none contain a saddler-specific 
tool icon (stitching awl, pricking iron, round knife, etc.).

`temaki-saddle` is the most accurate and meaningful representation of the 
Saddler preset as it directly depicts a saddle, which is the core output 
of a saddler's craft.

Fixes #2036

### DIFF
--- a/data/presets/craft/saddler.json
+++ b/data/presets/craft/saddler.json
@@ -1,5 +1,5 @@
 {
-    "icon": "temaki-tools",
+    "icon": "temaki-saddle",
     "geometry": [
         "point",
         "area"


ak8abhinay:
Here is how the icon looking in preview. 

image

matkoniecz:
huh, is this saddle supposed to be horse saddle or saddle as in pass between mountains?

because I am not entirely sure is it resembling either

matkoniecz:
lets see what others think here (I am not a horse expert)

tyrasd:
The `tamki-saddle` icon is definitely meant for [`natural=saddle`](https://wiki.openstreetmap.org/wiki/Tag:natural=saddle). It does not work for a saddler.

I could not find a suitable icon for a horse's saddle in the icon sets we use, so we'd have to [create one](https://github.com/rapideditor/temaki?tab=readme-ov-file#design-guidelines) from scratch. For a preset with currently only about 500 uses, this seems not worth the effort, (except of course if you really like this particular kind of craft, in which case feel free to work on it).

> toolkit icon

That one is essentially equivalent to the current `temaki-tools` icon, so I don't see the benefit of replacing one generic icon with another equally generic one.

PR #2062, opened 2026-03-15
Add preset for `man_made=geoglyph`

### LINKED ISSUE #14: Consider adding man_made=geoglyph to presets
https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dgeoglyph

ak8abhinay:
Adds a new preset for `man_made=geoglyph`.

wiki - https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dgeoglyph

Fixes #14

### DIFF
--- a/data/presets/man_made/geoglyph.json
+++ b/data/presets/man_made/geoglyph.json
@@ -1,0 +1,25 @@
+{
+    "fields": [
+        "name",
+        "description"
+    ],
+    "moreFields": [
+        "{man_made}",
+        "wikipedia",
+        "wikidata"
+    ],
+    "geometry": [
+        "point",
+        "line",
+        "area"
+    ],
+    "terms": [
+        "earthwork",
+        "hill figure",
+        "land art"
+    ],
+    "tags": {
+        "man_made": "geoglyph"
+    },
+    "name": "Geoglyph"
+}
--- a/data/presets/man_made/geoglyph.json
+++ b/data/presets/man_made/geoglyph.json
@@ -21,5 +21,9 @@
     "tags": {
         "man_made": "geoglyph"
     },
+    "reference": {
+        "key": "man_made",
+        "value": "geoglyph"
+    },
     "name": "Geoglyph"
 }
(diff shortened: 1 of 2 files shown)


data/presets/man_made/geoglyph.json
```diff
@@ -0,0 +1,25 @@
...
+    ],
+    "moreFields": [
+        "{man_made}",
+        "wikipedia",
+        "wikidata"
+    ],
+    "geometry": [
+        "point",
+        "line",
+        "area"
```
k-yle, line 14: [`man_made=geoglyph`](https://osm.wiki/Tag:man_made=geoglyph) is also used on relations with [`type=multilinestring`](https://osm.wiki/Relation:multilinestring). Currently it doesn't really work:

Image

Maybe we need a corresonding PR on the iD side, to allow the `line` presets on relations with [`type=multilinestring`](https://osm.wiki/Relation:multilinestring)?

That's what iD currently does for `area` presets, so that they can be used on relations with [`type=multipolygon`](https://osm.wiki/Relation:multipolygon).

ak8abhinay: Should I remove `line` from the geometry for now, 
or keep it as is until the iD side is resolved?

matkoniecz: > Maybe we need a corresonding PR on the iD side, to allow the line presets on relations with [type=multilinestring](https://osm.wiki/Relation:multilinestring)?

it seems so? Though not sure should this a bit silly relation type should be supported

>  or keep it as is until the iD side is resolved?

no, in either case (ignore relation type, add special support in iD side) it should be kept

k-yle: lets go with https://github.com/openstreetmap/iD/issues/12086#issuecomment-4116949225 as the short term solution then

ak8abhinay: Understood. So, an extra preset which has `relation` and `type=multilinestring` should be added alongside this preset.

matkoniecz (approved):
missing icon is sad but I do not really have an idea for one

Image
Image
Image

matkoniecz:

PR #2067, opened 2026-03-15
Update piste difficulty_downhill options to include symbols

andrewharvey:
Similar to #462 use symbol icons within the options

### DIFF
--- a/data/fields/piste/difficulty_downhill.json
+++ b/data/fields/piste/difficulty_downhill.json
@@ -6,10 +6,10 @@
     "strings": {
         "options": {
             "novice": "Novice (instructional)",
-            "easy": "Easy (green circle)",
-            "intermediate": "Intermediate (blue square)",
-            "advanced": "Advanced (black diamond)",
-            "expert": "Expert (double black diamond)",
+            "easy": "🟢 Easy (green circle)",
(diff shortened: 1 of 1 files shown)


andrewharvey:
image

tordans:
The alignment is pretty off in this. I wonder if this would look nicer if we expanded the logic behind https://github.com/ideditor/schema-builder?tab=readme-ov-file#icons to allow emoji/ascii "icons"…?

matkoniecz (changes requested):
this looks like a hack to me, and it is not some critical workaround

we have dedicated option for icons, and while it is impossible to define color for them I would prefer expending tagging schema over putting icons into translations which is hacky and may cause variety of side effects

admittedly it is much more complex and will require editors to add support - but maybe you would be interested in implementation

andrewharvey:
Fair points, happy if you want to close this if we feel dedicated icons should be used instead of these symbols within the text.

matkoniecz:
I am pretty sure. Emoji icons are tempting, but...

I created https://github.com/openstreetmap/id-tagging-schema/issues/2614 for the idea (which definitely makes sense, +1 to it and thanks for spotting and sharing it!)

PR #2078, opened 2026-03-17
Add `food=` field to pub and bar presets

### LINKED ISSUE #429: Presets for bars/pubs/cafes should ask if food is served
When looking for food, some apps suggest bars and pubs tagged with food=no.
- https://github.com/organicmaps/organicmaps/issues/2309
- https://github.com/osmandapp/OsmAnd/issues/13992

It seems that the food=no tag, while useful, is little used with [625 uses](https://taginfo.openstreetmap.org/keys/food#values)
Please consider adding an option for it to the presets for pubs, bars and cafes.

Quoting from https://wiki.openstreetmap.org/wiki/Key:food:

> The key food=yes/no is used to indicate whether a feature provides food.
> This is useful on features which do not always provide food, such as [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=[cafe](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcafe), [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=[bar](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbar) and especially [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=[pub](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dpub)

Thanks

ak8abhinay:
Added `food` to `moreFields` of the pub, bar and cafe presets.

Changes:

Created `data/fields/food.json` with label "Is food available?" 
Added `food` to `moreFields` in pub.json, bar.json and cafe.json

Fixes #429


ak8abhinay:
pub : 

image


bar : 

image


cafe : 

image

matkoniecz (approved):

data/presets/amenity/cafe.json
```diff
@@ -24,6 +24,7 @@
         "delivery",
         "diet_multi",
         "drive_through",
+        "food",
```
matkoniecz: see above, why it is added also here?

matkoniecz (changes requested):

data/presets/amenity/cafe.json
ak8abhinay: I added the food field following issue #429 and the wiki mentions food: = yes/no 

Should I keep it here in moreFields for those cases or  remove it from the cafe preset

codeinabox: What's a scenario where a cafe wouldn't serve food? Even a tiny coffee kiosk that doesn't serve meals would still fall under `cuisine=coffee_shop`.

ak8abhinay: Agree and I better remove `food` from `moreFields of cafe`

matkoniecz (approved):
Looks good to me,  but let's see whether there are other comments. 

I missed this cafe thing after all (thanks to @codeinabox for catching it!)

matkoniecz:

PR #2082, opened 2026-03-18
Add Rooftop Solar Thermal Collector preset

### LINKED ISSUE #762: generator:type=solar_thermal_collector
### OSM Tag(s)

generator:method=thermal & generator:type=solar_thermal_collector

As a Preset

### Label

Rooftop Solar Thermal Collector

### Aliases

Solar Thermal Heaters (not 100% accurate, it's the collector. But a decent amount of people would refer to it as such as well), solar hot water panels

### Terms

_No response_

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:generator:source%3Dsolar#Solar_Thermal_Collectors

### Status of the Tag

Approved

### Usage of the tag

12,151

### Replaces other Tag?

_No response_

### Regional Tag?

_No response_

### Further Information

We already have rooftop solar for regular solar panels (for electricity) but this would be for the thermal collector (heat)

ak8abhinay:
Adds a new preset for `generator:type=solar_thermal_collector`

Fixes #762

### DIFF
--- a/data/presets/power/generator/method/solar_thermal_collector.json
+++ b/data/presets/power/generator/method/solar_thermal_collector.json
@@ -1,0 +1,38 @@
+{
+    "icon": "fas-solar-panel",
+    "fields": [
+        "ref",
+        "operator"
+    ],
+    "moreFields": [
+        "{power/generator}",
+        "location"
+    ],
+    "geometry": [
+        "area",
+        "point",
+        "vertex"
+    ],
+    "tags": {
+        "power": "generator",
+        "generator:method": "thermal"
+    },
+    "addTags": {
+        "power": "generator",
+        "generator:method": "thermal",
+        "generator:source": "solar",
+        "generator:type": "solar_thermal_collector",
+        "location": "roof"
+    },
+    "reference": {
+        "key": "generator:method",
+        "value": "thermal"
+    },
+    "terms": [
+        "solar thermal collector",
+        "solar water heater",
+        "solar hot water panel",
+        "rooftop solar thermal"
+    ],
+    "name": "Rooftop Solar Thermal Collector"
+}
--- a/data/presets/power/generator/method/solar_thermal_collector.json
+++ b/data/presets/power/generator/method/solar_thermal_collector.json
@@ -28,7 +28,6 @@
         "value": "thermal"
     },
     "terms": [
-        "solar thermal collector",
         "solar water heater",
         "solar hot water panel"
     ],
--- a/data/presets/power/generator/method/solar_thermal_collector.json
+++ b/data/presets/power/generator/method/solar_thermal_collector.json
@@ -22,8 +22,8 @@
         "value": "thermal"
     },
     "terms": [
-        "solar water heater",
-        "solar hot water panel"
+        "solar hot water panel",
+        "solar water heater"
     ],
     "name": "Solar Thermal Collector"
 }
--- a/data/presets/power/generator/method/solar_thermal_collector.json
+++ b/data/presets/power/generator/method/solar_thermal_collector.json
@@ -18,8 +18,8 @@
         "generator:type": "solar_thermal_collector"
     },
     "reference": {
-        "key": "generator:method",
-        "value": "thermal"
+        "key": "generator:type",
+        "value": "solar_thermal_collector"
     },
     "terms": [
         "solar hot water panel",
--- a/data/presets/power/generator/method/solar_thermal_collector.json
+++ b/data/presets/power/generator/method/solar_thermal_collector.json
@@ -13,10 +13,6 @@
     ],
     "tags": {
         "power": "generator",
-        "generator:method": "thermal"
-    },
-    "addTags": {
-        "power": "generator",
         "generator:method": "thermal",
         "generator:source": "solar",
         "generator:type": "solar_thermal_collector"
--- a/data/presets/power/generator/method/solar_thermal_collector.json
+++ b/data/presets/power/generator/method/solar_thermal_collector.json
@@ -2,11 +2,11 @@
     "icon": "fas-solar-panel",
     "fields": [
         "ref",
-        "operator",
-        "location"
+        "operator"
     ],
     "moreFields": [
-        "{power/generator}"
+        "{power/generator}",
+        "location"
     ],
     "geometry": [
         "area",
--- a/data/presets/power/generator/method/solar_thermal_collector.json
+++ b/data/presets/power/generator/method/solar_thermal_collector.json
@@ -2,9 +2,10 @@
     "icon": "fas-solar-panel",
     "fields": [
         "{power/generator}"
+        "location"
     ],
     "moreFields": [
-        "location"
+        "{power/generator}"
     ],
     "geometry": [
         "area",
--- a/data/presets/power/generator/method/solar_thermal_collector.json
+++ b/data/presets/power/generator/method/solar_thermal_collector.json
@@ -1,8 +1,7 @@
 {
     "icon": "fas-solar-panel",
     "fields": [
-        "ref",
-        "operator"
+        "{power/generator}"
     ],
     "moreFields": [
         "{power/generator}",
--- a/data/presets/power/generator/method/solar_thermal_collector.json
+++ b/data/presets/power/generator/method/solar_thermal_collector.json
@@ -4,7 +4,6 @@
         "{power/generator}"
     ],
     "moreFields": [
-        "{power/generator}",
         "location"
     ],
     "geometry": [
--- a/data/presets/power/generator/method/solar_thermal_collector.json
+++ b/data/presets/power/generator/method/solar_thermal_collector.json
@@ -1,7 +1,7 @@
 {
     "icon": "fas-solar-panel",
     "fields": [
-        "{power/generator}"
+        "{power/generator}",
         "location"
     ],
     "moreFields": [
--- a/data/presets/power/generator/method/solar_thermal_collector.json
+++ b/data/presets/power/generator/method/solar_thermal_collector.json
@@ -22,6 +22,7 @@
         "value": "solar_thermal_collector"
     },
     "terms": [
+        "heating",
         "solar hot water panel",
         "solar water heater"
     ],
--- a/data/presets/power/generator/method/solar_thermal_collector/location/roof.json
+++ b/data/presets/power/generator/method/solar_thermal_collector/location/roof.json
@@ -23,9 +23,9 @@
         "value": "thermal"
     },
     "terms": [
-        "solar water heater",
+        "rooftop solar thermal",
         "solar hot water panel",
-        "rooftop solar thermal"
+        "solar water heater"
     ],
     "name": "Rooftop Solar Thermal Collector"
 }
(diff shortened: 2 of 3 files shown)


matkoniecz:
I expect it is likely safe given https://en.wikipedia.org/wiki/Solar_thermal_collector but occasionally Wikipedia has poor titles

matkoniecz (changes requested):
we definitely should add also regular variant - to avoid people tagging non-rooftop ones as rooftop ones

data/presets/power/generator/method/solar_thermal_collector_rooftop.json
```diff
@@ -0,0 +1,40 @@
...
+        "generator:source": "solar",
+        "generator:type": "solar_thermal_collector",
+        "location": "roof"
+    },
+    "reference": {
+        "key": "generator:method",
+        "value": "thermal"
+    },
+    "terms": [
+        "solar thermal collector",
```
matkoniecz:
```suggestion
```

no need to include substring

data/presets/power/generator/method/solar_thermal_collector.json
```diff
@@ -0,0 +1,36 @@
...
+        "generator:method": "thermal",
+        "generator:source": "solar",
+        "generator:type": "solar_thermal_collector"
+    },
+    "reference": {
+        "key": "generator:method",
+        "value": "thermal"
+    },
+    "terms": [
+        "solar thermal collector",
```
matkoniecz:
```suggestion
```
no need to include substring

matkoniecz:
and thanks for changes!

matkoniecz:
@1ec5 Maybe "Solar Water Heater" would be a better name here?

In such case  "solar thermal collector" would be worth having in terms so maybe not rush to remove them...

1ec5:
As I understand it, the collector is the only visible part of a solar water heating system. The water heater is the tank that’s typically located indoors. [This explainer](https://www.energy.gov/energysaver/solar-water-heaters) from the U.S. Department of Energy has some helpful diagrams.

matkoniecz:
@1ec5 Is it inaccurate enough to be misleading and confusing?

1ec5:
As a layperson, if I saw a preset named “[anything] Water Heater”, I would think of an indoor tank and wouldn’t think to apply it to anything on the roof. “Thermal collector” may be jargon, but it’s probably the best we can do until these systems become mainstream enough to have colloquial names.

data/presets/power/generator/method/solar_thermal_collector_rooftop.json
```diff
@@ -0,0 +1,39 @@
+{
+    "icon": "fas-solar-panel",
+    "fields": [
+        "ref",
+        "operator"
```
matkoniecz: is not referencing power/generator here intentional?

ak8abhinay: It was not intentional, but both presets have `{power/generator}` in `moreFields`. Should I move it to `fields`?"

matkoniecz: I expect to both `fields` and `moreFields` have this, so they will inherit both `fields` and `moreFields` from more general preset

unless there is a good reason to not do this?

ak8abhinay: > so they will inherit both `fields` and `moreFields` from more general preset

Alright!

data/presets/power/generator/method/solar_thermal_collector.json
```diff
@@ -0,0 +1,35 @@
+{
+    "icon": "fas-solar-panel",
+    "fields": [
+        "ref",
```
matkoniecz: is there reason to not inherit `"{power/generator}",` here?

data/presets/power/generator/method/solar_thermal_collector/location/roof.json
```diff
@@ -0,0 +1,38 @@
+{
+    "icon": "fas-solar-panel",
+    "fields": [
+        "{power/generator/method/solar_thermal_collector}"
+    ],
+    "moreFields": [
+        "{power/generator}"
```
matkoniecz:
```suggestion
        "{power/generator/method/solar_thermal_collector}"
```

good point that this sub-preset can inherit from its direct parent

lets do this for both fields

code change based on assumption it works for `fields`

data/presets/power/generator/method/solar_thermal_collector.json
ak8abhinay: `photovoltaic.json` also has `{power/generator}` in moreFields, so I followed the same pattern

matkoniecz: for reference https://github.com/openstreetmap/id-tagging-schema/blob/7f642043effe38a701ed0a77d070ef3aec367c12/data/presets/power/generator.json

https://github.com/openstreetmap/id-tagging-schema/blob/7f642043effe38a701ed0a77d070ef3aec367c12/data/presets/power/plant/source/solar/photovoltaic.json#L4

ak8abhinay: Should I replace `ref` and `operator` in fields with `{power/generator}`?

ak8abhinay: So, they get inherited without manually writing them again.

data/presets/power/generator/method/solar_thermal_collector/location/roof.json
```diff
@@ -0,0 +1,38 @@
...
+        "generator:type": "solar_thermal_collector",
+        "location": "roof"
+    },
+    "addTags": {
+        "power": "generator",
+        "generator:method": "thermal",
+        "generator:source": "solar",
+        "generator:type": "solar_thermal_collector",
+        "location": "roof"
+    },
```
matkoniecz:
```suggestion
```

this field is not needed at all, right?

data/presets/power/generator/method/solar_thermal_collector.json
```diff
@@ -0,0 +1,33 @@
...
+    "tags": {
+        "power": "generator",
+        "generator:method": "thermal"
+    },
+    "addTags": {
+        "power": "generator",
+        "generator:method": "thermal",
+        "generator:source": "solar",
+        "generator:type": "solar_thermal_collector"
+    },
```
matkoniecz, line 20:
```suggestion
    "tags": {
        "power": "generator",
        "generator:method": "thermal",
        "generator:source": "solar",
        "generator:type": "solar_thermal_collector"
    },
```

but not all `power=generator generator:method=thermal` are such

see https://wiki.openstreetmap.org/wiki/Tag:generator:method%3Dthermal#Solar_power_tower

I would rather do stricter matching

ak8abhinay: You're right!
I had generator:source and generator:type in addTags when they should've been part of the matching criteria in tags since generator:method=thermal alone isn't specific enough. Applying the suggestion.

data/presets/power/generator/method/solar_thermal_collector/location/roof.json
ak8abhinay: since location: roof is already in tags, this whole addTags block was just restating what gets applied automatically. Removing it.

data/presets/power/generator/method/solar_thermal_collector/location/roof.json
```diff
@@ -0,0 +1,31 @@
...
+    "reference": {
+        "key": "generator:method",
+        "value": "thermal"
+    },
```
matkoniecz, line 24:
```suggestion
    "reference": {
        "key": "generator:type",
        "value": "solar_thermal_collector"
    },
```

https://wiki.openstreetmap.org/wiki/Tag:generator:type%3Dsolar_thermal_collector is more specific

data/presets/power/generator/method/solar_thermal_collector.json
```diff
@@ -0,0 +1,29 @@
...
+    "tags": {
+        "power": "generator",
+        "generator:method": "thermal",
+        "generator:source": "solar",
+        "generator:type": "solar_thermal_collector"
+    },
+    "reference": {
+        "key": "generator:method",
+        "value": "thermal"
+    },
```
matkoniecz, line 24:
```suggestion
    "reference": {
        "key": "generator:type",
        "value": "solar_thermal_collector"
    },
```

https://wiki.openstreetmap.org/wiki/Tag:generator:type%3Dsolar_thermal_collector is more specific

data/presets/power/generator/method/solar_thermal_collector/location/roof.json
```diff
@@ -0,0 +1,31 @@
...
+    "terms": [
+        "rooftop solar thermal",
```
matkoniecz:
```suggestion
```

substring of name is not needed

data/presets/power/generator/method/solar_thermal_collector/location/roof.json
```diff
@@ -0,0 +1,30 @@
...
+        "generator:source": "solar",
+        "generator:type": "solar_thermal_collector",
+        "location": "roof"
+    },
+    "reference": {
+        "key": "generator:type",
+        "value": "solar_thermal_collector"
+    },
+    "terms": [
+        "solar hot water panel",
```
matkoniecz, line 27:
```suggestion
        "heating",
        "solar hot water panel",
```

data/presets/power/generator/method/solar_thermal_collector.json
```diff
@@ -0,0 +1,29 @@
...
+        "generator:method": "thermal",
+        "generator:source": "solar",
+        "generator:type": "solar_thermal_collector"
+    },
+    "reference": {
+        "key": "generator:type",
+        "value": "solar_thermal_collector"
+    },
+    "terms": [
+        "solar hot water panel",
```
matkoniecz, line 27:
```suggestion
        "heating",
        "solar hot water panel",
```

data/presets/power/generator/method/solar_thermal_collector.json
```diff
@@ -0,0 +1,29 @@
+{
+    "icon": "fas-solar-panel",
+    "fields": [
+        "{power/generator}"
+    ],
+    "moreFields": [
+        "location"
+    ],
```
matkoniecz, line 9:
```suggestion
    "fields": [
        "{power/generator}"
        "location"
    ],
    "moreFields": [
        "{power/generator}"
    ],
```

manufacturer and other seem relevant, `location` should be more accessible

matkoniecz:
I really hope that is the final round of nitpicking, reviewing and complaining from my side and next action will be merge :)

matkoniecz (changes requested):

data/presets/power/generator/method/solar_thermal_collector.json
ak8abhinay: Agree!

matkoniecz (approved):

matkoniecz:

PR #2171, opened 2026-04-09
Add Barracks Building as a value for building type dropdown

ak8abhinay:
Adds new preset for `building=barracks`

wiki : https://wiki.openstreetmap.org/wiki/Tag:building%3Dbarracks

Fixes #2156


ak8abhinay:
Here is a screenshot of the preset working in the iD editor.

image


Building type dropdown : 

image

matkoniecz:
I am not sure yet should it get own building preset or is dropdown in building type enough.

But tent icon is definitely bad.

ak8abhinay:
> I am not sure yet should it get own building preset or is dropdown in building type enough

Even I wasn't sure initially. But after looking at the building types like `Yurt` which has its own preset as well as dropdown and  even statistics wise where only 944 is the tag usage or  `Hanger Building` , `pigsty` which have both its own preset and dropdown as well.



> But tent icon is definitely bad

Agree. 
Another option we have is : 

image

matkoniecz:
Problem with barracks is that distinct building type is quite dubious and there is much greater risk of people misusing it for the current use. In fact the second icon (with flag) would make it even worse: if soldiers leave and flag is removed then it should be still tagged `building=barracks`. Or it should not be tagged as `building=barracks` in the first place.

Having it as dropdown of building type makes it more clear that it is about building type, not about building use.

In fact, do we have preset for things currently used as barracks?

ak8abhinay:
> do we have preset for things currently used as barracks?

`military=barracks` already exists as a preset in iD.

I totally agree the existing preset covers the situation where one can map an active military barracks and if wants to tags the building on top of that can be best served by the dropdown.

data/fields/building.json
```diff
@@ -4,10 +4,10 @@
     "label": "Building",
     "strings": {
         "options": {
-            "yes": "Unspecified Building Type",
             "allotment_house": "Allotment House",
             "apartments": "Apartment Building",
             "barn": "Barn",
+            "barracks": "Barrack Building",
```
matkoniecz:
```suggestion
            "barracks": "Barracks Building",
```

from my limited nonnative English and check online: s seems to be missing

ak8abhinay:
Building type dropdown :

image

matkoniecz (approved):

PR #2186, opened 2026-04-14
add missing colon in tag key of `plant:output` multiCombo field

tyrasd:
this still works in iD even without the fix, because there is a fallback in place to add the colon if missing, but our convention is to always include the `:` in the `key` property for all multiCombo fields.

### DIFF
--- a/data/fields/plant/output.json
+++ b/data/fields/plant/output.json
@@ -1,5 +1,5 @@
 {
-    "key": "plant:output",
+    "key": "plant:output:",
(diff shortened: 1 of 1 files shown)


matkoniecz (approved):
confirmed by https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema%20multiCombo&type=code

matkoniecz:
screen-2026-04-15-05-37-23

PR #2206, opened 2026-04-22
ignore patch release updates of prettier

matkoniecz:
we have piles and piles of PRs to process - spending time on stuff like #2204 is just silly

see https://github.com/prettier/prettier/blob/main/CHANGELOG.md

code taken from https://github.com/openstreetmap/tile-attribution/edit/main/.github/dependabot.yml where it worked well

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,4 +25,3 @@ updates:
         patterns:
           - "*"
         applies-to: version-updates
-
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,3 +22,6 @@ updates:
         patterns:
           - "*"
         applies-to: version-updates
+    ignore:
+      - dependency-name: "prettier"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -22,3 +22,6 @@ updates:
         patterns:
           - "*"
         applies-to: version-updates
+    ignore:
+      - dependency-name: "prettier"
+        update-types: ["version-update:semver-patch"]
```
FloEdelmann: `prettier` is an `npm` package, not a `github-actions` one, so this is in the wrong section

matkoniecz: @FloEdelmann can you take another look?

FloEdelmann: Fine by me now

tyrasd (approved):

matkoniecz:
no label strictly fits but this fits best

PR #2217, opened 2026-04-27
discourage reporting issues by commenting in closed PRs

matkoniecz:
comments like https://github.com/openstreetmap/id-tagging-schema/pull/1019#issuecomment-4324879974 are easily missed, if that problem is not reported already it should be a new issue in this or iD issue tracker


matkoniecz:
Maybe add also "But do not recreate issues for already rejected ideas."

tyrasd (approved):
👍

PR #2224, opened 2026-05-03
Introduce private swimming pool preset; modify general preset

### LINKED ISSUE #2221: Preset for private swimming pools
My understanding of the situation is: We have the tagging https://wiki.openstreetmap.org/wiki/Tag:leisure=swimming_pool that we use for pools and that does not differentiate between big, public pools and smaller private pools.

The separation is done via `access=private` (or `null`|`yes` for public).

The combination with `access` is preset ~1.8 Million times in the Data https://taginfo.openstreetmap.org/tags/leisure%3Dswimming_pool#combinations

I propose we add a separate Preset for private pools and rework the existing preset for public pools.

Preset currently (in GoMap)
> Image

tordans:
See https://github.com/openstreetmap/id-tagging-schema/issues/2221

**Notes:** 

The general presets has the same name. It is not the _public_ preset, because it does not set any access; And I don't think our "unspecified" terminology applies here, it is just the preset that is not the "private" one.

However, I move the access field to the top to signal it is important. And move `lit` and `length` to more fields because I just don't see them on the same level of importance as the other fields.

The private preset is simplified by design. But given it is also used for private hotel pools some of the fields are still relevant.

### DIFF
--- a/data/presets/leisure/swimming_pool_private.json
+++ b/data/presets/leisure/swimming_pool_private.json
@@ -1,0 +1,35 @@
+{
+    "icon": "fas-swimming-pool",
+    "fields": [
+        "swimming_pool",
+        "access_simple"
+    ],
+    "moreFields": [
+        "name",
+        "lit",
+        "location_pool",
+        "length",
+        "level"
+    ],
+    "geometry": [
+        "area",
+        "point"
+    ],
+    "terms": [
+        "backyard pool",
+        "home pool",
+        "residential pool",
+        "hotel pool",
+        "resort pool",
+        "water"
+    ],
+    "tags": {
+        "leisure": "swimming_pool",
+        "access": "private"
+    },
+    "reference": {
+        "key": "leisure",
+        "value": "swimming_pool"
+    },
+    "name": "Private Swimming Pool"
+}
--- a/data/presets/leisure/swimming_pool_private.json
+++ b/data/presets/leisure/swimming_pool_private.json
@@ -31,5 +31,5 @@
         "key": "leisure",
         "value": "swimming_pool"
     },
-    "name": "Private Swimming Pool"
+    "name": "Swimming Pool (private access)"
 }
--- a/data/presets/leisure/swimming_pool_private.json
+++ b/data/presets/leisure/swimming_pool_private.json
@@ -18,8 +18,8 @@
     "terms": [
         "backyard pool",
         "home pool",
-        "residential pool",
         "hotel pool",
+        "residential pool",
         "resort pool",
         "water"
     ],
--- a/data/presets/leisure/swimming_pool.json
+++ b/data/presets/leisure/swimming_pool.json
@@ -3,12 +3,12 @@
     "fields": [
         "name",
         "access_simple",
-        "lit",
         "location_pool",
-        "length",
         "swimming_pool"
     ],
     "moreFields": [
+        "lit",
+        "length",
         "address",
         "level",
         "gnis/feature_id-US",
@@ -21,7 +21,10 @@
     ],
     "terms": [
         "aquatics",
+        "community pool",
         "dive",
+        "municipal pool",
+        "public pool",
         "water"
     ],
     "tags": {


tordans:
# Testing

## Private

https://pr-2224--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=20.00/34.99138/33.97519&disable_features=boundaries&background=Mapbox&id=w1506892383

- image
- image
- Bildschirmfoto 2026-05-03 um 14 11 52
- Bildschirmfoto 2026-05-03 um 14 11 46
- Bildschirmfoto 2026-05-03 um 14 11 22

## Public

- image

andrewharvey:
I worry this may be misused as private as in it has privacy, or private as in it's privately operated (as opposed to publicly/government operated) or privately owned (as opposed to public/government owned), and not what `access=private` means which is private access (with individual permission) only.

A hotel/resort pool would usually be `access=customers`, a private owned/operated commercial swimming pool open to the general public would be `access=yes`.

At least if a mapper has to select from the Allowed Access field they can see that the field is for "allowed access" and see all the other access values when making the selection, and see the tag hint text when setting it.

andrewharvey:
> However, I move the access field to the top to signal it is important. And move lit and length to more fields because I just don't see them on the same level of importance as the other fields.

I agree with this.

tordans:
@andrewharvey yes, we had the same issue with the "private driveway" PR. 

Do you see a way to resolve this with the tools (manly icon and title) we have?

andrewharvey:
In my experience the biggest issue is residential pools that lack `access=private`, so a preset "Residential Swimming Pool" would be best, though the problem is we don't have a tag to say it's residential, apart from `leisure=swimming_pool` `access=private` within a `landuse=residential` area.

Perhaps the change to move access higher in the fields list would suffice?

I'm not sure...

tordans:
We could make it a bit verbose in the preset title, could we note?

* General preset:
  > Swimming Pool

* Additional preset:
  > Swimming Pool (private access)
  

The same would work for https://github.com/openstreetmap/id-tagging-schema/pull/2225

* General preset:
  > Driveway

* Additional preset:
  > Driveway (private access)
 
---

Update: I just made this change so we can [look at the result in the preview](https://pr-2224--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=20.00/34.99138/33.97519&disable_features=boundaries&background=Mapbox&id=w1506892383)

andrewharvey:
> Swimming Pool (private access)

I think this is clearer and less ambiguous than just "Private Swimming Pool".

Additional presets which specify the preset field values are good where there is a common term that distinguishes them, for example `highway=service` as "Service Road" benefits from the presets "Driveway" for `highway=service + service=driveway` and "Parking Aisle" for `service=parking_aisle".

In the same way that `swimming_pool=swimming/lap` could be "Lap Pool" and `swimming_pool=wave_pool" could be "Wave Pool", but once you start adding "(private access)" variants you open the door to many variations, e.g "Outdoor Lap Pool", "Indoor Lap Pool", "Outdoor Lap Pool (private access)", "Indoor Lap Pool (private access)", etc.

data/presets/leisure/swimming_pool_private.json
```diff
@@ -0,0 +1,35 @@
...
+        "area",
+        "point"
+    ],
+    "terms": [
+        "backyard pool",
+        "home pool",
+        "residential pool",
+        "hotel pool",
+        "resort pool",
+        "water"
```
matkoniecz, line 24:
```suggestion
        "backyard pool",
        "home pool",
        "hotel pool",
        "residential pool",
        "resort pool",
        "water"
```

based on CI busybody linter

data/presets/leisure/swimming_pool_private.json
```diff
@@ -0,0 +1,35 @@
...
+    ],
+    "tags": {
+        "leisure": "swimming_pool",
+        "access": "private"
+    },
+    "reference": {
+        "key": "leisure",
+        "value": "swimming_pool"
+    },
+    "name": "Swimming Pool (private access)"
```
matkoniecz, line 34: @1ec5 I remember you having Strong Opinions on capital letters in labels and it seem to introduce a new style, so flagging just in case

k-yle (approved):
lgtm as discussed earlier

andrewharvey (approved):
I think this is okay.

tyrasd (commented):
Sorry to chime in quite late, but I do have a small concern:

If we start introducing dedicated presets for existing presets that are just _private_, it would potentially open the door to do this for a whole variety of other features: e.g. why not also add `Footpath (private access)`, `Residential Garden (private access)`, `Garage (private access)`, `Entrance (private access)`, etc.? In my opinion introducing this preset is contrary to the existing mapping practices of having the `access` tag as an attribute of the respective features.

Yes, some mappers might be lazy and not set the `access` field on a residential pool (while just selecting it from the preset would be slightly more low effort), but isn't that how OSM works: if the attribute is important for you and missing: just map it (or create a maproulette challenge, etc.)? 🤷 

PS: don't see this as a veto against the PR, this is just my opinion and feel free to override me.

andrewharvey:
My thoughts, in reply to @tyrasd's comment, perhaps this could be seen as something iD itself could improve in the UX rather than via more presets?

Like a better way to re-apply the whole set of tags applied from the last feature, so if you create a swimming pool and tagged it as private, and location outdoor, then the next area you draw you could choose to have all of those tags set on the new way with one click.

Or somehow make it so rather than having to swap the Tags section to text mode and copy pasting tags, users could select a few tags in the Tags list/table view and copy selected tags (JOSM supports this). 

Or again like in JOSM how you can draw a bunch of untagged ways, then "select all new ways" then mass apply a set a preset and fields to them.

Though I realise these are more power user modes, and power users are likely already the ones setting the access tag for pools...

tordans:
>If we start introducing dedicated presets for existing presets that are just private, it would potentially open the door to do this for a whole variety of other features…

Yes, but in my mind that is the core idea of this repo: Find tagging combinations that are important and used a lot and make presets for them in order to improve the overall mapping experience and quality. So on a meta level this thread is not different from other tagging schema questions. 

The reason why I brought this and the private driveway up is, because I think that in those cases the majority of usage should apply `access` and access is a defining part of the preset, not an additional attribute.

Its not like a `surface` on a road but more like a type of object like a `service=driveway` vs. other.

The fact that it uses the `access` key as a differentiating factor – which is a bit unusual – only highlights the need for a preset that helps mappers to map those in a detailed and more correct way.

The tree presets are a good example of this as well. We did not have to make separate presets, but they make mapping trees sooo much easier and nicer and improve the data quality a lot.

Projects like https://bubatzkarte.de/ showed how problematic it is to not be able to distinguish between private and other kinds of pools.

In my book I am leaning towards adding this preset – but for example for the #2225 I am leaning towards not adding it.

---

Also, I think we should decouple the question of how to solve the emerging patterns.
- Are we going in a direction of sub-presets?
- Do we need a stronger priority sub tagging nudge on iD on tags that are more important to add than others?
- The construction-feature in iD is an example of a specific different (code/UX) solution to this topic
- https://github.com/openstreetmap/id-tagging-schema/pull/1207 is an example where the additional sub tag `bicycle_road=yes` is IMO the a defining tag. So we could argue this should get a separate preset as I was exploring in that PR. But today I would rather look into getting https://trafficsigns.osm-verkehrswende.org/ integrated into iD to give visibility to this kind of data in the existing preset…

Issue #236, opened 2021-09-15
Add ownership field


1ec5:
Some presets, such as the `highway` presets, should have an “Owner Type” field, a combo field tied to the [`ownership`](https://wiki.openstreetmap.org/wiki/Key:ownership) key. This field would complement the existing Operator field and reduce misuse of the Access field. For example, a road signposted as a “Private Drive” but with no legal or signposted access restriction would be tagged `ownership=private`. Unlike with the `owner` key, there would be less risk of a mapper overzealously tagging a feature with a private individual’s name.

matkoniecz:
@1ec5 Do you mean displaying it for all roads or where this tag was placed already?

tordans:
Just a ping to the related issue https://github.com/openstreetmap/id-tagging-schema/issues/1304

tordans:
TBH I don't think this tag is there yet in terms of clarity and usage. At least in Germany I never came across it. And I image there would be discussion on when the conditions for adding the tag which are not well defined on the wiki.

It would be great to postpone this until someone took the time to create a more detailed wiki page which likely involves more discussion with the community and at least a proposal draft / RFC IMO.

matkoniecz:
>  And I image there would be discussion on when the conditions for adding the tag which are not well defined on the wiki.

https://wiki.openstreetmap.org/wiki/Key:ownership seems relatively clear?

PR #2225, opened 2026-05-03
Preset for private driveways

tordans:
See https://github.com/openstreetmap/id-tagging-schema/pull/2224

**Notes:**

The general presets has the same name. It is not the _public_ preset, because it does not set any access; And I don't think our "unspecified" terminology applies here, it is just the preset that is not the "private" one.

Other than that nothing much special going on here.

### DIFF
--- a/data/presets/highway/service/driveway_private.json
+++ b/data/presets/highway/service/driveway_private.json
@@ -4,21 +4,19 @@
         "line"
     ],
     "fields": [
-        "{highway/service}"
+        "{highway/service/driveway}"
     ],
     "moreFields": [
-        "{highway/service}"
+        "{highway/service/driveway}"
     ],
     "tags": {
         "highway": "service",
-        "service": "driveway"
+        "service": "driveway",
+        "access": "private"
     },
     "reference": {
         "key": "service",
         "value": "driveway"
     },
-    "terms": [
-        "public driveway"
-    ],
-    "name": "Driveway"
+    "name": "Private Driveway"
 }


tordans:
# Testing

https://pr-2225--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=18.84/34.98989/33.97681&disable_features=boundaries&background=Bing&id=w1508082307

- info-i Bildschirmfoto 2026-05-03 um 14 17 05
- search Bildschirmfoto 2026-05-03 um 14 17 00
- private preste
  Bildschirmfoto 2026-05-03 um 14 16 51 Bildschirmfoto 2026-05-03 um 14 16 31
- general preset
   Bildschirmfoto 2026-05-03 um 14 16 20

matkoniecz (changes requested):
I am still really worried about "privately owned" vs "restricted access" confusion

if this is set in access field it is far more clear

I would lean toward not having such separate preset

tordans:
Yes, no objections on closing this...

Let's see if other feedback is given in the next few days.

Kovoschiz:
On top of `ownership=private` , worse the `access=` is unresolved. Without signs, some think `=destination` , me `=delivery` , as you can't drive and disembark there for no reason. But this may further depends on whether unsolicited advertising are prohibited in that community, which could be an argument for `=private` only selected deliveries allowed. 
I would stop at creating "Privately owned Driveway" for `ownership=private` (to let applications decide how to penalize it, not encouraging users to blindly set or judge `access=` without knowledge)

tordans:
@Kovoschiz IMO it is established tagging to map private/home driveways as access=private which is reflected IMO in the high usage numbers. 
Delivery routing apps will already have a way to handle those (eg include in route for last mile part but not for pass through or similar). 
If there is a different, documented a d established way please link those here for reference.

matkoniecz:
`access=delivery` would block access by homeowner, right?

`=destination` changes nothing if transit is not possible (as those ban transit only) - so only cases where single property has multiple driveways that can serve as shortcut (entering with one, leaving with another) are affected by setting it

or where one could otherwise walk on driveway and leave on some path going further

Kovoschiz:
@tordans  `access=private` was started by TIGER, and continued by users (and maybe even corporate organized edits, I forget) https://wiki.openstreetmap.org/wiki/TIGER_to_OSM_Attribute_Map#Road_as_other_thoroughfare https://osmus.slack.com/archives/C1US5SFUH/p1700365549104269?thread_ts=1700255554.134609&cid=C1US5SFUH

USA commonly discusses this 
* https://osmus.slack.com/archives/C2VJAJCS0/p1753717939917429
* https://osmus.slack.com/archives/C2VJAJCS0/p1642959602224400
*  https://osmus.slack.com/archives/C2VJAJCS0/p1744812368791899
* https://osmus.slack.com/archives/C2VJAJCS0/p1709335987157929

They have ~60% of the `access=private` on `service=driveway` https://taginfo.geofabrik.de/north-america:us/tags/service=driveway#combinations https://taginfo.openstreetmap.org/tags/service=driveway#combinations

Wiki

* https://wiki.openstreetmap.org/wiki/Skunked_tags#Examples
* https://wiki.openstreetmap.org/wiki/Tag:access=private#Clarification_of_the_term_“private”
* <https://wiki.openstreetmap.org/wiki/Talk:Tag:service=driveway#Why_access=private?>
* https://wiki.openstreetmap.org/wiki/US_Import_Problems#Louisiana
* https://wiki.openstreetmap.org/wiki/Vermont#Road_Ownership_and_Access
* https://wiki.openstreetmap.org/wiki/New_Hampshire#Ownership_and_Access

1ec5:
> IMO it is established tagging to map private/home driveways as access=private which is reflected IMO in the high usage numbers.

This is the established tagging for what _you_ want to map, but that isn’t called a “private driveway”. Calling any preset “Private Driveway” will exacerbate misuse of `access=private` when `ownership=private` is more correct. #236 tracks adding a field for `ownership=*` that would at least clarify the situation a little bit.

Kovoschiz:
@matkoniecz `=delivery` shouldn't block staff and official use. Same for homes?  
I don't agree with `=destination` , only mentioning this opinion. I haven't looked into it fully.

PR #2229, opened 2026-05-07
stop to suggest or to automatically add `surface=wood` to `bridge=boardwalk`

### LINKED ISSUE #2209: Problematic suggestion add `surface=wood` to boardwalks
This may not rise to the level of bug. So apologies if this is miscategorized.

**Describe the bug**
For any boardwalk that does not have a `surface=*` tag, an incomplete tags warning is displayed, suggesting that `surface=wood` should be added. 

**To Reproduce**
1. Select a boardwalk that does not have a surface tag

**Expected behavior**
No warning is displayed, as many board walks have wood surfaces but others do not, and specifying a surface should not be a required element. Displaying a warning if no surface is specified may lead to the addition of erroneous tags. 

**Screenshots**

Image

**Additional context**
data/presets/highway/path/boardwalk.json is I think where this is specified.

matkoniecz:

see #2209

### Related issues

fixes #2209

### Links and data

**Relevant OSM Wiki links:**
- …

**Relevant tag usage stats:**
> …

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/data/presets/highway/path/boardwalk.json
+++ b/data/presets/highway/path/boardwalk.json
@@ -13,11 +13,6 @@
         "highway": "path",
         "bridge": "boardwalk"
     },
-    "addTags": {
-        "highway": "path",
-        "bridge": "boardwalk",
-        "surface": "wood"
-    },
     "reference": {
         "key": "bridge",
         "value": "boardwalk"


matkoniecz:
by "bot edit" I mean here distributed edit performed by pushing people to perform tag change

I believe it fulfills "bot edit" definition and falls under https://wiki.openstreetmap.org/wiki/Automated_Edits_code_of_conduct

tordans:
Please don't frame this kind of thing as bot edit. It harms this project as well as any project that does human guided data improvements. 

I have nothing against removing the tagging, though.

matkoniecz:
the problem (and what I am trying to describe here) is that some fraction of users will mindlessly accept whatever validator suggests to change

even if vast majority is using their brain, and do not apply invalid changes - some are effectively performing bot edit due to not thinking at all, and even if nearly everyone correctly avoid applying suggestion sooner or later you will get someone who accepts all suggestions

problem is made worse as some interpret it as "osm told me to do this" - with say Osmose user needs to actively look for it at least

what would be better term here? if not "bot edit"?

tordans:
> what would be better term here? if not "bot edit"?

This repos goal is to define tag combinations that help users set the right combination of tags for a given feature. For me, this is a simple case of: When the preset was defined, we thought that a vast majority of those features use this tag combination. Now we re-evaluate this decision and are leaning towards something else. So we edit the preset to reflect this, keep (or add) the field for surface; remove the default-tag for surface … and with that the "missing tag" suggestion goes away.

There is nothing special going on here and IMO we don't need to find a name for it; especially not one that is associated with harmful mass editing of OSM – which is _not_ what is happening here.

matkoniecz:
> which is not what is happening here.

I guess that it is root of disagreement, I was fixing some result of buggy replacements - and there was decent chunk of ones completely invalid that was applied without checking.

I am sure that it extends to less blatantly invalid. And there is in fact plenty of harmful mass editing in cases which are not 100% applicable (and if they are, then it should be done as bot edit to save human time)

tyrasd (approved):
I think this is reasonable. Yes, it would help if the [wiki](https://wiki.openstreetmap.org/wiki/Tag:bridge=boardwalk) would have a better definition of when actually to use `boardwalk` instead of the generic `bridge=yes`, but I do also think that is becomes more and more common to see "boardwalks" (in the rough sense of shallow bridges over "muddy" terrains or still waters) to be made out of more durable materials like metal or special plastics. Yes, most are likely still constructed using a wooded surface, but I think it is fine to let the mappers to select the respectively correct surface value.

tordans:
@matkoniecz I don't want to move the discussion in https://github.com/ideditor/schema-builder/pull/259#issuecomment-4448803860 more off topic but those things are not the same and I really hope we can find a way to discuss those things with the nuance they require and deserve. What we are talking about here is a (strong) suggestion to users which they manually apply or not. On the other hand [`discarded.json`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/discarded.json) is a collection of keys (and now tags) that iD removes automatically without user intervention or even knowledge (based on triggers that are defined in iD; different editors may use this file differently or not at all). I hope we agree that those are very different things and that the context, business logic and user involvement is key evaluating the feature (in terms of its bot-edit-ness).

matkoniecz:
It could be interesting topic for iD meeting in June. 

And yes,  I agree that unstoppable invisible to user action is more powerful and needs to be treated accordingly.

PR #2233, opened 2026-05-08
make social media fields support URLs and usernames

### LINKED ISSUE #1795: Support facebook username link
Now it is not possible to visit fb page from iD if there is only facebook username without full link.

Image

k-yle:
Closes #1795

Follow up to https://github.com/openstreetmap/iD/pull/12306, see that PR for details.

By switching these fields to `type=identifier`, the link button will support both tagging styles (full URL or just username)

### DIFF
--- a/data/fields/contact/instagram.json
+++ b/data/fields/contact/instagram.json
@@ -1,6 +1,8 @@
 {
     "key": "contact:instagram",
-    "type": "url",
+    "type": "identifier",
+    "pattern": "^.+$",
+    "urlFormat": "https://instagram.com/{value}",
     "label": "Instagram account name or URL",
     "placeholder": "https://www.instagram.com/openstreetmap/"
 }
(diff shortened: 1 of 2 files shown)


k-yle:
## Test-Documentation

link works in all cases:

image
image
image

link does not work if the field is blank:

image

data/fields/contact/instagram.json
```diff
@@ -1,6 +1,8 @@
 {
     "key": "contact:instagram",
-    "type": "url",
+    "type": "identifier",
+    "pattern": "^.{2,}$",
```
matkoniecz: I am a bit confused by this pattern. Why reject single signs?

Or is it doing more and I do not see it?

k-yle: yeah, i guess we could just allow any non-empty string

matkoniecz (approved):

PR #2240, opened 2026-05-10
Add "log" as term for fallen tree barrier=log preset

### LINKED ISSUE #2228: Advertising sign appears before Fallen tree for "log"
Searching "log" for way node shows "Advertising Sign" preset before "Fallen Tree Obstruction".

Image

I believe this is since #2063

I had to check the source to figure out it's because it has "logo" in `terms`, whereas fallen tree doesn't have "log" in `aliases` or `terms` at all and so I assume it always matched before by key's value at some lower score.

I don't really know how search matching works for making an actual PR, but I assume one would just add "log" to `aliases` or `terms`?

en-GB

RudyTheDev:
As per title. Closes #2228.

### DIFF
--- a/data/presets/barrier/log.json
+++ b/data/presets/barrier/log.json
@@ -15,7 +15,8 @@
         "barrier": "log"
     },
     "terms": [
-        "downed tree"
+        "downed tree",
+        "log"
     ],
     "name": "Fallen Tree Obstruction"
 }


RudyTheDev:
image

matkoniecz (approved):

PR #2025, opened 2026-03-10
Add field for `content` tag to `man_made=silo` preset

matkoniecz:
"grain elevator" and "elevator" should find it

it is not only for crops, content= is also popular field

(maybe separate entry for grain elevator would make sense? but I am not sure how agricultural non-grain ones are called)

### Links and data

**Relevant OSM Wiki links:**
- …

**Relevant tag usage stats:**
> …

### Checklist and Test-Documentation Template

<details><summary>Read on to get your PR merged faster…</summary>

Follow these steps to test your PR yourself and make it a lot easier and faster for maintainers to check and approve it.

**This is how it works:**
1. After you submit your PR, the system will create a preview and comment on your PR:
   > 🍱 Your pull request preview is ready.
   If this is your first contribution to this project, the preview will not happen right away but requires a click from one of the project members. We will do this ASAP.

2. Once the preview is ready, use it to test your changes.

3. Now copy the snippet below into a new comment and fill out the blanks.

4. Now your PR is ready to be reviewed.

```
## Test-Documentation

### Preview links & Sidebar Screenshots

<!-- Use the preview to find examples, select the feature in question and **copy this link here**.
     Find examples of nodes/areas. Find examples with a lot of tags or very few tags. – Whatever helps to test this thoroughly.
     Add relevant **screenshots** of the sidebar of those examples. -->

<!-- FYI: What we will check:
     - Is the [icon](https://github.com/ideditor/schema-builder/blob/main/ICONS.md) well chosen.
     - Are the fields well-structured and have good labels.
     - Do the dropdowns (etc.) work well and show helpful data. -->

### Search

<!-- **Test the search** of your preset and share relevant **screenshots** here.
     - Test the preset name as search terms.
     - Also test the preset terms and aliases as search terms (if present). -->

### Info-`i`

<!-- **Test the info-i** for your fields and preset and share relevant **screenshots** here.
     The info needs to help mappers understand the preset and when to use it.
     [Learn more…](https://github.com/openstreetmap/id-tagging-schema/blob/main/CONTRIBUTING.md#info-i)
 -->

### Wording

- [ ] American English
- [ ] `name`, `aliases` (if present) use Title Case
- [ ] `terms` (if present) use lower case, sorted A-Z
<!-- Learn more in https://github.com/openstreetmap/id-tagging-schema/blob/main/GUIDELINES.md#2-design-the-preset -->
```

</details>

PR #2241, opened 2026-05-10
Add strings for `Content` field, add some new values

Geo-2695:

Added commonly used translatable value strings to the content field, representing contents in storage tank container.
After analysis, `autoSuggestions=false` was found unnecessary here, so it has been removed.

### Related issues

None

### Links and data

**Relevant OSM Wiki links:**

https://wiki.openstreetmap.org/wiki/Key:content

**Relevant tag usage stats:**

https://taginfo.openstreetmap.org/keys/content#values

### DIFF
--- a/data/fields/content.json
+++ b/data/fields/content.json
@@ -2,15 +2,17 @@
     "key": "content",
     "type": "combo",
     "label": "Content",
-    "options": [
-        "silage",
-        "water",
-        "oil",
-        "fuel",
-        "slurry",
-        "gas",
-        "manure",
-        "sewage"
-    ],
-    "autoSuggestions": false
+    "strings": {
+        "options": {
+            "fuel": "Fuel",
+            "gas": "Gas",
+            "hot_water": "Hot Water",
+            "oil": "Oil",
+            "sewage": "Sewage",
+            "silage": "Silage",
+            "slurry": "Liquid manure",
+            "water": "Water",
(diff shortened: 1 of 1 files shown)


Geo-2695:
### Testing
Screenshot_20260510_201412

Screenshot_20260510_201504

matkoniecz:
> After analysis, autoSuggestions=false was found unnecessary here, so it has been removed.

Was not sure here, but I have no opinion in the end

Added before history was tracked: https://github.com/Geo-2695/id-tagging-schema/commits/f5aa70182c631ba18a0ef83ecad821138c43a42b/data/fields/content.json

matkoniecz (approved):

data/fields/content.json
```diff
@@ -2,15 +2,19 @@
...
-        "fuel",
-        "slurry",
-        "gas",
-        "manure",
-        "sewage"
-    ],
-    "autoSuggestions": false
+    "strings": {
+        "options": {
+            "crop": "Crop",
```
matkoniecz, line 7: is `crop` stored in either of that rather than in https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dsilo ?

matkoniecz: though maybe silo should also start showing that field, see https://github.com/openstreetmap/id-tagging-schema/pull/2025

PR #2245, opened 2026-05-11
create a maritime access=* field

### LINKED ISSUE #2226: Waterway access field
### OSM Tag(s)

canoe=* boat=* motorboat=*

As a Field

### Label

Access

### Aliases

_No response_

### Terms

_No response_

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Key:boat

### Status of the Tag

De Facto

### Usage of the tag

500.000+

### Suggested Icon

_No response_

### Replaces other Tag?

_No response_

### Regional Tag?

global

### Further Information

I suggest adding a access field, similar to the one that already exists for highways.
This would allow for setting the access for various water vehicles. It should be added as an field to:

- [waterway=canal](https://wiki.openstreetmap.org/wiki/Tag:waterway=canal)
- [waterway=river](https://wiki.openstreetmap.org/wiki/Tag:waterway%3Driver)
- lock=yes

The multiple dropdowns, also used in the highway access field, seem ideal  for this.

I would suggest these vehicles:

- [canoe](https://wiki.openstreetmap.org/wiki/Key:canoe)=*
- [boat](https://wiki.openstreetmap.org/wiki/Key:boat)=*
- [motorboat](https://wiki.openstreetmap.org/wiki/Key:motorboat)=*
- [ship](https://wiki.openstreetmap.org/wiki/Key:ship)=*
- [sailboat](https://wiki.openstreetmap.org/wiki/Key:sailboat)

k-yle:
Closes #2226

This is just a test - I would rather proceed with https://github.com/openstreetmap/iD/pull/12011 but this PR could also be merged first


k-yle:
image

works nicely but needs code changes in iD, so we might as well wait for https://github.com/openstreetmap/iD/pull/12011

k-yle:
re-opening per discussion in https://github.com/openstreetmap/id-tagging-schema/issues/2226#issuecomment-4792528759

data/presets/waterway/stream.json
```diff
@@ -4,7 +4,8 @@
         "name",
         "structure_waterway",
         "width",
-        "intermittent"
+        "intermittent",
+        "access_maritime"
```
matkoniecz: especially for `waterway=stream` it seems that it should be in `moreFields`

at least from perspective of Poland where we (still) have many many many streams, with vast majority not being usable even for canoes

are there any streams anywhere where this would be truly relevant

wait, would it even work in moreFields?

k-yle:
now that schema-builder v7 is released, this works nicely:

image

I'm still not sure if we should focus on this first, or https://github.com/openstreetmap/iD/pull/12011 first

tordans:
> I'm still not sure if we should focus on this first, or openstreetmap/iD#12011 first

This PR looks like a clean and well contained improvement. Don't see a reason not to do it now.

data/fields/access_maritime.json
```diff
@@ -0,0 +1,15 @@
...
+        "canoe",
+        "kayak"
+    ],
+    "reference": {
+        "key": "access"
+    },
+    "type": "access",
+    "label": "Allowed Access",
+    "placeholder": "Not Specified",
+    "stringsCrossReference": "{access}"
```
tordans, line 14: If we keep the maritime strings here we can set up the access field in iD to only just what we have here? (But still need a way for specify the default list, right? 🤔)

tordans (approved):

data/fields/access_maritime.json
k-yle: `stringsCrossReference` applies to both `strings.options` and `strings.types`, and we want to avoid [repeating this big block of text](https://github.com/openstreetmap/id-tagging-schema/blob/7eebbf824bcf96f947f3f03adb3890fef2aa3990/data/fields/access.json#L26-L67). So i think translations for every mode of transport should be kept in the main `access.json` field.

---

and yeah, defaults vs the full list of keys is something we need to think about in https://github.com/openstreetmap/iD/pull/12011

data/presets/waterway/stream.json
k-yle: sure, moved to `moreFields`

matkoniecz:
1)

Got confused what "All Paddlecraft" may be about as opposed to canoes, followed wiki link from it to https://wiki.openstreetmap.org/wiki/Key:access

there is no word "Paddlecraft" on that page


2)

I worry a bit about people adding random and wrong guesses, like they do now with `horse=` field

Personally I have no idea about legal access status, even on some rivers where I was canoeing

I guess it is not a blocker but we may move it to moreFields in future

data/fields/access.json
```diff
@@ -15,6 +15,9 @@
     "strings": {
         "types": {
             "access": "All",
+            "boat": "Boat",
+            "canoe": "All Paddlecraft",
```
k-yle, line 19: _(replying to https://github.com/openstreetmap/id-tagging-schema/pull/2245#issuecomment-4935655409 in a thread)_

this was discussed recently in https://github.com/openstreetmap/id-tagging-schema/pull/1473#discussion_r3127808844

@quincylvania what do you think of dropping the `kayak=*` field, and labelling `canoe=*` as "Canoe/Kayak" instead of "Paddlecraft"?

Alternatively, I believe canoeing is much more common in the US, so maybe the `en-US` label could be "Canoe" and the `en-NZ`/`AU`/`SG` labels could be "Paddlecraft" or "Canoe/Kayak"?

quincylvania: Speaking from a North American perspective, canoe and kayak access is pretty much always identical. It's sort of like an SUV versus sedan situation… one might be better for certain types of trips, but very rarely are they restricted differently.

I would probably make the label "Canoe/Kayak" or maybe just "Canoe" for the US translation. Linguistically speaking, "Paddlecraft" is sort of a made up word that most people wouldn't recognize.

It would be safe to drop the `kayak` field. I documented the `kayak` wiki page since it had some usage and it's theoretically useful in edge cases, but it's nearly always redundant to ask people to tag it. Unless there is a region where `canoe` and `kayak` often differ.

quincylvania:
This field is cool! I do a lot of watercraft access mapping and the three main tags I would include are:

- `canoe`
- `motorboat`
- `sailboat`

The tag `boat` is common and should be included as well, but honestly it's not the most useful tag since boats range so widely in size and function.

data/fields/access.json
andrewharvey: At first glance I didn't recognise the term _paddlecraft_ and thought is this some new kind of craft, different to a canoe or kayak? It might be confused with a paddlewheeler like https://www.nepeanbelle.com.au/.

Second reading, after looking at all the comments, interpreting it as _paddle-craft_ as a craft you paddle, I suppose you paddle a canoe, paddle a kayak, paddle a "standup paddleboard (SUP)", paddle a dragon boat... but you can also paddle with your arms on a surfboard, or on a surf rescue board, so it depends on what the tag is meant to be used for, which leads to what I think the main issue is...

...is that https://wiki.openstreetmap.org/wiki/Key:access lists (and https://wiki.openstreetmap.org/wiki/Key:canoe confirms) the `canoe=*` access tag as

> boats without sail or motor, such as small dinghies, canoes, kayaks, etc.

which I think is problematic as it's describing access for small non-motorised, non-sail watercraft, but using the term of one specific type of watercraft. A bit like saying `bicycle=no` as an access value means any non-motorised vehicle including a skateboard or kick-scooter, which we don't do.

I realise it's not the place of the tagging schema to dictate tagging, but if people think the tagging should be improved (which I think it should at least be explored further) maybe we hold off on the field here until then?

andrewharvey: _pedalboat_ _paddleboat_  _pedalcraft_ _paddlecraft_ are all to easy to mix up, especially for dyslexics

quincylvania: @andrewharvey I broadly agree re: the "paddlecraft" label for the `canoe` tag. From what I've seen in OSM it's usually safest to make the English label match the tag value more or less exactly, or else you get semantic drift as different people interpret the label differently.

matkoniecz (changes requested):
Got confused what "All Paddlecraft" may be about as opposed to canoes, followed wiki link from it to https://wiki.openstreetmap.org/wiki/Key:access

there is no word "Paddlecraft" on that page

-----------------

I worry a bit about people adding random and wrong guesses, like they do now with horse= field

Personally I have no idea about legal access status, even on some rivers where I was canoeing

I guess it is not a blocker but we may move it to moreFields in future

k-yle:
thanks for the feedback everyone
- the _Paddlecraft_ label is gone, replaced with _Canoe_ , which I think is a good label for `en-US` given the comments above
- `kayak=*` was removed, it's too similar to `canoe=*`, there's no need for both
- `sailboat=*` added

the last open question is where `boat=*` should be replaced with `motorboat=*`, per @quincylvania's [comment above](https://github.com/openstreetmap/id-tagging-schema/pull/2245#issuecomment-5145639138). That seems logical to me, especially given that [CEVNI sign A.12](https://commons.wikimedia.org/wiki/File:A.12_Fahrverbot_f%C3%BCr_Fahrzeuge_mit_Maschinenantrieb.svg) means `motorboat=*`, not `boat=*`. However, `motorboat=*` is not used as much as `boat=*`

quincylvania:
@k-yle Thanks for your continued work on this. I'd say including `motorboat` would be expected. At least in the US, "motors allowed" versus "no motors" is the most common access restriction on waterways.

`boat` is tricky since it already has high usage and people would probably expect to see existing values. The vast majority is `boat=no`, which is useful since it implies `motorboat=no`, `canoe=no`, etc., unless those are otherwise tagged. However, `boat=yes` isn't very useful since consumers can't be sure which types of boats the mapper is talking about. It's a lot like `vehicle` in this way, which I don't think iD supports right now.

matkoniecz:
> At least in the US, "motors allowed" versus "no motors" is the most common access restriction on waterways.

also in Poland, AFAIK

> It's a lot like vehicle in this way, which I don't think iD supports right now.

though actually at least Europe would benefit from `vehicle`, the most common ban traffic sign maps to `vehicle=no` (or `vehicle=private` if used with typical exemptions). Though that is a different story and we have issue for it.

matkoniecz (commented):
May be overdoing things, but maybe set tag prerequisite to suppress it on `usage=spillway` ?

screen-2026-09-05-15-03-24

default view is confused by "All" defaulting to `yes` and other to "Not specified"

if nothing is set and "All" defaults to `yes`, then surely other should be `yes`, or "All" set also to "Not specified"?

quincylvania:
I would not interpret `access=yes` on a `waterway` to imply `motorboat=yes` or `canoe=yes`. Though I would interpret `access=no` to imply `canoe=no`. I guess this is different from `access=yes` on a `highway` feature. The major difference is that there is a very wide variety of physical characteristics of waterways since mostly they are not designed by humans for transport. So I might read `access=yes` on a stream to mean legal access is allowed but that doesn't mean it's physically possible to use a boat there.

quincylvania:
On further inspection the field label "Allowed Access" is a bit misleading since physical conditions are just as relevant for boat tagging.

matkoniecz:
Physical possibility is (mis)used already for road access...

For things like kayaks it gets more complicated, as for some tree trunk in river will be impassable obstruction and for some may be an attraction. The same for waterfalls.

matkoniecz:
In general I have doubts is `access=yes` on waterway making any sense at all.

andrewharvey:
> image

`waterway=river` or `waterway=stream` + `foot=*` or `swimming=*` aren't significant enough for taginfo to show anything, but I'm not sure if this "All" is going to be misused with someone tagging "All: no" because it's no boats, canoes, etc. but overlooking that swimming is allowed, and therefore can't be All: no (`access=no`).

> In general I have doubts is access=yes on waterway making any sense at all.

I would interpret it as "not `access=no`" meaning you can access it by at least one mode but it's unclear which specific modes are permitted.

andrewharvey:
Also it doesn't need to be in this PR, but should this maritime access field apply to any of the `water=*` values?

data/presets/waterway/stream.json
```diff
@@ -7,6 +7,7 @@
         "intermittent"
     ],
     "moreFields": [
+        "access_maritime",
```
andrewharvey: Sorry, another comment...

https://wiki.openstreetmap.org/wiki/Tag:waterway=stream says that `waterway=stream` can usually be jumped across. Even if you relax this to can be walked across without swimming, like something knee height, should `access_maritime` which asks about motorboat access apply here?

matkoniecz:
> I would interpret it as "not access=no" meaning you can access it by at least one mode but it's unclear which specific modes are permitted.

for me it would be one more reason to never ever use it or encourage or support

data/presets/waterway/stream.json
k-yle: I dind't realise the wiki claimed:

> "the commonly accepted rule for OpenStreetMap is that a `stream` can be jumped across by an active, able-bodied person)"

that would mean everything should `river` apart from tiny trickles of water. Even the photo on the wiki page disagrees with that claim... sure, you could jump the narrowest part of that photo, but not the rest of it...

---

but in any case, i'll remove it from stream from now, lets start with a few (hopefully) uncontroversial presets.

k-yle:
> On further inspection the field label "Allowed Access" is a bit misleading since physical conditions are just as relevant for boat tagging.

do you have any copywriting suggestions? How about just the word `Access`?

> for me it would be one more reason to never ever use it or encourage or support

sounds good, I've removed `boat=*` and kept `motorboat=*`, which matches iD's current behaviour for `vehicle`/`motor_vehicle`. Once we have https://github.com/openstreetmap/iD/pull/12011, we can reconsider both `boat=*` and `vehicle=*`.

> May be overdoing things, but maybe set tag prerequisite to suppress it on `usage=spillway` ?

actually, there are [spillways where you can paddle](https://en.wikipedia.org/wiki/Tacen_Whitewater_Course#/media/File:Tacen_Whitewater_Course_1.jpg) ;) 

> Also it doesn't need to be in this PR, but should this maritime access field apply to any of the `water=*` values?

Probably, but lets do in a different PR, if this one is released without complaints

PR #2249, opened 2026-05-12
Enable segregated surface field for cycle & foot presets

RudyTheDev:
This enables https://github.com/openstreetmap/iD/pull/12332 to show a custom grouped combo box for the preset. Please see details there.

<s>This requires https://github.com/ideditor/schema-builder/pull/302 update to support the new field and field's key.</s> _Edit: now part of this PR after schema repo merge._

### DIFF
--- a/data/fields/surface_segregated.json
+++ b/data/fields/surface_segregated.json
@@ -1,0 +1,25 @@
+{
+    "key": "surface",
+    "keys": [
+        "surface",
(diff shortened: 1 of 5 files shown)


matkoniecz:
See also https://github.com/ideditor/schema-builder/issues/282 - it may be possible to do it in more general way, without introducing very specified fields for specific tags

Issue #2049, opened 2021-04-18
Decrease the number of dropdowns


owiecc:
In many contexts a dropdown is not the best choice. If the default choice list is very short, buttons would be better in my opinion. As an example: direction of a yield sign. 

<img width="372" alt="Screenshot 2021-04-18 at 12 25 13" src="https://user-images.githubusercontent.com/6896639/115142429-fe052300-a041-11eb-8542-bb38f5d76b94.png">

A direction can be chosen with just one click and reversed also with one click.

matkoniecz:
Note that in meantime new option appeared: `"both": "Both / All"`

@FloEdelmann are you interested in scanning repo for `"type": "combo",` fields where option count is really low? (you made some scripts like that before, obviously feel free to ignore this ping - if you are note interested just ignore it or let me know)

matkoniecz:
New UI can be tested at https://github.com/openstreetmap/id-tagging-schema/pull/2081

I am curious is it better. Will take more space, and with three options it may be less obviously superior.

owiecc:
Wouldn't modern CSS allow to [style the radio buttons as labels](https://stackoverflow.com/a/75252962), as in my original picture? I do like the radio buttons HTML elements because they are semantically right for the job.

FloEdelmann:
I quickly adapted one of my preexisting scripts:

<details>
<summary>combo2radio.js</summary>

```js
#!/usr/bin/env zx

// NOTE: uses https://github.com/google/zx and has to be called from id-tagging-schema root directory

const fieldJsonFiles = await glob('./data/fields/**/*.json', {
    gitignore: true,
    absolute: true,
})

const files = {}

for (const file of fieldJsonFiles) {
    const field = JSON.parse(fs.readFileSync(file, 'utf-8'))

    if (field.type !== 'combo' || field.customValues) {
        continue
    }

    const optionsCount = field.strings && field.strings.options
        ? Object.keys(field.strings.options).length
        : field.options
            ? field.options.length
            : 0

    if (optionsCount === 0 || optionsCount > 9) {
        continue
    }

    files[optionsCount] ??= []
    files[optionsCount].push(file)
}

console.log(files)
```

</details>

These are the results:

<details>
<summary>Click to expand</summary>

```js
{
  '2': [
    '/home/flo/www/id-tagging-schema/data/fields/direction_clock.json',
    '/home/flo/www/id-tagging-schema/data/fields/direction_vertex_dual.json',
[...]

PR #2251, opened 2026-05-13
Remove `autoSuggestions`/`customValues` from radio fields

FloEdelmann:

Those properties are no longer needed for `radio` fields and were likely missed while converting the fields from `combo`.

### Related issues

See #2049.

CC @matkoniecz


matkoniecz (approved):

matkoniecz:
maybe schema builder should detect this as a lint problem?

FloEdelmann:
the "preset" label seems wrong; it should probably be "field"

PR #2255, opened 2026-05-14
update to schema-builder v7

### LINKED ISSUE #196: Major release planning version 7 (breaking changes)
(This issue is a stub for now; it will be updated).

- [ ] https://github.com/ideditor/schema-builder/pull/87

k-yle:
Closes https://github.com/ideditor/schema-builder/issues/196

we still need to wait for the npm release, this is just preparation to make sure it works.

I've committed the changes after running `npm run dist`, to show what files will immediately be changed. 

most of the diff is re-ordering noise caused by https://github.com/ideditor/schema-builder/pull/279


matkoniecz:
We also should wait for iD to adapt to v7 changes,  right?

Or is everything done now?

matkoniecz:
@k-yle would it be possible to split into two commits? One doing changes to human-editable parts and second running build script?

CONTRIBUTING.md
```diff
@@ -88,7 +88,7 @@ The following `npm` commands are used in this repository:
 * `npm test` – validates the source data
 * `npm run build` – validates the source data and builds some files which are used during development (e.g. strings to be supplied to the translation platform)
 * `npm run dist` – validates the source data and compiles output files for iD
-* `npm run translations` – fetches translations from transifex and compiles the translations files for iD
+* `npm run dist -- translations` – fetches translations from transifex and compiles the translations files for iD
```
tyrasd, line 91: If I remember correctly, the change in  6fc8f4dfe31cd2919422226a6296754cfb2e21fe only  added the `npm run dist translations` in order to be able to do a `npm run dist` without updating the translations, but still be able to do either `npm run translations` separately as well as `npm run dist translations` to do everything at once for improved performance.

```suggestion
* `npm run translations` – fetches translations from transifex and compiles the translations files for iD
* `npm run dist translations` – validates the source data, fetches translations from transifex, and  and compiles output files for both for iD
```

(and keep `scripts/translations.js` around along with the `npm run translations` command in package.json)

k-yle:
> @k-yle would it be possible to split into two commits? One doing changes to human-editable parts and second running build script?

sure, done. To do that, I had to force push over martin's commits (preserving those changes of course).

CONTRIBUTING.md
k-yle: > [...] but still be able to do either `npm run translations` separately [...]

in what situations would this command be useful?

This partial update command was removed in https://github.com/ideditor/schema-builder/pull/281#issue-4101266577, because fetching and generating translations now requires knowledge of the source files, as of https://github.com/ideditor/schema-builder/pull/281.

So it would be a bit wasteful to have a command that only updates `translations` and throws away that other changes. However, if there is a real use-case for this, then we can definitely add back that command. but as far as i can tell, it's not required

matkoniecz (approved):
though release should be coordinated

- done when iD is capable of supporting v7
- after previous release (with first v7 release made immediately after containing only and solely upgrade to v7 schema builder and no other changes)
- with translation update rerun before release is made

(any other concerns?)

PR #2271, opened 2026-05-16
Add deprecated support for  `storage=*`

Geo-2695:

Like  `contents=*` , these are non-standard tags and not recommended for use.
It is advised to use  `content=*`  instead.

### Related issues

None

### Links and data

**Relevant OSM Wiki links:**
[Key:content](https://wiki.openstreetmap.org/wiki/Key:content)

**Relevant tag usage stats:**
[Key:storage](https://taginfo.openstreetmap.org/keys/storage)

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1625,6 +1625,10 @@
     "replace": {"station": "train", "train": "yes"}
   },
   {
+    "old": {"storage": "*"},
+    "replace": {"content": "$1"}
+  },
+  {
     "old": {"stream": "intermittent", "waterway": "stream"},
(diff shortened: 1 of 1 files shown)


Geo-2695:
### Testing
Screenshot_20260516_114158

matkoniecz:
it seems to me that either there is agreement from community to mass move it and bot edit can be done or there is no such agreement and iD should not push it either

from quick look at taginfo it looks like it was mostly mass added AND top used value should be removed rather than migrated

Kovoschiz:
This should not be deprecated for `landuse=industrial` + `industrial=storage`

PR #2276, opened 2026-05-19
Add pedagogy as extra field for schools/preschools

RudyTheDev:
Add [`pedagogy` field](https://wiki.openstreetmap.org/wiki/Key:pedagogy) to more fields for schools, preschools <s>and colleges</s>. For now, it lists 3 of the most popular and tagged ones - `montessori`, `waldorf`, `dalton` (skipping others, because it's just more translation work for a tiny number of usages, at least at the moment) plus the default `mainstream` option.

image

The field is [approved on wiki](https://wiki.openstreetmap.org/wiki/Proposal:Key:pedagogy) and documented.

The [total uses](https://taginfo.openstreetmap.org/keys/pedagogy#values) of this are only about 2200 with about 40% mainstream, 40% one of the well-known models and the rest random assortment of values. So about 1000 of "real" values. This isn't a lot, but it's also growing fairly steadily over the last few years since approval (even though iD doesn't even offer the field). And for example many schools that are mapped and have literally stuff like "Montessori" in the name aren't yet tagged with `pedagogy`.

From my mild research on the subject, it's quite popular for preschools to have an educational philosophy/model, especially private ones. This also carries over to many schools. Even some colleges retain these. These institutions make it very clear when they follow a specific model including literally in the names. For example, it's estimated there are some 15-20K Montessori pre/schools worldwide, some 2-3K Waldorf/Steiner and <1K Dalton, so these are the ones I chose.

### DIFF
--- a/data/fields/pedagogy.json
+++ b/data/fields/pedagogy.json
@@ -1,0 +1,21 @@
+{
+    "key": "pedagogy",
+    "type": "combo",
+    "label": "Pedagogy",
+    "strings": {
+        "options": {
+            "mainstream": "Mainstream",
+            "montessori": "Montessori",
+            "waldorf": "Waldorf / Steiner",
+            "dalton": "Dalton"
+        }
+    },
+    "terms": [
+        "pedagogics",
+        "pedagogical model",
+        "educational model",
+        "educational philosophy",
+        "teaching method",
+        "learning approach"
+    ]
+}
--- a/data/fields/pedagogy.json
+++ b/data/fields/pedagogy.json
@@ -11,15 +11,15 @@
         }
     },
     "terms": [
-        "pedagogics",
-        "pedagogical model",
+        "dalton",
         "educational model",
         "educational philosophy",
-        "teaching method",
         "learning approach",
         "montessori",
-        "waldorf",
-        "dalton",
-        "steiner"
+        "pedagogical model",
+        "pedagogics",
+        "steiner",
+        "teaching method",
+        "waldorf"
     ]
 }
--- a/data/fields/pedagogy.json
+++ b/data/fields/pedagogy.json
@@ -16,6 +16,10 @@
         "educational model",
         "educational philosophy",
         "teaching method",
-        "learning approach"
+        "learning approach",
+        "montessori",
+        "waldorf",
+        "dalton",
+        "steiner"
     ]
 }
--- a/data/presets/education/college.json
+++ b/data/presets/education/college.json
@@ -12,7 +12,6 @@
         "{@templates/internet_access}",
         "{@templates/poi}",
         "denomination",
-        "pedagogy",
         "ref/edubase-GB",
         "religion"
     ],
--- a/data/presets/education/college.json
+++ b/data/presets/education/college.json
@@ -12,6 +12,7 @@
         "{@templates/internet_access}",
         "{@templates/poi}",
         "denomination",
+        "pedagogy",
         "ref/edubase-GB",
         "religion"
     ],
--- a/data/presets/education/kindergarten.json
+++ b/data/presets/education/kindergarten.json
@@ -22,6 +22,7 @@
         "not/name",
         "opening_hours",
         "payment_multi",
+        "pedagogy",
         "wheelchair"
     ],
     "geometry": [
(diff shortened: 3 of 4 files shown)


data/presets/education/college.json
```diff
@@ -12,6 +12,7 @@
         "{@templates/internet_access}",
         "{@templates/poi}",
         "denomination",
+        "pedagogy",
```
matkoniecz: Are there Montessori/Walforf etc colleges in any noticeable numbers? From my vague impression it is more for small children (but I may be wrong, I have not researched situation yet)

PR description mentions that some exist - but is it like 5 worldwide or more?

matkoniecz:
> Please use this preview to check your changes. Ideally use [the test documentation template](https://github.com/openstreetmap/id-tagging-schema/blob/main/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L38-L69) and document your test results by commenting on the PR. This will speed up the review process for everyone.

can you do this checking? Ideally linking places where this PR can be tested (linking preview site). This allows much quicker review

data/presets/education/college.json
RudyTheDev: I don't really know, I don't know where I could find data for this. It was also my impression that these methodologies are predominantly for preschools and schools. But when I was briefly researching there were mentions of colleges so I included it. Do you think colleges should be excluded?

tordans: @RudyTheDev you could use https://github.com/woodpeck/postpass#llm to get some numbers on this for current usage. And do some LLM research + spot check for a general websearch and knowledge summary.

RudyTheDev:
Preschool https://pr-2276--ideditor-presets-preview.netlify.app/id/dist/#map=19.00/52.36037/9.74780&id=w1382335831

School https://pr-2276--ideditor-presets-preview.netlify.app/id/dist/#map=19.00/52.09952/5.25337&id=w1506792869

College https://pr-2276--ideditor-presets-preview.netlify.app/id/dist/#map=18.00/52.32659/4.87979&id=w323875516

data/presets/education/college.json
RudyTheDev: @tordans Are we talking current OSM mapped data or actual real-life numbers? I thought matkoniecz was asking real-world.

On OSM, there are ~1300 schools, ~800 preschools and a handful of colleges with `pedagogy` tag.

I already did an LLM research and web search checks for real-world usage, which I summarized in the PR. There are a few studies on the subject and that's where the estimated numbers are from. At some point, LLM just starts hallucinating if you ask it any more details.

matkoniecz: If OSM numbers are tiny then noticeable amount of real-world features would be a good reason

RudyTheDev: I'll remove it from college for now since I cannot demonstrate wide use.

tordans:
This looks good to me and I don't see a reason not to merge this ATM.

>grafik

It would be ideal if we had osm wiki _data items_ for each of those main values, so the info-i is explicit for those. This would be a good place to clarify details like ""waldorf" also includes "steiner"" https://wiki.openstreetmap.org/wiki/Key:pedagogy

data/fields/pedagogy.json
```diff
@@ -0,0 +1,21 @@
...
+            "dalton": "Dalton"
+        }
+    },
+    "terms": [
+        "pedagogics",
+        "pedagogical model",
+        "educational model",
+        "educational philosophy",
+        "teaching method",
+        "learning approach"
```
matkoniecz: maybe add also Montessori etc as search term? I can easily imagine someone going "I want to mark it as Montessori school" and searching "Montessori" in fields

or is my imagination too wild?

RudyTheDev: Yeah, that does make sense now that you mention it. I can imagine someone thinking of it as a "montessori school", so trying to find a "montessori" field/checkbox is a natural attempt.

matkoniecz (approved):

data/fields/pedagogy.json
matkoniecz: maybe adding montessori etc to preset terms for school would make sense... But that is too much, probably

And "montessori school" should find it anyway

PR #2297, opened 2026-05-31
new outdoor_seating options

cuatim:

Specific options for `outdoor_seating` are well established now and are more informative then just `yes`. I have added the options that were listed in the wiki for now. All of these are already supported in [SCEE](https://github.com/Helium314/SCEE/blob/modified/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/seating/AddOutdoorSeatingTypeForm.kt) while a subset of these are supported by multiple other [projects](https://taginfo.openstreetmap.org/keys/outdoor_seating#projects).

### Links and data

**Relevant OSM Wiki links:**

- https://wiki.openstreetmap.org/wiki/Key:outdoor_seating

**Relevant tag usage stats:**

- Around 20k uses in total of the new options, see https://taginfo.openstreetmap.org/keys/outdoor_seating#values

### DIFF
--- a/data/fields/outdoor_seating.json
+++ b/data/fields/outdoor_seating.json
@@ -1,5 +1,22 @@
 {
     "key": "outdoor_seating",
-    "type": "check",
-    "label": "Outdoor Seating"
+    "type": "semiCombo",
+    "label": "Outdoor Seating",
+    "strings": {
+        "options": {
+            "yes": "Yes"
+            "no": "No"
+            "sidewalk": "Sidewalk"
+            "terrace": "Terrace"
+            "pedestrian_zone": "Pedestrian Zone"
+            "patio": "Patio"
+            "veranda": "Veranda"
+            "garden": "Garden"
+            "parklet": "Parklet"
+            "street": "Street"
+            "balcony": "Balcony"
+            "roof": "Roof"
+            "beach": "Beach"
+        }
+    }
 }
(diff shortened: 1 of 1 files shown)


cuatim:
## Test-Documentation

Looks good.

https://pr-2297--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=20.00/48.15105/11.56394&disable_features=boundaries&background=fr.ign.bdortho&id=n4311689762

image

image


imageimage

cuatim:
should we add the field `outdoor_seating` to `leisure=outdoor_seating` as well? See https://wiki.openstreetmap.org/wiki/Tag:leisure%3Doutdoor_seating

matkoniecz:
my first impression is that turning boolean yes/no field into listing various detail is always a bit annoying, and I am not a fan of it

on the other hand spreading it across two tags would have own problems...

and while use of values other than no/yes is not very very high it is not nonexisting either... See already linked https://taginfo.openstreetmap.org/keys/outdoor_seating#values

cuatim:
> my first impression is that turning boolean yes/no field into listing various detail is always a bit annoying, and I am not a fan of it

but that is the way it is documented and used, [for 10+ years](https://taginfo.openstreetmap.org/tags/outdoor_seating=sidewalk).

matkoniecz:
yeah, "I am not fan of it" is not used to imply "and I will filibuster it forever"

I will try to look at this PR soon

cuatim:
Thanks! I was not thinking you would :) I am just not always sure, if/when a reply from me is required :)

PR #2298, opened 2026-06-01
`highway=hitchhiking` in the United States

1ec5:
This extends the `highway=hitchhiking` preset to the United States. The U.S. is more or less [tied with Belgium](https://qlever.dev/osm-planet/4Ec7lg) for the country with the third most occurrences of this tag in OSM, far ahead of some other countries that were added prospectively on the basis that they speak German: https://github.com/openstreetmap/id-tagging-schema/pull/1705#discussion_r2380699635.

Simultaneously, I’ve renamed the preset to “Slugging Location”, to match the default development language (American English). This is crucial if the preset is to apply to an English-speaking country, because “hitchhiking” means something different in English. Hitchhiking is usually prohibited in localities where the `highway=hitchhiking` feature legitimately appears. It is signed – the sign says no. Yet if we leave the current name, mappers unaware of the history of this tag would inevitably use it to map actual hitchhiking spots. I’m pretty sure the community would agree that such a spot wouldn’t be observable or permanent enough to be mapped in OSM, as opposed to a _Mitfahrbank_. For safety reasons, we should distinguish the two activities if possible.

“Slugging Location” is a compromise. I considered the more colloquial “slug line”, but the user might think it needs to be mapped as a line rather than a point. On the other hand, official terms such as “carpool staging area” and “car stop” could confuse even native speakers, causing them to use it instead of other presets like “Car Pooling Station” and “Stop Sign”. I added all these terms as `term`s instead.

This change will also force translators to think twice before translating the preset name into the usual word for “hitchhiking” in each language. The German localization can of course call it _Mitfahrbank_ if that gets the point across. (Although I note that the English “car stop” does resemble the French and German “auto-stop”.)

### DIFF
--- a/data/presets/highway/hitchhiking.json
+++ b/data/presets/highway/hitchhiking.json
@@ -12,6 +12,9 @@
         "highway": "hitchhiking"
     },
     "terms": [
+        "carpooling pickup",
+        "carpool staging area",
+        "car stop",
         "hitchhiking",
         "slug line",
         "thumbing"
--- a/data/presets/highway/hitchhiking.json
+++ b/data/presets/highway/hitchhiking.json
@@ -13,6 +13,7 @@
     },
     "terms": [
         "hitchhiking",
+        "slug line",
         "thumbing"
     ],
     "locationSet": {
@@ -24,9 +25,10 @@
             "fx",
             "it",
             "li",
-            "nl"
+            "nl",
+            "us"
         ]
     },
-    "name": "Signed Hitchhiking Place",
+    "name": "Slugging Location",
     "icon": "far-thumbs-up"
 }


matkoniecz:
@paulklie Do you have opinion on this PR?

paulklie:
I approve adding the USA, however I don't like the idea of changing the name:

From reading https://en.wikipedia.org/wiki/Slugging and https://hitchwiki.org/en/Slugging it seems to me as "Slugging Location" would fit better to [amenity=car_pooling](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcar_pooling)  than highway=[hitchhiking](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dhitchhiking). Or how would you differentiate these two @1ec5 ?

Also pinging @tillwenke since you develop hitch wiki.

tillwenke:
Thanks for pinging @paulklie

Hitchhiking seems to be a superset of slugging.

I believe the misunderstanding comes from not being to familiar with hitchhiking or the US.
Maybe a US hitchhiker can chime in :) 

Am I correct that in the US there could be a hitchhiking sign that is not a slugging sign?

For me "Signed hitchhiking spot" hits the nail.

There has to be some sort of sign and one is allowed to do at least one of the activities described as hitchhiking in the Wikipedia article.

tillwenke:
On another note. The term slugging has an odd sound to me (as a German, also looking at the words that (to) slug translate to) it does not sound very neutral and slightly discriminative.

I would be curious to learn how osm deals with this in other instances.

1ec5:
Hitchwiki’s perspective seems to be that slugging is a subset of hitchhiking. I can see how that would seem logical, but it would be somewhat problematic here in the U.S. I’m under the impression that slugging is considered closer to carpooling than hitchhiking, but all three are distinct. `highway=hitchhiking` definitely is not for hitchhiking. I’ve already encountered cases along roads that should probably be tagged `hitchhiking=no`.

A slug line is always signposted, at least at the pickup zone if not also at the dropoff zone. It is by definition an impromptu, free arrangement with a stranger. I believe this matches every explanation of the tag that has been given so far. If nonetheless you still consider this `amenity=car_pooling`, then we have two overlapping tags. The only possible distinction I can think of is that slug lines are primarily intended for commuters.

“Slugging” is an informal term preferred in the eastern U.S. Yes, many decades ago, it was coined as a pejorative term by bus operators, but sluggers started using that term for themselves. If we want to avoid it, the obvious alternative would be “casual carpooling”, preferred in the western U.S.

(For full disclosure, I’ve barely ever gone hitchhiking, not enough to consider myself a hitchhiker. I used to walk by the casual carpooling signs in San Francisco fairly often, but I’ve never gone casual carpooling because I commuted in the opposite direction.)

matkoniecz:
> highway=hitchhiking definitely is not for hitchhiking.

then it should not have been merged and maybe we can still remove it?

1ec5:
Well, we’re in an awkward situation: as far as I can tell, the community approved `highway=hitchhiking` under a definition that excludes (what Americans would consider) hitchhiking, that conflicts with the presumed definition of `hitchhiking=*`. I take some responsibility for this, as the proposer asked me for input and I didn’t quite grasp the implications at the time. We could choose to cope with the misnomer, because OSM tags are just keywords after all, or we could go back and ask the community to reconsider the recent approval.

tillwenke:
> I’ve already encountered cases along roads that should probably be tagged `hitchhiking=no`

Do you have an example for this? The person using the tag wrongly perhaps did not have a look at https://wiki.openstreetmap.org/w/index.php?title=Tag:highway%3Dhitchhiking&oldid=2974978 (the latest revision leans a bit too much towards slugging for my taste).



> under a definition that excludes (what Americans would consider) hitchhiking

What is the definition in American English?

1ec5:
> > I’ve already encountered cases along roads that should probably be tagged `hitchhiking=no`
> 
> Do you have an example for this? The person using the tag wrongly perhaps did not have a look at https://wiki.openstreetmap.org/w/index.php?title=Tag:highway%3Dhitchhiking&oldid=2974978 (the latest revision leans a bit too much towards slugging for my taste).

See [this earlier forum discussion](https://community.openstreetmap.org/t/rfc-feature-proposal-tag-proposal-hitchhiking/108126/34). This is not a case of using the tag “wrongly”. It satisfies every criterion you’ve ever given, and yet hitchhiking per se is banned. `hitchhiking=no` came up during the RfC as the obvious tag for the standard No Hitchhiking sign. Even when that sign isn’t posted, state law implicitly prohibits hitchhiking in many cases, including on the same road that has the sign that `highway=hitchhiking` was intended for.

> > under a definition that excludes (what Americans would consider) hitchhiking
> 
> What is the definition in American English?

Standing beside the road at an _unsigned_ location and holding out your thumb to ask a passing motorist for a ride. If the location is designated even with a privately installed sign or pavement marking, then it’s slugging, not hitchhiking.

tillwenke:
> Standing beside the road at an _unsigned_ location and holding out your thumb to ask a passing motorist for a ride. If the location is designated even with a privately installed sign or pavement marking, then it’s slugging, not hitchhiking.

https://en.wikipedia.org/wiki/Slugging and https://en.wikipedia.org/wiki/Hitchhiking do not support this imo. Where did you draw you definition from? For me this reads as slugging is a subset of hitchhiking.

tillwenke:
> See [this earlier forum discussion](https://community.openstreetmap.org/t/rfc-feature-proposal-tag-proposal-hitchhiking/108126/34). This is not a case of using the tag “wrongly”. It satisfies every criterion you’ve ever given, and yet hitchhiking per se is banned. `hitchhiking=no` came up during the RfC as the obvious tag for the standard No Hitchhiking sign. Even when that sign isn’t posted, state law implicitly prohibits hitchhiking in many cases, including on the same road that has the sign that `highway=hitchhiking` was intended for.

Let's have a look at the example of Virginia, it is actually more nuanced. Hitchhiking is not prohibited, just standing on the roadway for that purpose (while the term hitchhiking is actually not mentioned) https://law.lis.virginia.gov/vacode/title46.2/chapter8/section46.2-929/ where roadway is defined here https://law.lis.virginia.gov/vacode/46.2-100/ excluding the shoulder. If there was a state where there is an area in which soliciting rides is prohibited and allowed at the same time then this is rather an inconsistency in the states's law.

1ec5:
> https://en.wikipedia.org/wiki/Slugging and https://en.wikipedia.org/wiki/Hitchhiking do not support this imo. Where did you draw you definition from? For me this reads as slugging is a subset of hitchhiking.

A Wikipedia article or [dictionary definition](https://en.wiktionary.org/wiki/slug#Verb) should not be read as a literal technical specification. But in any case, “essentially a variation of” in the lede means “like”, not “a kind of”. After all, “[organized hitchhiking](https://www.wsj.com/us-news/dc-carpooling-slugging-trump-return-to-office-dcff5c47)” is just a cute analogy, but “casual carpooling” is an official term.

> If there was a state where there is an area in which soliciting rides is prohibited and allowed at the same time then this is rather an inconsistency in the states's law.

I think the bigger inconsistency is that we have a frequently posted No Hitchhiking sign from the national standard. Even if it doesn’t occur on the same signpost as a `highway=hitchhiking` sign, we already agreed that `hitchhiking=no` would be a sensible tag for it. We shouldn’t reuse the same word for such different meanings. If slugging really is a subset of hitchhiking, then one would expect the primary feature tag to be more specific.

After all this, I hope it’s clearer that the situation is too nuanced to expect a layperson to choose the correct preset based on the name and icon alone. We can’t expect users to parse definitions in state law to understand whether the preset name is based on common language. Choosing the more specific term will help protect the dataset you’re trying to promote from misuse.

zekefarwell:
> highway=hitchhiking definitely is not for hitchhiking. 

How unfortunate that the value `hitchhiking` was chosen for this.  As an American English speaker, the [wiki page for highway=hitchhiking](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dhitchhiking) does not describe an activity that I would call or have heard described as "hitchhiking".  In my experience hitchhiking is an informal activity that doesn't have designated signed locations.  A signed location for sharing rides with strangers would be called something else.

Issue #2050, opened 2020-10-19
Grouped fields


quincylvania:
iD has a few special fields that group related tags together, like Name, Address, Structure, Allowed Access, and Bike Lanes. Grouping is a nice way to make the inspector simpler and more intuitive. There are a lot more tags that could benefit from grouping, for example:

- Internet connection/access/network name
- Wheelchair access/description (#7698)
- Population number/source/date (#6672)
- Fee presence/amount/payment types

Instead of coding a special field type for every grouping, we should look into a reusable grouped field type that can handle simple combinations.

PR #2300, opened 2026-06-02
Synergize opening_hours/drive_through

cuatim:
I moved `opening_hours/drive_through`, so its always right after `drive_through`, which is the case for all other appearances.

I think the same should be done for `indoor_seating`+`outdoor_seating` (maybe even in this PR?)


cuatim:
Works:

https://pr-2300--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=20.00/48.83827/2.59288&disable_features=boundaries&background=fr.ign.bdortho&id=n2705344356

image

cuatim:
Maybe both drive though keys should me systematically moved to `moreFields` also?

matkoniecz:
summoning @1ec5 @ZeLonewolf for US-local advise

I suspect that drive-in stuff may be there as popular and important property as stereotypes claim

ZeLonewolf:
I'm unclear on the question. But drive through (or informally, drive thru) is very different from drive in.

matkoniecz:
this question makes drive through opening hours field (yes, distinct from drive in) far less visible

instead of always appearing it would appear when filled already or when added manually by user

no idea is it a good change or not

1ec5:
> I suspect that drive-in stuff may be there as popular and important property as stereotypes claim

In general, extended drive-through hours are typical of `amenity=fast_food`. This is a defining characteristic of some burger chains like Wendy’s and White Castle. `amenity=cafe` may be affected too because we’re quite unsure which of the two tags is more appropriate for chains like [Dunkin’](https://community.openstreetmap.org/t/proposal-to-update-dunkin-category-to-cafe/128293), Krispy Kreme, and Tim Hortons.

I don’t know of any ice cream stand or ice cream parlor that has separate hours for the drive-through versus the dining room or walk-up window. Anything is possible, but I don’t think it would be very common, even at a gas station+convenience store+ice cream parlor like osmlab/name-suggestion-index#2234.

data/presets/amenity/cafe.json
```diff
@@ -24,6 +23,7 @@
         "delivery",
         "diet_multi",
         "drive_through",
+        "opening_hours/drive_through",
```
matkoniecz, line 26: is it deliberate effect so that first you need to manually add drive_through field, mark it as `yes` and then manually add opening_hours/drive_through ? Rather than it appearing on its own?

matkoniecz: because for me the current behavior makes more sense

matkoniecz: @cuatim

cuatim: is there some extra logic in the drive_through fields in iD? from a OSM point of view, it makes sense to set both keys?

matkoniecz: Not sure what you mean here by special logic

But https://wiki.openstreetmap.org/wiki/Key%3Aopening_hours%3Adrive_through field is hidden until drive-through field is set to indicate that drive through is present (as defined in its field code)

cuatim: Ah. Okay I was not aware of this (the `prerequisiteTag` feature does not work in EveryDoor, so I never noticed).

Well, not sure what is the best way and order to have drive_through+opening_hours:drive_through. I just made all presets the same.

matkoniecz: we probably need https://github.com/openstreetmap/id-tagging-schema/issues/2050

matkoniecz (changes requested):
please restore them to `fields` if they were already in `fields`

matkoniecz:
@cuatim ping

cuatim:
I think this whole PR can be put on hold, and then adapted once #2050 is merged?

data/presets/amenity/cafe.json
cuatim: Agreed!

matkoniecz:
> then adapted once https://github.com/openstreetmap/id-tagging-schema/issues/2050 is merged?

well, #2050 needs to be first designed and implemented, this is merely issue

given that it would need adapting anyway and is fairly small change, I will close this PR and it can be redone later

or just reorder in ice cream can be made (I would accept it)

PR #2307, opened 2026-06-03
Adding `delivery:partner` + wikidata if delivery != no

GA-Kevin-Codes:
Ever since COVID, more and more places are offering delivery via 3rd party services. These contract drivers to deliver an establishments food or products and facilitate ordering via mobile applications and websites. In OSM, we tag these with `delivery:partner` ([on the OSM wiki](https://wiki.openstreetmap.org/wiki/Key:delivery:partner) | [Wiki data item](https://wiki.openstreetmap.org/wiki/Item:Q20261)) to distinguish from first-party delivery options (expressed via `delivery=yes`.) To reduce spelling variations and provide additional information on these services, `delivery:partner:wikidata` ([Wiki data item](https://wiki.openstreetmap.org/wiki/Item:Q23779)) is documented to link these services to the establishment. These services are typically prominently displayed via signage or sticker inside or on the outward facing window of an establishment.

I have formed this PR with `prerequisiteTag` of delivery != no so that the partner fields remain hidden unless a delivery option is available. This should help retain minimal tags in the `fields` group while surfacing the relevant tags at the right time. I have added these fields to the restaurant, cafe, fast_food, and ice_cream presets as that is where `delivery` (the prerequisite) is currently but further improvements later can be added to the `moreFields` group of shops, convenience stores, grocery stores, or anywhere that would realistically be delivering via third party. This PR narrows the initial scope to only the highest-value adds of food related venues.

You can see an example of this tag usage in the city of Valdosta in the United States ([Overpass Turbo query](https://overpass-turbo.eu/s/2r5N)) including examples of locations that have multiple delivery partners such as [this Target](https://www.openstreetmap.org/way/838017770#map=19/30.843146/-83.316006) which has 4 delivery partners. There is a seperate PR out for parsing the `delivery:partner:wikidata` value on the OSM website openstreetmap/openstreetmap-website#7127.

### Links and data

**Relevant OSM Wiki links:**
- [`delivery=*` English wiki page](https://wiki.openstreetmap.org/wiki/Key:delivery)
- [`delivery:partner=*` English wiki page](https://wiki.openstreetmap.org/wiki/Key:delivery:partner)
- [`delivery (Q209)` OSM Wiki data item](https://wiki.openstreetmap.org/wiki/Item:Q209)
- [`delivery:partner (Q20261)` OSM Wiki data item](https://wiki.openstreetmap.org/wiki/Item:Q20261)
- [`delivery:partner:wikidata (Q23779)` OSM Wiki data item](https://wiki.openstreetmap.org/wiki/Item:Q23779)

**Relevant tag usage stats:**
- [delivery=*](https://taginfo.openstreetmap.org/keys/delivery) ~120,000 uses worldwide
- [delivery:partner=*](https://taginfo.openstreetmap.org/keys/delivery:partner) ~1,300 uses worldwide
- [delivery:partner:wikidata=*](https://taginfo.openstreetmap.org/keys/delivery:partner:wikidata) ~ 180 uses worldwide

### DIFF
--- a/data/fields/delivery/partner.json
+++ b/data/fields/delivery/partner.json
@@ -1,0 +1,9 @@
+{
+    "key": "delivery:partner",
+    "type": "text",
+    "label": "Delivery Partner",
+    "prerequisiteTag": {
+        "key": "delivery",
+        "valueNot": "no"
+    }
+}
--- a/data/fields/delivery/partner/wikidata.json
+++ b/data/fields/delivery/partner/wikidata.json
@@ -1,0 +1,13 @@
+{
+    "key": "delivery:partner:wikidata",
+    "keys": [
+        "delivery:partner:wikidata",
+        "delivery:partner:wikipedia"
+    ],
+    "type": "wikidata",
+    "label": "Delivery Partner Wikidata",
+    "prerequisiteTag": {
+        "key": "delivery",
+        "valueNot": "no"
+    }
+}
--- a/data/presets/amenity/cafe.json
+++ b/data/presets/amenity/cafe.json
@@ -10,7 +10,9 @@
         "{@templates/internet_access}",
         "phone",
         "website",
-        "opening_hours/drive_through"
+        "opening_hours/drive_through",
+        "delivery/partner",
+        "delivery/partner/wikidata"
     ],
     "moreFields": [
         "{@templates/internet_access}",
--- a/data/presets/amenity/fast_food.json
+++ b/data/presets/amenity/fast_food.json
@@ -11,7 +11,9 @@
         "drive_through",
         "opening_hours/drive_through",
         "phone",
-        "website"
+        "website",
+        "delivery/partner",
+        "delivery/partner/wikidata"
     ],
     "moreFields": [
         "{@templates/internet_access}",
--- a/data/presets/amenity/ice_cream.json
+++ b/data/presets/amenity/ice_cream.json
@@ -5,7 +5,9 @@
         "address",
         "building_area_yes",
         "opening_hours",
-        "outdoor_seating"
+        "outdoor_seating",
+        "delivery/partner",
+        "delivery/partner/wikidata"
     ],
     "moreFields": [
         "{@templates/internet_access}",
--- a/data/presets/amenity/restaurant.json
+++ b/data/presets/amenity/restaurant.json
@@ -8,7 +8,9 @@
         "building_area_yes",
         "opening_hours",
         "phone",
-        "website"
+        "website",
+        "delivery/partner",
+        "delivery/partner/wikidata"
     ],
     "moreFields": [
         "{@templates/internet_access}",


tordans:
@GA-Kevin-Codes I just startet the preview build so you can do some testing of this and share it. 

A few general thoughts:
- Having two related but disconnected semicolon based fields feels hacky. All those semicolon based "arrays" are not ideal and having two that are related means we have to be conscious about if we want to present this in this way at this point (given usage)
- Given the low usage (vs. high occurrence), we should have more clarity on how those fields came to be and what the community consensus is on them.
- We are talking about brand here, so this sounds like a NSI topic. We don't have a way yet to apply NSI to fields, but maybe we should look into this and explore what other fields would benefit from that and make this a separate field feature (expand the schema).
- Lastly: Is there any affiliation for this PR that we should be aware of – vs. pure community interest. Just so we understand the context better.

GA-Kevin-Codes:
Thanks @tordans! The preview is very helpful!

There does seem to be an issue of the partner fields displaying even if `delivery` is unset, despite the `prerequisiteTag` being set to !=no, but that may be moot depending on how we plan to proceed.

- Arrays in iD are typically of type `semiCombo` which may be useful precedent to go off of. To my knowledge, these do not typically provide a search function, unlike type `wikidata`. It would be very smooth to go IF `delivery!=no` THEN display a semiCombo - wikidata search to select carriers. If so, then `delivery:partner` and `delivery:partner:wikidata` can be filled with 1 selection. I am not sure how difficult combining semiCombo + wikidata would be in the schema but am happy to work on it if it is desired.
- History-wise, `delivery:partner` seems to have started around 2017, with a ramp up in usage around 2022. I don't see any red flags of maliciousness with usage, it's just one of those things that has increasingly become more popular as time goes on and OSM is catching up. Uber Eats, for example, didn't start until 2015 in earnest so that roughly tracks with our usage. Nowadays, this type of service is offered worldwide so usage should continue to see a healthy rise.
- @1ec5 mentioned this in the OSM US Slack about this tag, but as you say the NSI as it stands right now does not get applied to fields. This also may not be the best route to go as it is somewhat scattered what service an establishment uses even if they share a brand. For example, a McDonald's in Miami USA and Paris France likely uses different partners despite both having `delivery=yes`. It may be worth in the future considering a NSI-by-field, but I would consider that out of scope for this PR. Again, happy to work on it or discuss, I just don't want to derail this PR too much.
- No affiliation to disclose, I just am a part of the OSM community, it's my favorite hobby (don't tell Home Assistant I said that) and noticed these partner stickers around town and started mapping them :)

matkoniecz:
> These services are typically prominently displayed via signage or sticker inside or on the outward facing window of an establishment.

in my experience it is far from typical, and sticker coverage is hardly actually covering all options

also, it keep changing at high rate

GA-Kevin-Codes:
Thanks @matkoniecz, this discussion is on the implementation on the iD schema not the tag itself. The wiki talk pages would be the proper place to discuss tagging details (I know you know this, just trying to stay on topic!)

matkoniecz:
Judging whether tag is worth promoting is definitely on topic here. If tag is in some way problematic it may result it in not being accepted to be included into this preset.

Currently I have doubts and would lean toward rejecting it as the initial impression but I have not investigated it yet.

[ATYL](https://wiki.openstreetmap.org/w/index.php?title=ATYL) does not apply to inclusion into presets

GA-Kevin-Codes:
Please do, I have done quite a bit of investigating myself in this PR which is above but welcome independent verification!

GA-Kevin-Codes:
Hey @tordans, any thoughts on the above?

> Arrays in iD are typically of type semiCombo which may be useful precedent to go off of. To my knowledge, these do not typically provide a search function, unlike type wikidata. It would be very smooth to go IF delivery!=no THEN display a semiCombo - wikidata search to select carriers. If so, then delivery:partner and delivery:partner:wikidata can be filled with 1 selection. I am not sure how difficult combining semiCombo + wikidata would be in the schema but am happy to work on it if it is desired.

GA-Kevin-Codes:
Hey @tordans, just checking in on the progress of getting this reviewed

tordans:
> Hey @tordans, just checking in on the progress of getting this reviewed

Hey, I will not have the capacity to look int this more; I am focussed on other OSM things ATM. Personally I was hoping for some other input here. ATM it looks like it is just us talking and the Wiki is [also influenced by this thread](https://wiki.openstreetmap.org/w/index.php?title=Key:delivery:partner&action=history). Maybe Mateusz will have time to look at it soon but we have 120+ open PRs…
One thing that would help is if we get a better feel for if this tag is really smart to promoted based on feedback from others. Mateusz starts this forum threads from time to time to get a feel for what others say. ~1.4k usage is just not that high to make this something to merge without looking into it more…

matkoniecz:
> Mateusz starts this forum threads from time to time to get a feel for what others say.

yes, but currently I feel that I am out of forum thread posting quota :)

k-yle (changes requested):
in addition to the problems noted in https://github.com/openstreetmap/id-tagging-schema/pull/2307#issuecomment-4618726503:

5. [taginfo](https://taginfo.openstreetmap.org/keys/delivery%3Apartner%3Awikidata#values) and [the wiki](https://wiki.openstreetmap.org/wiki/Key:delivery:partner) suggest that multiple values will be very common. `"type": "wikidata"` does not support multiple values, so [#2722](https://github.com/openstreetmap/id-tagging-schema/issues/2722) would need to be implemented before we can consider this PR. 
6. this definitely needs to be discussed with the NSI maintainers, since tracking these sorts of brands is their area of expertise. NSI also [insists on adding annoying tags like `operator:type`](https://github.com/osmlab/name-suggestion-index/pull/10296) even when the local community opposes it. so maybe they'll require something like `delivery:partner:type=private;private;private` ? so it definitely needs a discussion and conclusion from their side first.

bhousel:
> 6\. NSI also [insists on adding annoying tags like `operator:type`](https://github.com/osmlab/name-suggestion-index/pull/10296) even when the local community opposes it.

I disagree with this framing.  Where on that thread did I "insist" on such a thing?

k-yle:
i don't think it was you, but this topic came up in that PR and a few times on discord iirc? i haven't heard any recent complaints, presumably because [1,452](https://overpass-turbo.eu/s/2vs9) of those 'errors' have been accepted already.

---

but in any case, this `delivery:patner` topic should ideally be discussed in the NSI repo to gather feedback before continuing with this PR

GA-Kevin-Codes:
Hey @k-yle thanks for the response. Please see https://github.com/openstreetmap/id-tagging-schema/pull/2307#issuecomment-4622556761 for my response to 1-4.

5. As Minh pointed out there, some fields do indeed accept multiple Wikidata fields already, as well as the OSM website gracefully handling multiple QIDs. I am unsure why this is a blocker.
6. As stated above, NSI tends to be a preset at the top level, not at individual fields. Do you mean Wikidata search instead? NSI would not handle this key as it can change region to region and it is primarily focused on the most precise set of tags that would be applicable to any location.

bhousel:
> NSI would not handle this key as it can change region to region and it is primarily focused on the most precise set of tags that would be applicable to any location.

I agree with this.  It doesn't make sense for `delivery:partner` to be in NSI - this is something that should be tagged per-restaurant.

k-yle:
okay, in that case the main blocker is #2722. 

> As Minh pointed out there, some fields do indeed accept multiple Wikidata fields already, as well as the OSM website gracefully handling multiple QIDs. I am unsure why this is a blocker.

unfortunately it's a blocker because iD's UI is completely broken for `*:wikidata` fields with multiple values. So it doesn't make sense to merge a PR knowing that it will be broken for the majority of cases. 

for `brand:wikidata=*` multiple values are very rare, for `deliver:panter:wikidata=*` the [majority of unique values](https://taginfo.openstreetmap.org/keys/delivery%3Apartner%3Awikidata#values) have a `;`. 

#2722 will be complex both technically and from a tagging perspective. for example, what should the UI do if `operator=a;b;c` has 3 items and `operator:wikidata=a;b;c;d` has 4? and `operator:type=a` has one value. etc. etc.

matkoniecz:
> okay, in that case the main blocker is https://github.com/openstreetmap/id-tagging-schema/issues/2722.

and also question whether this tagging is established enough to promote it via presets

Issue #2294, opened 2026-05-30
Stop bot edits on some traffic facilities


Geo-2695:
Some train stations and ferry terminals are not suitable for the tag `public_transport=station` . Certain railway station(distinct from  railway=yard ) and ferry terminals are exclusively for freight or military use, with no passenger services available. iD constantly flags these entries as having incomplete tags.

Some pages on the OSM Wiki contain relevant editing guidelines.

<img width="1080" height="1532" alt="Image" src="https://github.com/user-attachments/assets/a813cbf1-642d-4be6-8f3c-a58c83bfe616" />

<img width="1080" height="1283" alt="Image" src="https://github.com/user-attachments/assets/9c258889-74ca-43ce-aac7-62480c2c335d" />

matkoniecz:
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dferry_terminal

https://wiki.openstreetmap.org/wiki/Tag:railway%3Dstation

seems to not require them to be public transport objects

matkoniecz:
For ferries code is at https://github.com/openstreetmap/id-tagging-schema/blob/048ac09eac1ebdc37a79a930e4cf98d036b49546/data/presets/public_transport/station_ferry.json and https://github.com/openstreetmap/id-tagging-schema/blob/048ac09eac1ebdc37a79a930e4cf98d036b49546/data/presets/amenity/_ferry_terminal.json

maro-21:
I support.

Geo-2695:
The issue of bot-driven abusive edits involving `public_transport=station` appears to extend beyond these two types of facilities, and problematic edits may exist elsewhere as well.
See also #1633

matkoniecz:
> appears to extend beyond these two types of facilities

if other problem exist and issue or pull request does not exist for them, feel free to create one

matkoniecz:
@Geo-2695 how train station may be marked as not being part of a public transport? Do we have established tagging for that?

How it is done for ferry terminal?

Geo-2695:
I’ve looked through the relevant Wiki pages and documentation. When tagging non-passenger railway stations, the rule is to remove the `public_transport=station` tags while leaving all other tags unchanged. Some mappers add [public_transport=no](https://wiki.openstreetmap.org/wiki/Tag%3Apublic_transport%3Dno) to suppress validation warnings.
The same approach applies to ferry terminals.

andrewharvey:
[...]

PR #2309, opened 2026-06-05
Add unsearchable `aerialway=station` preset without public transport tags and update related icons

### LINKED ISSUE #1633: Do not suggest `public_transport=station` for goods aerialway

Geo-2695:
Close #1633
Some aerialway stations are not suitable for tagging with `public_transport=station`. The original preset tied these two tags together, which triggers warnings when mappers try to remove the public transport tag and may mislead them into incorrect edits. I have therefore created a new standalone preset for `aerialway=station` and adjusted its configuration to distinguish between passenger and non-public transport use cases. This ensures no warning will pop up if mappers remove the public transport related tags
<details><summary>Related Links</summary>

**Wiki**

https://wiki.openstreetmap.org/wiki/Tag:aerialway=station

https://wiki.openstreetmap.org/wiki/Tag:public_transport=station

**Taginfo**

https://taginfo.openstreetmap.org/tags/aerialway=station#combinations

https://taginfo.openstreetmap.org/tags/public_transport=station#combinations

</details>

### DIFF
--- a/data/presets/aerialway/_station.json
+++ b/data/presets/aerialway/_station.json
@@ -1,0 +1,20 @@
+{
+    "icon": "temaki-gondola_lift",
+    "fields": [
+        "{public_transport/station}"
+    ],
+    "moreFields": [
+        "{public_transport/station}"
+    ],
+    "geometry": [
+        "point",
+        "vertex",
+        "area"
+    ],
+    "tags": {
+        "aerialway": "station"
+    },
+    "matchScore": 0.95,
+    "name": "Aerialway Station",
+    "searchable": false
+}
--- a/data/presets/aerialway/_station.json
+++ b/data/presets/aerialway/_station.json
@@ -1,7 +1,9 @@
 {
     "icon": "temaki-gondola_lift",
     "fields": [
-        "{public_transport/station}"
+        "{public_transport/station}",
+        "aerialway/access",
+        "aerialway/summer/access"
     ],
     "moreFields": [
         "{public_transport/station}"
--- a/data/presets/aerialway/_station.json
+++ b/data/presets/aerialway/_station.json
@@ -1,12 +1,9 @@
 {
     "icon": "temaki-gondola_lift",
     "fields": [
-        "{public_transport/station}",
-        "aerialway/access",
-        "aerialway/summer/access"
-    ],
-    "moreFields": [
-        "{public_transport/station}"
+        "name",
+        "operator",
+        "address"
     ],
     "geometry": [
         "point",
@@ -18,6 +15,5 @@
     },
     "matchScore": 0.95,
     "name": "Aerialway Station",
-    "searchable": false,
-    "replacement": "public_transport/station_aerialway"
+    "searchable": false
 }
--- a/data/presets/aerialway/_station.json
+++ b/data/presets/aerialway/_station.json
@@ -18,5 +18,6 @@
     },
     "matchScore": 0.95,
     "name": "Aerialway Station",
-    "searchable": false
+    "searchable": false,
+    "replacement": "public_transport/station_aerialway"
 }
--- a/data/presets/aerialway/_station.json
+++ b/data/presets/aerialway/_station.json
@@ -9,7 +9,6 @@
         "{public_transport/station}",
         "{@templates/internet_access}",
         "network",
-        "vehicles",
         "building_area",
         "aerialway/access",
         "aerialway/summer/access"
--- a/data/presets/aerialway/_station.json
+++ b/data/presets/aerialway/_station.json
@@ -7,6 +7,10 @@
     ],
     "moreFields": [
         "{public_transport/station}",
+        "{@templates/internet_access}",
+        "network",
+        "vehicles",
+        "building_area",
         "aerialway/access",
         "aerialway/summer/access"
     ],
--- a/data/presets/aerialway/_station.json
+++ b/data/presets/aerialway/_station.json
@@ -5,6 +5,10 @@
         "operator",
         "address"
     ],
+    "moreFields": [
+        "aerialway/access",
+        "aerialway/summer/access"
+    ],
     "geometry": [
         "point",
         "vertex",
--- a/data/presets/aerialway/_station.json
+++ b/data/presets/aerialway/_station.json
@@ -14,6 +14,6 @@
         "aerialway": "station"
     },
     "matchScore": 0.95,
-    "name": "Aerialway Station",
+    "name": "{public_transport/station_aerialway}",
     "searchable": false
 }
--- a/data/presets/aerialway/_station.json
+++ b/data/presets/aerialway/_station.json
@@ -6,6 +6,7 @@
         "address"
     ],
     "moreFields": [
+        "{public_transport/station}",
         "aerialway/access",
         "aerialway/summer/access"
     ],
--- a/data/presets/aerialway/_station.json
+++ b/data/presets/aerialway/_station.json
@@ -8,6 +8,7 @@
         "building_area"
     ],
     "moreFields": [
+        "{public_transport/station}",
         "aerialway/access",
         "aerialway/summer/access"
     ],
--- a/data/presets/aerialway/_station.json
+++ b/data/presets/aerialway/_station.json
@@ -7,7 +7,7 @@
         "address",
         "building_area"
     ],
-    "moreFields": [ 
+    "moreFields": [
         "aerialway/access",
         "aerialway/summer/access"
     ],
--- a/data/presets/aerialway/_station.json
+++ b/data/presets/aerialway/_station.json
@@ -2,14 +2,12 @@
     "icon": "temaki-gondola_lift",
     "fields": [
         "name",
-        "operator",
-        "address"
-    ],
-    "moreFields": [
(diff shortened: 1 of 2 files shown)


Geo-2695:
### Testing
**Before**
Screenshot_20260606_084025
Screenshot_20260606_083921

**After**
Screenshot_20260606_083631
Screenshot_20260606_083646

data/presets/aerialway/_station.json
```diff
@@ -0,0 +1,23 @@
...
+        "vertex",
+        "area"
+    ],
+    "tags": {
+        "aerialway": "station"
+    },
+    "matchScore": 0.95,
+    "name": "Aerialway Station",
+    "searchable": false,
+    "replacement": "public_transport/station_aerialway"
```
matkoniecz: are you sure this property is needed here? 

See https://github.com/ideditor/schema-builder#replacement

Geo-2695: Actually, the `replacement` property isn't needed here. However, I've noticed similar issues exist in other preset. I plan to open another PR specifically to fix this problem in a few days, which makes it easier to categorize.

matkoniecz:
> 🍱 [Your pull request preview is ready](https://pr-2309--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=18.00/48.841708/2.587656)

have you done this testing? Can you post screenshots?

Geo-2695:
### Testing_2
AERIALWITHPUBLIC

aerial

This might incidentally resolve some of the issues in #2294

data/presets/public_transport/station_aerialway.json
```diff
@@ -14,6 +14,7 @@
         "area"
     ],
     "tags": {
+        "public_transport": "station",
         "aerialway": "station"
     },
     "addTags": {
```
matkoniecz: after this change we do not need `addTags` here, right?

data/presets/aerialway/_station.json
```diff
@@ -0,0 +1,19 @@
+{
+    "icon": "temaki-gondola_lift",
+    "fields": [
+        "name",
+        "operator",
+        "address"
```
matkoniecz, line 6: the twin preset has `"aerialway/access",` and `"aerialway/summer/access"`

is there reason to have them there and not here?

data/presets/aerialway/_station.json
```diff
@@ -0,0 +1,19 @@
...
+    "geometry": [
+        "point",
+        "vertex",
+        "area"
+    ],
+    "tags": {
+        "aerialway": "station"
+    },
+    "matchScore": 0.95,
+    "name": "Aerialway Station",
```
matkoniecz: if one is public transport and another not may distinct names make sense?

If name is intentionally exactly the same then it should be inherited to avoid extra work for translators

matkoniecz:
Sorry for being annoying but it looks like some additional changes are needed. Especially those not needed `addTags`

data/presets/aerialway/_station.json
matkoniecz: @Geo-2695

Geo-2695: It seems these two tags are primarily intended for passenger-facing information, I think.🤔 and there are likely many non-passenger aerialways that would not require distinguishing between them

matkoniecz: https://github.com/openstreetmap/id-tagging-schema/blob/52c7c22e6eb765a53f91c662662c67368ed04b95/data/fields/aerialway/access.json

https://github.com/openstreetmap/id-tagging-schema/blob/52c7c22e6eb765a53f91c662662c67368ed04b95/data/fields/aerialway/summer/access.json

these seems likely to be applicable also for ones which are not public transport?

Even if you count ski lifts as public transport

Though I guess these may be confusing for mappers for ones that are not available to general public at all

matkoniecz:
Actually, is it considered preferable to tag public transport tags on then? Which one should be primary and findable preset?

Note that https://wiki.openstreetmap.org/wiki/Tag:aerialway%3Dstation does not mention such tagging at all (not checked other pages and history and prevalent tagging)

matkoniecz:
@Geo-2695 please, mark it as ready for review once, well, it will be ready for review

matkoniecz (commented):
Actually, is it considered preferable to tag public transport tags on then? Which one should be primary and findable preset?

Note that https://wiki.openstreetmap.org/wiki/Tag:aerialway%3Dstation does not mention such tagging at all (not checked other pages and history and prevalent tagging)

Geo-2695:
It would be preferable to add these tags. There are many passenger‑carrying, publicly‑accessible aerialway stations at present, and this helps prevent missing‑tag 
`public_transport` tag can be manually removed if needed, and no validation errors will occur

I noticed that the page at https://wiki.openstreetmap.org/wiki/Tag:public_transport=station seems to mention relevant usage.
I have also edited the wiki to add a reference to this tag

data/presets/aerialway/_station.json
matkoniecz: Note that quite many ski lifts are tagged without `public_transport=station` - and omitting these fields would block their display

I think that confusion from showing them on goods aerialways is smaller than them missing on ski lifts.

At least they should be in `moreFields`

matkoniecz (changes requested):
https://github.com/openstreetmap/id-tagging-schema/pull/2309/changes#r3823854973

data/presets/aerialway/_station.json
Geo-2695: indeed.

data/presets/aerialway/_station.json
```diff
@@ -0,0 +1,23 @@
+{
+    "icon": "temaki-gondola_lift",
+    "fields": [
+        "name",
+        "operator",
+        "address"
+    ],
+    "moreFields": [
+        "aerialway/access",
```
matkoniecz, line 10:
```suggestion
        "{public_transport/station}",
        "aerialway/access",
```

matkoniecz: or is there really nothing to inherit?

data/presets/aerialway/_station.json
```diff
@@ -0,0 +1,23 @@
+{
+    "icon": "temaki-gondola_lift",
+    "fields": [
+        "name",
+        "operator",
```
matkoniecz, line 5: if https://github.com/openstreetmap/id-tagging-schema/blob/44eff2b82b8bc3e8a2e1d2135a7c58c9b9a08dcf/data/presets/public_transport/_station.json#L31 is not inherited from...

is missing `network`, `building_area`, `{@templates/internet_access}` intentional?

data/presets/aerialway/_station.json
Geo-2695: May be needed👍

Geo-2695: But it could mean nothing

data/presets/aerialway/_station.json
Geo-2695: `{@templates/internet_access}`
might not be needed

data/presets/aerialway/_station.json
matkoniecz: are you sure that none of 

```
        "{@templates/contact}",
        "{@templates/internet_access}",
        "opening_hours",
        "air_conditioning",
        "baby_feeding",
        "changing_table",
        "gnis/feature_id-US",
        "level",
        "toilets",
        "toilets/wheelchair",
        "toilets/menstrual_products_poi",
        "wheelchair"
```

is relevant here? I would think that those can and should be in general inherited

matkoniecz (changes requested):
see comments, starting from https://github.com/openstreetmap/id-tagging-schema/pull/2309/changes#r3837833296

data/presets/aerialway/_station.json
Geo-2695: >      "{@templates/contact}",
>        "{@templates/internet_access}",
>        "opening_hours",
>        "air_conditioning",
>        "baby_feeding",
>        "changing_table",
>        "gnis/feature_id-US",
>        "level",
>        "toilets",
>        "toilets/wheelchair",
>        "toilets/menstrual_products_poi",
>        "wheelchair"
feels like this does not make sense
nearly all features that these tags are public transit stations, which should already have these fields included in the visible presets and adding them also feels redundant

matkoniecz: The problem here is that part of those ARE public transport (or at least ski lifts), just not tagged with those public transport tags, in such case these tags should be shown so inheritance makes sense

matkoniecz (approved):

matkoniecz:
Thanks! And additional thanks for persisting through review rounds!

Issue #2243, opened 2026-05-11
Silo, bunker silo should have own content list, not listing liquids


matkoniecz:
Not high priority, but nice to have

see https://github.com/openstreetmap/id-tagging-schema/pull/2241#issuecomment-4415314102

see https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dbunker_silo

tyrasd:
see also #2025

BrianMacIntosh:
Closely related to PR #2313

matkoniecz:
> [matkoniecz](https://github.com/matkoniecz)
removed
[good first issueGood for newcomers](https://github.com/openstreetmap/id-tagging-schema/issues?q=state%3Aopen%20label%3A%22good%20first%20issue%22)
[28 minutes ago](https://github.com/openstreetmap/id-tagging-schema/issues/2243#event-27020071522)


as PR exists

matkoniecz:
done as much as could be done in #2313

PR #2313, opened 2026-06-07
Better handling of Bunker silo and Silo contents, including new Crop Silo preset

Bjoern94:

This adds a new list of only bulk goods for use in the `content` field of bunker silos (Issue #2243). List sorted from usage as per taginfo. Also added some more relevant bulk goods based on usage from taginfo:

- Manure (4117)
- Grain (1086)
- Cement (448)

### Related issues

Issue #2025 (already merged)

### Links and data
[Taginfo `content=`](https://taginfo.openstreetmap.org/keys/content#values)

## Test-Documentation
Already tested locally; Linter passed

id_bunker_silo_1


data/fields/content_bulk.json
```diff
@@ -0,0 +1,16 @@
...
+    "type": "combo",
+    "label": "Content",
+    "strings": {
+        "options": {
+            "silage": "Silage",
+            "manure": "Manure",
+            "crop": "Crop",
+            "grain": "Grain",
+            "salt": "Salt",
+            "cement": "Cement",
```
matkoniecz, line 12: is cement kept in bunker silos? I expect it requires more separation from weather.

Maybe it is done in arid areas? Or with some cover?

matkoniecz:
> Checklist and Test-Documentation Template
> Read on to get your PR merged faster…

Can you do this testing and post screenshots?

data/fields/content_bulk.json
matkoniecz: Maybe this list should be (per its name) shared with silos? See https://github.com/openstreetmap/id-tagging-schema/issues/2243

Bjoern94: Hey, I don't think that cement is kept in bunker silos - my bad. But I think it is a good idea to share it with silos. Should I add the additional usage for silos here or in a separate PR? I can then do the testing.

Bjoern94:
As requested I changed the content for the silos to the new list. This should now solve Issue #2243. I added a "Crop Silo" subtype with the crop field added.
Search:
search_silo

Bunker silo:
bs_content

Normal silo (without field for crops):
silo_1

Crop silo (with field for crops):
crop_silo_1

Sorry for the bit messy commits.

BrianMacIntosh:
In the future, if you want to avoid the merge commits, you can use the [rebase](https://stackoverflow.com/questions/2472254/when-should-i-use-git-pull-rebase) option when you pull remote changes. That helps keep the history tidy.

There's also a feature called "interactive rebasing" that lets you squash multiple commits into a single one for the pull request. I don't know if I would try to use it here because I don't know if it will interact well with the merges, but something to know about.

data/fields/content_bulk.json
matkoniecz: in this PR would make sense as cement can be kept here in list

matkoniecz:
> Sorry for the bit messy commits. 

do not worry, squashing will be done at merge

data/presets/man_made/silo.json
```diff
@@ -1,8 +1,7 @@
 {
     "icon": "temaki-silo",
     "fields": [
-        "content",
-        "crop",
```
matkoniecz, line 5: what about cases where `crop=` is tagged and `content=` is not?

Maybe move `crop` field to `moreFields` so it would appear in such case?

I know that wiki does not recommend such tagging but it is nevertheless quite popular

matkoniecz: @Bjoern94 ping? Have you tested such case?

Bjoern94: I added crop to more fields. It now shows if crop was previously tagged.
silo_crop_tagged

matkoniecz:
So, lets test it before merge...

What we want to check:

Silo finds "Crop Silo" - which is not really being known as term... https://duckduckgo.com/?t=ffab&q=%22crop+silo%22&ia=web - but maybe such osmism makes sense? I will check it.

-----------------

what I asked was:

There are silos ( https://en.wikipedia.org/wiki/Silo ) that can hold variety of bulk material.

There are grain elevators ( https://en.wikipedia.org/wiki/Grain_elevator ) for specifically grain.

Is there term for subset of silos holding crop of some kind? But not necessarily grain?
Or is it not a thing and non-grain crops are never stored in them? Either because they can be stored in bulk they are grain by definition or because there is no such crop that would be stored?
Or maybe when non-grain crop is stored the silo is called "Grain elevator" anyway?

"Crop silo" clearly does not exist as a name.

matkoniecz:
I will keep name question open for now, and test other parts.

So we want so support

`man_made=silo` with ability to add content and crop.

`man_made=silo crop=` with at least ability to display `crop=`

`man_made=silo content=` with at least ability to display `content=`

if `man_made=silo content= crop=` is tagged then both are shown

`man_made=bunker_silo` should no longer list liquids

matkoniecz:
well, `man_made=silo` now still lists `manure` in dropdown which does not make sense there AFAIK but at least outright liquids are not listed so improvements are clear anyway

and not sure do we want more complexity here

Bjoern94:
> There are grain elevators ( https://en.wikipedia.org/wiki/Grain_elevator ) for specifically grain.

From what I understand (as a non native English speaker) grain elevator is used to describe the whole facility (Silos, unloading bays, truckscales,..). Silos for Crops are normally just called Silo because from the context/location it should be clear what is meant. Proposal to solve this: Instead of `Crop Silo` change it to `Silo (Crops)`?

Bjoern94:
> well, `man_made=silo` now still lists `manure` in dropdown which does not make sense there AFAIK but at least outright liquids are not listed so improvements are clear anyway

I will have a look into it as well (together with the proposed changes above)

matkoniecz:
> From what I understand (as a non native English speaker) grain elevator is used to describe the whole facility (Silos, unloading bays, truckscales,..).

yep, confirmed to me by native speaker of US English

> Proposal to solve this: Instead of Crop Silo change it to Silo (Crops)?

I like that! I will ask them for feedback

matkoniecz:
And got more feedback that "crop silo" may  be actually fine...

So lets go with it I guess?

matkoniecz:
@1ec5 maybe you have a guidance and time to give it...

matkoniecz (approved):

Bjoern94:
I took the time and went over the list again. I saw you approved the PR so I will wait with changes for your feedback if here in a separate PR.

> `man_made=silo` with ability to add content and crop.

That's the `content_bulk.json` and crops as `moreFields`. As I don't see a possibility to remove `manure` without creating a new field only for `bunker_silo` I would not want to change the options.  => check

> `man_made=silo crop=` with at least ability to display `crop=`

Here a small cleanup would be good as `content_bulk` and `building_area_yes` should be passed on from `silo`. Also what I found would be the term `Grain Bin`. I would add this at least in the terms field => todo

> `man_made=silo content=` with at least ability to display `content=`

=> check

> if `man_made=silo content= crop=` is tagged then both are shown

=> yes, when `content=crop` is present it changes the preset to `crop_silo` 

> `man_made=bunker_silo` should no longer list liquids

=> check - see above

matkoniecz:
From my side everything seems well, just trying to get final English advise/review

1ec5:
> Silo finds "Crop Silo" - which is not really being known as term... https://duckduckgo.com/?t=ffab&q=%22crop+silo%22&ia=web - but maybe such osmism makes sense? I will check it.

Google does find some results, mostly from FarmVille, but there are a handful of [serious results](https://extapps.dec.ny.gov/docs/lands_forests_pdf/eloump.pdf#page=36) from the U.S. I suspect it just falls through the cracks: American English has “silo” for the broader concept and would normally refer to something like a “grain silo” for a narrower concept. We don’t have a common word for a silo for any crop including non-grains – after all, which non-grain crops would be stored in a silo? So “crop silo” is a common-sense term that we’d use if we had to refer to the same contrived semantics as `content=crop`.

matkoniecz:

If you are looking for some easy and useful task [look here](https://github.com/openstreetmap/id-tagging-schema/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22). If you would try one of those and got stuck and existing documentation is insufficient - will free to ask for help!

matkoniecz:
And I am really happy that these got (mostly) fixed now in presets, as far as we can.

PR #2321, opened 2026-06-11
Allowed "junction=yes" on isolated nodes

fibonacci-matrix:

This is necessary for named junctions with complex geometry, often found on roads with separated carriageways

### Links and data

**Relevant OSM Wiki links:**
- [junction=yes](https://wiki.openstreetmap.org/wiki/Tag:junction%3Dyes)

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/tags/junction=yes#overview

### DIFF
--- a/data/presets/junction.json
+++ b/data/presets/junction.json
@@ -5,6 +5,7 @@
     ],
     "geometry": [
         "vertex",
+        "point",
         "area"
     ],
     "tags": {


matkoniecz:
Is allowing it on areas insufficient?

fibonacci-matrix:
> Is allowing it on areas insufficient?

Sometimes it's difficult to draw the exact area, or doing so would be subjective

Kovoschiz:
Is there an example? `junction=` is on the `highway=` for direct retrieval. Polygons have a bounded spatial relationship, and are often connected. Perhaps no application would use proximity. Cf `=speed_camera` , there would need to be eg `type=junction` / `type=site` similar to `=enforcement` to database-link all relevant sections. 
Following `=traffic_signals` on all 4 intersecting vertex, I would expect `junction=yes` to be duplicated. In Japan, it's associated with signalized intersections, thus considered together, and often without `junction=yes` from misunderstanding/mistranslation before. 
Rendering is a renderer problem. Collision is automatically handled, or they can easily try to cluster them. 
There's no documentation on this method https://wiki.openstreetmap.org/wiki/Tag:junction=yes

matkoniecz:
>  There's no documentation on this method https://wiki.openstreetmap.org/wiki/Tag:junction=yes

if that is tagging method accepted by community consensus it should be first added to wiki

if it is not it should not be added to wiki and to iD presets (even if it is a good idea)

if unsure - ask on some OSM chat or on forums - just be clear that it is about freefloating node, note vertex like already documented

matkoniecz:
do you plan to resolve that?

(I am not disputing or confirming whether it is a good tagging schema, just that iD should not be pushing this change if it is not supported by community)

fibonacci-matrix:
I'm not rushing this because if I don't present convincing arguments, a good idea might not appeal to others

matkoniecz:
as amount of code here is minimal and it is pending discussion - I will close it for now

when it will be clear that this is considered as a good tagging it can be reopened

PR #2323, opened 2026-06-11
Created separate preset for sushi fast foods

fibonacci-matrix:

There are many establishments where sushi is the primary dish and full table service is not provided.

### Links and data

**Relevant OSM Wiki links:**
- [cuisine=sushi](https://wiki.openstreetmap.org/wiki/Tag:cuisine%3Dsushi)

**Relevant tag usage stats:**
>https://taginfo.openstreetmap.org/tags/cuisine=sushi#combinations

### DIFF
--- a/data/presets/amenity/fast_food/sushi.json
+++ b/data/presets/amenity/fast_food/sushi.json
@@ -14,6 +14,7 @@
         "drive-in",
         "drive-through",
         "eat",
+        "food truck",
         "lunch",
         "restaurant",
         "table",
--- a/data/presets/amenity/fast_food/sushi.json
+++ b/data/presets/amenity/fast_food/sushi.json
@@ -10,10 +10,15 @@
         "dine",
         "dining",
         "dinner",
+        "drive thru",
         "drive-in",
+        "drive-through",
         "eat",
         "lunch",
-        "table"
+        "restaurant",
+        "table",
+        "takeaway",
+        "takeout"
     ],
     "fields": [
         "{amenity/fast_food}"
--- a/data/presets/amenity/fast_food/sushi.json
+++ b/data/presets/amenity/fast_food/sushi.json
@@ -1,0 +1,33 @@
+{
+    "icon": "temaki-temaki",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "breakfast",
+        "canteen",
+        "dine",
+        "dining",
+        "dinner",
+        "drive-in",
+        "eat",
+        "lunch",
+        "table"
+    ],
+    "fields": [
+        "{amenity/fast_food}"
+    ],
+    "moreFields": [
+        "{amenity/fast_food}"
+    ],
+    "tags": {
+        "amenity": "fast_food",
+        "cuisine": "sushi"
+    },
+    "reference": {
+        "key": "cuisine",
+        "value": "sushi"
+    },
+    "name": "Sushi Fast Food"
+}


matkoniecz:
> Checklist and Test-Documentation Template
> Read on to get your PR merged faster…

Can you do this testing and post screenshots?

fibonacci-matrix:
# Preview links & Sidebar Screenshots
## Example with a lot of tags
https://pr-2323--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=20.00/51.52129/9.86583&disable_features=boundaries&background=fr.ign.bdortho&id=n5625651428

## Example with very few tags
https://pr-2323--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=20.00/52.27556/8.04628&disable_features=boundaries&background=fr.ign.bdortho&id=n3097392190

## Do the dropdowns (etc.) work well and show helpful data




## Search
### Test the preset name as search terms


### Also test the preset terms and aliases as search terms (if present)










## Info-i

data/presets/amenity/fast_food/sushi.json
```diff
@@ -0,0 +1,33 @@
...
+    "terms": [
+        "breakfast",
+        "canteen",
+        "dine",
+        "dining",
+        "dinner",
+        "drive-in",
+        "eat",
+        "lunch",
+        "table"
```
matkoniecz: terms from general fast food entry may be worth replicating here: https://github.com/openstreetmap/id-tagging-schema/blob/a551d5cca39db5f19850e17d5f3e3fe585d53e1f/data/presets/amenity/fast_food.json#L50

fibonacci-matrix: @matkoniecz there are no terms here that are duplicated from the base template.
https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/restaurant/sushi.json

I copied this file from "Sushi Restaurant" and kept only the most relevant terms

matkoniecz: meant that it may be worth taking also relevant terms from fast food base entry such as "takeaway"

unless there is no such thing as sushi takeaway?

fibonacci-matrix: @matkoniecz added everything from the base preset except "food truck"

matkoniecz: why exclude food truck?

fibonacci-matrix: I have never seen sushi made in food trucks. I can add if you think it's better

matkoniecz: https://duckduckgo.com/?t=ffab&q=sushi+food+truck&ia=images&iax=images definitely finds some (and it is only search helper, the only difference is whether it will be listed when user searches "food truck" or similar)

fibonacci-matrix: Added "food truck"

matkoniecz:
I wonder whether we should try something to change order here or just do not worry about it

and if we want to change it, what would be the proper way of doing it

screen-2026-07-01-14-06-47

matkoniecz:
@fibonacci-matrix Would you judge putting sushi fast food before sushi restaurant as reasonable?

(after all, sushi restaurants were added first so maybe they are more typical places? but I do not know much about either)

fibonacci-matrix:
@matkoniecz considering Taginfo [statistics](https://taginfo.openstreetmap.org/tags/cuisine=sushi#combinations) (18 612 restaurants and 5706 fast foods), both presets deserve to exist. It's clear that fast_food appears above in the search list due to the use of lexicographical sorting. I'm not an expert on this topic, but I can easily provide many examples where less popular presets rank higher in search results than more popular ones. Therefore, in my opinion, this is a general issue rather than something specific to this PR, and it should be addressed in the search implementation itself by taking tag popularity into account, rather than being handled here.

fibonacci-matrix:
Just compare the search results for "pizza":

Pizza Restaurant - 72 016 elements
Pizza Fast Food - 60 453 elements
https://taginfo.openstreetmap.org/tags/cuisine=pizza#combinations

matkoniecz (approved):

matkoniecz:
Everything looks fine! I will do one more final test before merge when I am not tired

matkoniecz (approved):

matkoniecz:
Thanks for PR! Looks like it got in before release that is coming soon.

PR #2326, opened 2026-06-13
Move the files to improve layout formatting

Geo-2695:

Adjusted the layout and added two folders, `station` and `stop_position`,to store dedicated preset files.
Synchronized and updated the two `replacement` references.
However, these two references look rather odd, as they may mislead users into adding incorrect tags.

### Related issues

none

### Links and data

**Relevant OSM Wiki links:**
- …

**Relevant tag usage stats:**
> …


Geo-2695:
### Testing
This change does not involve modifying the effect of the iD editor.
<details>
<summary>reference</summary>

Screenshot_20260614_164936
Screenshot_20260614_140943

</details>

matkoniecz:
on rechecking - such move will throw away existing translations for them if I understood it correctly

not sure is it worth doing

PR #2327, opened 2026-06-14
Change icon of `aeroway=terminal` preset

Geo-2695:
The `pinhead-airport_terminal` icon is now used for airport terminals, replacing the original generic `maki-airport` icon.

### DIFF
--- a/data/presets/aeroway/terminal.json
+++ b/data/presets/aeroway/terminal.json
@@ -1,5 +1,5 @@
 {
-    "icon": "maki-airport",
+    "icon": "pinhead-airport_terminal",
     "fields": [
         "name",
         "operator",


Geo-2695:
### Testing

Screenshot_20260710_152000

matkoniecz:
current one

screen-2026-07-10-16-20-37

matkoniecz (approved):
I guess that in theory new icon may confuse a bit people mapping airport with very primitive infrastructure that still have terminals?

But seems unlikely

lets give people few days to review and comment

1ec5:
In iD, the current preset combines the airplane glyph with a reddish square representing a building. This seems to communicate the concept well in the sidebar (and on the map, where it’s inside the building area). However, it is technically possible to clear the Building field, as in https://github.com/openstreetmap/id-tagging-schema/pull/2327#issuecomment-4936285439, making the icon somewhat misleading. This is rare in practice – [93.30%](https://taginfo.openstreetmap.org/tags/aeroway=terminal#combinations) of `aeroway=terminal` combine with `building=*`.

The proposed icon would choose an icon that on its own more accurately represents a terminal building, although iD still combines it with the reddish square. Any idea if we still need that?

Terminal icon inside a building area.

matkoniecz:
> although iD still combines it with the reddish square. Any idea if we still need that?

I think it is not bad, should be kept for consistency (shop icons do the same).

And would require extra effort to remove.

tyrasd (approved):

Issue #646, opened 2022-11-19
Standalone webpage for browsing presets


1ec5:
We should create a standalone webpage for browsing id-tagging-schema presets that’s easy to use without knowledge of JSON or this repository’s structure.

## Background

In tagging or onboarding discussions, I often find myself using [the standalone iD preview on Netlify](https://ideditor.netlify.app/) or this GitHub repository’s search box to translate a real-world term into a tagging suggestion or vice versa. This workflow requires some insider knowledge about how the repository is structured or how to use iD’s raw tag editor.

People often refer newcomers to the OSM Wiki’s “[Map features](https://wiki.openstreetmap.org/wiki/Map_features)” page because it offers a high-level overview of all the available tags on one page, a better starting point than using the OSM Wiki’s search engine. Ideally, an iD user should be able to access all the tagging documentation they need from within iD, but sometimes people need to refer to tags in other contexts. Users of other editors may gravitate toward documentation that lives outside the editor.

Unfortunately, the “Map features” page is reaching the limit of MediaWiki’s capabilities, because it has to transclude content and links from a very large number of independent pages. This problem is essentially the same whether the page is based on transcluding disparate content pages or data items, because MediaWiki’s core content model is a document-oriented database. Taginfo can make popular tags available, but it only traffics in raw tags, and the documentation it extracts from the wiki has large gaps, especially outside of a handful of languages, because it ignores the translated descriptions in data items. Besides, the wiki doesn’t maintain names of features, only descriptions.

“Map features” is the product of a good deal of curation. In principle, one could create a listing similar to “Map features” using a simple SPARQL query in Sophox, but even if Sophox/sophox#31 were fixed, the listing would be as untidy as the wiki itself. This repository is also curating tags, but at a higher level of abstraction. The presets in this repository would be more educational to mappers than a table that only lists one tag for each feature, even when a tag may only makes sense in combination with other tags.

## Proposal

A simple standalone webpage would list each preset in this repository, organized by category, listing any aliases and the tags applied. Selecting a preset could show a brief description and image based on either taginfo or the linked data item, like in iD, with a link out to the wiki. There could be an option to set the language and region of the presented presets.

Publishing this webpage probably doesn’t require anything more complex than GitHub Pages. id-tagging-schema’s built package is already distributed over the jsDelivr CDN, so it could be just a static webpage that loads the latest JSON files from there. A static webpage can be deployed anywhere. It could even be implemented as a gadget or user script on the wiki itself.

[...]

PR #2328, opened 2026-06-15
Netlify preview: Add CORS

tordans:
Our previews do not set CORS and as such prevent apps from reading the internal JSON files. For the tagging schema browser I need those files to render the compare state, see https://github.com/openstreetmap/id-tagging-schema/issues/646#issuecomment-4704720025
Right now I work around it with some CORS-workaround-helper-service; but I think we can just open this up in the Netlify config.
https://docs.netlify.com/manage/routing/headers/

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,13 @@
+# Netlify custom response headers — https://docs.netlify.com/routing/headers/
+#
+# The PR-preview deploy publishes the repo root (`netlify deploy --dir=.`), so
+# the built schema JSON is served from `/dist/`. Allow external tools (tagging
+# schema browsers, diff/preview viewers, etc.) to fetch it cross-origin; without
+# this header the browser blocks the request and the fetch fails with a CORS
+# error. The published npm/jsDelivr copy already sends `Access-Control-Allow-Origin: *`.
+# Deploy: .github/workflows/deploy-preview.yml
+
+[[headers]]
+  for = "/dist/*"
+  [headers.values]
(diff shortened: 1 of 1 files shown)


matkoniecz:
How this PR can be tested whether it works as intended?

Intent seems 100% fine as far as I can see, this folder should be safe to be shared as far as I can see

tordans:
@matkoniecz it looks like we cannot test this here. The preview uses the main branch always, so this config it not used during preview build and this is likely the reason why the cors headers don't show up.

We would have to change the way we build the preview (*) or merge this and see if it works.

I think merging this is the cleaner/easier path forward.

---

(*) According to Cursor we should try changing the ref in `deploy-preview.yml`
```
- uses: actions/checkout@v7
  with:
    ref: ${{ github.event.workflow_run.head_sha }}
```

matkoniecz (approved):
I am trusting @tordans that in the worst case it fails and in such case they will report back

tordans:
Thanks @matkoniecz it looks to me this all works as it should.
The latest builds worked… and with a latest build we now get the compare UI in the tool: https://osmberlin.github.io/tagging-schema-browser/comparison?dataUrl=https://pr-2382--ideditor-presets-preview.netlify.app/dist/&locale=de&q=&page=1&per_page=24&sort=name_asc&primaryTagKey=[]&geometry=[]&iconPrefix=[]&iconName=[]&fieldIds=[]&categoryNames=[]&hasIcon=[] because it can read the files

PR #2331, opened 2026-06-15
Prioritize peak preset slightly over related presets for double-tagged features

tyrasd:
so that when there are other primary tags on the same node (e.g. man_made=cross, or tourism=viewpoint), the peak preset and icon take precedence.

e.g. https://www.openstreetmap.org/node/1976901207 has
```
man_made=cross
natural=peak
```

current/new:

### DIFF
--- a/data/presets/natural/peak.json
+++ b/data/presets/natural/peak.json
@@ -30,5 +30,6 @@
         "tip",
         "top"
     ],
-    "name": "Peak"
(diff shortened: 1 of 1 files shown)


matkoniecz (approved):

matkoniecz:
though in proper mapping I would expect it to be rather two separate nodes

PR #2343, opened 2026-06-19
Set public transport presets with no defined vehicle mode (bus, tram etc.) to be non-searchable

### LINKED ISSUE #2368: Presets related to "Transit" without specified transport mode should be non-searchable
Specific categories—such as `Ferry Platform` and `Train Station`—already exist within the iD presets; users should not use generic presets like "Transit Platform".

Geo-2695:

Set the presets for `public_transport=platform`,`platform_point`,`station` and `stop_position` to unsearchable. These entries have no specified type and are misleading. 
<details>
Besides, they are not rendered by OpenStreetMap-Carto.
</details>

### Related issues

Close #2368

### DIFF
--- a/data/presets/public_transport/_platform.json
+++ b/data/presets/public_transport/_platform.json
@@ -1,25 +1,25 @@
 {
-    "icon": "temaki-sign_and_bench",
+    "icon": "temaki-board_transit",
     "fields": [
-        "name",
-        "ref_stop_position",
+        "ref_platform",
         "network",
         "operator",
         "vehicles",
         "departures_board",
-        "shelter"
+        "surface"
     ],
     "moreFields": [
-        "bench",
-        "bin",
-        "gnis/feature_id-US",
+        "access",
+        "covered",
+        "indoor",
+        "layer",
         "level",
         "lit",
-        "tactile_paving",
         "wheelchair"
     ],
     "geometry": [
-        "point"
+        "line",
+        "area"
     ],
     "tags": {
         "public_transport": "platform"
@@ -31,6 +31,7 @@
         "transit",
         "transportation"
     ],
-    "name": "Transit Stop / Platform",
-    "matchScore": 0.6
+    "name": "Transit Platform",
+    "matchScore": 0.6,
+    "searchable": false
 }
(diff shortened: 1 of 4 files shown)


Geo-2695:
### Testing
**Before**
Screenshot_20260619_192152
Screenshot_20260619_192533

**After**
Screenshot_20260619_192229
Screenshot_20260619_192549

matkoniecz:
> Besides, they are not rendered by OpenStreetMap-Carto.

this has exactly zero relevance, at least as long as `highway=busway` is not rendered there

matkoniecz:
> These entries have no specified type and are misleading.

while I am hardly fan of so called PTv2 (both as proposed, and as used) to hide it from iD presets more justification is needed, I will not go around and hide bad tagging schemes supported by community

how this specific entries are misleading?

what you mean by "no specified type"?

Kovoschiz:
I do use and agree with `public_transport=` features. The author means they always have a mode, which already have their own presets ("bus station" etc). Therefore users shouldn't be creating those alone directly, only through mode-specific presets.

matkoniecz:
@Geo-2695 would it make sense to retitle this PR to say "Set public transport presets with no defined vehicle mode (bus, tram etc.) to be non-searchable." or something similar?

Geo-2695:
> @Geo-2695 would it make sense to retitle this PR to say "Set public transport presets with no defined vehicle mode (bus, tram etc.) to be non-searchable." or something similar?

Indeed, this title is much better than the old one.

Geo-2695:
> Besides, they are not rendered by OpenStreetMap-Carto.

When using presets for specific modes ("Train Platform","Tram Station" etc.), OpenStreetMap-Carto will generally render them accordingly.
<details>
PLATF
*The rendering effect of "Train Platform" in OpenStreetMap-Carto*
</details>

However, presets without a defined transport type ("Transit Platform","Transit Station" etc.) that only carry the `public_transport` tag, do not appear on the map at all in OSM-Carto.
<details>
public
CA

*The "Transit Platform" preset is not rendered*
</details>

matkoniecz:
1) Whether it renders in OSM Carto is basically irrelevant

2) `railway=platform` is rendered, `public_transport=platform train=yes` is ignored by Carto

matkoniecz:
I was worried about translation being invalidated as result of file rename but the underscores in the file names are ignored for the translations keys.

matkoniecz (approved):

PR #2360, opened 2026-06-22
Added new preset for emergency suction_point

### LINKED ISSUE #94: Include emergency=suction_point?
See https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dsuction_point . A preferred point where firefighters can suck water out of a body of water.

The tag was already documented in 2011. Starting 2015‎, there were some discussions whether
`emergency=suction_point` or `emergency=fire_hydrant` + `fire_hydrant:type=suction(_point)` should be used.

Till then, the mappers seemed to very much vote by their feet and favoured `emergency=suction_point` over the hydrant-tagging. Also the discussion seemed to favour `emergency=suction_point`, arguing that just some point in a body of water does not qualify as a hydrant.

Note that the below graphic shows also `fire_hydrant:type=suction`. There are actually real hydrants that work like this, so an unknown number of the orange line are not actually an "alternative tagging to emergency=suction_point" but just a hydrant that works this way.

taghistory(7)
https://taghistory.raifer.tech/#***/emergency/suction_point&***/fire_hydrant:pressure/suction&***/fire_hydrant:pressure/suction_point

Bjoern94:

This solves an older Issue (#94) by adding a preset for `emergency=suction_point`. Since the ticket was opened in 2020 the usage steadily increased to now 19676. For now I used the same Icon (maki-water) as for the closely related `water_reservoir`.

### Related issues

Closes #94 

### Links and data

**Relevant OSM Wiki links:**
- [OSM Wiki](https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dsuction_point)

**Relevant tag usage stats:**
- [Taginfo (19676x)](https://taginfo.openstreetmap.org/tags/emergency=suction_point) 

<details>

[Example node](https://www.openstreetmap.org/node/8602884617#map=19/49.802018/8.481381)
suction_point_tagging

</details>

### Search

<details>

Search name:
suction_point_search_1

Search term:
suction_point_search_2

</details>

### Info-`i`

<details>
suction_point_info
</details>

### Wording

- Drafting Site
- terms
  - pond
  - reservoir
  - suction point


data/presets/emergency/suction_point.json
```diff
@@ -0,0 +1,24 @@
...
+    ],
+    "moreFields": [
+        "description",
+        "operator"
+    ],
+    "geometry": [
+        "point"
+    ],
+    "terms": [
+        "pond",
```
paulklie: How about "water" or "firefighting" as terms?

data/presets/emergency/suction_point.json
```diff
@@ -0,0 +1,24 @@
...
+        "reservoir",
+        "suction point"
+    ],
+    "tags": {
+        "emergency": "suction_point"
+    },
+    "name": "Drafting Site"
```
matkoniecz: @1ec5 @ZeLonewolf 

Do you know about a good term in US English for those?

Or even a short description?

my attempts failed to find any and test users entirely rejected "Drafting Site"

matkoniecz: Got feedback that "firefighting suction point" may make sense

matkoniecz: Overall, source of annoyance is that this does not really have name in nonspecialist language, descriptive names are too broad and does not appear as term in US English according to narrow pool of people I asked.

matkoniecz: But "Drafting site" sounds rather like place offering [tutoring to people learning to be an architect](https://en.wikipedia.org/wiki/Technical_drawing) or something similar...

1ec5: These features are extremely obscure. In the jargon of the firefighting industry, this is a “[drafting site](https://www.firefighternation.com/lifestyle/drafting-tools-techniques-2/)”, “[draft site](https://codelibrary.amlegal.com/codes/san_diego/latest/sandiego_regs/0-0-0-135436)”, “[fire site](https://www.firefighternation.com/firerescue/a-water-shuttle-drafting-drill/)”, “[filling site](https://www.usfa.fema.gov/downloads/pdf/publications/fa-248.pdf#page=104)”, or “[fill site](https://www.fireapparatusmagazine.com/equipment/highway-hydraulics-main-street-to-country-roads-part-3-routing-and-fill-sites/)”. Of these terms, “drafting site” appears to be the most common. To draft means to draw water, like from a well, but “drafting” can be confusing to a layperson out of context. To me it almost sounds like where you report for military conscription (the draft). How about something like “fire draft site” or “firefighting fill site” to avoid that particular confusion?

I would avoid using anything involving “suction”. That has nothing to do with this feature. Whoever calqued the raw tag value from German didn’t realize that most English speakers only know the word “suction” from suction cups. Otherwise, if the user makes the connection to “suck”, then it becomes a bit impolite. 😉

/ref https://github.com/openstreetmap/openstreetmap-website/pull/5029#discussion_r1694759488

matkoniecz: "firefighting fill site" seems to use the simplest words and I like it most...

(but I am not a native speaker of English)

> I would avoid using anything involving “suction”. That has nothing to do with this feature. Whoever calqued the raw tag value from German

I guess it is too late deprecate Germanglosmish tag value here. Especially as there is no obvious English term.

1ec5: @rsavoye knows a lot about drafting sites as a rural volunteer firefighter. He might have additional ideas about how to describe them succinctly and intuitively to laypeople.

Bjoern94: Hey, sorry for the late replay - I was away for 2 weeks. For Germany the term "Saugstelle" (suction point as per deepl) is used. This is also the term written on traffic signs to mark them to prohibit parking etc. How would such a point be marked in America on a traffic sign? Maybe this could be a term understood by many people?

1ec5: Indeed, _Saugstelle_ literally means “suction point”. Unfortunately, that isn’t a real word, just a last resort for the translation model. Maybe whatever we choose here will eventually result in an improvement. 🙂

A drafting site is a location on a stream or body of water. So far, I haven’t heard of any country marking these locations with an explicit sign except Germany and Russia.

In the U.S., we only have signs for standpipes (risers) connected to buildings and sometimes for fire hydrants, but as far as I know never for outdoor water sources. Whenever a locality doesn’t want parked cars to obstruct fire department access, they put up a “No Parking: Fire Lane” sign. But a fire lane doesn’t necessarily have anything to do with a water source. It’s more likely a staging point for the fire engine.

Bjoern94: Thank you for the insight. Just as an idea: would something like "Emergency water drafting site" or "Emergency drafting site" work (or even be correct english)? This would include the main tagging group (emergency) for this feature as well as the "theme" (water, drafting).

1ec5: Thanks, both suggestions sound more valid in English than “suction point”. We would be relying on users to understand “drafting” so that they don’t confuse it with whatever we would have to call [`emergency=fire_water_pond`](https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dfire_water_pond). Would “fire” be too specific for _Saugstelle_?

Incidentally, I stand corrected about the signs. [This paywalled article](https://www.fireengineering.com/firefighting-equipment/water-supply-preincident-intelligence-1477/) includes several photos of “Fire Dept. Drafting Site – No Parking Anytime” signs in the vicinity of Witmer, Pennsylvania. So at least one local fire department signposts them with this term. Others have [installed](https://www.gotbigwater.com/news/535/Dry-Fire-Hydrant-Design-49---Keeping-Out-Livestock---Letting-in-the-FD-) [drafting hydrants](https://www.primosupply.com/no-parking-sign-drafting-fire-hydrant-12-x-18/hf-sign-npdfh) so the signs might say that or simply “[water source](https://www.alamy.com/stock-photo/fire-hose-sign-fire-hose.html?pseudoid=9755ED22-3135-494B-8FFB-9770EEA61160)”. I did find one county in Oregon that posts “[suction water supply source](https://www.turnerfire.com/content/files/M_C%20App%20Guide%207-2024%283%29.pdf#page=19)” signs, but only at underground cisterns (`emergency=water_tank`).

Bjoern94: If I understand you correct you would suggest "Fire drafting site", correct? I Could life with that but would prefer "firefighting" or "fire department" instead. 
Also it might be worth considering _point_ instead of _site_ as to further differentiate between the specific point to place your engine/pump (`suction_point`; only valid for nodes) and the body of water your drafting from (`fire_water_pond`; nodes and areas).
So it would be something like: `Firefighting (water) drafting point` and `Firefighting water pond`

1ec5: So far I haven’t seen these officially described as “points”, since there seems to be some tolerance when there’s no fixed hydrant. But since we’re departing from official terminology anyways, I think it would be intuitive enough.

Bjoern94: One thing I noticed when testing: The `emergency=water_tank` which is very similar if not even the same as `emergency=fire_water_pond` already has a preset and is named _Emergency Water Reservoir_ or  _Emergency Water Reservoir (Underground)_ - so emergency might be better for contingency?

k-yle: we call these a _"Static Water Supply"_ [in New Zealand](https://www.fireandemergency.nz/building-fire-safety/building-and-designing-for-fire-safety/firefighting-water/). likewise [in Austr](https://www.rfs.nsw.gov.au/plan-and-prepare/static-water-supply-sws2)[alia](https://www.fire.nsw.gov.au/fire-safety/home-fire-safety/topics/static-water-supply), although I suspect most people will have never heard of this, unless they have [an "SWS" sign](https://www.facebook.com/2aaa.fm/posts/what-a-top-idea-sws-means-static-water-supply-telling-firies-you-have-a-pool-or-/3612696558769879/) on their letterbox

1ec5: Yes, it turns out that “emergency water” would be much more correct than [fire water](https://en.wikipedia.org/wiki/Firewater_(fire_fighting)).

Bjoern94: Ok, so I would suggest the following:
- Emergency Water Drafting Point (`emergency=suction_point`
- Emergency Water Reservoir (`emergency=water_tank`)
- Emergency Water Reservoir (Underground) (`emergency=water_tank`)

A bit off-topic: Reading the wiki page for `emergency=water_tank` I noticed that it also considers open waters (ponds) and has a pond as the main example image [Wiki](https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dwater_tank). For me a `water_tank` is a covered thing - especially to distinguish it from `emergency=fire_water_pond`. Is this something you would agree on? 
If so I am considering to bring this up in the forum and get a wider response and potentially change the wiki (and then the respective presets here) accordingly. Until then I would wait with changing the name and terms.

1ec5: The page appears to describe `emergency=water_tank` as a purely functional tag, to be combined with a physical tag such as `man_made=reservoir_covered` (aboveground), `natural=water` (surface), or `man_made=storage_tank` (also aboveground). The preset Emergency Water Tank unconditionally combines `emergency=water_tank` with `man_made=storage_tank`, so I guess the terminology works out. Maybe we need additional presets for the other combinations though.

https://github.com/openstreetmap/id-tagging-schema/blob/3c780a1910a6562ad0bbdd0bf907a1e68df52c7c/data/presets/emergency/water_tank.json#L27-L30

Bjoern94: I updated the code as described above.

data/presets/emergency/suction_point.json
```diff
@@ -0,0 +1,26 @@
...
+    "terms": [
+        "drafting",
+        "emergency",
+        "firefighting",
+        "water"
+    ],
+    "tags": {
+        "emergency": "suction_point"
+    },
+    "name": "Emergency Water Drafting Point"
```
matkoniecz, line 25:
```suggestion
    "name": "Firefighting Fill Site"
```

???

matkoniecz: I will not argue strongly for it but feels like best one to me. 

What you think about it now?

matkoniecz: @Bjoern94

PR #2244, opened 2026-05-11
Add preset for boundary=protected_area

k-yle:
### Related issues

Closes #1511, Closes #12 

### Links and data

- [wiki](https://osm.wiki/Tag:boundary=protected_area)
- [taginfo](https://taginfo.osm.org/tags/boundary=protected_area)

**Relevant tag usage stats:** 134679

<details><summary><h1>Test-Documentation</h1><sup>(click to expand)</sup></summary>

### Preview links & Sidebar Screenshots

image

### Search

image

### Info-`i`

image

image

### Wording

- [x] American English
- [x] `name`, `aliases` (if present) use Title Case
- [x] `terms` (if present) use lower case, sorted A-Z

PR #2376, opened 2026-06-25
code review changes for `protect_class`

k-yle:
follow up to https://github.com/openstreetmap/id-tagging-schema/pull/2244#pullrequestreview-4568739951

# Test Documentation

image

image

### DIFF
--- a/data/fields/protect_class.json
+++ b/data/fields/protect_class.json
@@ -1,7 +1,7 @@
 {
     "key": "protect_class",
     "type": "combo",
-    "label": "Protection Class",
+    "label": "IUCN Protected Area Management Category",
     "strings": {
         "options": {
             "1a": "1a: Strict Nature Reserve",
@@ -13,5 +13,5 @@
             "6": "6: Protected area with sustainable use of natural resources"
         }
     },
-    "autoSuggestions": true
+    "autoSuggestions": false
 }


data/fields/protect_class.json
```diff
@@ -1,7 +1,7 @@
 {
     "key": "protect_class",
     "type": "combo",
-    "label": "Protection Class",
+    "label": "IUCN Protected Area Management Category",
```
matkoniecz, line 4: not sure whether mappers care and should be made so strongly aware of from where this magic numbers are coming, but I am not strongly opposed either

1ec5: I figure that we should be transparent about what the numbers are supposed to be. But I also don’t think it’s strictly necessary to show the numbers in the labels. We hardly do that anywhere else.

matkoniecz: @k-yle What you think about potentially hiding numbers instead?

k-yle: i don't have a strong opinion, but i think they're useful to save time, just like [`mtb:scale`](https://osm.wiki/Key:mtb:scale). otherwise you need to waste time reading all the (unsorted) options

matkoniecz: oh, so it would be useful for more advanced mappers doing those over and over again

I guess it makes sense

matkoniecz (approved):

PR #2381, opened 2026-06-26
Add Planet include to exclude-only locationSets (temporary workaround)

### LINKED ISSUE #12491: Road Marking field appears twice
### URL

_No response_

### How to reproduce the issue?

Hello,
The “Road Marking” field appears twice.
Best regards

Image

### Screenshot(s) or anything else?

_No response_

### Which deployed environments do you see the issue in?

_No response_

### What version numbers does this issue affect?

2.41.1

### Which browsers are you seeing this problem on?

Firefox

nathanpixodeo:
﻿### Description, Motivation & Context

location-conflation v3 now treats locationSet objects without an include as invalid. Several schema entries used an exclude-only locationSet, which previously meant all locations except the excluded ones. When those excludes are ignored, country-specific crossing marking fields can appear alongside the global field.

This adds an explicit include: ["Planet"] to each exclude-only locationSet, preserving the existing scope while making the data valid for the newer parser.

### Related issues

Related: openstreetmap/iD#12491

### Links and data

**Relevant OSM Wiki links:**
- N/A; this is a schema locationSet compatibility fix, not a tagging change.

**Relevant tag usage stats:**
> N/A

### Testing

- npm test
- npx prettier --check on the changed JSON files
- Checked with a JSON parser script that no locationSet with exclude and without include remains under data/

### DIFF
--- a/data/fields/bicycle_road.json
+++ b/data/fields/bicycle_road.json
@@ -12,6 +12,9 @@
         "neighborway"
     ],
     "locationSet": {
+        "include": [
+            "Planet"
+        ],
         "exclude": [
             "BE",
             "NL"
--- a/data/fields/crossing/markings.json
+++ b/data/fields/crossing/markings.json
@@ -36,6 +36,9 @@
         "ladder:paired": "iD-crossing_markings-ladder_paired"
     },
     "locationSet": {
+        "include": [
+            "Planet"
+        ],
         "exclude": [
             "BG",
             "DE",
--- a/data/fields/crossing/markings_yes.json
+++ b/data/fields/crossing/markings_yes.json
@@ -6,6 +6,9 @@
     "stringsCrossReference": "{crossing/markings}",
     "iconsCrossReference": "{crossing/markings}",
     "locationSet": {
+        "include": [
+            "Planet"
+        ],
         "exclude": [
             "BG",
             "DE",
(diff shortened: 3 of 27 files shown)


matkoniecz:
@nathanpixodeo Can you do spot check using "🍱 [Your pull request preview is ready](https://pr-2381--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=18.00/48.841708/2.587656)" and post some screenshots?

tordans:
Are sure this is how we want to solve this? My first thought was that maybe an upstream change would be good because this seems like a change that does not really add a value. 
The other open is to add those in the schema-builder. 
I don't see value in adding them here...

I suggest to wait for Marin's input on which direction to take. 

But maybe someone want to ask upstream about this...

tordans:
FYI lets move the location conflation thread to https://github.com/openstreetmap/iD/issues/12496 to keep it easier to follow along.

nathanpixodeo:
## Test-Documentation

Spot-checked the PR preview with `post_box:type` examples:

- France, `n7985389798`, `amenity=post_box` with `post_box:type=lamp`: the sidebar shows the affected **Type** field.

France post box type field

- Great Britain, `n452215413`, `amenity=post_box` without `post_box:type`: the sidebar still offers the **Type** field in the preview.

GB post box type field

Given that result, I agree this probably needs Marin/upstream/schema-builder direction before pushing this pattern further. I will wait for that input rather than expanding the data-only change here.

matkoniecz:
I will merge it with intention of it being temporary workaround, with proper fix in schema builder. - which will enable revert here

matkoniecz (approved):

matkoniecz:

PR #2358, opened 2026-06-22
Change addr: neighbourhood placeholder to English (US) spelling, change takeaway to takeout to use English (US)

mugdhachalla:
### Description

Updates the addr:neighbourhood placeholder to use American English spelling.

### Related issues

Closes #2284 and closes https://github.com/openstreetmap/id-tagging-schema/issues/2344

### Links and data

**Relevant OSM Wiki links:**

* N/A

**Relevant tag usage stats:**

> N/A

PR #2387, opened 2026-06-27
Add plural presets names to terms

### LINKED ISSUE #1652: Add plurals as search terms
Most of id-tagging-schema’s clients are editors, in which the user is likely to search for a preset by a singular term such as “playground”, because they’re only mapping one playground at a time. But there are also times when the user thinks of a preset as a collection of features. We should add the plural form of the existing names to each preset’s search terms, at least in the default English localization.

For example, an editor could offer preset-based filters with human-readable names. A filter named “Playground” suggests that there’s only one to show, whereas “Playgrounds” refers to any number of playgrounds. Overpass turbo’s wizard accepts preset names. Often, when I show people this capability, they first try to search for something like “playgrounds” or “benches” and the wizard returns an error about not recognizing that term.

Some languages have different pluralization rules or use plurals in different context. It would be up to each localization to include plural forms or not.

matkoniecz:

fixes #1652

note: it was done with LLM generated script, but reviewed for problems and went through several iterations

can share that code but it is (as expected) horrific

[added_plurals.csv](https://github.com/user-attachments/files/29408590/added_plurals.csv) may be useful to review, easier to read than diff

see https://patch-diff.githubusercontent.com/raw/openstreetmap/id-tagging-schema/pull/2387.patch for full text diff

### Related issues

fixes #1652

### Links and data

**Relevant OSM Wiki links:**
- …

**Relevant tag usage stats:**
> …

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>


matkoniecz:
# known complex cases:
## " of " handling
"place of worship" should be changed to "places of worship"
"kingdom hall of jehovah's witnesses" should be changed to "kingdom halls of jehovah's witnesses"

## Fictional plurals
"childcare" has no real plural, "childcares" has a different meaning and does not refer to multiple childcare locations
"Police" must not be falsely pluralized as "polices"
"Greenhouse Horticulture" - "Horticultures" is not in real use 
"Boat Storage" - if it has plural form "Boat Storages" is not one of them
"Bungee Jumping" - "Bungee Jumpings" is not valid, in this context has no plural
"Bicycle Parking" - "bicycle parkings" is invalid. The same for "Motorcycle Parking", "Street-side Parking", "Underground Parking"
"golf" - "golves” is not an English plural of golf
"Horseback Riding" should not be pluralized to "Horseback Ridings"
"Snowshoeing / Winter Hiking Trail" - "snowshoeings" is not a valid plural

## multiple terms
"obstetrician / gynecologist" should be changed to two entries added to presets: "obstetricians" and "gynecologists", each added on its own
"nursery / childcare" should be changed to one entry added to presets: as "childcare" gets no plural - with each term changed on its own
But "Minor / Unclassified Road" should be pluralized as "Minor / Unclassified Roads"
"Anime / Manga Shop" should be pluralized to "Anime / Manga Shops"
"Train Station (Halt / Request)" should be pluralized as "Train Stations (Halt / Request)"
"Dojo / Martial Arts Academy" should be pluralized as "dojos" and "martial arts academies"
"Track / Land-Access Road" should be pluralized as "tracks / land-access roads"
"Fish Farm / Hatchery" should be pluralized as "fish farms / hatcheries"
"Gym / Fitness Center" should be pluralized as "gyms / fitness centers"
"Sports Center / Complex" should be pluralized as "sports centers / complexes"
"Attached Rock / Boulder" should be pluralized as "attached rocks / boulders"
"Guest Apartment / Condo" should be pluralized as "guest apartments" and "condos"
"Wet Dock / Dry Dock" should be pluralized as "wet docks" and "dry docks"
"Snowsports Trail / Piste (Unspecified Type)" should be pluralized as "Snowsports Trails / Pistes (Unspecified Type)"
"Test Prep / Tutoring School" should be pluralized as "Test Prep / Tutoring Schools"
"Radio / Electronic Component Store" should be pluralized as "radio / electronic component stores"
"Preschool / Kindergarten Building" should be pluralized as "preschool / kindergarten buildings"
"Preschool / Kindergarten Grounds" should be pluralized as "preschool / kindergarten grounds"
"Bedding / Mattress Store" should be pluralized as "bedding / mattress stores"
"Property Management / Leasing Office" should be pluralized as "property management / leasing offices"
"Cycle Crossing With Traffic Signals" should be pluralized as "cycle crossings with traffic signals"

## brackets

"needleleaved tree (evergreen)" should be changed to "needleleaved trees (evergreen)"

## Other cases where pluralized word is not at the end
"Road Under Construction" should be pluralized as "Roads Under Construction"
"Orchard Planted in a Meadow" should be pluralized as "Orchards Planted in a Meadow" or not at all

## Snobbery

"château" is pluralized "châteaux"

## Multiword mess

"Baking Oven" is pluralized as "Baking Ovens"
"Bicycle Shed" is pluralized as "bicycle sheds"
"Communal Kitchen" is pluralized as "communal kitchens"
"Senior Assisted Living Facility" is pluralized as "Senior Assisted Living Facilities"
"Radar Speed Sign" is pluralized as "Radar Speed Signs"
"Household Linen Shop" is pluralized as "household linen shops"

## naive solution fails for other reasons

"nursery / childcare" should not be changed to "nursery / childcares"
"bubble tea cafes" should not be changed to "bubble tea caves"
"mounted binoculars" should not be changed to "mounted binocularses"
"driver training grounds" should not be changed to "driver training groundses"
"research institute grounds" should not be changed to "research institute groundses"
"drinking water" should not be changed to "drinking waters"
"condo" is pluralized as "condos"

## Actually has plural
"Artwork" has plural "Artworks"
"Historic Aircraft" has plural "Historic Aircrafts"

bhousel:
Why do this? 

In almost all cases the user would have already typed their way through the singular form and would have seen the preset they are looking for then? 

And even if they somehow type "buildings" without first seeing "building", the levensthein distance is so small that the preset they are looking for should be the top match?

matkoniecz:
https://github.com/openstreetmap/id-tagging-schema/issues/1652 has some justification

> And even if they somehow type "buildings" without first seeing "building", the levensthein distance is so small that the preset they are looking for should be the top match?

this does not always work, especially when user uses substring rather than full term as specified

but overall I have some doubts do we really want it (I again underestimate complexity of "simple" grammar rules in a natural language, I expected some exceptions and complexity but...)

RudyTheDev:
I feel like this adds a lot of extra maintenance for only a tiny benefit given that the fuzzy search should match plurals to singular forms except some very rare exceptions, but I cannot even think of any examples (words like "child"/"children" or "person"/"people" don't seem to appear in any preset names.) And this would only work on English anyway. I don't think this approach is a solution to #1652, which is generally suggesting that clients just haven't implemented a fuzzy search like the mentioned levenshtein or something. I think that would need a separate field for clients that need a dedicated plural form, rather than mix it with regular terms/aliases that already match the plural forms.

tyrasd (approved):
it doesn't hurt to have them as well 👍

1ec5:
> Why do this?

As I mentioned in https://github.com/openstreetmap/id-tagging-schema/issues/1652#issue-3287513810, my main motivation is Overpass turbo. Every time I demo the wizard, people try to search for things like “buildings” and “benches”. There isn’t any autocompletion for the wizard.

bhousel:
> As I mentioned in [#1652 (comment)](https://github.com/openstreetmap/id-tagging-schema/issues/1652#issue-3287513810), my main motivation is overpass turbo. Every time I demo the wizard, people try to search for things like “buildings” and “benches”. There isn’t any autocompletion for the wizard.

This sounds like an issue to be raised with overpass turbo?

1ec5:
> This sounds like an issue to be raised with overpass turbo?

tyrasd/overpass-turbo#814 tracks autocompletion in the Overpass turbo wizard. I agree that it could help in English, less so in some other languages, but maybe preset names shouldn’t be tied to a particular editor affordance if the project is intended to be used more widely.

I can see how we might not want to mix the singular and plural forms in the same field. “Trees” is a good search term when querying OSM for `natural=tree`s but not a good name for a preset about a specific `natural=tree`. Fortunately, this PR is only adding the plural forms to terms, not aliases.

> And this would only work on English anyway.

Initially, yes, but everything gets translated eventually. On Transifex, some of the localization teams are already doing something similar: https://github.com/openstreetmap/iD/issues/3159#issuecomment-229955476 https://github.com/openstreetmap/iD/issues/8242#issuecomment-742220871. I do want to call out that merging this PR would massively reduce the completion rate of most localizations on Transifex. This would be an opportunity for some localizations to add the plural forms, but for others it would be mindless, repetitive work.

RudyTheDev:
I would argue that this should be a separate field. It would make it much clearer to any consumers. Translators would then know exactly what "alias" is intended. Search engines could either use it or skip it. UI could use it for displaying items. Overpass and such could accept it for natural language searches. Etc.

bhousel:
It's weird that this PR is even being considered.  

IMO #1652 should be closed as wontfix and you should have had your LLM fix the overpass turbo problem instead of auto-pluralizing every string in the presets as a workaround. It's a mistake to push this problem off onto your volunteer translators!

matkoniecz:
I want to clarify translation situation. As far as I know changing terms in base language (US English) does NOT invalidate translations of terms for other languages.

screen-2026-07-07-16-16-07

https://github.com/openstreetmap/id-tagging-schema/pull/2371 changed terms which have become listed as "Developer Notes"

It has not invalidated translation, see

screen-2026-07-07-16-16-59

base translation is set in this case to `<translate with synonyms or related terms for 'Bicycle Parking', separated by commas>` to prevent exactly this problem

otherwise changes like #2371 would be causing massive damage rather than potentially bringing tiny itty bitty improvement

matkoniecz:
Contrast

screen-2026-07-07-16-21-44

where #2358 invalidated translations

1ec5:
My bad, you’re right that this wouldn’t invalidate any translations then. So the only mechanism we have to encourage translators to provide plurals is to message them on Transifex. They won’t have any built-in way to track their progress in adding them.[^vi]

In my view, the things in `terms` are often so tangential to what the preset is actually called in real life that I don’t think this is going to cause significant disruption among consumers or users. To pick a random example, no one calls a dentist’s office [“a tooth” or even “a teeth”](https://github.com/openstreetmap/id-tagging-schema/blob/7f642043effe38a701ed0a77d070ef3aec367c12/data/presets/amenity/dentist.json#L20).[^overpass]

These would be good arguments for splitting plurals out as a separate field. After all, a hypothetical autocomplete for Overpass turbo’s wizard shouldn’t come up with “teeth” either. If we do that, we need to somehow accommodate languages that have no plural form or that also have a dual form etc.

Honestly, it would be a pain to review this massive PR manually. I would be surprised if there are many mistakes, but sometimes an English word has a different plural depending on the meaning. Maybe a spreadsheet would be easier to review.

[^vi]: When I did the diacritic-folding thing for Vietnamese, I had to download the translation file and write a script to detect missing diacritic-folded terms. That absolutely could be done in software, but the decision was made not to introduce language-specific logic into iD.
[^overpass]: This is not just about Overpass turbo. Projects like Cartes.app have been hard-coding POI category names because they need plurals.

RudyTheDev:
> Honestly, it would be a pain to review this massive PR manually [..] Maybe a spreadsheet would be easier to review.

see

> added_plurals.csv may be useful to review, easier to read than diff

PR #2395, opened 2026-06-28
Remove inappropriate fields and terms from the `attraction=swing_carousel` preset

### LINKED ISSUE #2394: The preset for `attraction=swing_carousel` contains incorrect content
See also https://github.com/openstreetmap/id-tagging-schema/data/presets/attraction/swing_carousel.json

The preset should not include `height`, `tower`, or any other tower-related contents. Since thispreset represents an amusement ride, adding these contents is meaningless - no one would associste them with tower-related info when editing

Geo-2695:
Close #2394

### DIFF
--- a/data/presets/attraction/swing_carousel.json
+++ b/data/presets/attraction/swing_carousel.json
@@ -1,8 +1,7 @@
 {
     "icon": "temaki-tower",
     "fields": [
-        "{attraction}",
-        "height"
+        "{attraction}"
     ],
     "moreFields": [
         "{attraction}",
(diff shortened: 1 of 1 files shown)


Geo-2695:
AA
S

data/presets/attraction/swing_carousel.json
```diff
@@ -16,9 +15,7 @@
     "terms": [
         "amusement ride",
         "carousel",
-        "carousel tower",
```
matkoniecz, line 19: from brief check "carousel tower" is in fact used for those (based on googling this term on DDG and Google)

it also seems reasonable term

why it is inappropriate term here?

data/presets/attraction/swing_carousel.json
```diff
@@ -1,8 +1,7 @@
 {
     "icon": "temaki-tower",
     "fields": [
-        "{attraction}",
-        "height"
```
matkoniecz, line 5: why `height` is inappropriate term?

PR #2412, opened 2026-07-01
Netlify preview: Add CORS for iD Staging deploy

tordans:
Follow up to https://github.com/openstreetmap/id-tagging-schema/pull/2328

With this we can use the latest release from iD to reference the latest id-tagging-schema in the schema browser.

See also https://github.com/openstreetmap/iD/pull/12515 for the other side of this.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -11,3 +11,11 @@
   for = "/dist/*"
   [headers.values]
     Access-Control-Allow-Origin = "*"
+
+# Staging deploys bundle id-tagging-schema under /id-tagging-schema/dist/ when
+# publishing iD to ideditor.netlify.app. Deploy: .github/workflows/staging.yml
+
+[[headers]]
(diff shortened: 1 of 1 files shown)


matkoniecz (approved):

tyrasd:
With 1777979853df8ab01c1ab8ad7df2efde58988ff2 the `dist` folder is available shortly after every commit in the `interim` branch, which is then available via gh-pages deployment including CORS headers (e.g. https://openstreetmap.github.io/id-tagging-schema/dist/fields.json).

I think this can be reverted now, as it should not needed anymore (#2328 is still required for individual PRs as we only have the built files for those in the respective netlify deployments).

tordans:
@tyrasd sounds good. Please have a look at https://github.com/openstreetmap/id-tagging-schema/commit/1777979853df8ab01c1ab8ad7df2efde58988ff2#r191454884 – I need a few more files.

Issue #2051, opened 2020-07-08
Are building type presets really necessary in iD?


AntMadeira:
Hi there.
iD has two presets that can be a bit confusing, specially for newbies: `amenity=place_of_worship` + `religion=christian` and `building=church`. (it could be extended to other types of buildings)
In a country with many thousands of churches, this is problematic, because most casual mappers don't read OSM wikis.
To avoid having dozens or hundreds of actual churches mapped only with building=church, why not remove those presets and use only the type dropdown selector that already exists when adding `amenity=place_of_worship` + `religion=christian`?

matkoniecz:
basically implemented by #2422 if I understand request right

AntMadeira:
> basically implemented by [#2422](https://github.com/openstreetmap/id-tagging-schema/pull/2422) if I understand request right

Yes. That's it. Thanks!

matkoniecz:
note that it is not merged yet - so closing it is premature

PR #2422, opened 2026-07-06
drop building presets which are too easily mistaken with "functional" presets

tyrasd:
When tagging a school, hospital, hotel, train station, or similar feature, it is easy to mistake the respective `building=school/hospital/hotel/…` preset with the one for the intended function of the feature (e.g. `education=school`, `healthcare=hospital`, `tourism=hotel`, `public_transport=station`, etc.).

In the majority of cases, a mapper actually wants to map the object including its function and is probably unaware about the connotation that a `xxx Building` preset (without the respective main primary tag) is actually only referring to the architectural style of the building instead of its current function.

As the regular Building preset already has a field which allows mappers to specify the architectural type of the building (e.g. `building=hotel` for a building with many small rooms and long corridors), it is not really necessary to keep the corresponding presets around for all those building types, especially if they cause potentially more confusion than benefit to the mappers.

An alternative to the complete removal would be to only make these presets non-searchable, but I think in that case they might as well be removed. :shrug: 

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Tag:building%3Dtransportation
- https://wiki.openstreetmap.org/wiki/Tag:building%3Dhotel
- and many similar pages

**Relevant tag usage stats:**
- these tags are typically used relatively frequently, but are much rarer to their "functional" counterparts, so it remains questionable if the inclusion of the tags as separate presets is beneficial

see below

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>


tyrasd:
comparison before/after:

matkoniecz:
Why "Unspecified Building Type" is removed (per before/after)?

https://github.com/openstreetmap/id-tagging-schema/pull/2422/changes/289800e436bf82e2a13707f2c7e9df4a1d4c181e has

> in order to make it clearer that the field should actually be filled in with a concrete building "architecture" type

but after change it seems to be less clear, not more clear

data/presets/building/school.json
k-yle, line 1: maybe keep `school`/`college`/`university`? since a _"Classroom Building"_ is clearly different from a _"School Campus"_, and it's unlikely that people will confuse the two concepts

matkoniecz:
> An alternative to the complete removal would be to only make these presets non-searchable, but I think in that case they might as well be removed. 🤷

I think there is noticeable benefit from keeping them - it is more clear that specific valid building value was selected from dropdown.

matkoniecz:
Also, if we decide to merge it and later reject - translations will not be lost.

matkoniecz:
> especially if they cause potentially more confusion than benefit to the mappers

is it something you encountered or people reported as happening? Or seen this confusion in action?

matkoniecz (changes requested):
needs to be changed given https://github.com/openstreetmap/id-tagging-schema/commit/a9ec6a9d32feb3780e4a834473b78b3760009d46

still thinking about main PR part, but now I am sure that those should be made unsearchable, not removed

AntMadeira:
As @matkoniecz referred about my post [here](https://github.com/openstreetmap/id-tagging-schema/issues/2051), I believe this will be very useful. 
I have doubts about removing the "Unspecified Building Type" field, because many mappers can just be mapping buildings from aerial images, for example, and don't know the type specifically.

Hufkratzer:
I don’t understand how deleting the building presets is supposed to be an improvement. I think it makes sense to list the presets that apply more frequently before those that apply less frequently (I’m not sure if this is always the case already). However, if we want to distinguish between building type and building use (function) in OSM, as these do not always correspond in reality, then this should be made clear to the mapper rather than being hidden.

If you remove the building presets and then, for example, someone wants to map a school building for a school that has already been mapped, and he can not **easily** find the preset for `building=school`, but he finds the preset for `amenity=school`, then it will be more likely that he will (mis)use that preset for the building. No?

AntMadeira:
> If you remove the building presets and then, for example, someone wants to map a school building for a school that has already been mapped, and he can not **easily** find the preset for `building=school`, but he finds the preset for `amenity=school`, then it will be more likely that he will (mis)use that preset for the building. No?

That can happen, yes. Although I would argue that it's a lesser problem than using buildings to map amenities, as it was often the case.
From my experience, is rare to find cases like that. Instead, schools, places of worship and others, despite the effort from translators to distinguish them clearly in the presets, abound and have to be corrected routinely.

Maybe there can be other solutions to mitigate the problems that arise from this change, but I believe this is an improvement for mapping with iD.

Hufkratzer:
@AntMadeira How are these abound errors relating to schools normally corrected? By adding `amenity=school` to the buildings? Or by creating additional features (nodes or areas) with `amenity=school`? Is there a standard procedure for this (compare [forum discussion](https://community.openstreetmap.org/t/should-amenity-be-tagged-on-a-building-way/138827))? If so, would deleting the building presets lead to this standard procedure being used more often?

According to taginfo, less than 3 % of objects with `building=school` also have `amenity=school`, and less than   3 % of objects with `amenity=school` also have `building=school`. And less than 20.5 % of objects with `building=hotel` also have `tourism=hotel`, and less than 7.5 % of objects with `tourism=hotel`also have `building=hotel`.

AntMadeira:
It depends. Usually, it's done adding `amenity=school` to the building, but not all buildings have just one school and not all schools have a buildings that clearly differentiates from the others. (For example, a simple house, an old museum, a shipment container, etc.) Those percentages of the relation between buildings and amenities clearly translate this problem: not always a building must have a correspondent amenity and vice-versa.

What the removal of the building preset would do is avoid the confusion between building and amenity. New mappers are unaware of this difference and if they see a `building=school` or `building=church`, they tend to use those tags without the amenity. 

In Portugal, for example, all schools were mapped with their respective buildings by a recent community effort, so the problem was greatly reduced, but we still have many `building=school` that were mapped before this effort and still have the name of the school without the amenity, specially in rural areas, where old schools were abandoned to gather the students in urban school hubs.

matkoniecz:
> we still have many building=school that were mapped before this effort and still have the name of the school without the amenity, specially in rural areas, where old schools were abandoned to gather the students in urban school hubs.

This does not seem obviously wrong? IF school is no longer operating

PR #2429, opened 2026-07-07
Change name of `railway=signal_box` preset

Geo-2695:

Change the name of the `railway=signal_box` (Signal Box) preset; the original name could be ambiguous, so "Railway" has been added to clarify that it is a railway facility.

### Related issues

none

### Links and data

**Relevant OSM Wiki links:**

https://wiki.openstreetmap.org/wiki/Tag:railway=signal_box

**Relevant tag usage stats:**
> …

### DIFF
--- a/data/presets/railway/signal_box.json
+++ b/data/presets/railway/signal_box.json
@@ -16,7 +16,7 @@
     "tags": {
         "railway": "signal_box"
     },
-    "name": "Signal Box",
+    "name": "Railway Signal Box",
     "terms": [
         "interlock tower",
         "interlocking tower",


Geo-2695:
Signalbox

matkoniecz:
What kind of ambiguity was possible?

Geo-2695:
The search interface only displays `Signal Box`, so I worry that people might mistakenly use them for other purposes, such as power control boxes or electrical substations.
Local translations may omit some infomation.
St

matkoniecz:
> such as power control boxes or electrical substations

why either would be referred to as a signal box? Are there any dialects of English where it is happening? Are there known mistagging cases across multiple mappers caused by this?

Geo-2695:
🤔I guess it's not a good idea to edit the name directly here, It might be better to modify it in the local translation.

matkoniecz:
Yeah, if it is confusing in specific language then fixing translation may be good enough. 

Not specialist but these terms, in English at least seem reasonably distinct.

PR #2431, opened 2026-07-08
merge the schema-builder repository into this repository

### LINKED ISSUE #35: consider consolidating
Let's discuss the suggestion from https://github.com/ideditor/schema-builder/commit/d9c23fac2a3d1733e55c841c1e9e476ad78e2f7f#commitcomment-64100888:

> [This repository could just be] some build scripts in the id-tagging-schema repo.

Right out of the box, I would agree.
I thought that there was a perhaps a reason for the repositories being structured like this. Perhaps some forks are using it to maintain their own set of presets?! Maybe @quincylvania could explain the original reasoning?

k-yle:
> [!warning]
> Do not squash or rebase this PR when merging. That will destroy the git history.

~~blocked by #2255~~. Closes https://github.com/ideditor/schema-builder/issues/35

---

This PR merges the two repositories together: [schema-builder](https://github.com/ideditor/schema-builder) is now simply a folder within this repository 🎉

This will significantly simplify the development process, and also make it easier for forks who want to extend the tagging schema.

**Open question:** Should we continue to publish schema-builder to npm? I suggest **no**, because it's not generic enough to be re-usable for anything else.

There are several other things we need to do after this PR is merged. Outlined in https://github.com/ideditor/schema-builder/issues/35#issuecomment-4914078288

---

To preserve the commit history, this was performed by:
- temporarily renaming some files with clashing names, like `README.md`
- running [`git merge --allow-unrelated-histories`](https://git-scm.com/docs/git-merge#Documentation/git-merge.txt---allow-unrelated-histories)
- undoing the temporary rename, and making other minor tweaks so that the project compiles. 

image


tordans:
Is this also the opportunity to rename this repo to "tagging-schema"? Just checking to not miss the opportunity...

tyrasd:
> Is this also the opportunity to rename this repo to "tagging-schema"?

I'd say that could be done independently from this sometime between now and the next (probably major) release v8. I'd be open for it. Only downside would be that consumers would need to update e.g. npm package name. Shall we put this on the agenda of the next community chat?

matkoniecz:
> I'd say that could be done independently from this sometime between now and the next (probably major) release v8. I'd be open for it.

I think it can be done independently, but I want to register my opposition. I think it is fine to primarily develop it for iD, while taking needs of other data consumers into account but with iD having priority.

matkoniecz:
No label really first but I guess that prominent mention in the changelog is fine in this case?

matkoniecz:
@k-yle Can you update this PR so it includes changes made also after March?

matkoniecz (changes requested):
it seems that changes from last months (after https://github.com/ideditor/schema-builder/pull/274 was merged) are missing in this PR

also, this will require changes to README at least but can be done as a followup I guess

Also, SCHEMA.md link should be added to readme.

tyrasd (dismissed):
Thanks Kyle! As #2255 is merged now, this should be unblocked.

> Open question: Should we continue to publish schema-builder to npm? I suggest no, because it's not generic enough to be re-usable for anything else.

Yeah, I agree. 👍 Let's publish it as a single package with data plus build tooling.

<hr>

@k-yle could you please update the branch to the final v7.0.0 of the schema-builder repo? At first I thought it was only missing the changelog and last few dependency updates, but it seems like there is also some of the more recent code updates not included here (e.g. https://github.com/ideditor/schema-builder/pull/281 and everything merged after that), if I see it correctly.

And we should probably remove the old schema-builder from `package.json`, I guess. 😉 


PS: note to self: after this, we can do an `npm deprecate @ideditor/schema-builder "this has been integrated into @openstreetmap/id-tagging-schema"`

matkoniecz:
Also, old repo readme should be probably blanked and replaced with link to SCHEMA file in this repo?

matkoniecz:
And one more question: how now we can do change to schema builder code and test it?

k-yle:
- PR has been recreated to include the commits to schema-builder from 15 minutes ago 
- next steps for the old repo can be discussed at https://github.com/ideditor/schema-builder/issues/35#issuecomment-4914078288
- to review this PR, you only need to look at 5e6184d7cdbe44bed890b7cc0626ca663e3553cf, and the overall diff (using the standard GitHub UI). In 5e6184d7cdbe44bed890b7cc0626ca663e3553cf, I made a few changes to fix broken links in the docs. If preferred, that could be removed and done in a separate PR.
  - schema-builder's `README.md` is now called `SCHEMA.md`
  - schema-builder's `CHANGELOG.md` is now called `CHANGELOG-old.md`. can't think of a better name at the moment.
  image
- spot-checked a few files to ensure the git blame wasn't destroyed: [1](https://github.com/openstreetmap/id-tagging-schema/blame/kh/migrate-schema-builder/SCHEMA.md), [2](https://github.com/openstreetmap/id-tagging-schema/blame/kh/migrate-schema-builder/README.md), [3](https://github.com/openstreetmap/id-tagging-schema/blame/kh/migrate-schema-builder/schemas/preset.json), [4](https://github.com/openstreetmap/id-tagging-schema/blame/kh/migrate-schema-builder/scripts/lib/build.js)

matkoniecz:
> schema-builder's CHANGELOG.md is now called CHANGELOG-old.md. can't think of a better name at the moment.

I dislike this but it can be amended later (merged into CHANGELOG.md or deleted and old version linked or hidden somewhere rather than stuck at repo's top level)

matkoniecz:
> I made a few changes to fix broken links in the docs.

great, thanks for doing this

matkoniecz (dismissed):
And one more remain question: what is the proper way of changing schema builder code and running it to test it? Maybe it would be worth adding to docs?

(I also want to confirm it works before merging)

k-yle:
> And one more remain question: what is the proper way of changing schema builder code and running it to test it? Maybe it would be worth adding to docs?

now you can directly edit any file, and run `npm run dist`. and the changes will immediately take effect. for example:
- in the schema files, change `"type": "string"` to `"type": "boolean"`, run `npm run dist`, and it will fail.
- in the schema files, change the description of any schema field, run `npm run dist`, and the `*.d.json.ts` files will update immediately.
-  in the build scripts, change some trivial code like [this line](https://github.com/openstreetmap/id-tagging-schema/blob/5e6184d7cdbe44bed890b7cc0626ca663e3553cf/scripts/lib/build.js#L624), run `npm run dist`, and the taginfo file will immediately update

tyrasd (approved):

PR #2437, opened 2026-07-09
Automatic set location-set to include planet

tyrasd:
Replace #2381 with custom logic to automatically add implicit `include: ["Planet"]` to the `dist` output if not specified otherwise.

### DIFF
--- a/data/fields/bicycle_road.json
+++ b/data/fields/bicycle_road.json
@@ -12,9 +12,6 @@
         "neighborway"
     ],
     "locationSet": {
-        "include": [
-            "Planet"
-        ],
         "exclude": [
             "BE",
             "NL"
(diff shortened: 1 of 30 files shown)


matkoniecz:
@tyrasd have you tested it?

matkoniecz:
@tyrasd 

At least https://github.com/openstreetmap/id-tagging-schema/blob/d47f7c4549e0d68bc88c065e999bb1b4e6685352/data/fields/maxweight.json and https://github.com/openstreetmap/id-tagging-schema/blob/d47f7c4549e0d68bc88c065e999bb1b4e6685352/data/fields/maxweight_bridge.json also should be cleaned

matkoniecz (approved):

matkoniecz:
I am tempted to merge now and make followup PR for those two fields.

PR #2445, opened 2026-07-10
do not allow invalid references

k-yle:
to make https://github.com/ideditor/schema-builder/pull/281 less disruptive, some references to non-existant properties was allowed, to match iD's current behaviour. 

now we can be more strict, and reject references to fields that do not exist. 

no impact on downstream consumers. this is a good example of how simple it is to change the build scripts now

### DIFF
--- a/data/presets/public_transport/stop_position_aerialway.json
+++ b/data/presets/public_transport/stop_position_aerialway.json
@@ -3,9 +3,6 @@
     "fields": [
         "{public_transport/stop_position}"
     ],
-    "moreFields": [
-        "{public_transport/stop_position}"
-    ],
     "geometry": [
         "vertex"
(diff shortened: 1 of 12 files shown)


tordans (approved):

matkoniecz (changes requested):
I think it is not a great change, I would rather do opposite.'

If we add `moreFields` to something we should not need to check all children and add explicit inheritance there.

k-yle:
okay, then we can close this. I was thinking that people would waste time looking through many files to figure out what  `"moreFields": ["{public_transport/stop_position}"]` refers to. which they'll never figure out, because the build script silently skips these invalid references. Whereas in normal code, importing something that doesn't exist is always an error

PR #2446, opened 2026-07-10
fix minor bugs in the build scripts

k-yle:
see comments inline

### DIFF
--- a/scripts/lib/build.js
+++ b/scripts/lib/build.js
@@ -598,7 +598,6 @@ function generateTaginfo(presets, fields, deprecated, discarded, tstrings, proje
 
   Object.keys(presets).forEach(id => {
     let preset = presets[id];
-    if (preset.suggestion) return;
     if (id.startsWith('@')) return;
 
     /** @type {Record<string, Set<string>>} */
@@ -764,7 +763,7 @@ function generateTaginfo(presets, fields, deprecated, discarded, tstrings, proje
     if (['multiCombo', 'manyCombo', 'check', 'defaultCheck', 'onewayCheck'].includes(field.type)) {
       return false;
     }
-    if (field.autoSuggestion === false && field.customValues === false) {
+    if (field.autoSuggestions === false && field.customValues === false) {
       return false;
     }
     return true;
--- a/scripts/lib/build.js
+++ b/scripts/lib/build.js
@@ -73,8 +73,7 @@ function buildDist(options) {
       _currBuild = null;
     })
     .catch((err) => {
-      process.stderr.write(err);
-      process.stdout.write('\n');
+      console.error(err);
       _currBuild = null;
       process.exit(1);
     });
--- a/scripts/lib/translations.js
+++ b/scripts/lib/translations.js
@@ -140,7 +140,8 @@ function fetchTranslations(options, references) {
       process.stdout.write(`got resource language stats collection for ${resourceId}\n`);
       callback(null, result);
     } catch (err) {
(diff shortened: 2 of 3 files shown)


scripts/lib/build.js
```diff
@@ -598,7 +598,6 @@ function generateTaginfo(presets, fields, deprecated, discarded, tstrings, proje
 
   Object.keys(presets).forEach(id => {
     let preset = presets[id];
-    if (preset.suggestion) return;
```
k-yle, line 601: `suggestion` is not a valid attribute, so this is dead code

i'm not sure what the intention was, but i think every preset should be included in the taginfo file, even if [`searchable` = false](https://github.com/openstreetmap/id-tagging-schema/blob/main/SCHEMA.md#searchable)

scripts/lib/translations.js
```diff
@@ -140,7 +140,8 @@ function fetchTranslations(options, references) {
       process.stdout.write(`got resource language stats collection for ${resourceId}\n`);
       callback(null, result);
     } catch (err) {
-      process.stderr.write(`error while getting resource language stats collection for ${resourceId}\n`, err);
+      process.stderr.write(`error while getting resource language stats collection for ${resourceId}\n`);
```
k-yle, line 143: the second argument is for the encoding, not for extra data. so the raw error was never being shown.

also, logging an error via `process.std*.write` has suboptimal formatting, so changed it to a `console.error` statement.

scripts/lib/build.js
```diff
@@ -764,7 +763,7 @@ function generateTaginfo(presets, fields, deprecated, discarded, tstrings, proje
     if (['multiCombo', 'manyCombo', 'check', 'defaultCheck', 'onewayCheck'].includes(field.type)) {
       return false;
     }
-    if (field.autoSuggestion === false && field.customValues === false) {
+    if (field.autoSuggestions === false && field.customValues === false) {
```
k-yle, line 765: typo, which caused this block to never run. 

After this change, `example_key=*` will no longer be included if there are already fixed values for `example_key=`.
Image

tordans (approved):

k-yle:
just added another one, `process.stdout.error(...)` only works for strings, you can't log [`Error`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)s or other objects.

scripts/lib/translations.js
matkoniecz: > also, logging an error via process.std*.write has suboptimal formatting, so changed it to a console.error statement.

it seems not changed?

scripts/lib/build.js
matkoniecz: whops, thanks for catching this

how you found it?

scripts/lib/translations.js
k-yle: it's on the line below now

scripts/lib/build.js
k-yle: I have a branch from a few years ago which adds JSDoc annotations to these files, so that obvious errors can be detected. It was blocked by https://github.com/ideditor/schema-builder/pull/177 until recently. i noticed these errors while rebasing the branch... will send a PR for that soon

matkoniecz (approved):

scripts/lib/build.js
tordans: > I have a branch from a few years ago which adds JSDoc annotations to these files

Or, let's just go TS...

PR #2447, opened 2026-07-10
added icon for office/union

paulklie:

Adds the fas raised fist as an icon for [office=union](https://wiki.openstreetmap.org/wiki/Tag:office=union).

I used the [raised fist symbol](https://icones.js.org/collection/fa7-solid?s=hand&icon=fa7-solid:hand-back-fist). From [wikipedia](https://en.wikipedia.org/wiki/Raised_fist):

>  It is a common symbol representing a wide range of political ideologies, most notably socialism, communism, anarchism, and **trade unionism**,

### Related issues
image

### DIFF
--- a/data/presets/office/union.json
+++ b/data/presets/office/union.json
@@ -1,5 +1,5 @@
 {
-    "icon": "maki-suitcase",
+    "icon": "fas-hand-back-fist",
     "geometry": [
         "point",
         "area"


matkoniecz (approved):
personally I think it fits

lets wait (maybe week?) for what others think

matkoniecz:
thanks for finding it!

fibonacci-matrix:
https://pinhead.ink/#crowd_of_people

crowd_of_people

paulklie:
> https://pinhead.ink/#crowd_of_people
> crowd_of_people

Hmm, I think that one could also appropriate, but since its so generic, maybe it would be better to save it for other presets that are using common icons?

For example:

- [office=political_party](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/office/political_party.json)
- https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/community_centre/community_hall.json

matkoniecz:
> https://pinhead.ink/#crowd_of_people

this seems less specific to me

fibonacci-matrix:
I read up on "raised fist" icon and realized it's actually a more accurate choice than the "group of people" icon. The latter could instead replace the suitcase icon for all presets that represent membership organizations, such as political parties, associations, and similar entities, when no more suitable icon is available

PR #2452, opened 2026-07-10
add terms for apartments

paulklie:

"flats" and "Condominiums" are commonly used terms for [apartments](https://wiki.openstreetmap.org/wiki/Tag:residential%3Dapartments)

image
image

### DIFF
--- a/data/presets/building/apartments.json
+++ b/data/presets/building/apartments.json
@@ -13,6 +13,9 @@
     "tags": {
         "building": "apartments"
     },
+    "terms": [
+        "flats"
+    ],
     "matchScore": 0.5,
     "name": "Apartment Building"
 }
--- a/data/presets/building/apartments.json
+++ b/data/presets/building/apartments.json
@@ -13,8 +13,10 @@
     "tags": {
         "building": "apartments"
     },
+    "aliases": [
+        "condominium"
+    ],
     "terms": [
-        "condominium",
         "flats"
     ],
     "matchScore": 0.5,
--- a/data/presets/building/apartments.json
+++ b/data/presets/building/apartments.json
@@ -14,7 +14,7 @@
         "building": "apartments"
     },
     "aliases": [
-        "condominium"
+        "Condominium"
     ],
     "terms": [
         "flats"
--- a/data/presets/building/apartments.json
+++ b/data/presets/building/apartments.json
@@ -14,6 +14,7 @@
         "building": "apartments"
     },
     "terms": [
+        "condominium",
         "flats"
     ],
     "matchScore": 0.5,


matkoniecz (approved):

matkoniecz:
Thanks!

1ec5:
At first glance, “flats” would be out of place in the default American English localization, as opposed to British English and other localizations maintained on Transifex. But apparently it isn’t considered archaic in some parts of New England, so I guess it’s fine here too. Just don’t expect to always hard-code Commonwealth terms directly in the codebase in general.

1ec5:
By the way, the `tourism=apartment` preset shows up first for “condo” with the label “Guest Apartment / Condo”. I suspect that an inexperienced mapper intending to map a condominium building would easily choose this preset instead of the `building=apartments` preset, since it sounds more specific and we lack a field for `condo=yes`. Is the label supposed to mean “Guest Apartment or Guest Condo”?

matkoniecz:
> Is the label supposed to mean “Guest Apartment or Guest Condo”?

yes

1ec5:
Ah OK. It’s really subtle, but changing it to “Guest Apartment/Condo” would at least make the user think twice before selecting it instead of Apartment Building. We could add “Condominium Building” as an `alias` of `building=apartments` too, but it won’t show up in the UI due to openstreetmap/iD#6139.

matkoniecz:
Maybe “Guest Apartment or Guest Condo” ?

(BTW, if you think that some change is a clearly good idea, feel free to make a PR)

1ec5:
More this is making me question the wisdom of using slashes everywhere. 😅 But the actual feedback for this PR would be to move “Condominium” to an `alias`.

data/presets/building/apartments.json
```diff
@@ -13,6 +13,10 @@
     "tags": {
         "building": "apartments"
     },
+    "terms": [
+        "condominium",
+        "flats"
+    ],
```
matkoniecz, line 21:
```suggestion
    "aliases": [
        "condominium"
    ],
    "terms": [
        "flats"
    ],
```

this should implement @1ec5 suggestion (promoting this term stronger, though without visible effects for now)

hopefully typofree?

data/presets/building/apartments.json
```diff
@@ -13,6 +13,12 @@
     "tags": {
         "building": "apartments"
     },
+    "aliases": [
+        "condominium"
```
matkoniecz:
```suggestion
        "Condominium"
```

names and aliases are capitalized

matkoniecz (approved):

Issue #2449, opened 2026-07-10
Archive schema-builder in less confusing way


matkoniecz:
Currently, after https://github.com/ideditor/schema-builder/pull/341#issuecomment-4935993956 the https://github.com/ideditor/schema-builder repository https://github.com/ideditor/schema-builder is in read-only state

wile using this github feature is a standard method for abandoned projects, in this case it is quite confusing as this repo was merged into another, not simply abandoned

up to date content was moved elsewhere and there should be at the very least note where current documentation is located, as currently it is quite hard to find

while deleting files (and keeping them in history) is not a typical method of archiving repositories, I think it is appropriate here to prevent confusion, though maybe just adding info about current status and where to find a current files would be better

Pinging @bhousel 

(opened issue here as the other repo is now read-only)

bhousel:
Let's discuss here how you want to proceed with archiving the `schema-builder` repo.

I disagree that "deleting all the files" is an appropriate way to archive a repository, particularly one that holds json schema documents that other projects are using to validate their presets files.

For now I've set the archive flag on the repo to prevent damage.  @tyrasd can un-archive it if he wants to.

tordans:
My recommendation would be
- put a Markdown callout in the readme (`> [INFO]` or something like this) https://github.com/ideditor/schema-builder/blob/main/README.md
- change the pinned message at https://github.com/ideditor/schema-builder/issues to link to here
- close open PRs with link to here https://github.com/ideditor/schema-builder/pulls

Then we have a nice migration path from there to here…

matkoniecz:
Archiving without explanation also is not great. 

Maybe blank only human-facing Markdown files and keep machine-readable code unchanged to allow for graceful migration (or indefinite referencing of old repo)?

matkoniecz:
@bhousel And thanks for pointing out that some files residing there are being actively used by other projects! I admit that I missed it.

mnalis:
> * put a Markdown callout in the readme (`> [INFO]` or something like this) https://github.com/ideditor/schema-builder/blob/main/README.md
> 
> * change the pinned message at https://github.com/ideditor/schema-builder/issues to link to here
> 
> * close open PRs with link to here https://github.com/ideditor/schema-builder/pulls

Agreed. [GitHub docs for archiving](https://docs.github.com/en/repositories/archiving-a-github-repository/archiving-repositories) recommend as much:

[...]

PR #2454, opened 2026-07-11
record expected vertex on power=line/minor_line and houses (and setup for of a more general solution, review welcome)

### LINKED ISSUE #9489: Don't suggest any non power related presets for vertexes of power lines
### URL

_No response_

### How to reproduce the issue?

1. Create a node on power line (`power=line` or `power=minor_line`) by double clicking on any place along the line
2. Select the node
3. (notice the preset suggestions on the left)

### Screenshot(s) or anything else?

For vertexes of lines, only `power` related features should be suggested. In the screenshot below you can see that iD suggests rooftop solar panels, truck scale, flags and crossings (which I've used in my recent changesets). I assume that there would be a need to change code in iD, id-tagging-schema and in schema-builder in order to fix this issue.

image

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### What version numbers does this issue effect?

2.24.1

### Which browsers are you seeing this problem on?

_No response_

matkoniecz:

This is sketch of a solution for https://github.com/ideditor/schema-builder/issues/96 / https://github.com/openstreetmap/iD/issues/9489 

This is intended to support special behavior for vertex nodes. After all, when vertex node in power line is selected it is `power=tower` rather than `highway=motorway_junction`. Currently we have no support in preset or iD for doing anything smart here. Lets change it!

Note that it is only toy change of presets, no documentation or build code was written and therefore it is not ready at all for merging.
But wanted to propose this specific syntax.

It is intentionally covering a tiny part of vertex presets, where it is the most obviously useful.
I have WIP for many more but maybe let's discuss such minimal plan first. 

Plan presented here would be to list those first when point within power line is selected, possibly in order as defined here. When point is within multiple lines at once - all their lists would count (maybe initially list first entries from all eligible list, then second positions and so on, skipping duplicates)

Other vertex presets could be still listed by editor, but maybe they can be hidden? Though I am still reviewing 300+ vertex presets, so lists here may be incomplete...

Possible changes: 
- different labels for attributes in code
- cut down this plan to MVP and list solely `power=tower` on `power=line` and nothing else
- define it on children nodes instead like I initially did (listing valid parent ways, rather than way listing valid children vertexes as here)
    - but then we lose ability to order them and it is harder to see what is listed for children of a given way
         - actually ordering them would be possible but would require multiple groups like `bestParentWays`, `moreParentWays` and so on
- recommend to block merging of vertex with incompatible parent ways or warn about such existing cases ( https://github.com/openstreetmap/iD/issues/4173 / https://github.com/openstreetmap/iD/issues/9488 ) - but it would require full coverage and even then seems, sadly, more dubious than I initially hoped for. But, if possible, change proposed here would be the first step toward it. And https://github.com/openstreetmap/iD/issues/8263 would require entirely different listing but I guess that introducing `related` would be at least starting in direction of such smartness.
- document that if line is multipolygon member then vertex on it is being counted as vertex on multipolygon (irrelevant for now as power lines cannot be areas)

In future more properties can be added to `related` field, see https://github.com/openstreetmap/iD/issues/4138 and https://github.com/openstreetmap/iD/issues/4139 that have many more interesting ideas.

### Next steps

Discuss syntax and how it is defined here. I can share WIP for fuller coverage if people are curious how it looks when used on a wider scale. I can also share tools that I have setup for analyzing OSM data.

If that syntax is liked I would add documentation into SCHEMA.md and build code ( ability to inherit lists is necessary for defining those more widely ).

Then next step would be release this non-breaking expansion of schema.

Then implementation of support in iD (and maybe elsewhere!)

Then adding wider coverage for other lines/areas with children presets in their vertexes.

### QA created as a side effect

BTW, I have some seemingly clearly invalid cases listed at https://wiki.openstreetmap.org/wiki/User:Mateusz_Konieczny/bad_vertex

If anyone is interested to look at those - feel free! Just do it with brain engaged, some may be valid, have even worse nearby problems etc. 
If you would be interested in looking at cases like this but you dislike this specific problem or many were done already and refresh is needed,  let me know. 

### Related issues

first step toward fixing https://github.com/ideditor/schema-builder/issues/96

### Links and data

**Relevant OSM Wiki links:**
- …

**Relevant tag usage stats:**
> …

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>


matkoniecz:
Pinging @mnalis @westnordost @tordans @Dimitar5555 (here as https://github.com/ideditor/schema-builder/issues/96 is unavailable due to https://github.com/openstreetmap/id-tagging-schema/issues/2449 )

matkoniecz:
Why I have went away from defining valid parents on node?

It is impossible to clearly define that say `power=tower` should be listed before anything else in `power=line`. I guess that having priority field or multiple named groups would work, but if defined on way you simply get list that will be shown to user and can be put into order

Spreading single list components across several fields will make things more annoying to adjust, debug and edit in future

sekerob:
Fortunately JOSM flags this none power connection, but only after you edit the line. See this often flagged when a  hiker or Mtb rider has drawn a new way and set a node too close upon which it snaps to the power object. Osmose will than ping the last powerline editor.

theraihanrakibb (commented):
Thanks for the exploratory PR and for inviting review.

**Blocking — schema validity:**
`related.expected_vertex_children` is not a defined field. I checked `schemas/preset.json` and the whole repo: there is no `related` object on presets and the string `expected_vertex_children` appears nowhere. The iD preset build will reject `power/line.json` / `power/minor_line.json` and CI will fail. Either extend `schemas/preset.json` first (add a `related` object with `expected_vertex_children`), or keep this as a pure discussion sketch until the schema question is settled.

**Content:**
- `minor_line.json` looks copy-pasted from `line.json`: `power/tower` and `power/generator/source/nuclear` as expected vertices on a *minor* line (street-lighting / low-voltage spur) is implausible. Curate the lists per preset rather than duplicating.
- Redundancy: `power/generator` is listed alongside its children (`power/generator/source/*`, `power/generator/method/photovoltaic`). A vertex matching the parent also matches a child — pick one level.
- `power/transformer` as a vertex child of a line is debatable (transformers are usually nodes, not line vertices).

**Question:** What is `expected_vertex_children` meant to drive — auto-connect hints, validation, or editor UX? Nailing that down will make the schema shape obvious.

data/presets/power/line.json
```diff
@@ -19,5 +19,18 @@
     "tags": {
         "power": "line"
     },
-    "name": "Power Line"
+    "name": "Power Line",
+    "related": {
+        "expected_vertex_children": [
```
tyrasd:
```suggestion
        "expectedVertices": [
```

Dimitar5555:
> Why I have went away from defining valid parents on node?

I see your point but the way it is defined, it would be possible to add say `power=tower` node on a `highway=track`.

What we also need is a way to say that a vertex can only have a parent way of x, y and z. That way when you come across the above mentioned situation, the validator can complain. This can be added at a later stage but it is a thing to consider.

matkoniecz:
> it would be possible to add say power=tower node on a highway=track.

yes

in future also `highway=track` would get its list of good vertex that would be promoted and we may at some point stop allowing cases outside good list

but as I keep investigating you have a looooot of weird special cases

> but the way it is defined

if it would be defined on points, adding them as vertex on objects outside that list would be still allowed anyway, at least initially

matkoniecz:
> What we also need is a way to say that a vertex can only have a parent way of x, y and z. That way when you come across the above mentioned situation, the validator can complain. This can be added at a later stage but it is a thing to consider.

I wanted to do this but I am getting skeptical is it achievable. I have seen some truly weird but valid cases at this point.

But I would love to reach this point. Anyone interested may help by processing some cases from https://wiki.openstreetmap.org/wiki/User:Mateusz_Konieczny/bad_vertex (with brain being used - and reporting back invalid ones)

matkoniecz:
So basically I would suggest to

- first introduce list of suggested matches and still show all vertex presets later
- later stop showing extra presets outside our list
- later possibly introduce list of valid but not shown ones, and start complaining about unexpected ones

matkoniecz:
interesting, I though I will need to change also `build.js` but it seems not needed :)

also, it is not a breaking change so it can be released without major version upgrade

matkoniecz:
Based on https://github.com/openstreetmap/id-tagging-schema/wiki/Projects-that-are-using-this-tagging-schema I am pinging @tyrasd @Zverik @bryceco @bhousel @westnordost @1ec5 who would be quite likely to use it in editors.

Feedback would be quite welcome, especially if current setup would be suboptimal for your uses.

Note: adding handling of more presets is planned, `power=line` and `power=minor_line` is just a start.

@simonpoole would you be interested in adding functionality like that to Vespucci? I could try exporting it to https://github.com/simonpoole/beautified-JOSM-preset somehow (no idea how hard it will be to match presets) if that project would gain ability to specify such relations.

matkoniecz:
Pinging also @GA-Kevin-Codes @pietervdvn @mvexel @PanierAvide @joto - based on description at https://github.com/openstreetmap/id-tagging-schema/wiki/Projects-that-are-using-this-tagging-schema you are far less likely to be using this new functionality, but feedback is still welcome!

If anyone of pinged people would prefer to not be pinged, please let me know and I will adjust pinging method in future.

matkoniecz:
Pinging also @kaligrafy (separate comment as they were added to https://github.com/openstreetmap/id-tagging-schema/wiki/Projects-that-are-using-this-tagging-schema minutes ago) - feedback on this format is welcome, especially if you plan on using this data.

tyrasd (approved):
Technically, the build scripts could add the inverse information (i.e. which presets are allowed as a parent of a particular vertex), but for the time being, it is better to define it only on the line preset only (as it allows to define the suggested order/"importance" of the vertex presets).

SCHEMA.md
```diff
@@ -222,6 +222,30 @@ Alternatively, `locationSetCrossReference` can be use to reference the `location
...
+Hints about related presets that are related to this one, can be optionally used for a context-aware handling.
+
+###### `expectedVertices`
+
+A list of preset IDs that are expected to appear as vertices on a line or area marked with this preset. For example `power=line` may have
+
+```
+    "related": {
+        "expectedVertices": [
+            "power/tower",
```
k-yle: the build script should check
- if these are valid preset IDs
- if the referenced preset actually supports `geometry: ["vertex"]`

otherwise it's a waste of time for code reviewers to manually check this

SCHEMA.md
```diff
@@ -222,6 +222,30 @@ Alternatively, `locationSetCrossReference` can be use to reference the `location
...
+            "power/portal",
```
k-yle: the standard convention is to wrap preset/field IDs in `{` + `}`. Should we follow this convention?

schemas/preset.json
```diff
@@ -141,6 +141,23 @@
             "pattern": "^\\{.+\\}$",
             "description": "An string referencing another preset which has a locationSet"
         },
+        "related": {
+            "description": "Hints about related presets that are related to this one, can be optionally used for a context-aware handling",
+            "type": "object",
+            "properties": {
+                "expectedVertices": {
+                    "description": "A list of preset IDs that are expected to appear as vertices on a line or area marked with this preset",
```
k-yle: To me, this phrasing sounds like a vertex with no tags is banned. Is that the intention?

For your power line example, I agree it makes sense that every vertex must have a `power=*` tag. But for `waterway`s and `railway`s, this is not true.

schemas/preset.json
k-yle, line 1: this is a many:many relationship, so it will quickly get very verbose. Every linear `waterway` preset (like `stream`, `drain`) will need to list every `waterway` vertex (like `waterfall`, `weir`)

Therefore, i think we will need to support cross-referencing. If we implement cross-referencing, would it be
- `relatedCrossReference`; or
- `relatedExpectedVerticesCrossReference`; or
- `related.expectedVerticesCrossReference` ?

data/presets/power/line.json
k-yle, line 1: > Other vertex presets could be still listed by editor, but maybe they can be hidden? 

I think we should clarify the purpose of this data before continuing, and add that conclusion to the docs.

Should editors:
1. restrict verticies to **only** these presets, and ban other presets
2. or; promote these presets, but still allow any other preset?

data/presets/power/minor_line.json
```diff
@@ -14,5 +14,19 @@
     "tags": {
         "power": "minor_line"
     },
-    "name": "Minor Power Line"
+    "name": "Minor Power Line",
+    "related": {
+        "expectedVertices": [
```
k-yle, line 19: > [...] but then we lose ability to order them [...]

If the order is important, should the docs mention that? 

I'm guessing you want the preset selector to order them in this order by default?

schemas/preset.json
```diff
@@ -141,6 +141,23 @@
             "pattern": "^\\{.+\\}$",
             "description": "An string referencing another preset which has a locationSet"
         },
+        "related": {
```
k-yle, line 168: since this a sub-property, do you have a plan to define more rules in the future?

I'm not sure if it's sustainable to hardcode preset IDs for all types of relationships. There are some good examples in https://github.com/openstreetmap/iD/issues/8263#issuecomment-749279400 about how it will be painful to attempt to list every possible combination. 

So, do we need to think about the long-term plan now?

schemas/preset.json
matkoniecz: oh, definitely

it was supposed to be included even if not needed for this subset

what about

`"related": { "expectedVertices": "{waterway/river}" }`

syntax? Or would it be too cryptic?

Though maybe we would want to reference multiple in some cases....

matkoniecz: If separate field and we keep `related` group then I would go with `related.expectedVerticesCrossReference`

schemas/preset.json
matkoniecz: > To me, this phrasing sounds like a vertex with no tags is banned. Is that the intention?

no and should be rephrased to avoid implying it

SCHEMA.md
matkoniecz: I guess I should wait for https://github.com/openstreetmap/id-tagging-schema/pull/2545 to be completed before adding this part?

SCHEMA.md
matkoniecz: I thought about `{}` as being used to inherit values, rather than list preset.

But I have no strong opinion either way.

data/presets/power/line.json
matkoniecz: I am hoping for (1) to be achievable but I am not entirely sure is it possible - and full review of preset and data will take enormous time and is not finished yet.

(2) is definitely achievable and would recommend it at least at start, we may also get feedback about missing ones from mappers

data/presets/power/minor_line.json
matkoniecz: > I'm guessing you want the preset selector to order them in this order by default?

yes, though editor may also take into account for example which were used recently

schemas/preset.json
matkoniecz: for all? no

for some? may be feasible

see for example https://github.com/openstreetmap/iD/issues/4138 and https://github.com/openstreetmap/iD/issues/4139 and https://github.com/openstreetmap/iD/issues/4173 and possibly https://github.com/openstreetmap/iD/issues/10757 and https://github.com/openstreetmap/iD/issues/5977

if those are doomed/unwanted these tickets should be closed

but I think that at least subset is doable and would use data recorded in `related` structure

> So, do we need to think about the long-term plan now?

I expect at least one more such field to be existing and I see benefit in gathering them together

I do not have strong opinion about that

matkoniecz: note: for cases mentioned above and similar ones I do not have proof of concept comparable to this one

mostly because this one eaten through hilarious amount of time, so decided to go with minimum viable version in the form of this PR to avoid fate of some other overly large projects

matkoniecz:
@k-yle thanks for review!

data/presets/power/line.json
matkoniecz: see also https://github.com/openstreetmap/id-tagging-schema/pull/2454#issuecomment-4969335736

SCHEMA.md
k-yle: I also have no strong opinion, but I would lean towards following the existing convention

schemas/preset.json
k-yle: > what about `"related": { "expectedVertices": "{waterway/river}" }`

we should avoid using the same field name, because it makes the types confusing. With this approach, the field is `string | object` internally, but for data consumers, it's always `object`.

---

Another alternative is to remove the `related.*` sub-object

---

> Though maybe we would want to reference multiple in some cases....

luckily, references are private, they're not exposed to data consumers anymore. so if we want to support more complex syntax in the future, that can be done without impacting data consumers

data/presets/power/line.json
k-yle: okay, should we mention this in the docs? 

maybe it could be like the `unsquare_corners` validator which is off by default, and uses the lowest-severity category ("`suggestion`")

data/presets/power/minor_line.json
k-yle: shall we mention that in the docs? there are too many comments on this PR now, it'll be a pain if we need to reference any decisions made in the comments above

matkoniecz: yes, I will move this into actual docs (at the same time keeping it clear that how editor uses is quite optional/freeform, though lists will be build with specific use in mind)

SCHEMA.md
k-yle: no need to wait, i'm happy to resolve conflicts and help with the javascript side

matkoniecz: OK, so I have no excuse to delay :)

schemas/preset.json
matkoniecz: hmm, maybe "A list of preset IDs that are expected to appear as vertices on a line or area marked with this preset **when selecting a preset on a vertex**" ?

k-yle: I think it could be even simpler? 
```suggestion
                    "description": "A list of preset IDs that may be used as a vertex of this preset",
```

or more explicitly:


```suggestion
                    "description": "A list of preset IDs that may be used as a vertex of this preset, when this preset is used on a line or area",
```

matkoniecz:
when node is on multiple ways (for example on `highway=residential` and `waterway=stream`) - what should be considered as a proper handling?

- boost entries on both ways stronger 
- show entries listed only for one of ways (???)
- do not complain if it has vertex listed for only one way

matkoniecz:
@k-yle done some doc extension taking into account some feedback, polishing is definitely needed for those

matkoniecz:
I marked some as resolved, feel free to unresolve them if more is needed

schemas/preset.json
matkoniecz: defined `expectedVerticesCrossReference` for now, code needed to actually support it still needed

simonpoole:
JOSM presets can match based on a filter expression so you could in theory just suggest the tower or pole presets for the nodes. 

The devil tends to be in the details with all these things though. Essentially this would only work with workflows in which you choose what you want to map first and then draw the geometry, not in the customary, draw first, then apply a preset, fashion.

matkoniecz:
> not in the customary, draw first, then apply a preset, fashion.

in my workflow, especially with Vespucci, I often add say traffic calming nodes to already mapped roads

in such case such context-aware listing of relevant features seems potentially useful

schemas/preset.json
k-yle: i'm happy to help with the code side. but lets see if there are further comments first

matkoniecz:
will add entrance for houses, to get also live test of inheritance (it was the simplest case I found, not the most useful one)

SCHEMA.md
matkoniecz: I though about those doing something else: these are not referencing and expanding into a property of a listed preset - but rather listing them

may be overly subtle and too confusing...

dist/presets.d.json.ts
```diff
@@ -98,6 +98,21 @@ export interface Preset {
    * An string referencing another preset which has a locationSet
    */
   locationSetCrossReference?: string
+  /**
```
matkoniecz, line 101: revert edit to `dist` before merge

matkoniecz: @k-yle is there some smarter way to get linter working while updating build scripts?

SCHEMA.md
matkoniecz: seems to be working

matkoniecz:
as far as I see, sole raised not done thing is whether listed presets should be wrapped in `{}` like other referenced ones

matkoniecz:
> as far as I see, sole raised not done thing is whether listed presets should be wrapped in {} like other referenced ones

now done to make @k-yle happy :)

matkoniecz:
Linter failed due to https://github.com/openstreetmap/id-tagging-schema/pull/2454/changes/0460e45594924288a38376a3910570a68fd5a6d4

I think that dist should be included when new version is released, right?

scripts/lib/build.ts
```diff
@@ -1021,6 +1021,28 @@ function validatePresetFields(presets: AllPresets, fields: AllFields) {
       }
     }
 
+    let relatedKeys = ['expectedVertices'] as const;
+    for (let relatedKeyIndex in relatedKeys) {
+      let relatedKey = relatedKeys[relatedKeyIndex];
+      if (!preset.related?.[relatedKey]) continue;
+
+      for (let vertexIndex in preset.related[relatedKey]) {
+        const regexOutput = betweenBracketsRegex.exec(preset.related[relatedKey][vertexIndex])
```
k-yle, line 1051: we have a helper function for this (`isReference`)

even better, we could enforce this in the JSON schema using [the `pattern` attribute](https://json-schema.org/understanding-json-schema/reference/string#regexp). then there's no custom code required

scripts/lib/references.ts
k-yle, line 1: yes, i think you'll need to commit the changes to `dist/*.d.json.ts` if you plan to immediately reference that. 

just like what the measurement PR does (#1988)

matkoniecz: should I do in this PR? I though that `dist` is supposed to contain code of a latest release

k-yle: yeah i would commit the changes to just that file (in this PR). there's no other workaround because of the weird setup where we commit the `dist` folder

matkoniecz: actually, why we even commit `dist` folder?

k-yle: ideally we wouldn't be committing that folder, since we already have an [`interim` branch](https://github.com/openstreetmap/id-tagging-schema/commits/interim/) for this purpose.

but now it's too late to change, because it will break downstream apps who ignore our releases and pull directly from github.

matkoniecz: I guess we could do a breaking change at some point. Though it may be not worth it.

scripts/lib/build.ts
matkoniecz: we need it later to confirm that referenced preset exists, and give useful error if it does not

I see no benefit from first using `isReference` and then running regexp anyway

==========================

I added pattern attribute.

matkoniecz: and note that those are left as references in published materials, not dereferenced and expanded at build time

data/presets/building/semidetached_house.json
k-yle, line 1: what is a data consumer supposed to do in this situation?

Image

 a `power=terminal` node is vertex of 3 ways:
- `building=house`, which explicitly does NOT allow `power=terminal` as a vertex
- `power=minor_line`, which explicitly does allow `power=terminal` as a vertex
- `landuse=residential`, which has no vertex definitions.

is this valid?

data/presets/building/house.json
```diff
@@ -20,5 +20,10 @@
         "residence"
     ],
     "matchScore": 0.5,
-    "name": "House"
+    "name": "House",
+    "related": {
+        "expectedVertices": [
+            "{entrance}"
```
k-yle, line 26: how are data consumers meant to interpret this? 

- by linking to  `entrance.json`, are we saying that a sub-preset like `{entrance/main}` are not allowed? 
- or do you want to use an implicit heirachy based on the file name?

data/presets/building/house/terraced.json
```diff
@@ -32,5 +32,8 @@
     "name": "Townhouse",
     "aliases": [
         "Row House"
-    ]
+    ],
+    "related": {
```
k-yle, line 36: this is my biggest concern with this PR. now we're saying that `entrance=*` is the only vertex allowed on a terraced house, but there are hundreds of other valid options.

for example, if i walk down the street, i will find [this kind of `house=terraced`](https://www.heritage.org.nz/list-details/568/Listing) ([wiki page](https://en.wikipedia.org/wiki/Terrace_houses_in_Australia)) with vertices such as:
- `memorial=plaque`
- `amenity=public_bookcase`
- `emergency=fire_service_inlet`
- `amenity=parking_entrance`
- `advertising=billboard`
- `fire_hydrant:type=wall`
- `barrier=height_restrictor`
- anything with `support=wall`
- `addr:*`
- `power=terminal`
- and probably many more


given https://github.com/openstreetmap/id-tagging-schema/pull/2454#discussion_r3622351441, i think we should probably run a script against the entire planet first, to see what tags are common. or are there any QA tools/scripts (like your own tool?) that already collected this data?

SCHEMA.md
k-yle: some people configure their IDEs to hyperlink common patterns like `JIRA-1234` to open http://jira.example.net/1234, or `{a/b/c}` to open the corresponding `a/b/c.json` file. 

i don't know if anyone has bothered to setup this for id-tagging-schema. that's the only minor benefit i can think of, for using the existing `{}` syntax.

data/presets/building/house/terraced.json
matkoniecz: > i think we should probably run a script against the entire planet first, to see what tags are common

I am doing it already :) Can share either code for it or debug output or cases where we have clearly bogus data

> is the only vertex allowed on a terraced house

my intention was to claim it is the only privileged vertex - after all, house may have life-size shark model sticking out of it, or tree, or power tower, or monument, or locomotive or aircraft or aquarium, or lion enclosure or meteorite, or cannon ball, or cannon, or air-defense installation, or 3M54-1 Kalibr UXO, or information terminal or turnstile or wayside shrine or cross or...

data/presets/building/semidetached_house.json
matkoniecz: in this theoretical case at least one of parent ways is allowing so it should be listed among promoted ones and treated as a clearly valid

(currently `power=terminal` is not on any list as it is not in iD presets as PR https://github.com/openstreetmap/id-tagging-schema/pull/1637 is stuck on naming, but if it would be listing it like above would make sense)

I will take look at docs are we handling it clearly

data/presets/building/house/terraced.json
matkoniecz: I rewrote documentation 

> A list of preset IDs that may be used as a vertex of this preset

to

> A list of preset IDs that are expected to be used as a vertex of this preset

is the good summary of what I expressed

k-yle: > my intention was to claim it is the only privileged vertex [...]

this somewhat contradicts the statement in the docs:

> In addition QA warning may be shown if presets from outside of such list are present.

If we are intentionally not including some valid options, then this data is not suitable for validation warnings (yet). so i think we should clarify that in the docs?

```diff
- In addition QA warning may be shown if presets from outside of such list are present.
+ In the future, the list of vertices will be expanded, so that it could be used for QA warnings. Currently the data coverage is not suitable for QA warnings.
```

something like this?

data/presets/building/semidetached_house.json
matkoniecz: tried to improve it, I think it is a good way to do this, but feedback welcome

k-yle: we should mention this in the docs, it's a very important decision. i originally hoped that this PR could prevent situations like this:

image

However, since you answered "at least one of parent ways", this dragged-node is considered valid. 

There is no good answer to this question, both options have annoying drawbacks. so we should clearly document what our approach is. 

also because it affects whether `power=terminal` needs to be listed in every single `man_made=*` structure, so this topic will keep coming up in code reviews.

data/presets/building/house.json
matkoniecz: Hmm. "A list of preset IDs that are expected to be used as a vertex of this preset" seemed to be clear enough, but I need to change it.

Should I do "A list of preset IDs that are expected to be used as a vertex of this preset (this specific presets, does not include sub-presets - if `{entrance}` is listed it includes only this specific preset, not also `{entrance/main}`)" or similar? Maybe as an extra example in schema.md? Extra sentence at the end?

data/presets/building/house/terraced.json
matkoniecz: Right.

Or maybe introduce concept of closed/open sets?

For example listing all cases what can be validly attached to `power=line` as vertex seems achievable. While doing it for roads and buildings is at least much harder and will take longer to work reliably.

Maybe marking `power=line` listing with "you can complain about anything unlisted" would allow to start using it for validation there while using it only for suggestions elsewhere?

data/presets/building/semidetached_house.json
matkoniecz: > i originally hoped that this PR could prevent situations like this

the same but I am getting endless parade of special cases, many valid.

For example you would think that `highway=motorway` will not get valid `highway=stop` - but https://www.openstreetmap.org/node/144570623 is not blatantly unreasonable.

Or `pipeline/valve` on `man_made/pier` - see https://www.openstreetmap.org/node/4218396016'

Or trees as vertex of water area - see https://www.openstreetmap.org/node/352778180

matkoniecz: > However, since you answered "at least one of parent ways", this dragged-node is considered valid.

Maybe additional concept can be setup here, where point feature can be considered as exclusive use as vertex solely on ways that listed it as a valid vertex?

And while art installation can be used as vertex of buildings, cliffs, waterways, dams, roads and everything else, `power=pole` should be only used where it is listed as valid.

But I already have seen path going under power tower, with `power=tower` as vertex of both `power=line` and `highway=path` - and this not being blatantly invalid. Though maybe QA should not complain about such cases, while iD not allowing gluing of `power=pole` into "illegal" ways?

matkoniecz:
I added it to https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2026#Community_Chat_August_5%2C_12%3A00_-_13%3A00_%28UTC%29 discussion, if anyone wants to join feel free!

data/presets/building/house.json
k-yle: okay, if that's your intention, then i guess thedocumentation is fine. But then we also need to fix this problem by adding `{entrance/main}` and `{entrance/emergency}` to this file right?

and for regional presets, you'll need to list every variant

data/presets/building/semidetached_house.json
k-yle: > Or maybe introduce concept of closed/open sets?

> Maybe additional concept can be setup here, [...]

that's why i strongly prefer the original approach, where this data is defined on the vertex, not the on the way. 

- This allows you to make some verticies strict (like `power=pole`) but others can be weak (like `tourism=artwork`, `memorial=plaque`). no hacks required.
- It also seems much easier to tackle this problem 1 vertex at a time, rather than trying to define an impossibly long list of children. Both from a data perspective, and intuitively.

data/presets/building/house/terraced.json
k-yle: that works, but an even easier solution is https://github.com/openstreetmap/id-tagging-schema/pull/2454#discussion_r3719440221

data/presets/building/house.json
matkoniecz: I expect single family houses to typically have neither of those. 

(`building=house` used as a church may have `entrance=main` but it is an edge case, maybe emergency exits are more popular elsewhere)

data/presets/building/semidetached_house.json
matkoniecz: How you would define order of entries then? For example `power/line` should list `power/tower` first, before `power/generator` variants.

> This allows you to make some verticies strict (like power=pole) but others can be weak (like tourism=artwork, memorial=plaque). no hacks required.

you still would need per-preset option anyway, not sure why with current solution a per-preset option would be worse

matkoniecz: (I am in general not caring much about exact syntax as long as it gets job done and would be happy to test alternatives)

data/presets/building/house/terraced.json
matkoniecz: replied there

data/presets/building/semidetached_house.json
matkoniecz: Benefit of per-vertex option:

- strict/weak matching is defined together with which presets are matching in vertex relationship - rather one defined on ways and another on vertices
- it is a bit easier to think about valid parents for given preset than about valid children

Benefit of per-parent option:

- it is possible to define order in suggestions in a clear way
- if something goes wrong it will be easier to debug and change list of suggested vertex presets


anything else?

matkoniecz:
I was thinking yesterday how to handle it and I propose to add option, defined on a vertex presets

`onlyAsExpectedVertex` - defines is vertex matching strict or weak for this preset when used as vertex, mattering when vertex would be placed on more than way at the same time. Defaults to false. When set to true it indicates that all parent ways must list it in `expectedVertices` to make this preset valid as vertex. Editor may prevent actions that would violate this rule. In practical terms it may be blocking of attaching nodes to ways, blocking merging nodes, not listing as valid preset when tagging vertex nodes when not way match.

This can be set on for example `power/tower` preventing merging nodes with this preset onto rivers or roads.

Note that some objects like statues may be promoted as vertex in some cases, and be valid in many more.  For example statues may be valid vertex of rivers and who knows what else. In such case this value should remain as default false. 

-----------------------------

Also, I was thinking about `ListOnlyApprovedVertices` that could be additional property added to ways and would instruct editors to further discourage use of presets other than listed.

Mappers may be still able to click "show also very unusual values" to map say artwork attached to `power=line` but in typical use when selecting vertex of `power/line` only presets declared in `expectedVertices` would be shown and nothing else.

`ListOnlyApprovedVertices` set to true could also prevent merging say `shop=whatever` onto `power=line`

data/presets/building/semidetached_house.json
matkoniecz: expanded on my idea in https://github.com/openstreetmap/id-tagging-schema/pull/2454#issuecomment-5190205124

presumably the same kind of weak/strict setting would be used when defining valid parents? Or have you thought about something else?

matkoniecz:
These two options are adding extra complexity, but not sure how to otherwise handle complexity that for some presets we can express "and we listed everything valid, everything else is at least suspect" and some are more "anything goes"

For example `building=` objects can have amazing managerie of things attached to them, while `power=line` has far more enumerable list. Which makes sense as buildings will not actively kill when touching unprepared objects, buildings existed for far longer and in far wider variety etc..

So some behavior makes sense for some objects and is useful there while being pointless elsewhere.

And even if solely power lines would benefit it would be already worth implementing, I would say.

data/presets/building/house/terraced.json
matkoniecz: > something like this?

tried in https://github.com/openstreetmap/id-tagging-schema/pull/2454/changes/55d62fbf5af95cf3e55a6677afbe17a3e6c93310

Dimitar5555:
> Also, I was thinking about `ListOnlyApprovedVertices` that could be additional property added to ways and would instruct editors to further discourage use of presets other than listed.

Would `allowedVertices` (for closed lists) and `preferredVertices`/`suggestedVertices` (for open lists) work?

matkoniecz:
idea from meeting: special wildcard entry

matkoniecz:
> Would allowedVertices (for closed lists) and preferredVertices/suggestedVertices (for open lists) work?

at least for me it is not immediately clear which is which

though in general different names would be possible

matkoniecz:
@tyrasd @k-yle 

I am considering this version ready for review.

scripts/lib/build.ts
```diff
@@ -1021,6 +1021,28 @@ function validatePresetFields(presets: AllPresets, fields: AllFields) {
...
+        if (regexOutput == null) {
+          process.stderr.write('got' + preset.related[relatedKey][vertexIndex] + '" in "expectedVertices" in "' + presetID + '" (' + preset.name + ') where referenced preset in {} was expected - this syntax is invalid\n');
+          process.stdout.write('\n');
+          process.exit(1);
+        }
+        let foreignPresetID = regexOutput[0];
+        if (presets[foreignPresetID] === undefined) {
```
k-yle, line 1058: from https://github.com/openstreetmap/id-tagging-schema/pull/2454#discussion_r3621785884 - we're not checking the geometery here. currently you can list features that only support `point`, not `vertex`

SCHEMA.md
```diff
@@ -224,6 +224,69 @@ Alternatively, `locationSetCrossReference` can be use to reference the `location
...
+* then `natural/tree/needleleaved/deciduous`
+* then other vertex presets
+
+If `power/generator/source/wind` had only two recent uses then probably `power/tower` should stay as the first option.
+
+Note that as world is complex this is not exhaustive. For example `tourism/artwork/statue` may be not listed as expected vertex of `waterway/river` as it is not an expected situation. But it does not make [this statue](https://www.openstreetmap.org/node/1374406687/history/11) invalidly mapped.
+
+In some cases single vertex may be member of multiple ways. In such case expected strategy would be to group listed options by how many parent ways list them as valid vertices.
+
+For example if `highway/track` and `highway/stream` intersect and both list `ford` preset as a valid vertex, then it should be listed before ones appearing on only one of parent ways.
```
k-yle: `highway=stream` ?

Image

MIGRATION_GUIDE.md
```diff
@@ -1,3 +1,7 @@
+# unreleased version Migration Guide
+
+`related.expectedVertices` and `related.expectedVerticesCrossReference` were added, editors may use this to enhance editing experience of vertices.
```
k-yle, line 3: this file is only for breaking changes right?

scripts/lib/build.ts
```diff
@@ -1021,6 +1021,28 @@ function validatePresetFields(presets: AllPresets, fields: AllFields) {
...
     }
 
+    let relatedKeys = ['expectedVertices'] as const;
+    for (let relatedKeyIndex in relatedKeys) {
+      let relatedKey = relatedKeys[relatedKeyIndex];
+      if (!preset.related?.[relatedKey]) continue;
+
+      for (let vertexIndex in preset.related[relatedKey]) {
+        const regexOutput = betweenBracketsRegex.exec(preset.related[relatedKey][vertexIndex])
+        if (regexOutput == null) {
```
k-yle, line 1052: given https://github.com/openstreetmap/id-tagging-schema/pull/2454#discussion_r3703655728, this whole if-statement is impossible and could be deleted

k-yle (commented):
since this data won't be used for QA/validation, all my implementation questions are resolved. one last round of comments, sorry 😅

MIGRATION_GUIDE.md
matkoniecz: We listed nonbreaking changes there iirc

For me it makes sense for things where editors may or need to take action

scripts/lib/build.ts
matkoniecz: I will check,  not sure is typescript smart enough to detect that null is impossible

SCHEMA.md
tyrasd: > the standard convention is to wrap preset/field IDs in { + }. Should we follow this convention?

well, interestingly the convention started initially for the `fields` property which can list both fields or presets, in order to distinguish them. Then, I just reused the `{…}` in the `*CrossReference` properties (where the brackets would not actually be needed!). So, it's a bit of a mix at the moment.

One one argument could be: everything that is resolved during build-time should have the brackets (e.g. all _CrossReference_ props, but also `{…}` in `fields/moreFields`), but eveything that is included as is in the `dist` output should not need to be decorated. Does that make sense @k-yle ?

tyrasd (commented):
see https://github.com/openstreetmap/id-tagging-schema/pull/2454#discussion_r3805004233

dist/presets.d.json.ts
```diff
@@ -98,6 +98,21 @@ export interface Preset {
...
+    /**
+     * An list of preset IDs that may be used as a vertex of this preset, when this preset is used on a line or area. List is ordered, starting from ones more likely to be selected by mapper.
+     *
+     * @minItems 1
+     */
+    expectedVertices?: string[]
+    /**
+     * An string referencing another preset which has a related.expectedVertices to be used also here
+     */
+    expectedVerticesCrossReference?: string
```
matkoniecz, line 114: `expectedVerticesCrossReference` `expectedVertices` should raise error if both used for the same preset

SCHEMA.md
```diff
@@ -226,6 +226,69 @@ Alternatively, `locationSetCrossReference` can be use to reference the `location
...
+
+###### `expectedVertices`
+
+A list of preset IDs that are expected to be used as a vertex of this preset, when this preset is used on a line or area. 
+
+For example `power=line` may have
+
+```
+    "related": {
+        "expectedVertices": [
```
tyrasd, line 241: thinking about it, we could allow this list to work similarly to how the `fields` property works: i.e. that you can reference either the list from another preset by writing `{presetID}` or by including a vertex-preset by just writing `presetID`. That allows to extend the list of expected presets for a sub-preset of another preset, e.g. let's say we have a preset for via ferrata, which can have in addition to generic hiking-path vertices (e.g. a `ford`), also special kind of climbing-related vertices. That could be modeled by:

```json
    "related": {
        "expectedVertices": [
            "{highway/path/hiking}",
            "climbing/hook"
        ]
    }
```

(not the best example, but I hope it does get the idea across).

As this is very similar to how `fields` already work, this would not add new complexity to the schema, and allow us to not need the separate `expectedVerticesCrossReference` property for referencing.

SCHEMA.md
k-yle: yeah I don't mind either way, that sounds reasonable

matkoniecz:
> one last round of comments, sorry 😅

after relation structure PR circus I definitely do not get to complain

scripts/lib/build.ts
matkoniecz: oh right, I can use !! ending with `regexOutput!![0]`

scripts/lib/build.ts
matkoniecz: done, though I have run into a bug with `preset.name` missing

while debugging it I opened https://github.com/openstreetmap/id-tagging-schema/pull/2817 for another build-time bug

scripts/lib/build.ts
```diff
@@ -1024,6 +1024,29 @@ function validatePresetFields(presets: AllPresets, fields: AllFields) {
...
+      for (let vertexIndex in preset.related[relatedKey]) {
+        const regexOutput = betweenBracketsRegex.exec(preset.related[relatedKey][vertexIndex])
+        let foreignPresetID = regexOutput!![0];
+        if (presets[foreignPresetID] === undefined) {
+          process.stderr.write('Unknown preset "' + foreignPresetID + '" referenced in "expectedVertices" array of preset "' + presetID + '" (' + preset.name + ')\n');
+          process.stdout.write('\n');
+          process.exit(1);
+        }
+        if (!presets[foreignPresetID].geometry.includes("vertex")) {
+          process.stderr.write('preset not valid as a vertex "' + foreignPresetID + '" referenced in "expectedVertices" array of preset "' + presetID + '" (' + preset.name + ')\n');
```
matkoniecz, line 1058: `preset.name` here is undefined

to reproduce: edit `data/presets/power/minor_line.json` to add `highway=primary` to `expectedVertices`

matkoniecz: this seems to affect also other error messages, see also #2817

scripts/lib/build.ts
matkoniecz: now linter fails PR with comment `Error:   1034:31  error  Forbidden extra non-null assertion  @typescript-eslint/no-extra-non-null-assertion`

data/presets/building/house.json
matkoniecz: @k-yle I plan to mark it as resolved, let me know if you think that work remains to be done

data/presets/building/semidetached_house.json
matkoniecz: @k-yle I plan to mark it as resolved, let me know if you think that work remains to be done

SCHEMA.md
matkoniecz: changed, will convert `expectedVerticesCrossReference` property to use `{}` wrapped

scripts/lib/references.ts
```diff
@@ -111,6 +111,39 @@ export function dereferenceUntranslatedContent(presets: AllPresets, fields: AllF
...
+
+          if (!referencedPreset.related.expectedVertices) {
+            throw new Error(
+              `Preset “${presetID}” references “${otherPresetID}” in related.expectedVertices, but this preset has no related.expectedVertices field.`,
+            );
+          }
+
+          // replace the reference with every field. decrement i to reprocess this array index.
+          // this is necessary as it can also be a reference
+          preset.related.expectedVertices.splice(i--, 1, ...referencedPreset.related.expectedVertices);
```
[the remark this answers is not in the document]
matkoniecz: the same as in few other places, I think that not adding loop detection is fine

SCHEMA.md
```diff
@@ -226,6 +226,72 @@ Alternatively, `locationSetCrossReference` can be use to reference the `location
...
+List is ordered, entries expected to be used more often by mappers listed first. 
+Editor software may use this list to show more relevant presets when selecting preset for `power=line` vertex.
+
+And following syntax may be used to reference values from other preset.
+
+```json
+    "related": {
+        "expectedVertices": [
+            "{building/house}",
+            "emergency=emergency_ward_entrance"
```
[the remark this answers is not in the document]
matkoniecz: damm, impressive

data/presets/power/line.json
```diff
@@ -19,5 +19,19 @@
     "tags": {
         "power": "line"
     },
-    "name": "Power Line"
+    "name": "Power Line",
+    "related": {
+        "expectedVertices": [
```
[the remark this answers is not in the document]
matkoniecz: this is done knowingly and may happen elsewhere IIRC, and detecting and aborting in advance would be quite annoying

matkoniecz:
@k-yle @tyrasd @tordans Review would be appreciated!

PR #2468, opened 2026-07-11
Use "Lungs" icon for pulmonologists

fibonacci-matrix:

MakiDoctor
Fa7SolidLungs

### Related issues

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Tag:healthcare:speciality%3Dpulmonology

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/tags/healthcare:speciality=pulmonology

### DIFF
--- a/data/presets/amenity/doctors/pulmonology.json
+++ b/data/presets/amenity/doctors/pulmonology.json
@@ -1,5 +1,5 @@
 {
-    "icon": "maki-doctor",
+    "icon": "fas-lungs-virus",
     "geometry": [
         "point",
         "area"
--- a/data/presets/amenity/doctors/pulmonology.json
+++ b/data/presets/amenity/doctors/pulmonology.json
@@ -1,5 +1,5 @@
 {
-    "icon": "fas-lungs-virus",
+    "icon": "fas-lungs",
     "geometry": [
         "point",
         "area"


Dimitar5555:
I'm not sure if this one is suitable here. Pulmologists also treat non-contagious diseases so the virus icon might be confusing.

Ideally there would be an icon with just lungs, and this icon could be used for infectionist/infectious diseases.

matkoniecz (changes requested):
is there available version without that cogs? (virus apparently)

matkoniecz:
BTW, thanks for review of available icons! Though if there are many more body part possibilities maybe better to diversify for now? And do a second wave of those later?

I expect at least most of them to be merged, but lets see what other will say.

And thanks again for taking look. I had it on my TODO list but now I will be able to do something else :)

fibonacci-matrix:
> Though if there are many more body part possibilities maybe better to diversify for now? And do a second wave of those later?

@matkoniecz agreed. After PR2471, I reached a logical stopping point, and for now I don't plan to add any more healthcare-related icons, although I still have a few ideas left

matkoniecz (approved):

matkoniecz:
As seen - body icons for doctors were discussed and liked!

Are you potentially interested in finding more or adding them also to dropdown selector ? See https://github.com/openstreetmap/id-tagging-schema/issues/2491

matkoniecz:
And thanks for finding them!

PR #2479, opened 2026-07-13
allow typescript to be used for the build scripts

k-yle:
follow up to https://github.com/openstreetmap/id-tagging-schema/pull/2446#discussion_r3559490532

This PR updates the configuration so that the build scripts can be written in TypeScript. Also renamed 1 file as an example. The other files can be converted in follow up PRs, since they're more complicated


matkoniecz (commented):
I do not understand TS-related things as much as I would like, but this looks fine

and it would be clearly useful, see stupid bug I made recently that would be spotted by TSization

matkoniecz:
@tyrasd @tordans are you be any chance interested in reviewing this? Maybe you can give better feedback on TS than me

tordans (approved):
I had Cursor Composer compare the `dist` output from `main` with this and it is identical. Since this is a good base for other changes, I think we should merge this now and address all future improvements in follow up PRs.

<details><summary>Full LLM Compare Results</summary>

> Yes — if `dist/` is identical after a clean build on the same input data, the build output hasn't changed. That's the right way to validate a refactor like this.
> 
> | | Commit | Description |
> |---|---|---|
> | **Old** | `3c780a19` | Last commit before the TS migration (includes the #2446 bug fixes) |
> | **New** | `0a2ff33c` (`kh/ts1`) | TS migration commit |
> 
> Both were built from isolated git worktrees on the same `data/`, with `rm -rf dist interim` before each run, then `node scripts/dist.js` (no Transifex fetch).
> 
> **Result: identical.** 23 files in each `dist/`, `diff -r` reported no differences, and SHA-256 checksums matched for every file.
> 
> So the TypeScript migration does not change build output. The runtime behavior of the build pipeline is the same on this data snapshot.
> 
> **Caveats worth keeping in mind:**
> - This compares only the TS commit on top of the bug-fix commit — not `main` vs `kh/ts1`, since `main` has moved ahead with unrelated data changes.
> - Builds were done without the `translations` flag, so only the core dist artifacts were compared (JSON, minified JSON, type defs, `en` translations) — not the full set of Transifex-downloaded locale files.

</details>

PR #2480, opened 2026-07-13
replace confusing mix of callbacks and promises

k-yle:
the translation fetching logic has a mix of callbacks and promises which is a bit hard to follow, and doesn't work well with typescript.

since these functions already use promises, it's easy to simplify the code and remove the callback syntax

tested by running `npm run dist translations` before & after, and confirming that there was no change to `dist/*` or `interim/*`


tordans (approved):
I tested the build output which did not change so I think we can merge this now and address changes - if any - in follow ups.

PR #2482, opened 2026-07-14
Replace icon for preset `natural=spring`

Geo-2695:

Replaced the original generic `maki-water` icon with the [Pinhead icon](https://github.com/waysidemapping/pinhead/pull/256) .

An update to the Pinhead release may be required for proper display

### Related issues

#41


matkoniecz:
very similar was rejected already in https://github.com/openstreetmap/id-tagging-schema/pull/2316#issuecomment-4778059981 when I proposed it

PR #2533, opened 2026-07-14
validate `locationSet` at compile-time

k-yle:
follow up to https://github.com/rapideditor/location-conflation/pull/78

Currently we don't test if `locationSet` is valid. Now, if you use an invalid value like [`zz`](https://r12a.github.io/app-subtags/index?lookup=ZZ) or [`Q3916957`](https://wikidata.org/wiki/Q3916957), it will throw an error.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "ISC",
       "devDependencies": {
         "@eslint/js": "^10.0.1",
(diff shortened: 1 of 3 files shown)


matkoniecz:
https://github.com/rapideditor/location-conflation/blob/414644019f623703f114863f726ed0e30c216b23/API.md#validatelocationset

matkoniecz (approved):
thanks!

PR #2538, opened 2026-07-15
Allow cross referencing `icon` and `icons` entries

tordans:
Closes #2503 (originally [ideditor/schema-builder#275](https://github.com/ideditor/schema-builder/issues/275))

Port of [openstreetmap/schema-builder#296](https://github.com/openstreetmap/schema-builder/pull/296), recreated here after schema-builder was merged into this repository.

The idea is: Now that we build the final file without references, we can add some more to make handling preset `icon` and field `icons` easier.

- Reference the `icon` from one preset in another
- Reference the `icon` from a preset in the `icons` list of a field

---

Is this a real issue? – Yes, I even build https://osmberlin.github.io/tagging-schema-browser/?reference=release&iconMismatch=[%22mismatch%22] just to find those cases, so it would be great to have a way to think about this less.

A follow up could be to enforce this in the build script to remove the need for external checks like the one above (which make that app slower) and tighten the contract during build. – But that is for a follow up discussion.

---

I built this using Cursor LLM and reviewed the code and docs.

---

**Testing**

Unit tests were added in `scripts/lib/__tests__/schema-builder.test.js` (T1–T9): preset icon resolution, chained references, field `icons` values, `iconsCrossReference` + `{presets/…}`, validation errors (invalid prefix, cycle, missing preset), and interim `icons.json` listing only resolved ids.

`npm run lint` and `npm test` pass locally.

Manual end-to-end check:

> ## Steps to reproduce
>
> **Goal:** Compare two full `dist/` builds so the only difference is the resolved `playground` combo icon for `slide` after switching from a literal icon to `{presets/playground/slide}`.
>
> **Baseline (before your change)**
>
> 1. Check out `main`.
> 2. Ensure `data/fields/playground.json` defines the combo **`icons`** map with a **literal** slide icon that is **not** the same as the Slide preset, e.g.  
>    `"slide": "temaki-slide2"`  
>    while `data/presets/playground/slide.json` keeps  
>    `"icon": "roentgen-slide"`.
> 3. `npm run build` and `npm run dist`.
> 4. Copy the output somewhere stable, e.g.  
>    `cp -a dist /tmp/dist-before`.
>
> **After your change (cross-reference)**
>
> 5. Check out this branch.
> 6. In `data/fields/playground.json`, set  
>    `"slide": "{presets/playground/slide}"`  
>    (preset id for `data/presets/playground/slide.json` is `playground/slide`; the wrapper must be **`{presets/…}`**, not `{preset/…}`.)
> 7. `npm run build` and `npm run dist`.
> 8. `cp -a dist /tmp/dist-after`.
>
> **Compare**
>
> 9. `diff -rq /tmp/dist-before /tmp/dist-after`  
>    Expect differing files: `fields.json`, `fields.min.json` (and `.d.json.ts` schema doc lines if you rebuild those).
>
>    ### Full diff — `dist/fields.json` (expected hunk)
>
>    ```diff
>    -            "slide": "temaki-slide2",
>    +            "slide": "roentgen-slide",
>    ```

Fazit:

This works as expected for this test case; the icon changes because we had a mismatch before.


SCHEMA.md
```diff
@@ -669,7 +677,9 @@ Combo field types can accept key-label pairs in the `options` value of the `stri
 
 ##### `iconsCrossReference`
 
-An optional property to reference to the icons of another field, indicated  by using that field's name contained in brackets, like `{field}`. This is for example useful when there are multiple variants of fields for the same tag, which should all use the same icons.
+An optional property to copy the entire `icons` object from another field by giving that field's id in brackets (no `presets/` prefix)—for example `{kerb}` copies the field whose id is `kerb`. Useful when several field variants share the same option icons.
```
matkoniecz:
```suggestion
An optional property to reference to the `icons` object of another field, indicated  by using that field's name contained in brackets, like `{field}`. `presets/` prefix is not included - for example `{kerb}` copies the field whose id is `kerb`. This is useful for example when there are multiple variants of fields for the same tag, which should all use the same icons.
```

Trying shorter sentences like it was before. Other changes that were introduced should be preserved, except `—`

scripts/lib/references.ts
```diff
@@ -4,6 +4,100 @@ export function isReference(string: string) {
   return string.startsWith('{') && string.endsWith('}');
 }
 
+/**
+ * Resolve "{presets/<id>}" to a concrete icon id (following chains). Only
```
matkoniecz: later docs claim that `presets/` part is supposed to be omitted

it sounds like it resolves `{<id>}` and id must represent preset it? Or is it getting here id enriched by `presets/` prefix that at some point got automatically added?

matkoniecz: oh wait, so it has syntax different than `iconsCrossReference`

SCHEMA.md
```diff
@@ -650,7 +654,11 @@ For `identifier` fields, the regular expression that valid values are expected t
 
 ##### `icons`
 
-For combo and radio fields, the `icons` object might contain the name of icons which represent the different values of the field. More information about available icon sets and usage of icons can be found on the [icons subpage](ICONS.md).
+On [combo / dropdown](#combodropdown-fields) and [radio](#radio-buttons) fields, the `icons` object maps each option key to an icon id shown beside that value in the editor, e.g. `"zebra": "iD-crossing_markings-zebra"` in the snippet below. More information about available icon sets and usage of icons can be found on the [icons subpage](ICONS.md).
+
+You may set each value to `"{presets/<preset-id>}"` instead of a literal id, with the same rules as [preset `icon`](#icon).
```
matkoniecz:
```suggestion
You may also instead use `"{presets/<preset-id>}"` references instead of a literal id, with rules as for [preset `icon`](#icon).
```

if we want `presets` prefix then we cannot claim that rules are the same

"may set each value" also sounded like you cannot reference only for some

SCHEMA.md
```diff
@@ -669,7 +677,9 @@ Combo field types can accept key-label pairs in the `options` value of the `stri
 
 ##### `iconsCrossReference`
 
-An optional property to reference to the icons of another field, indicated  by using that field's name contained in brackets, like `{field}`. This is for example useful when there are multiple variants of fields for the same tag, which should all use the same icons.
+An optional property to copy the entire `icons` object from another field by giving that field's id in brackets (no `presets/` prefix)—for example `{kerb}` copies the field whose id is `kerb`. Useful when several field variants share the same option icons.
+
+If a field uses `iconsCrossReference`, the builder copies the other field’s `icons` map first, then expands every `{presets/…}` value in **all** field `icons` maps (including the copy).
```
matkoniecz: does order actually matter here?

scripts/lib/__tests__/schema-builder.test.js
```diff
@@ -245,4 +245,216 @@ describe('schema-builder', () => {
...
+        outDirectory: _workspace + '/dist',
+        taginfoProjectInfo: taginfo,
+      });
+      const presets = JSON.parse(
+        fs.readFileSync(_workspace + '/dist/presets.json', 'utf8'),
+      );
+      expect(presets.t_chain_c.icon).toBe('maki-x');
+    });
+
+    it('T3 resolves field icons value from {presets/id}', async () => {
```
matkoniecz:
```suggestion
    it('T3 resolves value from {presets/id} for field icons', async () => {
```

I think it is clearer, previously it sounded to me like it was possible to reference icon specified for given field value

scripts/lib/__tests__/schema-builder.test.js
```diff
@@ -245,4 +245,216 @@ describe('schema-builder', () => {
...
+        outDirectory: _workspace + '/dist',
+        taginfoProjectInfo: taginfo,
+      });
+      const fields = JSON.parse(
+        fs.readFileSync(_workspace + '/dist/fields.json', 'utf8'),
+      );
+      expect(fields.t_fld_combo.icons.yes).toBe('iD-bus');
+    });
+
+    it('T4 resolves {presets/…} inside icons after iconsCrossReference', async () => {
```
matkoniecz:
```suggestion
    it('T4 iconsCrossReference works if referenced icons contains icon references', async () => {
```

in theory resolving them could happen before and be implemented in way that keeps it working

matkoniecz:
I did some nitpicking, hopefully useful one. While this function extraction makes sense it is quite annoying that it is in the same comment.

I feel the most confident about https://github.com/openstreetmap/id-tagging-schema/pull/2538#pullrequestreview-4712154051

also, pinging @k-yle

matkoniecz:
In general: why that `presets/` prefix? Do we want to add it to other references in followup PR (definitely not this one)? If not, why specifically here?

SCHEMA.md
matkoniecz: are you expecting to have time to look into it?

I am considering making own PR based on this one, or making own one manually, as ideally we would have this functionality to allow making https://github.com/openstreetmap/id-tagging-schema/pull/2633 and similar in a smarter way

tordans:
> In general: why that `presets/` prefix? Do we want to add it to other references in followup PR (definitely not this one)? If not, why specifically here?

Same-entity refs omit the type prefix (`iconsCrossReference: "{kerb}"`, `fields: ["{shop}"]`). Cross-entity refs use a type prefix (`locationSetCrossReference`, option labels, and now icon refs). Icon refs always pull a **preset’s** `icon`, including when used from field `icons`, so they use `{presets/…}`. 

I think that setup is fine as is but happy to rework it based on further feedback.

tordans:
@matkoniecz I just applied most of the changes with small adjustments and then merged develop in (which broke the comment-references…). I would squash this into one commit before we merge to make this a clean PR.

No hurt feelings if you want to do something different / your own.

schemas/field.json
```diff
@@ -402,15 +402,15 @@
             ]
         },
         "icons": {
-            "description": "For combo and radio fields: Name of icons which represents different values of this field",
+            "description": "For combo and radio fields: names of icons that represent different values of this field. Values may be icon ids or \"{presets/<preset-id>}\".",
```
matkoniecz, line 405:
```suggestion
            "description": "For combo and radio fields: icons that represent different values of this field. Values may be icon ids or \"{presets/<preset-id>}\" referencing their icons.",
```

reference is not icon name

scripts/lib/__tests__/schema-builder.test.js
```diff
@@ -46,7 +46,7 @@ describe('schema-builder', () => {
     expect(schemaBuilder && schemaBuilder.validate).not.toBeUndefined();
   });
 
-  it('runs validate', () => {
+  it('runs validate', async () => {
```
matkoniecz, line 49: do we need this seemingly unrelated change?

if yes, why?

scripts/lib/__tests__/schema-builder.test.js
```diff
@@ -56,7 +56,7 @@ describe('schema-builder', () => {
         name: 'Natural Feature'
       }
     });
-    schemaBuilder.validate({
+    await schemaBuilder.validate({
```
matkoniecz, line 59: do we need this seemingly unrelated change?

if yes, why?

scripts/lib/references.ts
```diff
@@ -169,6 +264,9 @@ export function dereferenceUntranslatedContent(presets: AllPresets, fields: AllF
       delete field.locationSetCrossReference;
     }
   }
+
+  // 11. preset `icon` and field `icons` values may use "{presets/<id>}" for another preset's icon id
```
matkoniecz, line 268:
```suggestion
  // preset `icon` and field `icons` values may use "{presets/<id>}" for another preset's icon id
```

these mysterious numberings got dropped upstream, right?

scripts/lib/references.ts
```diff
@@ -4,6 +4,101 @@ export function isReference(string: string) {
...
+  if (isReference(next)) {
+    return resolveIconRefToId(presets, next, contextMessage, visitedPresetIds);
+  }
+  return next;
+}
+
+function dereferencePresetIconStrings(presets: AllPresets, fields: AllFields) {
+  for (const presetID in presets) {
+    const preset = presets[presetID];
+    if (typeof preset.icon === 'string' && isReference(preset.icon)) {
```
matkoniecz, line 77: why `typeof preset.icon === 'string' ` would be needed? I think it is enforced already?

scripts/lib/references.ts
```diff
@@ -4,6 +4,101 @@ export function isReference(string: string) {
...
+  }
+  if (!id) {
+    throw new Error(
+      `Invalid icon reference “${ref}” in ${contextMessage}: missing preset id after “presets/”.`,
+    );
+  }
+
+  if (visitedPresetIds.has(id)) {
+    throw new Error(
+      `Cycle detected while resolving icon reference “${ref}” in ${contextMessage}: preset “${id}” appears more than once in the chain.`,
```
matkoniecz, line 49:
```suggestion
      `Cycle detected while resolving icon reference “${ref}” in ${contextMessage}: preset “${id}” appears in the loop.`,
```

well, with loop all appear infinite numbers of times

scripts/lib/references.ts
```diff
@@ -4,6 +4,101 @@ export function isReference(string: string) {
   return string.startsWith('{') && string.endsWith('}');
 }
 
+/**
+ * Resolve "{presets/<id>}" to a concrete icon id (following chains).
+ * Icon refs always require the `presets/` prefix (unlike iconsCrossReference,
+ * which uses a bare field id). `{fields/…}` and bare `{id}` throw.
+ */
+function resolveIconRefToId(
```
matkoniecz, line 12: why exactly we need to allow going through multiple references?

maybe allowing at most one step would be fine and would really simplify the code?

matkoniecz: are other referencing allowing such multistep-indirect references?

matkoniecz (changes requested):
see comments

note, there may be valid reasons to dismiss them

matkoniecz:
> I would squash this into one commit before we merge to make this a clean PR.

oh definitely

> No hurt feelings if you want to do something different / your own.

though succeeding with this one would be nice

scripts/lib/build.ts
```diff
@@ -28,15 +28,15 @@ let _currBuild: Promise<void> | null = null;
 const jsonschema = new Validator();
 const locationConflation = new LocationConflation();
 
-function validateData(options?: Options) {
+async function validateData(options?: Options) {
```
matkoniecz: why it gets turned into async?

scripts/lib/__tests__/schema-builder.test.js
tordans: The short answer: We need it because we rely on stuff that has to be build first. Some test don't use it and get away with it…

<details><summary>Longer LLM answer</summary>

> 
> Both `buildDev` and `buildDist` return a **Promise**. The `await`s are not a special rule of icon references; they are how you wait for that Promise (and, for T5–T8, how you assert it **rejects**).
> 
> **`buildDist` (smoke test + T1–T4)**  
> `buildDist` always returns a Promise. Dist output is finished only after that Promise settles (`processData` even `await`s directory cleanup). The existing `runs buildDist` test already used `await`. T1–T4 do the same so they read `presets.json` / `fields.json` after they exist.
> 
> **`buildDev` (T5–T9 vs `runs buildDev`)**  
> `buildDev` is `async` too. The older `runs buildDev` test never awaits it. That often still works because the interim files are written with **sync** `fs.writeFileSync` before `processData` returns, so the files are already on disk by the time the test checks them. It is still a small race, just one the old test got away with.
> 
> T5–T8 cannot do that. Failures (`{fields/…}`, cycles, missing presets) happen **inside** the Promise. A sync `expect(() => …).toThrow()` never sees them (`schemaBuilder.validate` used to throw immediately; that API is gone). You need:
> 
> ```js
> await expect(schemaBuilder.buildDev(…)).rejects.toThrow(/…/)
> ```
> 
> T9 awaits because it then reads `interim/icons.json`. Without `await`, that read can run before dereference finishes.
> 
> So: **await when the assertion depends on the build finishing or failing**. The smoke `buildDev` test does not, because it only checks that a file exists and the sync writes usually already happened.

</details>

scripts/lib/references.ts
tordans: Reworked to `if (preset.icon && isReference(preset.icon)) {`

scripts/lib/references.ts
tordans: We have to handle them somehow. Resolving them is better than adding error handling here. So even if we don't encourage it or even use it, we should handle it. And given that, why not do it in a way that resolves them…

scripts/lib/build.ts
tordans: See [above](https://github.com/openstreetmap/id-tagging-schema/pull/2538#discussion_r3989991785); closing here.

scripts/lib/__tests__/schema-builder.test.js
tordans: Same as https://github.com/openstreetmap/id-tagging-schema/pull/2538#discussion_r3989991785 - it is more right this way.

tordans:
@matkoniecz FYI I did a cleanup run here and then resolved some of the now outdated notes because it was too hard to match them to the current code.

scripts/lib/references.ts
matkoniecz: > We have to handle them somehow.

I think that not handling them at all is also fine, up to and including infinite loops in build process
I have not verified it but AFAIK you can get them already in few places

seems unironically better than overly complex code

schemas/field.json
```diff
@@ -443,7 +443,7 @@
             }
         },
         "iconsCrossReference": {
-            "description": "A field can reference icons of another by using that field's identifier contained in brackets, like {field}.",
+            "description": "Reference the icons object of another field by that field's id in brackets, e.g. `{other_field}`. Do not use a `presets/` prefix.",
```
matkoniecz, line 446: old description here seems better for me

schemas/preset.json
```diff
@@ -71,7 +71,7 @@
             }
         },
         "icon": {
-            "description": "Name of preset icon which represents this preset",
+            "description": "Name of the icon that represents this preset, or a cross-reference `{presets/<preset-id>}`.",
```
matkoniecz, line 74:
```suggestion
            "description": "Name of the icon that represents this preset, or a reference `{presets/<preset-id>}`.",
```

or is a cross-reference and reference something else?

scripts/lib/__tests__/schema-builder.test.js
```diff
@@ -205,16 +213,259 @@ describe('schema-builder', () => {
...
+          tags: { f: 'b' },
+          icon: 'iD-bus',
+        }),
+        'data/fields/t_fld_combo.json': {
+          key: 't_fld',
+          type: 'combo',
+          label: 'T fld',
+          universal: true,
+          strings: { options: { yes: 'Yes' } },
+          icons: { yes: '{presets/t_fld_base}' },
```
matkoniecz, line 298:
```suggestion
        'data/presets/example.json': minimalPreset({
          tags: { f: 'b' },
          icon: 'iD-bus',
        }),
        'data/fields/t_fld_combo.json': {
          key: 't_fld',
          type: 'combo',
          label: 'T fld',
          universal: true,
          strings: { options: { yes: 'Yes' } },
          icons: { yes: '{presets/example}' },
```

preset name field (or fld) is fairly confusing

schemas/field.json
matkoniecz: not entirely sure why, at least in part I keep reading id as iD

PR #2541, opened 2026-07-15
Build: add dist manifest with build provenance

tordans:
This adds a new file to the build:

- Consumers can read manifest.json to see schema version, package version, commit, and ref for a dist build
- Avoids inferring v6 vs v7 or stale PR builds from preset JSON shape alone

In https://github.com/osmberlin/tagging-schema-browser I have to do some sniffing to see which version a deploy branch is using. Something like this would make things a bit clearer.

Are there other apps that would find this usefull?

### DIFF
--- a/dist/manifest.json
+++ b/dist/manifest.json
@@ -1,0 +1,8 @@
+{
+    "schemaVersion": "7",
+    "packageVersion": "7.1.0-dev",
+    "builtAt": "2026-07-15T07:06:12.060Z",
+    "repository": "https://github.com/openstreetmap/id-tagging-schema",
+    "commit": "8a5e65b8278a46bd273d442826e7c618c9542319",
+    "ref": "add-build-version-to-dist"
+}
--- a/scripts/lib/manifest.ts
+++ b/scripts/lib/manifest.ts
@@ -1,0 +1,40 @@
+import shell from 'shelljs';
+import packageInfo from '../../package.json' with { type: 'json' };
+
+export type Manifest = {
+  schemaVersion: string;
+  packageVersion: string;
+  builtAt: string;
+  repository: string;
+  commit: string;
+  ref: string;
+};
+
+export function generateManifest(): Manifest {
+  // CI builds (PR previews, staging) set GITHUB_SHA and GITHUB_REF automatically.
+  // npm releases ship committed dist/ from local `npm run dist`, so fall back to git.
+  const commit = process.env.GITHUB_SHA ?? execGit('git rev-parse HEAD');
+  const ref = process.env.GITHUB_REF ?? execGit('git rev-parse --abbrev-ref HEAD');
+
+  return {
+    schemaVersion: String(packageInfo.version).split('.')[0],
+    packageVersion: packageInfo.version,
+    builtAt: new Date().toISOString(),
+    repository: npmRepositoryUrlToHttpsUrl(packageInfo.repository.url),
+    commit,
+    ref,
+  };
+}
+
+/** Convert an npm package.json repository URL to a normal HTTPS web URL. */
+function npmRepositoryUrlToHttpsUrl(repositoryUrl: string): string {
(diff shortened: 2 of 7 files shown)


dist/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "schemaVersion": "7",
+    "packageVersion": "7.1.0-dev",
```
matkoniecz, line 3: dist should presumably match last release, right?

Can you generate using state from last released version?

tordans: Maybe we should remove those files in this PR and then let the regular processes generate them… – Mainly the Github Actions first and later the first release run by Martin.

dist/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "schemaVersion": "7",
```
k-yle: personally, i would not include this, because it suggests that the `schemaVersion` is independent to the major `packageVersion`

tsconfig.json
k-yle, line 1: should we also change `package.json`'s [main field](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#main) to point to `dist/manifest.json`? currently it points to a non-existant file

and/or should this file be called `index.json` if it's conceptually the entrypoint to the package?

dist/manifest.json
k-yle, line 2: personally, i would not include this, because it suggests that the `schemaVersion` is independent to the major `packageVersion`

scripts/lib/manifest.ts
```diff
@@ -0,0 +1,40 @@
...
+    builtAt: new Date().toISOString(),
+    repository: npmRepositoryUrlToHttpsUrl(packageInfo.repository.url),
+    commit,
+    ref,
+  };
+}
+
+/** Convert an npm package.json repository URL to a normal HTTPS web URL. */
+function npmRepositoryUrlToHttpsUrl(repositoryUrl: string): string {
+  return repositoryUrl.replace(/^git\+/, '').replace(/\.git$/, '');
```
k-yle, line 31: technically you can get this directly from npm thanks to [hosted-git-info](https://npm.im/hosted-git-info), for example `npm repo --browser=false --json`. but probably not worth the effort

scripts/lib/manifest.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+function execGit(command: string): string {
```
k-yle, line 34: sorry, there'll be merge conflicts here because the build scripts and deps have changed. e.g. shelljs is gone

k-yle (commented):
ideally we would have automated releases, then all this information would be automatically available from the npm API ([example](https://registry.npmjs.org/-/npm/v1/attestations/pdf-to-img@6.3.0)) thanks to [npm provenance](https://docs.npmjs.com/viewing-package-provenance)

but since automated releases [have been ruled out](https://github.com/openstreetmap/iD/issues/12757#issuecomment-5341573573), i think this is useful. also for PR preview branches which aren't on npm

tordans:
Thanks for the feedback. FYI I will wait for a clear signal that his has changes to get merged before doing the next rework. Let me know if you think you will want to merge this or if we want Martin to sign of first.

matkoniecz:
@k-yle can you take look at that?

k-yle:
imo this will be useful[^1], but let's wait for Martin's feedback first

[^1]: disclaimer: i'm [now a data consumer](https://github.com/k-yle/OpenSeaMap-vector/commit/bbb13c674b696a0ec41e71128b13e625fb4bed13) of this repo, so i guess i'm a bit biased

tyrasd (commented):
@tordans @k-yle can you maybe point out what the concrete benefits are from a data consumer point of view? At first glance, it seems to be a bit unnecessary to me: version (and thus schema-version?) and repository can be read from the `package.json` already, so that should be redundant. Commit / ref only seem to be necessary when you're doing something with git branches, in which case you should already know the commit and branch name already? Maybe I'm overlooking something…

PR #2544, opened 2026-07-15
add `reference` to presets where the main tag is ambiguous

k-yle:

Ideally, every preset with >1 tag in the `tags` object should use the `reference` attribute to explicitly indicate which tag is the 'main' tag to use for documentation. 

Otherwise, iD has to guess based on the order of the tags in the object, which is not really documented anywhere. 

There are genuine cases where we cannot pick a reference tag. For example, the preset for _"Utility Marker"_ matches on `marker=*` + `utility=*`. Therefore, this cannot be a compile-time check. Maybe it could be a warning in the tagging-schema- browser?

---

anyway, this PR adds `reference` to some manually reviewed cases, where the main tag is obvious.

### Info-`i`

<table>
<tr>
<td>Good</td>
<td>Bad</td>
</tr>
<tr>
<td>image
image
image
image
image
image
image
image
image
image
image
image</td>
<td>image
image
image
image
image
image
image
image
image
image
image
</td>
</tr>
</table>

### DIFF
--- a/data/presets/amenity/fast_food/cafeteria.json
+++ b/data/presets/amenity/fast_food/cafeteria.json
@@ -14,6 +14,10 @@
         "amenity": "fast_food",
         "fast_food": "cafeteria"
     },
+    "reference": {
+        "key": "fast_food",
+        "value": "cafeteria"
+    },
     "fields": [
         "{amenity/fast_food}",
         "access_simple"
--- a/data/presets/amenity/parking_space/disabled.json
+++ b/data/presets/amenity/parking_space/disabled.json
@@ -12,6 +12,10 @@
         "amenity": "parking_space",
         "parking_space": "disabled"
     },
+    "reference": {
+        "key": "parking_space",
+        "value": "disabled"
+    },
     "terms": [
         "disability",
         "disabled parking stall",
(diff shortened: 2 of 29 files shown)


matkoniecz:
> Otherwise, iD has to guess based on the order of the tags in the object, which is not really documented anywhere.

and these are technically not ordered, so it works only by accident...

matkoniecz:
is there some reason for keeping it in draft state? is it not finished/checked or otherwise unready in some way?

matkoniecz:
(adding labels is useful as it affects changelog - if it is unclear which labels should be applied, please file issue so their descriptions can be improved)

k-yle:
> is there some reason for keeping it in draft state? is it not finished/checked or otherwise unready in some way?

i was going to manually check the ℹ️ button for each preset, but i ran out of time. will come back to this

k-yle:
test-documentation updated. 

off topic, but i wonder if we should have a script which detects presets that have no wikibase item (or maybe this could be part of the [tagging-schema-browser](https://osmberlin.github.io/tagging-schema-browser)?

matkoniecz:
Personally I am considering wikibase as dysfunctional due to lack of ability to watchlist or otherwise monitor entries, so I would prefer to not be forced to garden it.

matkoniecz (approved):

PR #2545, opened 2026-07-15
convert the remaining build scripts to typescript

k-yle:
this was surprisingly easy, only a small number of runtime changes.

confirmed that the `īnterim` and `dist` folders are unchanged after running `npm run dist translations`


tordans (commented):
How should we go about the many `as` ind the code? Look into cleaning them now or later or never?

And is the re-export really needed?

k-yle:
thanks, i removed the reëxport. 

for the remaining typecasts, i have no good solution:
- most are due to fundamental limitations like `for...in` or `Object.keys()` always returning `string[]`
- the types from transifex's SDK is also quite limited and somewhat wrong
- the rest could be fixed, but the diff would be huge, and so far this PR has managed to convert everything with close to 0 runtime changes

matkoniecz:
> but the diff would be huge, and so far this PR has managed to convert everything with close to 0 runtime changes

diff as in code changed by this PR or would it introduce changes in output?

k-yle:
> diff as in code changed by this PR

yeah, this. either way it's not a big deal, i've just been trying to implement these typescript PRs without changing any actual javascript runtime logic, if possible

matkoniecz:
I feel a bit bad about merging this without getting it as good review as it deserves (not TS expert) but can we get one? I guess I can try to summon @tyrasd who definitely knows more...

scripts/lib/build.ts
```diff
@@ -776,11 +771,11 @@ function generateTaginfo(presets, fields, deprecated, discarded, tstrings, proje
 
   taginfo.tags.forEach(elem => {
     if (elem.description) {
-      elem.description = elem.description.join(', ');
+      elem.description = elem.description.join(', ') as never;
```
tyrasd: 🤔 wouldn't it be clearer / "more honest" if we work around it like the following?

```suggestion
      // @ts-expect-error -- assign string to string[]
      elem.description = elem.description.join(', ');
```

scripts/lib/types.def.ts
```diff
@@ -1,7 +1,15 @@
...
+import type * as Taginfo from 'taginfo-projects';
+import type { Field, Preset, PresetCategory } from '../../dist';
 
 export type * from '../../dist';
 
+/**
+ * this library doesn't follow the type-definitions strictly, it uses `string[]` instead
+ * of `string` for the description field.
+ */
+export type TaginfoTag = Omit<Taginfo.Tag, 'description'> & { description?: string[] };
```
tyrasd: wouldn't either of the following two approaches work too?

```suggestion
export type TaginfoTag = Omit<Taginfo.Tag, 'description'> & { description?: string[] | string };
```

```suggestion
export type TaginfoTag = Taginfo.Tag & { _descriptions?: string[] };
```

k-yle: good idea, done

matkoniecz:
@k-yle 

:(

k-yle:
conflicts fixed again

matkoniecz (approved):

PR #2546, opened 2026-07-15
Fix build script on Windows assuming path separator for Regex

RudyTheDev:
So the issue is in `build.js`: 
* `fs.globSync` returns OS-preferred separator (`/` or `\`)
* `.match` Regex pattern assumes `/` path separator

The result is that stuff like `npm test` or `npm run build` throws an exception on Windows because Regex match fails. (Specifically, trying to get the `[1]` capture group throws on a null Regex result.)

<details>

<summary>Previous version</summary>

<s>The fix just always replaces `\` with `/` for the glob's resulting file path used in Regex in `generateFields()` and `generatePresets()`.

An alternative would be to use proper file/path functions to strip the top folders and file extension instead of Regex. I don't think it makes any functional difference, but arguably adds more complexity and multiple function calls. And this would just end up having `\` in ids on Windows, which doesn't make anything cleaner in the end.

Technically, `generateCategories()` also does a glob and will end up with `\` slashes, but it doesn't parse/Regex them, so it's not affected on Windows, so I didn't alter that one.

I also thought about adding a current `os.platform()` check, but no other script uses `os`, so I didn't want to import it (although I see some packages use it). And since this is a build-only script, there's probably no need to complicate it.</s>

</details>

The fix passes the path and leading folder used in `generateFields()` , `generatePresets()` and `generateCategories()` to a new "standardized" `extractIdFromPath()` function, which then generates the id using proper path functions to strip the leading folder, strip extensions, strip underscores and normalize to posix path separator. So Regex with hard-coded path separator is avoided. And all three functions now use a common function.

So on *nix this would work like this:
Preset `data/presets/addr/_interpolation.json` to id `addr/interpolation`

While on Windows it would work like this:
Preset `data\presets\addr\_interpolation.json` to id `addr/interpolation`

I checked the output from before and after and it produces the exact same results for all 2491 category, preset and field ids.

### DIFF
--- a/scripts/lib/build.js
+++ b/scripts/lib/build.js
@@ -240,13 +240,13 @@ function validateSchema(file, instance, schema) {
 function generateCategories(dataDir, tstrings) {
   let categories = {};
 
-  fs.globSync(dataDir + '/preset_categories/*.json', {
-    posix: true,
-  }).forEach(file => {
+  const categoriesDir = path.posix.join(dataDir, 'preset_categories');
+  fs.globSync(categoriesDir + '/*.json').forEach(file => {
     let category = read(file);
+    const id = 'category-' + extractIdFromPath(categoriesDir, file);
+
     validateSchema(file, category, categorySchema);
 
-    let id = 'category-' + path.basename(file, '.json');
     tstrings.categories[id] = { name: category.name };
     delete category.name;
 
@@ -260,11 +260,10 @@ function generateCategories(dataDir, tstrings) {
 function generateFields(dataDir, tstrings, searchableFieldIDs, references) {
   let fields = {};
 
-  fs.globSync(dataDir + '/fields/**/*.json', {
-    posix: true,
-  }).forEach(file => {
+  const fieldsDir = path.posix.join(dataDir, 'fields');
+  fs.globSync(fieldsDir + '/**/*.json').forEach(file => {
     let field = read(file);
-    let id = stripLeadingUnderscores(file.match(/fields\/([^.]*)\.json/)[1]);
+    const id = extractIdFromPath(fieldsDir, file);
 
     validateSchema(file, field, fieldSchema);
 
@@ -335,10 +334,21 @@ function generateFields(dataDir, tstrings, searchableFieldIDs, references) {
 }
 
 
-function stripLeadingUnderscores(str) {
-  return str.split('/')
-    .map(s => s.replace(/^_/,''))
-    .join('/');
+/**
+ * @example `data/presets/addr/_interpolation.json` -> `addr/interpolation`
+ */
+function extractIdFromPath(parentDir, file) {
+  // Strip the leading directory
+  const relativePath = path.relative(parentDir, file);
+  const dirName = path.posix.dirname(relativePath);
+  // Strip the extension, i.e. `.json`
+  const extension = path.extname(relativePath);
+  const fileName = path.posix.basename(relativePath, extension);
+  // Build the id from only the subdirectories and base file name
+  const id = path.posix.join(dirName, fileName);
+  // Remove leading underscores from any directory or file name
+  // Also enforce posix path separator '/' in case the path was with Windows '\' path separator
+  return path.posix.join(...id.split(path.sep).map(s => s.replace(/^_/, '')));
 }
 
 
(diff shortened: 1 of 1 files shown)


scripts/lib/build.js
```diff
@@ -266,6 +264,7 @@ function generateFields(dataDir, tstrings, searchableFieldIDs, references) {
     posix: true,
   }).forEach(file => {
     let field = read(file);
+    file = file.replace(/\\/g, '/'); // potentially replace Windows path separator '\' with *nix '/' since we want to Regex it
```
matkoniecz:
```suggestion
    file = file.replace(/\\/g, '/'); // useful on Windows - replaces '\' path separator with standard *nix '/' since we want to Regex it
```

scripts/lib/build.js
```diff
@@ -363,6 +352,7 @@ function generatePresets(dataDir, tstrings, searchableFieldIDs, listReusedIcons,
     posix: true,
   }).forEach(file => {
     let preset = read(file);
+    file = file.replace(/\\/g, '/'); // potentially replace Windows path separator '\' with *nix '/' since we want to Regex it
```
matkoniecz:
```suggestion
    file = file.replace(/\\/g, '/'); // useful on Windows - replaces '\' path separator with standard *nix '/' since we want to Regex it
```

scripts/lib/build.js
matkoniecz: at least for me something like this is clearer...

not demanding such change, but lets see what others think

RudyTheDev: Yeah, it looks fine. I didn't spend too long thinking about the comment wording since nothing else is commented there anyway.

Only point is that it's not "useful", but required. Useful implies you don't have to do it.

matkoniecz:
Lets wait for @k-yle - maybe they know smarter way to do this.

But otherwise looks like a good idea. Maybe we should document somewhere that `/` and `\` are not supposed to be in text part of identifiers? (if not documented already)

matkoniecz:
@RudyTheDev and thanks for testing on exotic OS :)

RudyTheDev:
> Maybe we should document somewhere that `/` and `\` are not supposed to be in text part of identifiers?

These values are from file names, so neither `/` or `\` (or any other disallowed/special path characters) are possible without completely breaking the repo checkout (at least on Windows). The `/` in the final id only comes from the json file being in a subfolder like:

File `data/fields/water.json` to id `water`
File `data/fields/tower/platforms.json` to id `tower/platforms`
File `data/fields/railway/signal/position.json` to id `railway/signal/position`
etc.

matkoniecz:
Well, `\` is entirely allowed on ext4 filesystems. The same as newlines, `*` or `?` etc. 

screen-2026-07-15-16-58-53


actual filename

RudyTheDev:
Yes, you can technically do it, but once you do that, you cannot check out the repo on Windows at all. Hence, as I said, "[not] possible without completely breaking the repo checkout (at least on Windows)".

matkoniecz:
And we definitely want to prevent also that :)

scripts/lib/build.js
```diff
@@ -266,6 +264,7 @@ function generateFields(dataDir, tstrings, searchableFieldIDs, references) {
     posix: true,
   }).forEach(file => {
     let field = read(file);
+    file = file.replace(/\\/g, '/'); // required on Windows - replaces '\' path separator with standard *nix '/' since we want to Regex it
```
k-yle: this works fine, but it might be more idiomatic to use node's builtin API. something like this:

```suggestion
    file = path.posix.join(...file.split(path.sep));
```

scripts/lib/build.js
```diff
@@ -266,6 +264,7 @@ function generateFields(dataDir, tstrings, searchableFieldIDs, references) {
     posix: true,
```
k-yle: we should remove this `{ posix: true }` option everywhere, since it's misleading. this option is invalid and does nothing. that's how this bug got introduced in the first place. 
```suggestion
```

k-yle (commented):
the 2 suggestions inline are untested, i don't have access to a windows laptop this week sorry. so feel free to ignore these comments

RudyTheDev:
@k-yle I went ahead and rewrote the approach. Please see the new PR description.

I feel like if this is going to use path functions to change the path separator, then it may as well do the whole id generation properly with OS-independent path functions from start to finish and avoid the Regex completely while having a standard approach for all three functions. So this also makes that `posix: true` unnecessary (not that it apparently worked).

I don't know if anything I did needs TS changes for #2545.

scripts/lib/build.js
RudyTheDev: Oooh, so that's what that does, I didn't realize. I guess it's not functional anymore for some reason? Anyway, I removed these in the new proposed version.

k-yle (approved):
thanks for the PR. i don't mind what approach we use (this or the original one)

tested on windows and confirmed that it worked

PR #2558, opened 2026-07-18
Use "Yin and Yang" icon for traditional chinese medicine facilities

fibonacci-matrix:

Fa7SolidSpa
yin_yang

Alternative icon (fas):
Fa7SolidYinYang

### Related issues

None

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Tag:healthcare%3Dalternative

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/tags/healthcare%3Aspeciality=traditional_chinese_medicine

### DIFF
--- a/data/presets/healthcare/alternative/traditional_chinese_medicine.json
+++ b/data/presets/healthcare/alternative/traditional_chinese_medicine.json
@@ -1,5 +1,5 @@
 {
-    "icon": "fas-spa",
+    "icon": "pinhead-yin_yang",
     "geometry": [
         "point",
         "area"


fibonacci-matrix:

Kovoschiz:
This alone only suggests tai chi. Yin yang isn't unique to TCM. The most signature TCM symbols are gourd, and apricot forest, from classic literature. 
* https://www.healthbureau.gov.hk/CMDevBlueprint/en/
* https://www.cmhhk.org.hk/en/about_us/cmculturedisplays/classicalanecdotesoftraditionalchinesemedicine/

Ginkgo is a herb encompassing yin yang https://classicalchinesemedicine.org/ginkgo-cultural-background-medicinal-usage/

For ginseng, TCM is usually white ones, dried ones. Red steamed ones is termed "Korean".

fibonacci-matrix:
@Kovoschiz could you share a link to an image that you think would be a better fit? Any image from the internet is fine. I'll then try to find a matching icon in the supported icon sets myself

matkoniecz:
interesting

I have no idea how much of it would be accurate and matching symbol and how much it would be misleading and not fitting.

From cursory look at https://en.wikipedia.org/wiki/Traditional_Chinese_medicine it seems like potentially a good match? But not sure enough to merge it...

Kovoschiz:
@fibonacci-matrix It's somewhat difficult, as TCM shops are traditionally signposted text-based. Can't think of any map or software that has TCM icons either. 
Canva PRC has a content farm quality looking article with a few company logos, and I can't say they are immediately obvious https://www.canva.cn/learn/traditional-chinese-medical-logo/
PRC's TCM academy does use a gourd https://zh.wikipedia.org/wiki/File:China_Academy_of_Chinese_Medical_Sciences_logo.jpg
Do the other `healthcare=alternative` have any examples I can refer to?

matkoniecz:
TCM will be hard to find a single icon for as it a group of practices, not a single one

yin-yang seems important in many (most?) of them but not sure is it covering everything and is this specific icon applicable to all variants, or at least vast majority of them

Kovoschiz:
Yin yang is a very broad concept, eg imagine if we use "scientific method" symbols to represent medical science. My personal judgement is on how yin yang doesn't immediately convey TCM, while it may be included conceptually. 
Five Elements, or qi energy seem equally important. Unfortunately those don't have a well-known icon either. 
In Wiki, I would rather start with the topic template's aura-surrounded human figure https://commons.wikimedia.org/wiki/File:Outline-body-aura.svg
Or this more `=acupuncture` logo https://commons.wikimedia.org/wiki/File:Alternative_Medicine_-_The_Noun_Project.svg
Yin yang could be inserted or replacing the generic symbol in them respectively though

matkoniecz:
acupuncture is confusing, like in general using specific preset icon for more general one would be 

I think that current one may be close to optimal :/

Kovoschiz:
I prefer including the gourd, for the traditional and formal roots. But to be fair, I don't know whether non-Asians understand it. 
I asked in my community. Let's wait a while.

fibonacci-matrix:
It turns out the "Yin and Yang" icon is already in use. However, I didn't find it immediately because its name "temaki-taoism" isn't entirely accurate:

Vectorial1024:
With all due respect to the idea of "traditional Chinese medicine", that system can largely be categorized as "herbal medicine" / herbology, so we should start with whether we have an icon for these herbal medicine shops, and then decide what to do specifically with TCM.

I largely echo Kovoschiz's points of "tai chi is not specific to TCM".

matkoniecz:
currently we have some vaguely herbal icon

I will close this one, I think

matkoniecz:
though big thanks for @fibonacci-matrix and your icon hunting had decent hit rate! If you would find more cases in ocean of Pinhead (and other) icons further PRs are highly welcome!

Vectorial1024:
To slightly add to this (for future readers), in fact, the concept of "traditional Chinese medicine" is actually ambiguous between e.g.:

- natural herbology (I think this is what this PR is trying to say)
- "powdered" herbology (example company https://zh.wikipedia.org/wiki/%E5%9F%B9%E5%8A%9B%E8%BE%B2%E6%9C%AC%E6%96%B9)
- various forms of acupuncture
- other "ancillary" shops

Ignoring the herbology icons, this PR is risky anyway. Even the underlying OSM tagging schema may be inaccurate.

PR #2559, opened 2026-07-18
Selected a more appropriate icon for adoption agencies

fibonacci-matrix:

MakiSuitcase
Fa7SolidHandsHoldingChild

### Related issues

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Tag:office%3Dadoption_agency

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/tags/office=adoption_agency

### DIFF
--- a/data/presets/office/adoption_agency.json
+++ b/data/presets/office/adoption_agency.json
@@ -1,5 +1,5 @@
 {
-    "icon": "maki-suitcase",
+    "icon": "fas-hands-holding-child",
     "geometry": [
         "point",
         "area"


fibonacci-matrix:

matkoniecz (approved):
looks fitting, hard to find something less abstract here, thanks for looking at office icons where it would be nice to have less of the same suitcases for each value

westnordost:
The small version of the icon looks a bit like a lotus flower. (But it is anyway better than the suitcase)

PR #2571, opened 2026-07-21
Add line as valid type for speed hump

catgirlseraid:
As per https://wiki.openstreetmap.org/wiki/Tag:traffic_calming%3Dhump speed hump on a line is valid mapping

### DIFF
--- a/data/presets/traffic_calming/hump.json
+++ b/data/presets/traffic_calming/hump.json
@@ -5,7 +5,8 @@
         "direction_vertex"
     ],
     "geometry": [
-        "vertex"
+        "vertex",
+        "line"
     ],
     "terms": [
         "bump",


matkoniecz:
How it looked before and after? I would expect that before it also would be working fine, but maybe there were some bad effects?

(my guess is that mapping section of road with `traffic_calming=` is not improved by this PR but maybe I guess badly)

Kovoschiz:
While this drawing of `traffic_calming=` itself is fine, it's not completely clear whether wiki could have meant the roadway `highway=` + `traffic_calming=hump` line as in other `traffic_calming=` , which might be used for a road section with a series of humps. Applications usually expect `traffic_calming=` vertex, then you would have to topologically double-feature it on the intersecting vertex, cf `barrier=` viz `=bollard` situation.  
`=bump` is not implemented on lines yet either. In general, fundamentally these are not "strongly typed", being context/combination-dependent.

PR #2575, opened 2026-07-21
fix field inheritance bug

### LINKED ISSUE #2535: shop=cheese should not have shop= field
it is inherited from generic shop preset where it makes sense

it makes sense to inherit other fields

not sure yet how to fix it...

define shared shop fields template and inherit from it, rather than from generic shop= preset?

define `fields` and do not inherit any?

https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/shop/cheese.json https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/shop.json

likely other presets are also affected

Image

k-yle:
Closes #2535

this [block of code](https://github.com/openstreetmap/iD/blame/d0b561bee187907e1b04864fec136c5dbc869290/modules/presets/preset.js#L360-L375) was lost when the referencing logic was migrated into this repo.

**The diff from this PR is 60678cc0447664c9038ff8023f7a210abbbfe5a2**. This diff has two changes:
1. `moreFields` currently repeats fields like `phone`, even if they're already in `fields`
2. for presets like `office=yes`, `office=*` is no longer a field.

sample:

image

### DIFF
--- a/scripts/lib/references.ts
+++ b/scripts/lib/references.ts
@@ -37,8 +37,42 @@ export function dereferenceUntranslatedContent(presets: AllPresets, fields: AllF
             continue;
           }
 
+          // Skip `fields` for the keys which define the preset.
+          // These are usually `typeCombo` fields like `shop=*`
+          function shouldInherit(fieldId: string) {
+            // shouldInherit is called recursively as references are expanded.
+            // if this field is reference, skip it for now. It will be
+            // processed in the next loop iteration.
+            if (isReference(fieldId)) return true;
+
+            const field = fields[fieldId];
+            if (!field.key) return true;
+
+            if (
+              preset.tags[field.key] &&
+              // inherit anyway if multiple values are allowed or just a checkbox
+              field.type !== 'multiCombo' &&
+              field.type !== 'semiCombo' &&
+              field.type !== 'manyCombo' &&
+              field.type !== 'check'
+            ) {
+              return false;
+            }
+
+            if (
+              preset.fields?.some(originalField => field.key === fields[originalField]?.key) ||
+              preset.moreFields?.some(originalField => field.key === fields[originalField]?.key)
+            ) {
+              // current preset already has a field for this field
+              return false;
+            }
+
+            return true;
+          }
+
           // replace the reference with every field. decrement i to reprocess this array index.
-          preset[prop].splice(i--, 1, ...referencedPreset[prop]);
+          // this is necessary as it can also be a reference
+          preset[prop].splice(i--, 1, ...referencedPreset[prop].filter(shouldInherit));
         }
       }
     }


k-yle:
tested with the PR preview:

image

matkoniecz (commented):
wait, do we have no unit tests for this script? maybe we should have some...

k-yle:
build scripts typically don't have unit tests, but we could add some if that's desired?

matkoniecz:
not entirely sure why build tests have no unit tests, but I think it may make sense in case of ours where we basically transfrom data structures and they are human and machine readable both as input and as output, with quite stable output (unlike compiling to executable binary file)

scripts/lib/references.ts
```diff
@@ -35,8 +35,40 @@ export function dereferenceUntranslatedContent(presets: AllPresets, fields: AllF
...
+              // current preset already has a field for this field
+              return false;
+            }
+
+            return true;
+          }
+
           // replace the reference with every field. decrement i to reprocess this array index.
-          preset[prop].splice(i--, 1, ...referencedPreset[prop]);
+          preset[prop].splice(i--, 1, ...referencedPreset[prop].filter(shouldInherit));
```
matkoniecz, line 75:
```suggestion
          // Skip `fields` for the keys which define the preset.
          // These are usually `typeCombo` fields like `shop=*`
          function shouldInherit(fieldId: string) {
            // references can be recursive. return true, and reprocess
            // it in the next loop iteration.
            if (isReference(fieldId)) return true;

            const field = fields[fieldId];
            if (!field.key) return true;

            if (
              preset.tags[field.key] &&
              // inherit anyway if multiple values are allowed or just a checkbox
              field.type !== 'multiCombo' &&
              field.type !== 'semiCombo' &&
              field.type !== 'manyCombo' &&
              field.type !== 'check'
            ) {
              return false;
            }

            if (
              preset.fields?.some(originalField => field.key === fields[originalField]?.key) ||
              preset.moreFields?.some(originalField => field.key === fields[originalField]?.key)
            ) {
              // current preset already has a field for this field
              return false;
            }

            return true;
          }

// replace the reference with every field. decrement i to reprocess this array index.
// this is necessary as it can also be a reference
          preset[prop].splice(i--, 1, ...referencedPreset[prop].filter(shouldInherit));
```

matkoniecz: maybe it is just me but I was confused by first comment in shouldInherit and I was quite confused looking for where "next loop iteration" is

Or maybe it should not call it "next loop iteration" but rather "and this index will be processed again"? I was looking for some additional loop (and I admit that to figure out how it works I needed to ask LLM which was right...)

k-yle: i applied your suggestion and reworded the first comment

matkoniecz (approved):

matkoniecz:
adding testing I think would make sense, but no reason to block merge over this

and I may be wrong about testing being useful

matkoniecz:
And @k-yle - thanks for fixing this release-blocker!

PR #2576, opened 2026-07-22
add relation role definitions for type=boundary and type=multipolygon

k-yle:
cherrypicked from #1538, that PR is too big to review. 

ideally, this PR could be merged sooner, then we can create PRs in iD to actually use this data.

### DIFF
--- a/data/presets/area.json
+++ b/data/presets/area.json
@@ -12,5 +12,30 @@
         "polygon"
     ],
     "name": "Area",
-    "matchScore": 0.1
+    "matchScore": 0.1,
+    "relation": {
+        "reference": "multipolygon",
+        "allowDuplicateMembers": false,
+        "role_labels": {
+            "inner": "Inner",
+            "outer": "Outer"
+        },
+        "members": [
+            {
+                "role": "inner",
+                "geometry": [
+                    "line",
+                    "area"
+                ]
+            },
+            {
+                "role": "outer",
+                "geometry": [
+                    "line",
+                    "area"
+                ],
+                "min": 1
+            }
+        ]
+    }
 }
--- a/data/presets/type/boundary/administrative.json
+++ b/data/presets/type/boundary/administrative.json
@@ -18,5 +18,58 @@
         "key": "boundary",
         "value": "administrative"
     },
-    "name": "Administrative Boundary"
+    "name": "Administrative Boundary",
+    "relation": {
+        "reference": "boundary",
+        "allowDuplicateMembers": false,
+        "role_labels": {
+            "inner": "Inner",
+            "outer": "Outer",
+            "label": "Label",
+            "admin_centre": "Admin Center"
+        },
+        "members": [
+            {
+                "role": "inner",
+                "geometry": [
+                    "line",
+                    "area"
+                ]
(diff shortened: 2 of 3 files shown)


matkoniecz:
> I also created relation-roles - any thoughts on that ?

not sure whether we need so many labels... I would be tempted to use just "presets", but not opposed to it either as long as people are not complaining when "presets" was used

matkoniecz (approved):

matkoniecz:
@tordans @tyrasd Are you interested in taking look at this one? (and maybe also #1538)

k-yle:
okay, i deleted the label. to make searching possible, i'll include the phrase `relation role` in every PR

.github/release-drafter.yml
```diff
@@ -16,6 +16,7 @@ categories:
       - 'new-icon'
       - 'new-label'
       - 'category-related'
+      - 'relation-roles'
```
matkoniecz:
```suggestion
```

iff we are going without label

data/presets/type/boundary.json
```diff
@@ -10,5 +10,58 @@
...
+                ],
+                "matchTags": [
+                    {
+                        "place": "*"
+                    }
+                ],
+                "max": 1
+            },
+            {
+                "role": "admin_centre",
```
matkoniecz: "[boundary](https://wiki.openstreetmap.org/wiki/Key:boundary)=* [relations](https://wiki.openstreetmap.org/wiki/Relations) are used to define political, administrative, nature protection and similar boundaries." https://wiki.openstreetmap.org/wiki/Relation:boundary

should nature protection ones got admin_centre role also listed?

matkoniecz: maybe put definition in administrative boundary instead and at most inherit it here, pending potential future changes and improvements?

matkoniecz: or define only boundary=administrative ?

k-yle: good point, changed to `boundary=administrative`

matkoniecz (approved):
for the record: I am really confused why `label` is not https://wiki.openstreetmap.org/wiki/Tagging_for_the_renderer but it is widely accepted, and I failed to convince people...

As far as I see things are fine here, I guess there is risk of something wrong that will be revealed once iD tries to use this code (you cannot live-test it), but in such case we can just fix it

matkoniecz:
@k-yle thanks for your persistence in dealing over multiple PRs, piles of review and not abandoning it despite being stuck

1ec5:
Ah, I missed this PR and went for https://github.com/openstreetmap/id-tagging-schema/pull/1538#pullrequestreview-4791982482 instead. I agree that “Label” is a confusing label for `label` (heh). We definitely don’t want people mapping for the renderer. This is a legitimate geographic concept that suffers from a badly named keyword.

matkoniecz:
maybe make an issue then for this?

I also admit that I do not really get difference presented in https://github.com/openstreetmap/id-tagging-schema/pull/1538#discussion_r3661318002 and I always thought that label is entirely about manual label placement for renderers via osm editing

PR #2577, opened 2026-07-22
re-use translations for `strings.options.*`

### LINKED ISSUE #1152: Duplicated entries in Transifex for `traffic_calming`

k-yle:
closes #1152. part two of https://github.com/openstreetmap/schema-builder/issues/162 - we implemented the schema-builder side ages ago. now we can use the new feature.

this PR removes 291 duplicate translations, replacing them with references. 

---

While working on this PR, i discovered quite a few inconsistencies, like `Dance School` vs `Dancing School`. These can be dealt with in a follow-up PR, so that it doesn't create extra work for translators.


Dimitar5555:
Resolves #1152

matkoniecz:
hmm, testing all of it manually would be quite annoying

I hope that it can be test by building before and after and confirming output has not changed, then editing something and confirming that build output has changed

matkoniecz:
a bit obnoxiously dist file CHANGES as it reorders values in output

is it maybe possible to avoid this?

k-yle:
updated to preserve the object order in `dist`, so this is easier to review. There is no diff for the base language (`en.json`)

The diff is interesting for other languages, where translators have been inconsistent. example from `de`:

image


full diff: c50bbce5f819264aa0672f6721566dc733f42441

matkoniecz (approved):

matkoniecz:
> These can be dealt with in a follow-up PR, so that it doesn't create extra work for translators.

PR would be welcome for ones where divergence is not needed!

PR #2581, opened 2026-07-22
assign permissions specifically in deploy-preview.yml

matkoniecz:
this would bring us step closer to disabling global "read and write permission" that is currently enabled for actions and limit potential damage

would also make action testing on separate repository less painful, by moving proprietary github config from opaque untracked settings into proprietary github config tracked as part of repo

part of https://github.com/openstreetmap/id-tagging-schema/pull/1284 testing

was tested on https://github.com/matkoniecz/test_manual_rerun

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -5,7 +5,8 @@ on:
     workflows: ["Build Preview"]
     types:
       - completed
-
+permissions:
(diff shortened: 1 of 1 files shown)


matkoniecz:
disclaimer: github actions are annoying and I may have missed something

tyrasd (approved):

PR #2583, opened 2026-07-22
allow deploy preview to run (potentially) outside PR context

matkoniecz:
prerequisite to #1284 but additional changes are needed

tested in https://github.com/matkoniecz/test_manual_rerun/pull/13

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -10,7 +10,7 @@ jobs:
   deploy-preview:
     environment: pr-previews
     runs-on: ubuntu-latest
-    if: ${{github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success'}}
+    if: ${{github.event.workflow_run.conclusion == 'success'}}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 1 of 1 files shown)


matkoniecz:
https://github.com/openstreetmap/id-tagging-schema/commit/f955b26875d854e68e7acdeff13ab19849ac0125 has something that looks better

.github/workflows/deploy-preview.yml
k-yle, line 28: even if we make this change, this will still fail if the artifact has expired right? (90 days)

When i attempted this in https://github.com/openstreetmap/id-tagging-schema/commit/f955b26875d854e68e7acdeff13ab19849ac0125, i realised that the easiest option would be trigger the workflow when you post a PR comment like `/rebase`. But at that point, closing/reopening the PR an easier solution

matkoniecz: > But at that point, closing/reopening the PR an easier solution

main negative here is that contributors may be sad/confused why maintainer is closing their PR (and immediately reopening)

though I guess I can have prepared reply and post it as an explanation?

matkoniecz: > even if we make this change, this will still fail if the artifact has expired right? (90 days)

IIRC plan is to rerun entire preview preparation, which would trigger this job after files are built (but I am still trying to wrap my head around how it works)

PR #2586, opened 2026-07-23
make linter complaining about minor issue less annoying

matkoniecz:
by making possible to copy-paste output rather than fiddle with spaces

### DIFF
--- a/scripts/lib/build.js
+++ b/scripts/lib/build.js
@@ -1067,7 +1067,7 @@ function validateTerms(terms, where) {
   process.stdout.write('expected terms\n\n');
   const jsonExpected = JSON.stringify(expectedTerms, null, 4);
   const indentedJsonExpected = jsonExpected.split('\n').map(line => '    ' + line).join('\n');
-  process.stdout.write(indentedJsonExpected); // make possible to directly copy CI output into code 
+  process.stdout.write(indentedJsonExpected); // make possible to directly copy CI output into code
   process.stdout.write('\n\ndiffer from actual ones\n\n');
   process.stdout.write(`${terms}`);
   process.stdout.write('\n');
--- a/scripts/lib/build.js
+++ b/scripts/lib/build.js
@@ -1065,7 +1065,9 @@ function validateTerms(terms, where) {
   process.stderr.write(`Expected terms in ${where} to be lowercase and sorted alphabetically.`);
   process.stdout.write('\n');
   process.stdout.write('expected terms\n\n');
-  process.stdout.write(JSON.stringify(expectedTerms, null, 2));
+  const jsonExpected = JSON.stringify(expectedTerms, null, 4);
+  const indentedJsonExpected = jsonExpected.split('\n').map(line => '    ' + line).join('\n');
+  process.stdout.write(indentedJsonExpected); // make possible to directly copy CI output into code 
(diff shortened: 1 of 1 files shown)


matkoniecz:
triggered by https://github.com/openstreetmap/id-tagging-schema/pull/2224

screen-2026-07-23-13-31-05

k-yle (commented):
do you have [prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) installed in your IDE and [format-on-save](http://vscode://settings/editor.formatOnSave) enabled?

if you do that, then you never need to think about formatting, when you press <kbd>Ctrl</kbd> <kbd>S</kbd> all formatting issues are automatically fixed

matkoniecz:
I do this kind of things directly in browser (review PR, spot failing CI, go into CI report, copy output, propose edit via github interface like at https://github.com/openstreetmap/id-tagging-schema/pull/2224#pullrequestreview-4763630597 )

matkoniecz:
and AFAIK proposing edits cannot be done via git - unless I would do it by forking branch and opening PR in their repo

scripts/lib/build.js
```diff
@@ -1065,7 +1065,9 @@ function validateTerms(terms, where) {
   process.stderr.write(`Expected terms in ${where} to be lowercase and sorted alphabetically.`);
   process.stdout.write('\n');
   process.stdout.write('expected terms\n\n');
-  process.stdout.write(JSON.stringify(expectedTerms, null, 2));
+  const jsonExpected = JSON.stringify(expectedTerms, null, 4);
+  const indentedJsonExpected = jsonExpected.split('\n').map(line => '    ' + line).join('\n');
+  process.stdout.write(indentedJsonExpected); // make possible to directly copy CI output into code
```
matkoniecz:
```suggestion
  process.stdout.write(indentedJsonExpected); // make possible to directly copy CI output into code
```

k-yle (changes requested):
i still think sorting terms using this manual approach is a big waste of everyone's time, and this patch doesn't solve the root cause.

why don't we make this sorting thing a [custom lint rule](https://eslint.org/docs/latest/extend/custom-rules)? it would be slightly more lines of code, but not much more. 

main benefit: 
- the sorting issues would be automatically fixed by your IDE when you press <kbd>Ctrl</kbd> <kbd>S</kbd>
- every sorting issue in every file would be auto-fixed when you run `npm run lint:fix`.
- GitHub PR view would show you exactly which line caused the sorting issue, no need to trawl through CI logs

matkoniecz:
> why don't we make this sorting thing a [custom lint rule](https://eslint.org/docs/latest/extend/custom-rules)?

mostly because I forgot/was unaware those exist

from description seem preferable, I will try to implement with that approach instead

k-yle:
if you prefer, i can send a PR for both that and https://github.com/openstreetmap/id-tagging-schema/pull/2663#pullrequestreview-4874026788. but not tonight sorry

matkoniecz:
@k-yle that would be really really welcome so I do not need to dig into another tech stack (however small it would be)

definitely not an urgent case

matkoniecz:
(this above is to indicate offer to maker superior replacement, let me know if you would prefer me to implement it instead)

Issue #2208, opened 2026-04-22
Package: Bundle Icons as part of the npm packge (or subpackage)


tordans:
This was a side topic in https://github.com/ideditor/schema-builder/pull/278#issuecomment-4083703709 and we talked about it a bit in one of the community meetups.

To gives this some more room, here is a separate issue…

(I am not sure is this is for this repo or for something that needs to happen in the schema-builder repo. I think here, but we can move this of course.)

**The current situation:**
- Whenever we add an icon library upstream users need to add this new source
- Each upstream user has to build the icon fetching separately
- GoMap and SC for example do this already
- This makes adding an Icon library more like a breaking change of the schema (unless we accept fallback icons…)
- Whenever a new icon is added, data users have to orchestrate pulling the latest icon library package that fits the current usage in the tagging-schema — AFAIK this is the reason behind https://github.com/openstreetmap/id-tagging-schema/issues/2207#issuecomment-4298000819

**Possible different situation:**

We package the icons as part of the tagging-schema-release or as a related subpackage. Whenever we release the tagging schema, we also package all the used icons.
Data consumers can then use this one package as their source of truth and simplify their implementation to just one source.
Note that SC and GoMap already have their own processing of icons to format them in a way that fits their apps. This part will not change; but they can rely on one stable source to do this.

We could also look into running the icons through the same [svgo](https://github.com/svg/svgo) pipeline so they are all formatted in the same way.

---

**Other scope: Also bundle unused icons**

We could also consider to create a bundle of all possible icons, which would make it easier to create preview of available icons as part of the schema preview https://github.com/openstreetmap/id-tagging-schema/issues/646

matkoniecz:
It could help to prevent issues like https://github.com/openstreetmap/id-tagging-schema/issues/2207 I guess.

matkoniecz:
> we also package all the used icons.

note that we would need to deal with some licensing fun

tordans:
> > we also package all the used icons.
> 
> note that we would need to deal with some licensing fun

I don't see this as an issue. We would have a Readme about sources and licenses and also could use svgo to add a data attribute or similar to the files themselves ([via a plugin](https://svgo.dev/docs/plugins/)). The change for consuming projects is that they need to look at one file for this info and not research it for each separately.

k-yle:
at the same time we should add some compile-time checks to prevent bugs like #2335

that would also allow the spritesheet to contain only the used icons, not every single icon.

PR #2596, opened 2026-07-24
Selected a more appropriate icon for Fertility Clinics

fibonacci-matrix:

MakiHospital
Fa7SolidPersonPregnant

### Related issues

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Tag:healthcare:speciality%3Dfertility

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/tags/healthcare:speciality=fertility

### DIFF
--- a/data/presets/amenity/clinic/fertility.json
+++ b/data/presets/amenity/clinic/fertility.json
@@ -1,5 +1,5 @@
 {
-    "icon": "maki-hospital",
+    "icon": "fas-person-pregnant",
     "geometry": [
         "point",
         "area"


fibonacci-matrix:
@k-yle @tyrasd is it expected that the icon is clipped, or is this an iD rendering issue?

k-yle:
> is it expected that the icon is clipped

for some bizzare reason font awesome v7 uses incorrect [`viewBox`](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/viewBox) values....
image

(`M0 0 H384 V512 H0 Z` was my own addition for debugging)


pullings icons from a standardised icon aggregator like iconify would solve this, since they normalise the weird sizes (likewise for temaki). since we [don't want to do this](https://github.com/openstreetmap/schema-builder/pull/278#ref-issue-3093176583), i guess we have to calculate the real bbox ourselves and then centre it vertically and horizontally?

tordans:
> i guess we have to calculate the real bbox ourselves and then centre it vertically and horizontally?

Or do https://github.com/openstreetmap/id-tagging-schema/issues/2208 and use svgo to streamline the SVGs in a way that works for us.

matkoniecz:
> i guess we have to calculate the real bbox ourselves and then centre it vertically and horizontally?

or get such icon in pinhead/roentgen but less broken?

Sadly, https://pinhead.ink/#pregnant_person seems far less readable to me...

matkoniecz:
Another option: report bug to whoever makes font awesome v7

fibonacci-matrix:
> Another option: report bug to whoever makes font awesome v7

I plan to do this soon

fibonacci-matrix:
Looks like it’s not the only icon with a cropped head. `fas-running` is already in the repo

PR #2601, opened 2026-07-25
mark some strings as strings

matkoniecz:
in-editor linter wanted to do this, and it looked reasonable and scripts still work


tordans (approved):

k-yle (changes requested):
this is sympton of a bigger problem. If you're seeing that error, it means that the defintions for `fs.globSync` were not loaded. 

do you need to run `npm install` again? the version [was updated 2 weeks ago](https://github.com/openstreetmap/id-tagging-schema/blame/41c7c0fe09ca00b68aec68ceb221440c124f2b89/package-lock.json#L836)

matkoniecz:
`npm install` fixed it, sorry for a noise

PR #2608, opened 2026-07-27
fix bug with stringsCrossReference

k-yle:
fixes a bug discovered in https://github.com/openstreetmap/id-tagging-schema/pull/1014#issuecomment-5083398835

If `field.options` does not exist, it is set to `Object.keys(strings.options)`. This logic already exists and works. 

However, it doesn't work when using `stringsCrossReference`. Currently, `stringsCrossReference` only copies `strings.options`. It also needs to copy `options`. 

---

The diff of `npm run dist` for this PR is 69ef6d2be5079cdf1e3dd6728a5763f8d0d0c9fe

## Sidebar screenshots

The easiest way to test this is to draw an area tagged with `power=plant`. The _Energy Source_ field (`plant:source`) now works correctly.

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>image
 <td>image
</table>

### DIFF
--- a/SCHEMA.md
+++ b/SCHEMA.md
@@ -559,6 +559,11 @@ An optional property to reference to the strings of another field, indicated  by
 
 This would inherit all translations from the `sport` field but keep only the defined options.
 
+If you omit [`options`](#options), then it will include every option from the referenced field. For example:
+```json
+  "stringsCrossReference": "{sport}",
+```
+
 ##### `autoSuggestions`
 
 For combo fields, the most common tag values will be fetched from TagInfo and shown
--- a/scripts/lib/references.ts
+++ b/scripts/lib/references.ts
@@ -9,11 +9,11 @@ export function isReference(string: string) {
  * For example, `fields` can reference the list of field IDs from another
  * preset.
  */
-export function dereferenceUntranslatedContent(presets: AllPresets, fields: AllFields) {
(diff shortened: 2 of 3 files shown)


SCHEMA.md
```diff
@@ -559,6 +559,11 @@ An optional property to reference to the strings of another field, indicated  by
 
 This would inherit all translations from the `sport` field but keep only the defined options.
 
+If you omit [`options`](#options), then it will include every option from the referenced field. For example:
+```json
+  "stringsCrossReference": "{sport}",
+```
```
k-yle, line 565: this was already documented [on line 485](https://github.com/openstreetmap/id-tagging-schema/blame/674599e2d795f746704c4522f7d55c392faddd66/SCHEMA.md#L485), so it's not a behaviour change.

i just want to mention it here, because it wasn't clear to me if this was a bug or a feature

scripts/lib/references.ts
```diff
@@ -130,6 +130,25 @@ export function dereferenceUntranslatedContent(presets: AllPresets, fields: AllF
       delete field.iconsCrossReference;
     }
 
+
+    // 7b. if a field uses `stringsCrossReference`, we need to copy `options` if it
```
matkoniecz: Why it is 7b and later 7a? Overall numbering here looks weird

k-yle: numbers removed now

tyrasd (approved):

PR #2611, opened 2026-07-28
Use 'Physical Therapist' instead of 'Physiotherapist'

### LINKED ISSUE #2425: Use "Physical Therapist" instead of "Physiotherapist"
"Physical Therapist" should be used instead of "Physiotherapist" as this project uses US English.

mugdhachalla:

Updates the user facing preset name from **"Physiotherapist"** to **"Physical Therapist"** to align with the project's use of American English.

This change only affects the displayed preset name. 

### Related issues

Closes #2425

### Links and data

**Relevant OSM Wiki links:**
- N/A

**Relevant tag usage stats:**
> N/A

### DIFF
--- a/data/presets/healthcare/physiotherapist.json
+++ b/data/presets/healthcare/physiotherapist.json
@@ -18,5 +18,8 @@
     "tags": {
         "healthcare": "physiotherapist"
     },
-    "name": "Physical Therapist"
+    "name": "Physical Therapist",
+    "terms": [
+        "physiotherapist"
+    ]
 }
--- a/data/presets/healthcare/physiotherapist.json
+++ b/data/presets/healthcare/physiotherapist.json
@@ -18,5 +18,5 @@
     "tags": {
         "healthcare": "physiotherapist"
     },
-    "name": "Physiotherapist"
+    "name": "Physical Therapist"
 }
--- a/data/presets/healthcare/physiotherapist.json
+++ b/data/presets/healthcare/physiotherapist.json
@@ -19,7 +19,7 @@
         "healthcare": "physiotherapist"
     },
     "name": "Physical Therapist",
-    "terms": [
-        "physiotherapist"
+    "aliases": [
+        "Physiotherapist"
     ]
 }


data/presets/healthcare/physiotherapist.json
```diff
@@ -18,5 +18,5 @@
     "tags": {
         "healthcare": "physiotherapist"
     },
-    "name": "Physiotherapist"
+    "name": "Physical Therapist"
```
1ec5: Should we add “physiotherapist” to `terms`, or is that already implied by the raw tag value?

mugdhachalla: I tested this in the preview, and searching for physiotherapist still returns the Physical Therapist preset, so it looks like the raw tag value is already indexed. Happy to add it to terms as well if you think that's preferable.

mugdhachalla:
## Test Documentation

Verified in the preview that the preset name has been updated from **Physiotherapist** to **Physical Therapist** while the underlying tag value remains `healthcare=physiotherapist`.

Preview link:
https://osmberlin.github.io/tagging-schema-browser/preset/healthcare/physiotherapist?dataUrl=https://pr-2611--ideditor-presets-preview.netlify.app/dist/

Screenshots:
image 

Searching for physiotherapist returns Physical Therapist preset:
image

Searching for physical therapist returns Physical Therapist preset:
image


### Search

Verified the following in the preview:

- Searching for `Physical Therapist` returns the correct preset.
- Searching for `physiotherapist` also returns the correct preset.

### Info-`i`

No changes to the info panel were introduced by this PR.

### Wording

- [x] American English
- [x] `name` uses Title Case
- [x] `terms` use lower case and remain sorted A-Z

data/presets/healthcare/physiotherapist.json
matkoniecz: it is indexed specifically by iD having special rule for that

though if we would want to ensure that raw-tag-value as term is present/not present in `terms` we can do it separately

matkoniecz (approved):

data/presets/healthcare/physiotherapist.json
matkoniecz: On the other hand it IS actual term, even if it is listed via special mechanism in iD.

I would also list it then, not caring that search at least in iD finds in anyway

see also https://github.com/openstreetmap/id-tagging-schema/issues/2618 and https://github.com/openstreetmap/id-tagging-schema/pull/2619

matkoniecz:
```suggestion
    "name": "Physical Therapist",
    "terms": [
        "physiotherapist"
    ]
```

this should work

mugdhachalla: done, thanks!

mugdhachalla: I just noticed the suggestion replaces the existing terms array. Should I keep the existing terms and add physiotherapist, or replace them entirely?

matkoniecz: Add new value if terms array exists already - sorry that I missed existing one,  I commented on mobile device from train

1ec5: Yeah, come to think of it, it would even be a reasonable `alias`. I think it could prompt translators to check if they’ve got a similar split between a layperson’s term and specialist jargon.

matkoniecz (approved):

data/presets/healthcare/physiotherapist.json
matkoniecz: I would not be against getting such PR either in future, if you think it makes sense,  though I would be fine with this one as is right now

matkoniecz: @mugdhachalla though upgrading it to alias from terms may be included also in this PR

mugdhachalla: Just to clarify, when you mention upgrading it to an alias, would you prefer moving Physiotherapist from terms to aliases, or keeping it in both terms and aliases?

matkoniecz: Moving to it being an alias

mugdhachalla: Done! I've moved it to aliases

matkoniecz:
Thanks! I expected it to be a bit simpler so thanks for going through whole process!

matkoniecz:

PR #2616, opened 2026-07-28
format JSON schema files

k-yle:
schema-builder and this repo used different code formatters. 

after merging the repos, there are some files which don't follow the formatting rules

so it would be nice to fix that. This commit can be added to [`.git-blame-ignore-revs`](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#bypassing-git-blame-ignore-revs-in-the-blame-view) after it's merged

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,5 +8,5 @@ trim_trailing_whitespace = true
 indent_style = space
 indent_size = 4
 
-[{data/deprecated.json,data/discarded.json,package.json,.package-lock.json}]
+[{data/deprecated.json,data/discarded.json,package.json,.package-lock.json,*.ts}]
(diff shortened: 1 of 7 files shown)


tordans (approved):
Good idea to do this now. There should be no pending PRs that would be touched by those changes, so don't see a reason not to align this.

tordans:
> This commit can be added to [`.git-blame-ignore-revs`](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#bypassing-git-blame-ignore-revs-in-the-blame-view) after it's merged

@k-yle yes, why not. Not many changes, though, so IMO this does not have a high priority.

PR #2630, opened 2026-07-31
Add translations for Scuba Diving Services

fibonacci-matrix:
### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Tag:shop%3Dscuba_diving

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/search?q=scuba_diving

### Wording

- [x] American English

### DIFF
--- a/data/fields/scuba_diving.json
+++ b/data/fields/scuba_diving.json
@@ -2,16 +2,18 @@
     "key": "scuba_diving:",
     "type": "multiCombo",
     "label": "Services",
-    "options": [
-        "air_filling",
-        "courses",
-        "filling",
-        "hobby",
-        "nitrox_filling",
-        "oxygen_filling",
-        "rental",
-        "repair",
-        "trimix_filling"
-    ],
+    "strings": {
+        "options": {
+            "air_filling": "Air Filling",
+            "courses": "Courses",
+            "filling": "Filling",
+            "hobby": "Hobby",
+            "nitrox_filling": "Nitrox Filling",
+            "oxygen_filling": "Oxygen Filling",
+            "rental": "Rental",
+            "repair": "Repair",
+            "trimix_filling": "Trimix Filling"
+        }
+    },
     "autoSuggestions": false
 }
--- a/data/fields/scuba_diving.json
+++ b/data/fields/scuba_diving.json
@@ -2,12 +2,22 @@
     "key": "scuba_diving:",
     "type": "multiCombo",
     "label": "Services",
+    "options": [
+        "air_filling",
+        "courses",
+        "filling",
+        "hobby",
+        "nitrox_filling",
+        "oxygen_filling",
+        "rental",
+        "repair",
+        "trimix_filling"
+    ],
     "strings": {
         "options": {
             "air_filling": "Air Filling",
             "courses": "Courses",
             "filling": "Filling",
-            "hobby": "Hobby",
             "nitrox_filling": "Nitrox Filling",
             "oxygen_filling": "Oxygen Filling",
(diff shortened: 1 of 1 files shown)


fibonacci-matrix:

data/fields/scuba_diving.json
```diff
@@ -2,16 +2,18 @@
...
-        "rental",
-        "repair",
-        "trimix_filling"
-    ],
+    "strings": {
+        "options": {
+            "air_filling": "Air Filling",
+            "courses": "Courses",
+            "filling": "Filling",
+            "hobby": "Hobby",
```
matkoniecz: where this one is documented on wiki?

fibonacci-matrix: https://wiki.openstreetmap.org/wiki/Tag:shop%3Dscuba_diving

matkoniecz: there is no word "hobby" on this page from what I see

fibonacci-matrix: Is it better to remove it?

matkoniecz: no idea, would require some research

matkoniecz: Marking as [waitfor-documentation ](https://github.com/openstreetmap/id-tagging-schema/issues?q=state%3Aopen%20label%3Awaitfor-documentation) as it is not entirely clear what it means, and translators should not be asked to translate undocumented tags

fibonacci-matrix: Would it make sense to add translations for the fields that are not in doubt and discuss the remaining ones with the community?

matkoniecz: it makes sense to me

matkoniecz: https://github.com/openstreetmap/id-tagging-schema/issues/2648

data/fields/scuba_diving.json
```diff
@@ -13,5 +13,17 @@
         "repair",
         "trimix_filling"
     ],
+    "strings": {
+        "options": {
+            "air_filling": "Air Filling",
+            "courses": "Courses",
+            "filling": "Filling",
+            "nitrox_filling": "Nitrox Filling",
```
matkoniecz: this looks reasonably, but is not documented on wiki

no idea whether OSM Wiki is missing established value or is it case of iD in past inventing new tagging scheme

wiki has [scuba_diving:air_filling](https://wiki.openstreetmap.org/wiki/Key:scuba_diving:air_filling)=* nitrox;trimix;heliair;heliox;hydrox;hydreliox;...

maybe it also should be discussed? maybe iD tagging scheme needs to be adjusted? maybe wiki should be edited

requires research

fibonacci-matrix: Considering all of this, can we add translations for the three keys: "courses", "rental", and "repair" for now?

matkoniecz: definitely

matkoniecz: obvious meaning and documented

fibonacci-matrix:
So, at the moment there are two open questions: 

1. tagging scheme for `scuba_diving:filling`
2. whether `scuba_diving:hobby` is needed

matkoniecz:
> tagging scheme for scuba_diving:filling

my wild guess is that iD invented new tagging scheme as it was incapable then of supporting documented one (personally, it actually looks better to me, but this should not be pushed via presets)

matkoniecz (approved):

matkoniecz:
Thanks again!

PR #2633, opened 2026-08-01
Add icons for 'craft' combobox

fibonacci-matrix:

This will help choose an icon from the combobox

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Key:craft

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/keys/craft


fibonacci-matrix:








Screenshot 2026-08-01 215505

data/fields/craft.json
```diff
@@ -2,6 +2,64 @@
...
+        "distillery": "temaki-storage_fermenter",
+        "dressmaker": "temaki-dress",
+        "electrician": "temaki-power",
+        "electronics_repair": "fas-screwdriver",
+        "floorer": "temaki-brick_trowel",
+        "gardener": "maki-garden-centre",
+        "glaziery": "temaki-window",
+        "grinding_mill": "temaki-tools",
+        "handicraft": "temaki-vase",
+        "hvac": "temaki-tools",
```
paulklie, line 30: how about https://pinhead.ink/#snowflake ?

data/fields/craft.json
```diff
@@ -2,6 +2,64 @@
...
+        "agricultural_engines": "temaki-tools",
+        "basket_maker": "temaki-vase",
+        "beekeeper": "roentgen-bee",
+        "blacksmith": "temaki-anvil_and_hammer",
+        "boatbuilder": "temaki-boat_repair",
+        "bookbinder": "maki-library",
+        "brewery": "temaki-storage_fermenter",
+        "builder": "temaki-tools",
+        "carpenter": "fas-hammer",
+        "carpet_layer": "temaki-tools",
```
paulklie, line 15: https://pinhead.ink/#carpet is not super recognisable, but maybe still better?

data/fields/craft.json
```diff
@@ -2,6 +2,64 @@
...
+        "dressmaker": "temaki-dress",
+        "electrician": "temaki-power",
+        "electronics_repair": "fas-screwdriver",
+        "floorer": "temaki-brick_trowel",
+        "gardener": "maki-garden-centre",
+        "glaziery": "temaki-window",
+        "grinding_mill": "temaki-tools",
+        "handicraft": "temaki-vase",
+        "hvac": "temaki-tools",
+        "insulation": "temaki-tools",
```
paulklie, line 31: how about fa7-solid:temperature-arrow-up ?

data/fields/craft.json
```diff
@@ -2,6 +2,64 @@
     "key": "craft",
     "type": "typeCombo",
     "label": "Type",
+    "icons": {
+        "agricultural_engines": "temaki-tools",
```
paulklie, line 6: how about fa7-solid:tractor ?

fibonacci-matrix:
@matkoniecz what do you think about the suggested icons? Would it make sense to add them to the presets in a separate PR?

data/fields/craft.json
matkoniecz: I see it in past and had mixed feelings (especially as it is carpet layer, not selling carpets)

either way, would require adding icon to preset, right? So maybe better as a separate PR

data/fields/craft.json
matkoniecz: tentatively: makes sense to me, but it is of variety where I would leave it to others for review rather than merge immediately

and would be better as a separate PR

data/fields/craft.json
matkoniecz: a bit abstract but I really like the idea :)

(again: either way, would require adding icon to preset, right? So maybe better as a separate PR)

data/fields/craft.json
matkoniecz: even more abstract, but may be a good idea

I have more mixed feelings here

matkoniecz:
> Would it make sense to add them to the presets in a separate PR?

I think so, see comments above (with varying degree of enthusiasm)

data/fields/craft.json
```diff
@@ -2,23 +2,81 @@
...
+        "gardener": "maki-garden-centre",
+        "glaziery": "temaki-window",
+        "grinding_mill": "temaki-tools",
+        "handicraft": "temaki-vase",
+        "hvac": "temaki-tools",
+        "insulation": "temaki-tools",
+        "jeweller": "maki-jewelry-store",
+        "joiner": "fas-hammer",
+        "key_cutter": "fas-key",
+        "locksmith": "maki-marker-stroked",
```
matkoniecz, line 35:
```suggestion
```

given it is effectively empty icon, I think that leaving empty would make sense

(and BTW maybe key icon would make sense?)

matkoniecz: sorry for delay in review

fibonacci-matrix:
It looks like the build failed for an unknown reason. Could you help restart it? I don't have the necessary permissions.

matkoniecz:
close reopen dance should wake it up, except cases of new contributors

PR #2635, opened 2026-08-02
Add `office=translator` to `office` combobox

fibonacci-matrix:

This is the only `office` value for which there is a preset but no in combobox

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Tag:office%3Dtranslator

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/tags/office=translator

### Wording

- [x] American English

### DIFF
--- a/data/fields/office.json
+++ b/data/fields/office.json
@@ -49,6 +49,7 @@
             "physician": "{presets/office/physician}",
             "surveyor": "{presets/office/surveyor}",
             "quango": "{presets/office/quango}",
+            "translator": "{presets/office/translator}",
             "private_investigator": "{presets/office/private_investigator}",
             "adoption_agency": "{presets/office/adoption_agency}"
         }


fibonacci-matrix:

matkoniecz (approved):
Mixing places where you can go to get translation handled and mixing internal-only places where translation happens but no clients can walk in is annoying, but it is known unsolved and generally accepted tagging schema problem

and `office=translator` is in presets already

matkoniecz:
again, thanks for your contributions!

matkoniecz:
@fibonacci-matrix have you got script to detect such cases or have you found it manually?

(not sure whether it is worth looking for more cases like this)

fibonacci-matrix:
@matkoniecz

Not yet. I tried to write something similar for populating the `fields`, including those that are currently completely empty, but I ran out of time.

The idea is very simple: we go through all the presets, take the `key` from the `reference` of preset, check whether this `key` exists in `fields`, and then populate them.

I was stopped by a few issues:

Some fields are defined as an `options` array, while others are defined as `string.options`.
Sorting issues — some presets are sorted lexicographically, while others are sorted by usage frequency.

PR #2644, opened 2026-08-03
Reuse translations for crossings

fibonacci-matrix:

Now, crossing vertices and lines will use the same translations. Previously, the translations were identical, but they used different translation keys. This change will help reduce the workload for translators

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Crossings

### Wording

- [x] American English

### DIFF
--- a/data/presets/highway/footway/crossing/_marked.json
+++ b/data/presets/highway/footway/crossing/_marked.json
@@ -28,6 +28,6 @@
         "key": "crossing",
         "value": "marked"
     },
-    "name": "{highway/footway/crossing/uncontrolled}",
+    "name": "{highway/crossing/uncontrolled}",
     "searchable": false
 }
--- a/data/presets/highway/footway/crossing/_zebra.json
+++ b/data/presets/highway/footway/crossing/_zebra.json
@@ -22,6 +22,6 @@
         "key": "crossing",
         "value": "zebra"
     },
-    "name": "{highway/footway/crossing/uncontrolled}",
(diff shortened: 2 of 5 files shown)


fibonacci-matrix:

matkoniecz:
> and the size of the output package

AFAIK this will have no change as references are expanded at build time (maybe as result of more complete coverage it will increase)

matkoniecz (approved):
tested, looked at code, looks fine

leaving to review for a bit more as code for crossings is complex

tyrasd:
these are separate on purpose, because in some languages one might need a different word for "(generic) crossing" and "crossing in a pedestrian-only context" (etc.). 

For example, in the German translations the two  `*/traffic_signals` presets are translated as _Überweg mit Ampelregelung_ and _Gehweg-Querung mit Ampel
_. The first might technically work also for the second, but is less specific and not the ideal choice.

fibonacci-matrix:
> these are separate on purpose, because in some languages one might need a different word for "(generic) crossing" and "crossing in a pedestrian-only context" (etc.).
> 
> For example, in the German translations the two `*/traffic_signals` presets are translated as _Überweg mit Ampelregelung_ and _Gehweg-Querung mit Ampel _. The first might technically work also for the second, but is less specific and not the ideal choice.

Do you mean that the translation differs because a vertex can represent either a pedestrian crossing or a bicycle crossing or a combined one, while the line specifically represents a pedestrian crossing?

PR #2646, opened 2026-08-03
Reuse translations for unsearchable presets that have an unambiguous replacement

fibonacci-matrix:

This change will help reduce the workload for translators

### Wording

- [x] American English

### DIFF
--- a/data/presets/landuse/_farm.json
+++ b/data/presets/landuse/_farm.json
@@ -12,6 +12,6 @@
     "tags": {
         "landuse": "farm"
     },
-    "name": "{landuse/farmland}",
+    "name": "Farmland",
     "searchable": false
 }
--- a/data/presets/landuse/_farm.json
+++ b/data/presets/landuse/_farm.json
@@ -12,6 +12,6 @@
     "tags": {
         "landuse": "farm"
     },
-    "name": "Farmland",
+    "name": "{landuse/farmland}",
     "searchable": false
 }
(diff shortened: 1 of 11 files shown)


data/presets/landuse/_farm.json
```diff
@@ -12,6 +12,6 @@
     "tags": {
         "landuse": "farm"
     },
-    "name": "Farmland",
```
matkoniecz, line 15: not sure about this one

`landuse=farm` was deprecated because it was unclear is it `landuse=farmland` or farmyard

and it basically went extinct, see https://taginfo.openstreetmap.org/tags/landuse=farm

I think it should be removed altogether instead - can you drop it from this PR as this name should NOT be the same?

matkoniecz: opened https://github.com/openstreetmap/id-tagging-schema/issues/2647 as it should be removed rather name being fixed

matkoniecz:
it should avoid wasting translator time, thanks! Before I will recheck - I spotted one case that should not be the same (but probably should be removed, not fixed)

fibonacci-matrix:

matkoniecz (approved):

matkoniecz:
I wonder what else remains duplicated in such way...

fibonacci-matrix:
There are a few more, but there are still some questions about them:

1. `landuse=basin` and `natural=water;water=basin` (https://github.com/openstreetmap/id-tagging-schema/issues/1804)
2. `embankment=yes` and `man_made=embankment`

matkoniecz:
> landuse=basin and natural=water;water=basin

I think this is clearly an exact duplicate

PR #2649, opened 2026-08-03
Reuse translations for `direction` fields

fibonacci-matrix:

Most of these fields are the same; they only differ in the source tag. This change will help reduce the workload for translators

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Key:direction

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/search?q=direction

### Wording

- [x] American English

### DIFF
--- a/data/fields/camera/direction.json
+++ b/data/fields/camera/direction.json
@@ -1,7 +1,7 @@
 {
     "key": "camera:direction",
     "type": "number",
-    "label": "Direction (Degrees Clockwise)",
+    "label": "{direction}",
     "increment": 5,
     "placeholder": "45, 90, 180, 270"
 }
--- a/data/fields/railway/signal/direction.json
+++ b/data/fields/railway/signal/direction.json
@@ -1,12 +1,11 @@
 {
     "key": "railway:signal:direction",
     "type": "radio",
-    "label": "Direction Affected",
-    "strings": {
-        "options": {
-            "forward": "Forward",
-            "backward": "Backward",
-            "both": "Both / All"
-        }
-    }
+    "label": "{direction_vertex}",
+    "options": [
+        "forward",
+        "backward",
+        "both"
+    ],
+    "stringsCrossReference": "{direction_vertex}"
 }
--- a/data/fields/traffic_sign/direction.json
+++ b/data/fields/traffic_sign/direction.json
@@ -1,14 +1,13 @@
 {
     "key": "traffic_sign:direction",
     "type": "radio",
-    "label": "Direction Affected",
-    "strings": {
-        "options": {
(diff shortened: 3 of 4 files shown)


fibonacci-matrix:

matkoniecz:
@1ec5 are you aware of any weirdness that may justify it?

1ec5:
Not off the top of my head, but it does seem kind of risky because of “Both / All”. Some languages might go “both what?” Can someone check if any existing translations differ between these fields?

fibonacci-matrix:
> Not off the top of my head, but it does seem kind of risky because of “Both / All”. Some languages might go “both what?” Can someone check if any existing translations differ between these fields?

Thanks for the suggestion @1ec5. For these languages, the translations are not the same, so when I have a bit more time, I’ll try to check them:

1. **BG**
2. **CA**
3. **IR**
4. **IT**
5. **JA**
6. **KO**
7. **LV**
8. **NL**
9. **PL**
10. **TR**
11. **YUE**

For these ones, they’re almost identical — the differences are only in capitalization and spaces around “/”:

1. **DA**
2. **FA**
3. **GL**
4. **NO**

matkoniecz:
> PL

I can check this one, maybe even before I will leave train, if someone would share content

fibonacci-matrix:
> > PL
> 
> I can check this one, maybe even before I will leave train, if someone would share content

For key `direction`:
**W obu/wszystkich kierunkach**

For  other keys (`railway/signal/direction`, `traffic_sign/direction`, `traffic_signals/direction`):
**Oba / Wszystkie**

matkoniecz:
for PL unifying it to "W obu/wszystkich kierunkach" actually sounds like an improvement

matkoniecz:
@fibonacci-matrix have you looked into other cases?

(I guess that reaching out to local communities makes sense, have you done any of that? it would be nice to avoid duplication but I would be happy to help with it)

matkoniecz:
@fibonacci-matrix ?

fibonacci-matrix:
I have investigated all the mentioned languages except CJK using AI tools. For most languages, the differences are minor and similar to Polish, but Italian and Catalan use different grammatical genders. This needs to be discussed with native speakers.

Sorry for the delay. I’m a little short on time at the moment

fibonacci-matrix:
Do we need to write to each community separately, or can we use one general topic? Perhaps there are contributors who are familiar with the mentioned languages, and we could get an answer directly here?

matkoniecz:
I would use dedicated thread at least for Italian community, and maybe wait for reply there.

If that is the blocker we have an answer

PR #2663, opened 2026-08-05
linter: mention actual command to improve formatting

matkoniecz:
fixes #2148

fixes #2148

### Related issues

fixes #2148

### Links and data

**Relevant OSM Wiki links:**
- …

**Relevant tag usage stats:**
> …

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "main": "build.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "lint": "eslint && tsc && prettier --check data",
+    "lint": "eslint && tsc && prettier --check data || (echo && echo 'Formatting issues found. Run `npm run lint:fix` to fix them.' && exit 1)",
     "lint:fix": "eslint --fix && tsc && prettier --write data",
     "build": "node scripts/build.ts",
     "dist": "node scripts/dist.ts",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "main": "build.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "lint": "eslint && tsc && prettier --check data || (echo && echo 'Formatting issues found. Run `npm run lint:fix` to fix them.' && exit 1)",
+    "lint": "eslint && tsc && prettier --check data || (echo '::error Formatting issues found. Run `npm run lint:fix` to fix them.' && exit 1)",
     "lint:fix": "eslint --fix && tsc && prettier --write data",
     "build": "node scripts/build.ts",
     "dist": "node scripts/dist.ts",


matkoniecz:
yes, it is attempt to summon @k-yle with a better idea of fixing it

this solution always assumes that failure was caused by linting issues and only appends extra message

(but current `Code style issues found in the above file. Run Prettier with --write to fix.` is distinctly unhelpful, so improving it would help)

matkoniecz:
Obviously, prettier does not allow configuring such errors ( https://prettier.io/docs/configuration ), but double-checked to be sure

k-yle:
going back to the original problem: people are editing code without using an IDE, so they're not seeing the warnings as they type. There are 2 ways to meaningfully improve this:

1. maybe we should suggest editing via [https://vscode.dev](https://vscode.dev/github/openstreetmap/id-tagging-schema) for those who really don't want to install an IDE? the DX is already much better: 

image


2. currently, you have to look at the raw CI logs to see what failed. this can be solved by configuring [Workflow Annotations](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands) properly. Then you can see the errors attached to the exact line of code, while reviewing the PR:

image

this is already configured for the spellchecker, but not for the formatter or the JSON schema validator

matkoniecz:
If I interpret things correctly I changed it to a far less silly fix. It should be also passable when used in CLI, not by Github

> maybe we should suggest editing via [https://vscode.dev](https://vscode.dev/github/openstreetmap/id-tagging-schema) for those who really don't want to install an IDE?

I am not fan of encouraging even greater lock-in and dependency on proprietary solutions, but I type it on github so...

(though locally I use VSCodium limited to actually open-source parts of vscode)

k-yle (changes requested):
similar to https://github.com/openstreetmap/id-tagging-schema/pull/2586#pullrequestreview-4873978945 - i think we should fix the root cause, instead of working around the annoying setup.

specifically: prettier can be [run via eslint](https://github.com/prettier/eslint-plugin-prettier), instead of standalone. if we do that, then [`actions/setup-node` will detect](https://github.com/actions/setup-node/blob/master/.github/eslint-stylish.json) the errors, and show them inside the PR diff, on the exact line with the issue:

Image

matkoniecz:
https://github.com/openstreetmap/id-tagging-schema/pull/2586#issuecomment-5203963787

matkoniecz:
(this above is to indicate offer to maker superior replacement, let me know if you would prefer me to implement it instead)

k-yle:
I opened #2683. after that one's merged, this PR might still be useful if you think people need a more prominent reminder about `npm run lint:fix`

PR #2645, opened 2026-08-03
Merge US and global presets for post boxes

matkoniecz:
push US-specific limit to fields

### Description, Motivation & Context

see #2643

### Related issues

fixes #2643

### Links and data

**Relevant OSM Wiki links:**
- …

**Relevant tag usage stats:**
> …

### Checklist and Test-Documentation Template

<details><summary>Read on to get your PR merged faster…</summary>

Follow these steps to test your PR yourself and make it a lot easier and faster for maintainers to check and approve it.

**This is how it works:**
1. After you submit your PR, the system will create a preview and comment on your PR:
   > 🍱 Your pull request preview is ready.
   If this is your first contribution to this project, the preview will not happen right away but requires a click from one of the project members. We will do this ASAP.

2. Once the preview is ready, use it to test your changes.

3. Now copy the snippet below into a new comment and fill out the blanks.

4. Now your PR is ready to be reviewed.

```
## Test-Documentation

### Preview links & Sidebar Screenshots

<!-- Use the preview to find examples, select the feature in question and **copy this link here**.
     Find examples of nodes/areas. Find examples with a lot of tags or very few tags. – Whatever helps to test this thoroughly.
     Add relevant **screenshots** of the sidebar of those examples. -->

<!-- FYI: What we will check:
     - Is the [icon](https://github.com/openstreetmap/id-tagging-schema/blob/main/ICONS.md) well chosen.
     - Are the fields well-structured and have good labels.
     - Do the dropdowns (etc.) work well and show helpful data. -->

### Search

<!-- **Test the search** of your preset and share relevant **screenshots** here.
     - Test the preset name as search terms.
     - Also test the preset terms and aliases as search terms (if present). -->

### Info-`i`

<!-- **Test the info-i** for your fields and preset and share relevant **screenshots** here.
     The info needs to help mappers understand the preset and when to use it.
     [Learn more…](https://github.com/openstreetmap/id-tagging-schema/blob/main/CONTRIBUTING.md#info-i)
 -->

### Wording

- [ ] American English
- [ ] `name`, `aliases` (if present) use Title Case
- [ ] `terms` (if present) use lower case, sorted A-Z
<!-- Learn more in https://github.com/openstreetmap/id-tagging-schema/blob/main/GUIDELINES.md#2-design-the-preset -->
```

</details>

Issue #1720, opened 2025-09-11
`drive_through` field for ATMs, banks and pharmacies


Dimitar5555:
Currently the presets for [banks](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/bank.json), [ATMs](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/atm.json) and [pharmacies](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/pharmacy.json) have a default field for `drive_through`. I know that in some countries it is common for such objects to have drive throughs but for most of Europe it is quite uncommon.

The obvious solution is to have two presets, one for countries where `drive_through` is in `fields` and one where it in `moreFields` (or even removed).

P.S. There was a tool that could show tag usage for each country. Can someone share the link for it?

1ec5:
The U.S. is the land of drive-through everything, so I can believe that some other countries would find it extraneous. The codebase is full of such imbalances, not only in favor of the U.S. Is the only downside to showing the field in Europe is that it takes up extra space in the sidebar? Or is this a reaction to the [10%](https://taginfo.geofabrik.de/europe/tags/amenity=atm#combinations) of `amenity=atm`s there that have `drive_through=no`?

For what it’s worth, drive-through banks and ATMs are common throughout the Americas. If it doesn’t show up in the statistics, it’s only because of a coverage gap. We also have the Drive-Through field on the `shop=alcohol` preset. That’s probably more specific to the U.S.

matkoniecz:
> Is the only downside to showing the field in Europe is that it takes up extra space in the sidebar?

this, and increased risk of invalid data added by confused users (I have yet to see drive-through banks or ATM).

In at least my area every single `drive_through` added on such objects would be pointless or wrong.

-------------------

I guess that proper fix would be to have area-limited field?

1ec5:
It does sound like drive-in banks have [declined  in the UK](https://web.archive.org/web/20080623004716/http://www.ulsterbank-group.com/drive-through-banking/) due to the rise of ATMs in general. (Here in the U.S., the trend instead is to install a drive-through ATM where the teller window used to be.)

matkoniecz:
> TLDR: For all 3 main tags, USA is in the top 2, for pharmacies Morocco is first which was surprising to see. I don't see a point in having this field for any other country except these two.

[...]

PR #2668, opened 2026-08-05
Limit `drive_through` field for ATMs to specific countries (please create issue if more countries should have this field as prominent one)

Dimitar5555:

Limit `drive_through` field for ATMs to specific countries.

I'm not sure if it should appear in moreFields like it does in #2645.

### Related issues

Reference #1720 

### Links and data

**Relevant OSM Wiki links:**
- …

**Relevant tag usage stats:**
> …

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/data/fields/drive_through_common_for_atm.json
+++ b/data/fields/drive_through_common_for_atm.json
@@ -1,0 +1,24 @@
+{
+    "key": "drive_through",
+    "type": "check",
+    "label": "{drive_through}",
+    "locationSet": {
+        "include": [
+            "ca",
+            "us",
+            "mx",
+            "cr",
+            "ec",
+            "gt",
+            "hn",
+            "ni",
+            "pr",
+            "do",
+            "ve",
+            "au",
+            "nz",
+            "id",
+            "my"
+        ]
+    }
+}
--- a/data/presets/amenity/atm.json
+++ b/data/presets/amenity/atm.json
@@ -5,13 +5,14 @@
         "network",
         "cash_in",
         "currency_multi",
-        "drive_through"
+        "drive_through_common_for_atm"
     ],
     "moreFields": [
         "branch_brand",
         "ref",
         "brand",
         "covered",
+        "drive_through",
         "height",
         "indoor",
         "level",


Dimitar5555:
Something has gone south and iD is throwing console errors.

matkoniecz:
If I understood correctly info from https://github.com/openstreetmap/id-tagging-schema/issues/1720 it should be kept enabled in few other places at least

matkoniecz:
> Something has gone south and iD is throwing console errors.

hard to help without knowing specific errors

matkoniecz:
May be related to https://github.com/openstreetmap/iD/pull/12685

matkoniecz:
forcing rebuild

Dimitar5555:
The deploy works as expected now.

> If I understood correctly info from #1720 it should be kept enabled in few other places at least

That would require maybe two more drive through country specific fields. The easier option would to keep it as `moreFields`. iD is smart enough to not show it twice in the US when it has some value.

matkoniecz:
> That would require maybe two more drive through country specific fields. 

and that is entirely fine if we offer as result better experience in for example Canada

drawbacks are minor, and can be balanced by removal of `landuse=pond` or similar pointless preset

1ec5:
So far we know of 15 countries that would need to keep a drive-through field on at least one of these presets. (Drive-through pharmacies are orthogonal to drive-through banking, though they use similar pneumatic tube technology. Ideally we’d handle those in a separate PR.)

Dimitar5555:
I've updated the PR to only cover ATMs. The 15 countries are split by region in 3 groups (North America, Central/South America, Australia/Southeast Asia). I can merge them in a single field but it feels more natural this way.

data/presets/amenity/atm.json
```diff
@@ -5,13 +5,16 @@
         "network",
         "cash_in",
         "currency_multi",
-        "drive_through"
+        "drive_through-CA-US-MX-CR",
```
matkoniecz: why three separate fields rather than a single field?

matkoniecz: > I've updated the PR to only cover ATMs. The 15 countries are split by region in 3 groups (North America, Central/South America, Australia/Southeast Asia). I can merge them in a single field but it feels more natural this way.

I see no benefit from three separate fields right now

Dimitar5555: At the very least, it is obvious that those are 3 different regions and not a list of seemingly randomly selected countries. You also save yourself from having to see a file named `drive_through-CA-US-MX-CR-EC-GT-HN-NI-PR-DO-VE-AU-NZ-ID-MY.json`. If you still prefer it as a single field, I will change it.

1ec5: I think the only reason the countries we’ve identified are clustered this way is just that they’re the countries we’re personally familiar with, directly or indirectly. So it is basically a random list of countries, probably incomplete. 😅

Is there a hard requirement to list all the affected countries in the file name? Maybe we can call it something like drive_through_atm.json for simplicity.

matkoniecz: especially as it is likely to keep changing

data/fields/drive_through_common_for_atm.json
```diff
@@ -0,0 +1,24 @@
+{
```
matkoniecz, line 1: maybe `drive_through_common_for_atm.json` would be better? Or is it too verbose?

1ec5: It probably doesn’t matter much. My suggestion was based on existing file names like data/fields/kerb/kerb_barrier.json and data/fields/hazard_highway.json.

matkoniecz:
CI died with

> Error: Service Unavailable
> Error: Failed to resolve action download info.

so I will trigger restart

data/fields/drive_through_common_for_atm.json
Dimitar5555: I also prefer the current version.

Dimitar5555:
| [n11369410398](https://pr-2668--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=20.00/39.24735/-94.65598&disable_features=boundaries&background=fr.ign.bdortho&id=n11369410398) (USA)| [n13262137309](https://pr-2668--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=21.37/42.92223/22.92999&disable_features=boundaries&background=fr.ign.bdortho&id=n13262137309) (Bulgaria) |
|-|-|
|image | image |

data/fields/drive_through_common_for_atm.json
matkoniecz: @k-yle do you have opinion here?

(or would you be prefer to not be pinged about this kind of stuff?)

k-yle: i don't really have an opinion. 

since there are no translations in this file, and NSI won't be referencing this file, the name doesn't really matter and can be changed without impacting anyone

matkoniecz (approved):

matkoniecz:
looks fine, the only aspect that makes me a bit unsure is https://github.com/openstreetmap/id-tagging-schema/issues/1720#issuecomment-5076726020

(I have not traveled enough to see drive through in my life, but maybe it globally is more typical)

matkoniecz (changes requested):
On second look I am requesting change to `drive_through_common_for_atm.json`

reason for that is that it is otherwise confusing and looking like it is disabled otherwise, when inspecting field list

(and others were neutral)

(hopefully it is the last change)

matkoniecz (approved):

matkoniecz:
`drive_through=yes amenity=atm` in USA displays field twice

see https://github.com/openstreetmap/id-tagging-schema/pull/2787 for how it needs to be fixed :(

see #2762 for bug report

PR #2672, opened 2026-08-06
Add `width` to the `waterway=drain`

Geo-2695:
This will be very useful when representing drain waterways

See https://taginfo.openstreetmap.org/tags/waterway=drain#combinations

https://wiki.openstreetmap.org/wiki/Tag%3Awaterway%3Ddrain

Perhaps the `waterway=ditch` tag also needs such fields added

### DIFF
--- a/data/presets/waterway/drain.json
+++ b/data/presets/waterway/drain.json
@@ -1,7 +1,6 @@
 {
     "icon": "iD-waterway-ditch",
     "fields": [
-        "name",
         "structure_waterway",
         "width",
         "intermittent"
--- a/data/presets/waterway/drain.json
+++ b/data/presets/waterway/drain.json
@@ -1,7 +1,9 @@
 {
     "icon": "iD-waterway-ditch",
     "fields": [
+        "name",
         "structure_waterway",
+        "width",
         "intermittent"
     ],
     "moreFields": [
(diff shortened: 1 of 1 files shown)


Geo-2695:
### Testing

Screenshot_20260806_090618

matkoniecz:
are validly tagged named drains really so popular? (there is already universal `moreFields` name field)

data/presets/waterway/drain.json
```diff
@@ -1,7 +1,9 @@
 {
     "icon": "iD-waterway-ditch",
     "fields": [
+        "name",
         "structure_waterway",
+        "width",
```
matkoniecz, line 5: I am not sure should it be pushed strongly, maybe may make more sense in `moreFields` ?

Geo-2695: drains generally have standard widths (man-made), so I think adding the `width`  field serves a practical purpose
it will be easier to add if displayed within the `fields` property

andrewharvey: Personally I would agree with including `width` in fields, but excluding `name` since all drains can have a width but few would have a name.

Though https://taginfo.openstreetmap.org/tags/waterway=drain#combinations says 9% have a `name` but only 1% have a width.

matkoniecz (changes requested):
can you add width to moreFields instead?

I worry about pushing mappers to fill it, and as result garbage being filled there

matkoniecz (approved):

matkoniecz:
Thanks! Ended a bit smaller than initially planned but still nice.

PR #2679, opened 2026-08-06
Add dedicated surface parking preset

### LINKED ISSUE #2676: Make dedicated preset for surface parking / parking lot
> Just one more signal: It is one of those things that I will create a user preset for as soon as we have this feature (with sync preferences to use in GoMap). It is quite annoying to add the "parking=surface" - nearly all my mapping when i Map those are surface…

see also https://github.com/openstreetmap/id-tagging-schema/pull/2380#issuecomment-4888163974

idea from #2380

ak8abhinay:
Adds a dedicated preset for Surface Parking Lot (amenity=parking + parking=surface). This allows mappers to directly select surface parking with the correct tags applied, rather than manually adding the parking=surface tag to a generic lot.

Fixes #2676.

### DIFF
--- a/data/presets/amenity/parking_surface.json
+++ b/data/presets/amenity/parking_surface.json
@@ -33,11 +33,11 @@
         "parking": "surface"
     },
     "terms": [
-        "surface parking",
-        "parking lot",
-        "car park",
-        "open air parking",
         "at-grade parking",
-        "ground level parking"
+        "car park",
+        "ground level parking",
+        "open air parking",
+        "parking lot",
+        "surface parking"
     ]
 }
--- a/data/presets/amenity/parking_surface.json
+++ b/data/presets/amenity/parking_surface.json
@@ -1,36 +1,36 @@
 {
-  "icon": "temaki-car_parked",
-  "fields": [
-    "operator",
-    "operator/type",
-    "capacity_parking",
-    "capacity/disabled_parking",
-    "access_simple",
-    "fee",
-    "payment_multi_fee",
-    "charge_fee",
-    "surface"
-  ],
-  "moreFields": [
-    "{amenity/parking}"
-  ],
-  "geometry": [
-    "area",
-    "point",
-    "vertex"
-  ],
-  "name": "Surface Parking Lot",
-  "tags": {
-    "amenity": "parking",
-    "parking": "surface"
-  },
-  "terms": [
-    "at-grade parking",
-    "car park",
-    "ground level parking",
-    "open air parking",
-    "parking lot",
-    "surface lot",
-    "surface parking"
-  ]
+    "icon": "temaki-car_parked",
+    "fields": [
+        "operator",
+        "operator/type",
+        "capacity_parking",
+        "capacity/disabled_parking",
+        "access_simple",
+        "fee",
+        "payment_multi_fee",
+        "charge_fee",
+        "surface"
+    ],
+    "moreFields": [
+        "{amenity/parking}"
+    ],
+    "geometry": [
+        "area",
+        "point",
+        "vertex"
+    ],
+    "name": "Surface Parking Lot",
+    "tags": {
+        "amenity": "parking",
+        "parking": "surface"
+    },
+    "terms": [
+        "at-grade parking",
+        "car park",
+        "ground level parking",
+        "open air parking",
+        "parking lot",
+        "surface lot",
+        "surface parking"
+    ]
 }
--- a/data/presets/amenity/parking_surface.json
+++ b/data/presets/amenity/parking_surface.json
@@ -1,0 +1,43 @@
+{
+    "icon": "temaki-car_parked",
+    "fields": [
+        "operator",
+        "operator/type",
+        "capacity_parking",
+        "capacity/disabled_parking",
+        "access_simple",
+        "fee",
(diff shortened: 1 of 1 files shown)


data/presets/amenity/parking_surface.json
```diff
@@ -0,0 +1,43 @@
+{
+    "icon": "temaki-car_parked",
+    "fields": [
+        "operator",
```
matkoniecz: is it possible to reference part at least from more general preset?

data/presets/amenity/parking_surface.json
```diff
@@ -0,0 +1,43 @@
...
+        "capacity_parking",
+        "capacity/disabled_parking",
+        "access_simple",
+        "fee",
+        "payment_multi_fee",
+        "charge_fee",
+        "surface"
+    ],
+    "moreFields": [
+        "{@templates/contact}",
```
matkoniecz: is it possible to reference part at least from more general preset?

data/presets/amenity/parking_surface.json
ak8abhinay: I considered referencing the general parking preset, but it includes the parking field. Since this preset fixes `parking=surface`, that would expose an unnecessary parking field dropdown to the UI.

data/presets/amenity/parking_surface.json
ak8abhinay: I considered referencing the general parking preset, but it includes the parking field. Since this preset fixes `parking=surface`, that would expose an unnecessary parking field dropdown to the UI.

matkoniecz: For start, I would not expect this field to be in both `fields` and `moreFields`. At least one of them should be free completely from this issue

ak8abhinay: I doubt we can reference only `moreFields` like that.

matkoniecz: we can

https://github.com/openstreetmap/id-tagging-schema/blob/de64f337107037a1ba6420085aabe2548d85cea0/data/presets/natural/reef.json#L11

data/presets/amenity/parking_surface.json
```diff
@@ -0,0 +1,43 @@
...
+        "parking": "surface"
+    },
+    "terms": [
+        "at-grade parking",
+        "car park",
+        "ground level parking",
+        "open air parking",
+        "parking lot",
+        "surface parking"
+    ]
```
1ec5, line 27: “Surface lot” is a common informal word for these.

data/presets/amenity/parking_surface.json
ak8abhinay: I understand that `fields` and `moreFields` can reference the same group, as in `reef.json`. My concern is that the general parking preset includes the parking field, which would expose the parking dropdown even though this preset fixes parking=surface. How would you recommend avoiding that while still reusing the general preset?

data/presets/amenity/parking_surface.json
```diff
@@ -0,0 +1,43 @@
...
+        "amenity": "parking",
+        "parking": "surface"
+    },
+    "terms": [
+        "at-grade parking",
+        "car park",
+        "ground level parking",
+        "open air parking",
+        "parking lot",
+        "surface parking"
```
matkoniecz, line 26:
```suggestion
        "surface lot",
        "surface parking"
```
applicable suggestion from @1ec5

data/presets/amenity/parking_surface.json
matkoniecz: as https://github.com/openstreetmap/id-tagging-schema/blob/de64f337107037a1ba6420085aabe2548d85cea0/data/presets/amenity/parking.json `parking` field is in `fields` referencing `moreFields` will not pull it

maybe https://github.com/openstreetmap/id-tagging-schema/blob/main/SCHEMA.md#fieldsmorefields should be more clear, but using reference in `moreFields` gets only `moreFields` entries

using reference in `fields` would pull `fields` entries

ak8abhinay: Thanks for clarifying that! I didn't realize `moreFields` references only map to the parent's `moreFields` array. I have been thinking reference can only be used in `morefields` when reference used in `fields`.

data/presets/amenity/parking_surface.json
matkoniecz: https://github.com/openstreetmap/id-tagging-schema/blob/main/SCHEMA.md#fieldsmorefields

> Fields for keys that define the preset via tags are generally not inherited. E.g. the shop field is not inherited despite specifying as inheriting from {shop} presets.

this is a bit vaguely stated and confuses myself, but I tested it in https://github.com/openstreetmap/id-tagging-schema/pull/2689 and should be safe to inherit

matkoniecz: https://github.com/openstreetmap/id-tagging-schema/pull/2691 attempts to make docs clearer

matkoniecz (changes requested):
`fields` should also reference parent preset - processing is smart enough to drop `parking` field when inherinting

matkoniecz (approved):

matkoniecz:
Thanks for the PR!

PR #2682, opened 2026-08-07
replace jsonschema validator library with ajv

k-yle:
part 1 of 3 independant changes to make the DX better for contributors who don't use IDEs.

see https://github.com/openstreetmap/id-tagging-schema/pull/949#issuecomment-5188560933 for context

---

by switching from https://npm.im/jsonschema to https://npm.im/ajv we get access to the `line`+`column` number of each validation issue.

This means the errors can be shown in pull request UI, so no need to dig through the CI logs. Example from this PR:

image


matkoniecz:
for some reasons error messages appeared twice

matkoniecz:
screen-2026-08-07-12-52-46

matkoniecz:
or should we consider this as github being github?

k-yle:
that's because we have two parallel CI jobs which both run schema validation:
- `build-preview.yml` runs `npm run dist`
- `test.yml` runs `npm run test`

CI minutes are free, so the extra 30 seconds of compute power is not costing us anything. but we could optimise this separately

matkoniecz:
I care less about minutes (so far we are not running into any limits AFAIK) but rather about duplication.

Are there some errors that will appear in one of them but another and reverse? Maybe we can just run more noisy one?

Though that may lead to missing some bugs in build code itself.

matkoniecz:
BTW, would it be possible to now remove that fail commit or open new mergeable PR without it?

k-yle:
> [...] Maybe we can just run more noisy one?

done in #2686

> BTW, would it be possible to now remove that fail commit or open new mergeable PR without it?

~~i'll do that after #2686 is accepted or rejected, because github's stack diffs can't form a Y shape~~ edit: nevermind, done now

matkoniecz:
Why space appears as dot AND is quoted?

screen-2026-08-08-07-38-10

probably not fixable but still weird

scripts/lib/ci.ts
```diff
@@ -0,0 +1,44 @@
+import type { IOutputError } from 'better-ajv-errors';
+
+const escapeData = (str: string) => str
+  .replace(/%/g, '%25')
```
matkoniecz, line 4: maybe more general-purpose escape would be better?

(I also did escape this way few times and almost often it was later broken by unexpected weird inputs)

matkoniecz (approved):
as long as failing commit is removed

(my other comments should be treated as non-blocking nitpicking)

scripts/lib/ci.ts
k-yle: as far as i can tell, these are the only ones that need to be escaped, [`encodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) would be escaping too much

k-yle:
demo commit removed. 

> Why space appears as dot AND is quoted?

i guess for the same reason that [taginfo does it](https://taginfo.osm.org/tags/inscription=24%20%26%2025%20VICT%0ACAP%2042). the backticks are markdown `code blocks` which i guess is not supported in that UI

PR #2683, opened 2026-08-07
run prettier via eslint

k-yle:
part 2 of 3 independant changes to make the DX better for contributors who don't use IDEs.

see https://github.com/openstreetmap/id-tagging-schema/pull/2663#pullrequestreview-4874026788 for context. 

---

currently we run prettier as a standalone tool. instead, we can run prettier via eslint, then [`actions/setup-node` will detect](https://github.com/actions/setup-node/blob/master/.github/eslint-stylish.json) the errors, and show them inside the PR diff, on the exact line with the issue:

image


matkoniecz:
> part 2 of 3 independant changes to make the DX better for contributors who don't use IDEs.

what have happened to part 1? (or was it merged some time ago?)

matkoniecz:
Also, big big thanks for implementing this!

FloEdelmann:
The PR title is the wrong way around: this PR makes ESLint run Prettier

eslint.config.js
```diff
@@ -125,5 +127,19 @@ export default defineConfig(
...
       globals: vitest.environments.env.globals,
     },
+  },
+  {
+    files: ['**/*.json'],
+    ignores: [
+      'package.json',
+      'package-lock.json',
+      'data/deprecated.json',
+      'data/discarded.json',
```
matkoniecz, line 137: are these having some special syntax or formatting? (deprecated, discarded)

matkoniecz (approved):
as long as that failing commit is removed from PR

eslint.config.js
k-yle: yes, these ones don't follow the standardised format

k-yle:
demo commit removed

matkoniecz:
Has some merge conflicts

k-yle:
> Has some merge conflicts

fixed, but that will keep happening because of the neverending spam of @dependabot PRs

matkoniecz:
And I merged more of those, dammit.

PR #2684, opened 2026-08-07
move the sorted-terms validator to a custom lint rule

k-yle:
part 3 of 3 independant changes to make the DX better for contributors who don't use IDEs.

see #2586 for context

---

this PR moves the [sorting terms validator](https://github.com/openstreetmap/schema-builder/pull/243) into a custom lint rule. it requires a little bit more code, but it has major benefits:
- you can see the issues as you type
- pressing <kbd>Ctrl</kbd> <kbd>S</kbd> in your IDE will automatically reorder the terms to fix the problem
- when you run `npm run lint`, you'll see the warnings for all files at once, instead of only seeing the first issue.
- running `npm run lint:fix` will automatically fix the sorting issues in all files 
- you can see the errors in the Github PR view:

image

image

### DIFF
--- a/scripts/lib/build.ts
+++ b/scripts/lib/build.ts
@@ -289,7 +289,6 @@ function generateFields(dataDir: string, tstrings: TStrings, searchableFieldIDs:
     // @ts-expect-error -- deleting a non-optional property
     delete field.label;
 
-    validateTerms(field.terms, `field "${id}"`);
     tstrings.fields[id].terms = Array.from(new Set(
       (field.terms || [])
         .map(t => t.toLowerCase().trim())
@@ -416,7 +415,6 @@ function generatePresets(
     ));
     preset.aliases.forEach(a => names.add(a.toLowerCase()));
 
-    validateTerms(preset.terms, `preset "${id}"`);
     preset.terms = Array.from(new Set(
       (preset.terms || [])
         .map(t => t.toLowerCase().trim())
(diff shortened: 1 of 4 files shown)


matkoniecz:
Maybe it would be nice to split "Terms must be lowercase and sorted alphabetically" into two warning to make errors immediately clear, but maybe this can be done separately.

scripts/lib/build.ts
```diff
@@ -1051,22 +1049,6 @@ function validatePresetFields(presets: AllPresets, fields: AllFields) {
...
-function validateTerms(terms: string[] | undefined, where: string) {
-  if (!terms) return;
-
-  const expectedTerms = terms.map(term => term.toLowerCase().trim()).sort();
-  if (terms.every((term, index) => expectedTerms[index] === term)) return;
-
-  process.stderr.write(`Expected terms in ${where} to be lowercase and sorted alphabetically.`);
-  process.stdout.write('\n');
-  process.stdout.write('expected terms\n\n');
-  process.stdout.write(JSON.stringify(expectedTerms, null, 2));
```
matkoniecz, line 1068: one thing that is now seemingly lost is that if you review PR on github and see this error you cannot copy anymore correct text from CI

you need to add remote locally, fetch, switch to branch and run linter

matkoniecz: or is it possible to feed `fixer` outut into CI interface somehow?

k-yle: what do you think of this? i think it's way more clunky, but it supports your copy-paste workflow.

image

it also means that the specific line is no longer highlighted, instead the entire block is highlighted:
 
image

matkoniecz: direct copy-paste is not possible as it does not have newlines, but probably good enough to satsfy maybe unusual workflow

matkoniecz: BTW, ATP has bot that autofixes formatting in PRs and pushes own fix comment automatically

maybe it would make sense for us?

k-yle: > direct copy-paste is not possible as it does not have newlines

newlines won't work unfortunately, cause this is a weird hack that very few other lint rules do

> BTW, ATP has bot that autofixes formatting in PRs and pushes own fix comment automatically
> maybe it would make sense for us?

i think that was previously suggested in some other thread, but yes that would be nice. 

this PR is a prerequisite step, so that `npm run lint:fix` will automatically fix these issues

matkoniecz (approved):
on condition that failing commit is removed

k-yle:
failing commit removed

PR #2690, opened 2026-08-07
drop node-fetch dependency

### LINKED ISSUE #2677: Drop `node-fetch` dependency
The fetch API is part of NodeJS 18 (released in 2022). `node-fetch` is used only in a single place and came from the schema builder repo. It should be safe to drop it.

https://github.com/openstreetmap/id-tagging-schema/blob/de64f337107037a1ba6420085aabe2548d85cea0/scripts/lib/translations.ts

nightcityblade:

Use the Fetch API built into the supported Node.js runtime instead of importing `node-fetch`. This removes the direct dependency and prunes its unused transitive packages from the lockfile.

Tests:
- `npx eslint scripts/lib/translations.ts`
- `npm run lint`
- `npm test` (4 tests passed)

### Related issues

Fixes #2677

### Links and data

**Relevant OSM Wiki links:**
- N/A — this does not change tagging schema data.

**Relevant tag usage stats:**
> N/A — this does not change tagging schema data.

**This is how it works:**

```
## Test-Documentation

This is a dependency-only maintenance change; it does not affect presets or the preview UI.

N/A — no preset or field behavior changed.

### Search

N/A — no preset search data changed.

### Info-`i`

N/A — no field information changed.

### Wording

- [x] N/A — no user-facing wording changed.
```

</details>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,6 @@
         "json-schema-to-typescript-lite": "^15.0.0",
         "jsonschema": "^1.1.0",
         "marky": "^1.2.4",
-        "node-fetch": "^3.2.10",
         "prettier": "^3.2.5",
         "shelljs": "^0.10.0",
         "typescript": "^6.0.3",
@@ -1443,15 +1442,6 @@
         "node": ">= 8"
       }
     },
-    "node_modules/data-uri-to-buffer": {
-      "version": "4.0.1",
-      "resolved": "https://registry.npmjs.org/data-uri-to-buffer/-/data-uri-to-buffer-4.0.1.tgz",
-      "integrity": "sha512-0R9ikRb668HB7QDxT1vkpuUBtqc53YyAwMwGeUFKRojY/NWKvdZ+9UYtRfGmhqNbRkTSVpMbmyhXipFFv2cb/A==",
-      "dev": true,
-      "engines": {
-        "node": ">= 12"
-      }
-    },
     "node_modules/debug": {
       "version": "4.4.3",
       "resolved": "https://registry.npmjs.org/debug/-/debug-4.4.3.tgz",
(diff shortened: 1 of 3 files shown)


matkoniecz:
it only removes dependency? I was not expecting it to be enough, I thought that code using it also needs to be changed

matkoniecz:
I need to double check when this code (fetching translations) is run to ensure it is tested

nightcityblade:
Thanks for checking — the code using it did change, but only at the import boundary. The PR removes `import fetch from 'node-fetch'` from `scripts/lib/translations.ts`; the existing `await fetch(url).then(d => d.text())` call remains unchanged because it now resolves to Node's global Fetch API, which has the same interface.

I also checked when this path runs. `scripts/dist.ts` supplies the translation options only for `npm run dist -- translations`, and `scripts/lib/build.ts` calls `fetchTranslations` only when both `translOrgId` and `translProjectId` are set. So you are right that the ordinary `npm test` / `npm run dist` path does not exercise the authenticated translation download. The staging workflow runs the translation form with `TX_TOKEN`.

I verified the affected path separately on Node v22.22.0 (the package requires Node `>=22`):

- `fetch`, `Response`, and `Headers` are all available globally.
- `npm run lint` passes under Node 22, including TypeScript checking of the global `fetch` call.
- A focused local integration check invoked the exported `fetchTranslations` function, mocked only the Transifex language/download metadata, and served a YAML translation from a local HTTP server. This executed the actual `fetch(url).text()` line and completed with:

  ```text
  got project languages
  got translations for presets, language fr
  got resource language stats collection for presets
  translation fetch passed: 1 HTTP request, fr.json written
  ```

So no additional source change is needed: removing the import is what switches that existing call from `node-fetch` to the built-in implementation.

tordans (approved):

PR #2693, opened 2026-08-08
Update GUIDELINES.md allowing non-US English in search terms

### LINKED ISSUE #2675: Consider allowing search terms in English which is not US English, in US English base
this appeared in https://github.com/openstreetmap/id-tagging-schema/pull/2665#pullrequestreview-4863638136

The [guidelines](https://github.com/openstreetmap/id-tagging-schema/blob/main/GUIDELINES.md?rgh-link-date=2026-08-05T10%3A36%3A44Z#2-design-the-preset) say that also search terms should be American English only

impact is limited as at least in iD searching for "youth centre" still finds the preset

see also say https://github.com/openstreetmap/id-tagging-schema/commit/79c9d5366bdada18074bbb303093a7993532829c for example where it had larger impact

I think this guideline can be loosened a bit and allow non-US search terms

I motivate it by fact that en-US is often also used by people outside US, outside places with actual English dialects like Australia/India/etc. For example using English interface is fairly popular in Poland, especially among more techy people. And it would be useful to have some limited support for reasonable weird search terms to support them.

matkoniecz:

fixes #2675

I motivate it by fact that en-US is often also used by people outside US, outside places with actual English dialects like Australia/India/etc. For example using English interface is fairly popular in Poland, especially among more techy people. And it would be useful to have some limited support for reasonable weird search terms to support them.

Including for example UK spelling or terms.

For example data/presets/leisure/pitch/soccer.json has "football" search terms and this is entirely fine.

It does not mean that any dialect term and emojis and terms from other languages not present in English should be added.

-----------------------

<s>also remove mention of category (what it is?)</s> (not anymore)

### Related issues

fixes #2675

### Links and data

**Relevant OSM Wiki links:**
- …

**Relevant tag usage stats:**
> …

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -76,7 +76,7 @@ The user interface must be clear, concise, and easy to use, leaving no room for 
 
 - Define the tags required on an object to trigger the preset.
 - Choose a name label using American English.
-- Add search terms - these should typically be American English, but they may include some terms from other English dialects.
+- Add search terms - these should typically be American English, but they may rarely include terms from other English dialects.
 - Use Title Case for the the preset `name` and [`aliases`](./SCHEMA.md#aliases) as well as the field [`label`](./SCHEMA.md#label) property. Use lower case for the preset [`terms`](./SCHEMA.md#terms) (sorted A-Z) and Title Case or sentences for preset's [`strings`-`options`](./SCHEMA.md#strings).
 - Check the search functionality to ensure other presets do not cause confusion.
 - Select an icon or start the process to create a new one.
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -81,6 +81,7 @@ The user interface must be clear, concise, and easy to use, leaving no room for 
 - Check the search functionality to ensure other presets do not cause confusion.
 - Select an icon or start the process to create a new one.
 - Define which fields to show (`fields`) and suggest (`moreFields`), considering the order of fields.
+- If needed add preset to categories
 - Check the [`(i)` documentation](./CONTRIBUTING.md#info-i) and add or update the OSM Wiki data item if needed to provide a helpful short text.
 - Use the PR preview to add test cases with deep links to OSM objects that demonstrate the preset in use.
 
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -76,7 +76,7 @@ The user interface must be clear, concise, and easy to use, leaving no room for 
 
 - Define the tags required on an object to trigger the preset.
 - Choose a name label using American English.
-- Add search terms - these should typically be American English, but they may rarely include terms from other English dialects.
+- Add search terms - these should typically be American English, but they may rarely include terms from other English dialects. Terms from other dialects must not cause confusion.
 - Use Title Case for the the preset `name` and [`aliases`](./SCHEMA.md#aliases) as well as the field [`label`](./SCHEMA.md#label) property. Use lower case for the preset [`terms`](./SCHEMA.md#terms) (sorted A-Z) and Title Case or sentences for preset's [`strings`-`options`](./SCHEMA.md#strings).
 - Check the search functionality to ensure other presets do not cause confusion.
 - Select an icon or start the process to create a new one.
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -75,7 +75,8 @@ There are, however, alternatives to consider:
 The user interface must be clear, concise, and easy to use, leaving no room for misunderstandings.
 
 - Define the tags required on an object to trigger the preset.
-- Choose a name, category, and define a list of American English search terms.
+- Choose a name label using American English.
+- Add search terms - these may include limited English content outside American English.
 - Use Title Case for the the preset `name` and [`aliases`](./SCHEMA.md#aliases) as well as the field [`label`](./SCHEMA.md#label) property. Use lower case for the preset [`terms`](./SCHEMA.md#terms) (sorted A-Z) and Title Case or sentences for preset's [`strings`-`options`](./SCHEMA.md#strings).
 - Check the search functionality to ensure other presets do not cause confusion.
 - Select an icon or start the process to create a new one.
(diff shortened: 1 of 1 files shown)


GUIDELINES.md
```diff
@@ -75,7 +75,8 @@ There are, however, alternatives to consider:
 The user interface must be clear, concise, and easy to use, leaving no room for misunderstandings.
 
 - Define the tags required on an object to trigger the preset.
-- Choose a name, category, and define a list of American English search terms.
+- Choose a name label using American English.
+- Add search terms - these may include limited English content outside American English.
```
1ec5: “Category” dropped out of this change – is it still required?

matkoniecz: Yes,  as I have no idea what category is in this context and it is not mentioned in schema.md

Unless I missed something?

We have categories but are these set in presets?

(I think it was mentioned in commit description)

1ec5: I assumed that it was a suggestion to add the preset to a relevant [preset category](https://github.com/openstreetmap/id-tagging-schema/tree/1e7cbfd4d5d78b098bccda043d77cf240bdf224d/data/preset_categories). Not every preset should be part of a category, but it wouldn’t hurt to build out the categories more. Most new mappers don’t know they can search for something that isn’t in the initial listing.

k-yle:
```suggestion
- Add search terms - these should typically be American English, but they may include limited terms from other English dialects.
```

?

k-yle (approved):
the documentation looks good to me, assuming that the idea itself is accepted (#2675). one minor phrasing suggestion below:

GUIDELINES.md
matkoniecz: I restored mention of categories, but without its naming as it is about preset creation not about category creation

1ec5: OK, sounds good. (I don’t have permission to review this PR as requested.)

matkoniecz: I think you should be still able to leave a comment

matkoniecz:
@tordans @k-yle @tyrasd @FloEdelmann @1ec5 - do you have comments/preferences on the idea itself?

1ec5:
I agree with the problem statement, but https://github.com/openstreetmap/id-tagging-schema/pull/2707#issuecomment-5301233276 makes me warier of this change. Client editors will need a corresponding enhancement that prefers earlier terms over later terms. Otherwise, there’s plenty of potential for mistagging, potentially more over time.

The reverse of my example in https://github.com/openstreetmap/id-tagging-schema/pull/2707#issuecomment-5288352991 applies here: searching for “trolley” should turn up Tram Stop over Cart Corral. It already doesn’t due to the raw tag value, so putting it in terms might exacerbate the issue. One would expect a user to notice that they aren’t tagging a cart corral, but if they aren’t a native American English speaker then they may not spot the difference before choosing the preset in haste. Other differences are even subtler, like campsite being swapped with camp pitch.

Our presets make distinctions that aren’t necessarily the usual lay distinctions. It’s hard enough to pick up on that without other English dialects throwing curveballs too. I experience this a lot with Wikidata’s autocompletion menus, as they suggest items based on matching labels in any language. But at least they weight same-language and same-dialect matches over more tangential matches.

That said, I think a manual addition of some judiciously chosen terms would be fine. That’s what you mean by “rare”, but I’d also qualify it with: as long as it doesn’t cause confusion among speakers of both dialects.

GUIDELINES.md
```diff
@@ -75,11 +75,13 @@ There are, however, alternatives to consider:
 The user interface must be clear, concise, and easy to use, leaving no room for misunderstandings.
 
 - Define the tags required on an object to trigger the preset.
-- Choose a name, category, and define a list of American English search terms.
+- Choose a name label using American English.
+- Add search terms - these should typically be American English, but they may rarely include terms from other English dialects.
```
matkoniecz:
```suggestion
- Add search terms - these should typically be American English, but they may rarely include terms from other English dialects. Terms from other dialects must not cause confusion.
```

matkoniecz: @1ec5 would it resolve your concern?

matkoniecz:
> It already doesn’t due to the raw tag value

I am pretty sure it has no impact at all in iD for preset ordering. It only has impact on value suggestions coming from taginfo, as far as I know.

GUIDELINES.md
1ec5: Yes, that’s probably the best we can do purely in this repository.

tyrasd:
> I’d also qualify it with: as long as it doesn’t cause confusion among speakers of both dialects.

I removed this qualifier, because it should be a general (even if not formally written down) rule that a preset should always be designed to be not confusing. :wink: 

IMHO, potential corner cases such as including [_dairy_](https://en.wikipedia.org/wiki/Dairy_(New_Zealand)) in the convenience store preset should be discussed on an individual case by case basis in the respective PRs during the review.

GUIDELINES.md
```diff
@@ -75,11 +75,13 @@ There are, however, alternatives to consider:
 The user interface must be clear, concise, and easy to use, leaving no room for misunderstandings.
 
 - Define the tags required on an object to trigger the preset.
-- Choose a name, category, and define a list of American English search terms.
+- Choose a name label using American English.
```
matkoniecz, line 78:
```suggestion
- Choose a name label using American English. This dialect is used for base version of text (translation for other languages and dialects are available).
```

matkoniecz: but it may be too wordy to be worth including

matkoniecz:
@1ec5 what you think now?

1ec5:
There are a lot of opportunities for confusion among speakers of different dialects of English. I think this is greatly underappreciated within the OSM community, as seen in the discussion in openstreetmap/openstreetmap-website#5029. I suggested adding a reminder about avoiding confusion specifically among different dialects so that the contributor would double-check this possibility. However, at this point we’re probably spending more than enough effort perfecting the wording of a guideline that we won’t be treating as binding law anyways.

PR #475, opened 2022-05-31
Add aliases and terms for `shop=*` presets

westnordost:
Also, corrected `shop=deli` preset name. In American English, "Deli" means something else as what `shop=deli` is originally defined with.

Sources: English Wikipedia, OpenStreetMap Wiki

PR #2696, opened 2026-08-09
Reused the translation key for `pharmacy` in the combo box

fibonacci-matrix:

After this change, the name in the combobox will change from "Pharmacy" to "Pharmacy Counter", but in most other languages the translations are still the same.

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Tag:amenity=pharmacy

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/tags/amenity=pharmacy

### Wording

- [x] American English

### DIFF
--- a/data/fields/healthcare.json
+++ b/data/fields/healthcare.json
@@ -32,7 +32,7 @@
     },
     "strings": {
         "options": {
-            "pharmacy": "Pharmacy",
+            "pharmacy": "{presets/amenity/pharmacy}",
             "doctor": "{presets/amenity/doctors}",
             "clinic": "{presets/amenity/clinic}",
             "hospital": "Hospital",


fibonacci-matrix:

matkoniecz:
if "Pharmacy Counter" even and US term for pharmacy? If it is then what "Pharmacy" is?

@1ec5 as I understand and remember "Pharmacy Counter" was used because nearly all pharmacies in USA are shop-in-shop rather than standalone shops. Is it really accurate?

1ec5:
The vast majority of retail pharmacies in the U.S. consist of a department within a drugstore, which in turn may be a department within a supermarket or hypermarket (`shop=supermarket`). Most of the remaining pharmacies are located near the entrance of a hospital or clinic, primarily serving patients. Standalone pharmacies are rare, mostly specialty pharmacies like [compounding pharmacies](https://www.health.harvard.edu/healthy-aging-and-longevity/thinking-of-using-a-compounding-pharmacy-what-you-need-to-know). We would call one of those “pharmacy” and [tag it as `amenity=pharmacy`](https://www.openstreetmap.org/node/6862922607) too.

Originally, OSM only used `shop=chemist` for what British English calls a “chemist’s shop”: a store that sells personal care items and over-the-counter drugs but does not dispense prescription drugs. Such shops are very rare in the U.S., so we avoided the tag here. But this was also at a time when we rarely mapped departments within stores, at most just the bank branch inside a supermarket. In 2016, the wiki and iD reoriented the tags so that `shop=chemist` would also cover a drugstore, as long as there’s an `amenity=pharmacy` inside it to represent the pharmacy counter: openstreetmap/iD#3213.

Confusing matters, in American English, “pharmacy” is [the most common word](https://books.google.com/ngrams/graph?content=pharmacy+aisle%2Cdrugstore+aisle&year_start=1800&year_end=2022&corpus=en-US&smoothing=3) for a drugstore that has a pharmacy inside it, by metonymy. By comparison, “drugstore” is fairly old-fashioned and regional. Personally, I was almost unaware of the word until iD started using it. So in the first few years after the switch, we saw lots of buildings tagged `amenity=pharmacy` without a corresponding `shop=chemist`, despite naming the `amenity=pharmacy` preset “Pharmacy Counter”. name-suggestion-index and especially AllThePlaces have helped us avoid discrepancies among chain stores, but one still sees orphaned pharmacy counters or vice versa about as much as one sees gas station buildings without building tags. openstreetmap/iD#4138 is one possible solution for improving the discoverability of drugstore micromapping.

Any store department can have “counter” after it to indicate behind-the-counter service: “meat counter”, “seafood counter”, “deli counter”, etc. We don’t have presets for meat and cheese counters, but deli counters have contributed to all kinds of confusion and controversy over the years: https://github.com/openstreetmap/id-tagging-schema/pull/475#discussion_r886136776. By comparison, “pharmacy counter” accurately describes the vast majority of instances in the U.S. Even if one encounters a standalone pharmacy either here or abroad, “pharmacy counter” seems accurate, if trivializing. It implies that the customer can only walk up to the counter to interact with the pharmacist or their staff – no aisles to peruse. It’s like how a `shop=ticket` would be accurately called a “ticket counter”.

matkoniecz (approved):

matkoniecz:
Still sounds weird to me, but I guess it actually makes sense

PR #2702, opened 2026-08-10
Add "Bail Bond Agent" to `office` combobox

fibonacci-matrix:
### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Tag:office%3Dbail_bond_agent

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/tags/office=bail_bond_agent

### Wording

- [x] American English

### DIFF
--- a/data/fields/office.json
+++ b/data/fields/office.json
@@ -50,7 +50,7 @@
             "surveyor": "{presets/office/surveyor}",
             "quango": "{presets/office/quango}",
             "translator": "{presets/office/translator}",
-            "bail_bond_agent": "{presets/office/bail_bond_agent}",
+            "bail_bond_agent": "{presets/office/bail_bond_agent-US-PH}",
             "private_investigator": "{presets/office/private_investigator}",
             "adoption_agency": "{presets/office/adoption_agency}"
         }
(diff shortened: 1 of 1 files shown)


fibonacci-matrix:
I can't understand why it's available worldwide rather than only in the US and the Philippines

matkoniecz:
> rather than only in the US and the Philippines

because preset `presets/office/bail_bond_agent-US-PH` is limited to specific area, while field `fields/office.json` is global?

Note how after its selection it is marked as generic Office.

It looks like here new schema capability is needed. Or create a new field for USA and Canada with all entries and one for world except USA and Canada.

fibonacci-matrix:
> It looks like here new schema capability is needed. Or create a new field for USA and Canada with all entries and one for world except USA and Canada.

I thought this could be done more easily. But considering <400 uses this is low priority. Can we close this?

matkoniecz:
yes, in general feel free to close own PRs if it turns out that effort/benefit is not looking as well as it looked initially

pity that it has not worked here (happened for me few times) but your contributions are really welcome!

PR #2707, opened 2026-08-12
automatically include american english terms in other english dialects

### LINKED ISSUE #461: "parking lot" should find `amenity=parking` also in UK English
(I may implement it, after https://github.com/openstreetmap/id-tagging-schema/pull/337 is reviewed/rejected/merged)

EDIT: asked on osm IRC "Also, I have UK English question. I know that "parking lot" is not primary term for amenity=parking in UK English. But would it be reasonable to list is among options if user searches for it (and they have set UK English as a language)?"

EDIT:

Image

k-yle:
Closes #461

the English `name`/`label`/`terms`/`aliases` which are defined in this repo are now automatically included as **terms** in other english dialects (`en-{AU,NZ,SG,CA,IN,GB,IE}`)

This only applies to english dialects, not to any other languages. 

- For translatators of `en-*`, it means that we don't need to manually copy-paste the existing terms when we rename a preset. And we don't need to worry about the terms in transifex becoming out of sync with the terms in repository.
- For users, the benefit is pretty clear and has been discussed extensively in [#461](https://github.com/openstreetmap/id-tagging-schema/issues/461). it also means that incremental improvements to this repository will benefit all english dialects, not just `en-US`. For example [#2678](https://github.com/openstreetmap/id-tagging-schema/pull/2678).

<details>
<summary>to test: search for <code>Subway Entrance</code> <a href="https://pr-2707--ideditor-presets-preview.netlify.app/id/dist/#locale=en-AU&map=18.60/48.84188/2.58732">using <code>en-AU</code></a>, which now works. or look at the PR diff, I committed the changes to <code>dist/*</code></summary>

image

</details>


andrewharvey:
I haven't reviewed all the code, but in concept I agree with the change.

matkoniecz:
Are there any cases where US terms would be outright misleading and should not be used? Probably no, due to dominance and influence of US English.

andrewharvey:
> Are there any cases where US terms would be outright misleading and should not be used? Probably no, due to dominance and influence of US English.

The US term isn't being shown to the user and the locale translated name is still used for the label, this is just about search terms so I don't think that's an issue.

1ec5:
Lots of British English terms are confusing in American English, so I would assume the reverse can happen too. Off the top of my head: trolley, deli, suburb, campsite. Though if these terms end up appearing at the end of the terms list, these discrepancies should have a more muted impact on search quality.

andrewharvey:
> Though if these terms end up appearing at the end of the terms list, these discrepancies should have a more muted impact on search quality.

I thought this change only makes existing presets show up when searching by a term that wasn't originally on the translated list of terms. With this change, for US locales searching for "nappy" (the British term) would start showing results for "diaper" (the US term) when previously it would not have come back with any results (assuming there was a preset for that).

1ec5:
If I’m reading it right, this is about the reverse: you’ll be able to search for “diaper” when using the British English locale. Which is fine: I was just commenting on the need to ensure that any British term takes precedence in preset matching, so that for instance the Trolley Bay preset comes up on top of the Tram Stop preset when you search for “trolley”.

I haven’t tried this out in detail, but I think iD will mostly do the right thing. It prefers matches towards the beginning of a terms list before matches towards the end. There might be some weirdness if American English has terms for things that British English does not, but the impact should be minimal. iD’s search ranking isn’t perfect anyways.

k-yle:
yeah this PR only affects non-US english dialects, it has no impact on `en-US` (the language that is version-controlled in this repo).

The opposite idea is tracked by #2675

> It prefers matches towards the beginning of a terms list

are you sure? i believe the order [doesn't matter](https://github.com/openstreetmap/iD/blob/77574f6f051d58e221d272d5c34ee98aee112aa2/modules/presets/collection.ts#L166-L171) - and this repo sorts terms alphabetically anyway

1ec5:
> are you sure? i believe the order [doesn't matter](https://github.com/openstreetmap/iD/blob/77574f6f051d58e221d272d5c34ee98aee112aa2/modules/presets/collection.ts#L166-L171) - and this repo sorts terms alphabetically anyway

Oh huh, you’re right, that’s [only for preset names](https://github.com/openstreetmap/iD/blob/77574f6f051d58e221d272d5c34ee98aee112aa2/modules/presets/collection.ts#L104-L106), not terms. I thought it used to also favor matches early in terms as well. There goes my workaround for the lack of language-specific diacritic folding.

dist/translations/en-AU.json
```diff
@@ -203,34 +280,48 @@
...
+                        "position"
+                    ]
                 },
                 "fire_hydrant/type": {
                     "options": {
                         "pillar": "Pillar/Above Ground"
-                    }
+                    },
+                    "terms": [
+                        "shape"
```
matkoniecz: hmm, is "shape" legitimate term in US English?

(issue is not blocking this PR)

dist/translations/en-AU.json
```diff
@@ -417,48 +592,74 @@
...
                 },
                 "roller_coaster/track": {
                     "options": {
                         "station": {
                             "description": "A section of track where a train stops to load and offload passengers"
                         }
-                    }
+                    },
+                    "terms": [
+                        "type"
```
matkoniecz: I am a bit confused how this type ended here

Label seems not defined for AU dialect so it will inherit "Type" from US English so "type" does not need to be defined as a term, right?

scripts/lib/build.ts
k-yle, line 1: > There goes my workaround for the lack of language-specific diacritic folding.

off topic, but @1ec5 could you please open an issue for that? i don't know enough about vietnamese to understand the problem. but the browser now has this data, for example:

```js
const f = new Intl.Collator('vi', { sensitivity: 'base' }).compare;

f('ở', 'ơ') // 0
f('ở', 'o') // 1
f('ơ', 'o') // 1
```

so I wonder if we can write something to extract this data out of [`Intl.Collator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator), similar to [what you did with `Intl.NumberFormat`](https://github.com/openstreetmap/iD/blob/7095309926bf93c0137b4275559473d0f681f8cd/modules/core/localizer.ts#L649-L670)?

1ec5: Possibly, if we can run the comparison performantly: openstreetmap/iD#12759.

dist/translations/en-AU.json
k-yle: this was an unrelated issue because the `dist` folder was updated on the main branch. so the diff is gone now

dist/translations/en-AU.json
k-yle: yeah, see https://github.com/openstreetmap/id-tagging-schema/pull/2707#discussion_r3813092939 - fixed now

matkoniecz:
How can I convince https://pr-2707--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=19.93/48.84220/2.58820&disable_features=boundaries&background=fr.ign.bdortho&id=w-1 to show me page in UK English?

with browser's Accept-Language header: `pl,en-GB;q=0.9` it still shows in US English

1ec5:
Change the `locale=` parameter from `en` (which means `en-US`) to `en-GB`.

matkoniecz (changes requested):
> Closes https://github.com/openstreetmap/id-tagging-schema/issues/461

seems not? "Car Park" preset is not found, only this new not-yet-translated one

https://pr-2707--ideditor-presets-preview.netlify.app/id/dist/#locale=en-GB&map=19.93/48.84220/2.58820&disable_features=boundaries&background=fr.ign.bdortho&id=w-2

Image

k-yle:
> seems not? "Car Park" preset is not found, only this new not-yet-translated one

that's because the PR preview doesn't build translations. to work around this, i've commited the changes to `dist/translations/en-*.min.json`. so it can be tested now

matkoniecz:
screen-2026-08-22-12-52-04

matkoniecz (approved):
now the only question is whether this should be reduced to code change, or should PR commit include also rebuilds of /dist and pulling translation update

k-yle:
i think it's file to keep the partial updates to `dist/translations/en-*.json` in this PR, because it makes the git blame more meaningful. 

but that's not particuarly important, and i'm probably one of the only people who looks thru the git history of dist/translations/en-*.json`

matkoniecz:
> i'm probably one of the only people who looks thru the git history of dist/translations/en-*.json`

Do you prefer squashed or with merge commit or rebase + merge?

k-yle:
i don't think it really matters, maybe rebase+merge this one?

PR #2708, opened 2026-08-13
add boat rental types

andrewharvey:
### Description, Motivation & Context

Allow users to select the types of watercraft that may be rented at an `amenity=boat_rental` https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dboat_rental

There is already a StreetComplete quest about this, so it helps to have iD support it too.

StreetComplete Boat Rental Quest

### Links and data

https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dboat_rental

**Relevant tag usage stats:**
* https://taginfo.openstreetmap.org/keys/kayak_rental (800)

### Checklist and Test-Documentation Template

<details><summary>Read on to get your PR merged faster…</summary>

Follow these steps to test your PR yourself and make it a lot easier and faster for maintainers to check and approve it.

**This is how it works:**
1. After you submit your PR, the system will create a preview and comment on your PR:
   > 🍱 Your pull request preview is ready.
   If this is your first contribution to this project, the preview will not happen right away but requires a click from one of the project members. We will do this ASAP.

2. Once the preview is ready, use it to test your changes.

3. Now copy the snippet below into a new comment and fill out the blanks.

4. Now your PR is ready to be reviewed.

```
## Test-Documentation

### Preview links & Sidebar Screenshots

<!-- Use the preview to find examples, select the feature in question and **copy this link here**.
     Find examples of nodes/areas. Find examples with a lot of tags or very few tags. – Whatever helps to test this thoroughly.
     Add relevant **screenshots** of the sidebar of those examples. -->

<!-- FYI: What we will check:
     - Is the [icon](https://github.com/openstreetmap/id-tagging-schema/blob/main/ICONS.md) well chosen.
     - Are the fields well-structured and have good labels.
     - Do the dropdowns (etc.) work well and show helpful data. -->

### Search

<!-- **Test the search** of your preset and share relevant **screenshots** here.
     - Test the preset name as search terms.
     - Also test the preset terms and aliases as search terms (if present). -->

### Info-`i`

<!-- **Test the info-i** for your fields and preset and share relevant **screenshots** here.
     The info needs to help mappers understand the preset and when to use it.
     [Learn more…](https://github.com/openstreetmap/id-tagging-schema/blob/main/CONTRIBUTING.md#info-i)
 -->

### Wording

- [ ] American English
- [ ] `name`, `aliases` (if present) use Title Case
- [ ] `terms` (if present) use lower case, sorted A-Z
[...]

PR #2709, opened 2026-08-13
ban empty strings in the json schema

k-yle:
to prevent https://github.com/openstreetmap/id-tagging-schema/pull/2708#issuecomment-5274641670

empty strings [are falsy in javascript](https://developer.mozilla.org/en-US/docs/Glossary/Falsy), which causes bizarre errors in iD

### DIFF
--- a/schemas/deprecated.json
+++ b/schemas/deprecated.json
@@ -12,13 +12,15 @@
             "old": {
                 "type": "object",
                 "additionalProperties": {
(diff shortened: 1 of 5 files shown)


matkoniecz (approved):

matkoniecz (approved):
I just wonder is there a way to ensure we will not forget to set this property for future ones...

PR #2716, opened 2026-08-17
stop PRs about patch updates to eslint-plugin

matkoniecz:
in general I think we should reduce dependabot spam, and I plan to propose stopping patch-level updates in more of dependencies

### Links and data

**Relevant OSM Wiki links:**
- …

**Relevant tag usage stats:**
> …

### Checklist and Test-Documentation Template

<details><summary>Read on to get your PR merged faster…</summary>

Follow these steps to test your PR yourself and make it a lot easier and faster for maintainers to check and approve it.

**This is how it works:**
1. After you submit your PR, the system will create a preview and comment on your PR:
   > 🍱 Your pull request preview is ready.
   If this is your first contribution to this project, the preview will not happen right away but requires a click from one of the project members. We will do this ASAP.

2. Once the preview is ready, use it to test your changes.

3. Now copy the snippet below into a new comment and fill out the blanks.

4. Now your PR is ready to be reviewed.

```
## Test-Documentation

### Preview links & Sidebar Screenshots

<!-- Use the preview to find examples, select the feature in question and **copy this link here**.
     Find examples of nodes/areas. Find examples with a lot of tags or very few tags. – Whatever helps to test this thoroughly.
     Add relevant **screenshots** of the sidebar of those examples. -->

<!-- FYI: What we will check:
     - Is the [icon](https://github.com/openstreetmap/id-tagging-schema/blob/main/ICONS.md) well chosen.
     - Are the fields well-structured and have good labels.
     - Do the dropdowns (etc.) work well and show helpful data. -->

### Search

<!-- **Test the search** of your preset and share relevant **screenshots** here.
     - Test the preset name as search terms.
     - Also test the preset terms and aliases as search terms (if present). -->

### Info-`i`

<!-- **Test the info-i** for your fields and preset and share relevant **screenshots** here.
     The info needs to help mappers understand the preset and when to use it.
     [Learn more…](https://github.com/openstreetmap/id-tagging-schema/blob/main/CONTRIBUTING.md#info-i)
 -->

### Wording

- [ ] American English
- [ ] `name`, `aliases` (if present) use Title Case
- [ ] `terms` (if present) use lower case, sorted A-Z
<!-- Learn more in https://github.com/openstreetmap/id-tagging-schema/blob/main/GUIDELINES.md#2-design-the-preset -->
```

</details>

PR #2723, opened 2026-08-19
reduce dependabot spam to once a month

matkoniecz:

based on https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#interval

As k-yle in #2716 mentions @dependabot should not be the second most active contributor in this repo ( and as of 2026-08-19 it is https://github.com/openstreetmap/id-tagging-schema/graphs/contributors?from=8%2F16%2F2025 )

### Related issues

see #2716

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,7 @@ updates:
   - package-ecosystem: "npm"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "monthly"
     cooldown:
       semver-minor-days: 7
       semver-patch-days: 7
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,7 @@ updates:
     directory: "/"
     schedule:
       interval: "monthly"
+    open-pull-requests-limit: 100
     cooldown:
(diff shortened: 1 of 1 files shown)


k-yle (approved):
i think this is a good idea. we don't configure [`open-pull-requests-limit`](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#open-pull-requests-limit-), so it defaults to 5. if there are more than 5 PRs, it will wait until the next month for the remaining ones

https://github.com/openstreetmap/id-tagging-schema/pull/2716#pullrequestreview-4968684835 also proposes merging all dependencies into a single PR, which avoids this issue. 

if we don't want to do that, then `open-pull-requests-limit` should probably be set to `999` or something

matkoniecz:
large open-pull-requests-limit should not break anything anyway, so just changed it

matkoniecz:
@k-yle Can you merge this? This is a minor internal config change, so I see no reason to wait usual week.

And I do not want to start next day from reviewing more "hello, this is dependabot - some other project made commit and released new version just for it" PRs.

Even if it results in making some metrics impressive.

PR #1855, opened 2025-12-09
Explicitly inherit `fields`/`moreFields`

FloEdelmann:
### Description, Motivation & Context

See #1813 and https://github.com/ideditor/schema-builder/issues/204. Changes created by script in https://github.com/openstreetmap/id-tagging-schema/pull/1855#issuecomment-3631864589.

<details>
<summary>Changes made by the script</summary>

- `aerialway/cable_car.json` implicitly inherits `moreFields` from `_aerialway.json` (better explicitly reference `{aerialway}`)
- `aerialway/chair_lift.json` implicitly inherits `moreFields` from `_aerialway.json` (better explicitly reference `{aerialway}`)
- `aerialway/drag_lift.json` implicitly inherits `moreFields` from `_aerialway.json` (better explicitly reference `{aerialway}`)
- `aerialway/gondola.json` implicitly inherits `moreFields` from `_aerialway.json` (better explicitly reference `{aerialway}`)
- `aerialway/goods.json` implicitly inherits `moreFields` from `_aerialway.json` (better explicitly reference `{aerialway}`)
- `aerialway/j-bar.json` implicitly inherits `moreFields` from `_aerialway.json` (better explicitly reference `{aerialway}`)
- `aerialway/magic_carpet.json` implicitly inherits `moreFields` from `_aerialway.json` (better explicitly reference `{aerialway}`)
- `aerialway/mixed_lift.json` implicitly inherits `moreFields` from `_aerialway.json` (better explicitly reference `{aerialway}`)
- `aerialway/platter.json` implicitly inherits `moreFields` from `_aerialway.json` (better explicitly reference `{aerialway}`)
- `aerialway/pylon.json` implicitly inherits `moreFields` from `_aerialway.json` (better explicitly reference `{aerialway}`)
- `aerialway/rope_tow.json` implicitly inherits `moreFields` from `_aerialway.json` (better explicitly reference `{aerialway}`)
- `aerialway/t-bar.json` implicitly inherits `moreFields` from `_aerialway.json` (better explicitly reference `{aerialway}`)
- `aerialway/zip_line.json` implicitly inherits `moreFields` from `_aerialway.json` (better explicitly reference `{aerialway}`)
- `amenity/bar/lgbtq.json` implicitly inherits `fields` and `moreFields` from `amenity/bar.json` (better explicitly reference `{amenity/bar}`)
[...]

PR #2792, opened 2026-08-30
Offer the segregated field on shared refuge island presets

KTibow:
### Description & Context

Add `segregated` to [Cycle & Foot Refuge Island](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/highway/cycleway/traffic_island_shared.json) by inheriting from [Cycle & Foot Path](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/highway/cycleway/bicycle_foot.json) instead of [Cycle Path](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/highway/cycleway.json).

(Side effect: since Cycle Path had [Tunnel Name `tunnel:name`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/tunnel/name.json) but Cycle & Foot Path doesn't, Cycle & Foot Refuge Island no longer has Tunnel Name. Tunnel Name doesn't make sense and was never used on a traffic island, so this should be fine. There are no other side effects (verified via [Tagging Schema Browser](https://osmberlin.github.io/tagging-schema-browser/comparison?dataUrl=https://pr-2792--ideditor-presets-preview.netlify.app/dist/).))

### Related issues

No existing issue. Arguably this should've been caught by #1164.

### Links, data, motivation

**Relevant OSM Wiki links:**
- [`Key:segregated`](https://wiki.openstreetmap.org/wiki/Key:segregated) — "This key has no default value and should be tagged on all shared ways", so a data consumer cannot assume a value where it is missing.
- [`Tag:cycleway=traffic_island`](https://wiki.openstreetmap.org/wiki/Tag:cycleway%3Dtraffic_island) — in use, describes splitting a `cycleway=crossing` at the island and moving `crossing:island` handling onto the island way.
- [`Tag:footway=traffic_island`](https://wiki.openstreetmap.org/wiki/Tag:footway%3Dtraffic_island)

**Relevant tag usage stats:**

> [`segregated`](https://taginfo.openstreetmap.org/keys/segregated) is on ~1.87 M objects — 1,532,235 `no` (82.0%) and 336,198 `yes` (18.0%).
>
> [`cycleway=traffic_island`](https://taginfo.openstreetmap.org/tags/cycleway=traffic_island) 9,167 objects · [`path=traffic_island`](https://taginfo.openstreetmap.org/tags/path=traffic_island) 2,817 · [`footway=traffic_island`](https://taginfo.openstreetmap.org/tags/footway=traffic_island) 169,006.
>
> Mappers already tag `segregated` on refuge islands, and they tag it consistently. Of 8,641 `cycleway=traffic_island` ways worldwide, 3,499 (40.5%) already carry `segregated` — 3,016 `no`, 481 `yes`, 2 junk values.

I've also verified a few `segregated`-on-traffic-island cases to make sure that they actually make sense, and they do. It looks like it was just an accident to not include `segregated` originally (it already has references to [Cycle & Foot Path](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/highway/cycleway/bicycle_foot.json)) (relevant history: 874cea8b #1855).

### DIFF
--- a/data/presets/highway/cycleway/traffic_island_shared.json
+++ b/data/presets/highway/cycleway/traffic_island_shared.json
@@ -5,7 +5,7 @@
         "line"
     ],
     "fields": [
-        "{highway/cycleway}"
+        "{highway/cycleway/bicycle_foot}"
     ],
     "moreFields": [
         "{highway/cycleway}"


KTibow:
`segregated` is now available for shared refuge islands ✅
image

matkoniecz:
Have you verified presumably Claude's PR description? 

Can it be maybe shortened?

KTibow:
It's a long justification for a simple change (adding `/bicycle_foot`). The TLDR is that it's valid to include `segregated` on shared traffic islands, and people are already doing this in other editors including StreetComplete frequently and correctly, so iD should support this too.

andrewharvey (approved):
I didn't review the lengthy AI generated description, but going by the diff and screenshot, I agree with this change, a "Cycle & Foot Refuge Island" (a way tagged as `highway=cycleway` + `cycleway=traffic_island` + `bicycle=designated` + `foot=designated`) should still include the `segregated=*` field.

matkoniecz:
I am tempted to reject it on grounds of description being severe case of LLMitis.

Though sadly it actually makes sense, and making own commit doing exactly the same change seems wrong.

On the other hand I now did actual verification step, which was not done by human opening this PR.

matkoniecz:
Maybe I will treat it as unreviewed AI suggestion and open PR with commit with no stated author?

KTibow:
> On the other hand I now did actual verification step, which was not done by human opening this PR.

Ironically the reason it's so long is because I pushed my LLM to try to verify that the change actually makes sense (since "iD doesn't have this right now" is a strong precedent, and it could've been for a good reason, like believing that most shared traffic islands aren't segregated). But I can trim that back if you'd like.

matkoniecz:
yeah, or at least some human-written summary

in this specific case potential copyright LLM-related horrors are not a thing (one of many potential LLM traps, not sure how to do if anything about it)

But it would be very useful to have some simple human summary before - or instead of - LLM wordflow.

This is simply necessary as in many cases people generate similarly looking PRs without taking any effort at reviewing them, up to and including fully autonomously spamming thousands of repositories.

So it is more and more useful to (1) do human review (2) clearly indicate that human did more than copy-pasting of LLM output.

Sorry for being initially classified together with LLM PR spammers but it is quite effort intensive to distinguish LLM output making sense, wordy human output, reviewed LLM output, LLM output full of hallucinations and so on.

KTibow:
I've rewritten the PR description, hope it reads easier! Old version is still in the edit history if needed for any reason.

matkoniecz:
Is it causing any other fields to be added? Or any to be lost except `tunnel/name`? (last one seems fine)

from past look at past LLM it seems to be claiming that just `tunnel/name` will be lost, but it would be worth verifying it

> [Compare View on the Tagging Schema Browser (alpha)](https://osmberlin.github.io/tagging-schema-browser/comparison?dataUrl=https%3A%2F%2Fpr-2792--ideditor-presets-preview.netlify.app%2Fdist%2F) to see the changes of your PR against the main branch.

this should be able to help with checking if you rebase this branch, checking manually also would work

KTibow:
Yeah Tagging Schema Browser shows no other changes:
image

matkoniecz:
how you checked that? for me it shows also bunch of other changes, as some PRs were merged since then (that is why I suggested rebasing)

I guess I can also make disposable rebased PR just to test that if you prefer to not deal with rebasing

KTibow:
I hit "Sync fork" just to make sure and these are really the only changes, as expected given the tiny diff:
image

matkoniecz (approved):

matkoniecz:
oh, merging also works

PR #1921, opened 2026-01-31
Add `post_office:<service>` multiCombo field

FloEdelmann:
### Description, Motivation & Context

To describe which services a post office provides, there are several subkeys: <https://wiki.openstreetmap.org/wiki/Key:post_office#post_office:%3Cservice%3E=*>
This PR adds a multiCombo field that is used in the [`amenity=post_office`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dpost_office) preset to allow adding all approved subkeys.

I added `autoSuggestions: false` and `customValues: false` because the `post_office:*` namespace is also used for other tags following different formats, e.g. `post_office:opening_hours` and `post_office:brand`.

### Related issues

None.

### Links and data

**Relevant OSM Wiki links:** See above.
**Relevant tag usage stats:** <https://taginfo.openstreetmap.org/search?q=post_office:>

### Checklist and Test-Documentation Template

<details><summary>Read on to get your PR merged faster…</summary>

Follow these steps to test your PR yourself and make it a lot easier and faster for maintainers to check and approve it.

**This is how it works:**
1. After you submit your PR, the system will create a preview and comment on your PR:
   > 🍱 Your pull request preview is ready.
   If this is your first contribution to this project, the preview will not happen right away but requires a click from one of the project members. We will do this ASAP.

2. Once the preview is ready, use it to test your changes.

3. Now copy the snippet below into a new comment and fill out the blanks.

4. Now your PR is ready to be reviewed.

```
## Test-Documentation

### Preview links & Sidebar Screenshots

<!-- Use the preview to find examples, select the feature in question and **copy this link here**.
     Find examples of nodes/areas. Find examples with a lot of tags or very few tags. – Whatever helps to test this thoroughly.
     Add relevant **screenshots** of the sidebar of those examples. -->

<!-- FYI: What we will check:
     - Is the [icon](https://github.com/ideditor/schema-builder/blob/main/ICONS.md) well chosen.
     - Are the fields well-structured and have good labels.
     - Do the dropdowns (etc.) work well and show helpful data. -->

### Search

<!-- **Test the search** of your preset and share relevant **screenshots** here.
     - Test the preset name as search terms.
     - Also test the preset terms and aliases as search terms (if present). -->

### Info-`i`

<!-- **Test the info-i** for your fields and preset and share relevant **screenshots** here.
     The info needs to help mappers understand the preset and when to use it.
[...]

PR #2793, opened 2026-08-31
Add additional fields to climbing presets

elnappo:

This should make it easier to add climbing crags & routes to OSM with the correct tags. Tags are based on OSM wiki and my experience mapping climbing crags and alpine routes. I mainly use it in combination with Go Map!! to map climbing crags and routes, as the mobile reception is often to bad to edit it live. These tags are also used by the [OpenClimbing](https://openclimbing.org/) project which is the main motivation for this PR.

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Climbing

**Relevant tag usage stats:**
- https://taginfo.openstreetmap.org/search?q=climbing
- https://taghistory.raifer.tech/#***/climbing/

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>


matkoniecz:
That is quite a lot of tags.

Are these tags established in OSM tagging?

data/fields/climbing/aid.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "climbing:aid",
```
tordans, line 2: We need to have a good info-I help text for this one because it is apparently the right term for this, but still very niche.

data/fields/climbing/dry.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "climbing:dry",
```
tordans, line 2: We need a very good info-I for this one as well. Apparently, this means to use tooling for ice climbing, but on a dry rock.

data/fields/climbing/grade/hueco.json
```diff
@@ -0,0 +1,34 @@
+{
+    "key": "climbing:grade:hueco",
+    "type": "combo",
+    "label": "Grade (Hueco / V-grade)",
```
tordans, line 4: The others have a good name that connects them to their yes/no field, but this one does not connect to data/fields/climbing/boulder.json yet.

data/fields/climbing/mixed.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "climbing:mixed",
+    "type": "check",
+    "label": "Mixed Climbing"
```
tordans, line 4: So this is ice+rock? We need a better name for this. Also are those routes seasonal?

data/fields/climbing/rock.json
```diff
@@ -0,0 +1,22 @@
+{
+    "key": "climbing:rock",
```
tordans: Is this really the schema and good practice to prefix this? Material is a standard field...
Those sub schema are a bit problematic is some cases when they establish not in fact just copy a main tag...

data/presets/climbing/crag.json
```diff
@@ -1,10 +1,25 @@
...
     "moreFields": [
-        "website"
+        "description",
+        "website",
+        "climbing/rock",
+        "climbing/hazard",
+        "climbing/inclination",
+        "covered",
+        "climbing/family",
+        "climbing/beginner"
```
tordans: Not sure about those. They feels very subjective. Like "smoothness", which required years of discussion to become useful and is still hard to tag. Are they critical here? How are the decisions being made?
IMO this is more of a value that belongs into the user feedback category of an app, like the personal star rating and those flags...

data/presets/climbing/route_bottom.json
```diff
@@ -7,17 +7,33 @@
     "fields": [
         "name",
         "climbing/sport",
+        "climbing/boulder",
         "climbing/trad",
-        "climbing/bolts"
+        "climbing/multipitch",
+        "climbing/ice",
+        "climbing/mixed",
+        "climbing/dry",
```
tordans, line 15: When we list them like that users will likely start adding values to each. Do we want that? And if not, can we discourage it...
We could make some separate presets. 
We could try to use different fields that encode "nothing" as "no". 
Or make some conditional?

Also when users fill out all, we will very likely end up with non logical combinations.

tordans (changes requested):

tordans:
Can you please make a list of usage per tag. And of the wiki data item per tag. And of the wiki desc/help/clarity per tag. We will have to look at each to see if this is for the schema now...

This is the first mayor push for those tagging afaik so now is when alls the errors happen that we have to live with forever...

data/fields/climbing/aid.json
elnappo: As a climber it's quite clear in my opinion but yes. Do you have an idea how to do that?

data/fields/climbing/grade/hueco.json
elnappo: You mean more then?
```
"prerequisiteTag": {
        "key": "climbing:boulder",
        "value": "yes"
    },
```

data/fields/climbing/mixed.json
elnappo: Yes, thats the normal term for that, see https://en.wikipedia.org/wiki/Mixed_climbing Some but not all routes are seasonal. Higher in the mountains you can do that all year round.

data/fields/climbing/rock.json
elnappo: That I don't know. I can only say that this tag is documented in the wiki, widely used and interpreted by openclimbing.org

data/presets/climbing/crag.json
elnappo: I can remove them, in climbing it's easy to agree on them in my experience. Currently they are described as:
* The climbing crag/area is safe and suitable for families with children (easy access, safe terrain, space to play).
* The climbing crag/area is suitable for beginners with easy access, safe terrain, good protection and mostly easy routes.

data/presets/climbing/route_bottom.json
elnappo: I see your point but every "bigger" preset has man "no" "-" "yes" of type `check` and I don't see many things taged with multiple "no". I also haven't seen it in climbing related nodes yet. Is there a technical solution to that? Something like `prerequisiteTag `? I'm open to add more logic here.

elnappo:
@matkoniecz your'e absolutely right, climbing has a lot of tags... (see [wiki](https://wiki.openstreetmap.org/wiki/Climbing)) One problem is that climbing has multiple major styles like [lead](https://en.wikipedia.org/wiki/Lead_climbing), [boulder](https://en.wikipedia.org/wiki/Bouldering), [ice](https://en.wikipedia.org/wiki/Ice_climbing), [dry](https://en.wikipedia.org/wiki/Dry-tooling), [trad](https://en.wikipedia.org/wiki/Traditional_climbing), ... even [deep water soloing](https://en.wikipedia.org/wiki/Deep-water_soloing) and each has it's own [grading system](https://en.wikipedia.org/wiki/Grade_(climbing)). Some style also have multiple grading system like bouldering (2) and  sport climbing (>5) which are all commonly used in different regions of the world.

@tordans In my opinion we can't to anything wrong with climbing styles and their grading as they are fundamental for tagging climbs and should not be omited. Here are stats for the other proposed tags:

| Tag                  | Count |
|----------------------|-------|
| climbing:hazard      | 168   |
| climbing:inclination | 1129  |
| climbing:pitches     | 1237  |
| climbing:rock        | 7625  |
|climbing:family_friendly | 75 |
|climbing:beginner_friendly | 6 |

I observed on me that not having the tags in e.g. Go Map!! leads to not use these tags at all because it's to annoying to write them every time while adding/updating multiple crags. Furthermore is feels like a chicken & egg problem establishing tags.

Just tagging @jvaclavik and @zbycz, creator of the first version of this preset and maintainers of OpenClimbing.

matkoniecz:
`climbing:beginner_friendly` is barely used, subjective, https://wiki.openstreetmap.org/wiki/Climbing is not specifying clearly what would be threshold and is basically review in its current state...

at least this should be removed and other values also reviewed

elnappo:
@matkoniecz agreed, it's removed. Thanks for your input!

matkoniecz:
look like `climbing:family_friendly` has similar issue

matkoniecz:
`climbing:hazard` is also having suspiciously low use, with some values used just few times worldwide and potentially quite subjective (though I am getting less sure here)

data/fields/climbing/grade/hueco.json
```diff
@@ -0,0 +1,34 @@
+{
+    "key": "climbing:grade:hueco",
```
matkoniecz, line 2: https://wiki.openstreetmap.org/wiki/Climbing describes it as US-specific

no idea whether it should be appearing only in US, nowhere or actually everywhere

matkoniecz:
> at least this should be removed and other values also reviewed

I am worried that while I can spot say `climbing:family_friendly` also being problematic, I cannot do this for other tagging schemas: please, review also other parts of PR and include only and solely established tagging schemas

elnappo:
I also removed `climbing:family_friendly`. 

> climbing:hazard is also having suspiciously low use, with some values used just few times worldwide and potentially > quite subjective (though I am getting less sure here)

I guess that's a good thing in this case ;) As this is kind of a "safety" feature, I would prefer to make the barrier to add these as low as possible.

data/fields/climbing/grade/hueco.json
elnappo: There are two grading system for bouldering. heuco (mainly us) and france (mainly europe). The rest of the world is somewhere between. So is should be available worldwide.

data/fields/climbing/aid.json
andrewharvey: http://wiki.openstreetmap.org/wiki/Key:climbing:aid is empty, it should contain information about the tag, when it applies, possible tag values and what they mean. The info help text I believe comes from the KeyDescription template on that wiki page.

matkoniecz: > The info help text I believe comes from the KeyDescription template on that wiki page.

sadly it comes from data items

elnappo: Where can I find these?

matkoniecz: these are linked using

screen-2026-09-03-09-36-52

in infoboxes

see https://wiki.openstreetmap.org/wiki/Data_items

elnappo: Do I have to create a wiki page and data item for every tag used here?

tordans: For iD just the Data Items are enough. https://github.com/openstreetmap/id-tagging-schema/blob/main/CONTRIBUTING.md#info-i explains more I hope (or we should improve that)

For the Wiki I think it would be best to add redirect pages that then link to the section of the main article where the sub tags are described, is such a thing exists.

I know all those docs tasks are tedious but I consider them very important whenever we start promoting schema wie an app like the climbing app and this schema. All this friction now will mitigate pain that we will otherwise have down the road.

elnappo:
I created data items for all keys and wiki pages witch redirect to [Climbing](https://wiki.openstreetmap.org/wiki/Climbing) as suggested by @tordans 

Please check if [climbing:hazard](https://wiki.openstreetmap.org/wiki/Item:Q23977) and [climbing:inclination](https://wiki.openstreetmap.org/wiki/Item:Q23978) are correct as they have multiple keys.

I also added a few more details to existing climbing data items. Examples:
* https://wiki.openstreetmap.org/wiki/Item:Q9444
* https://wiki.openstreetmap.org/wiki/Item:Q3038
* https://wiki.openstreetmap.org/wiki/Item:Q23973
* https://wiki.openstreetmap.org/wiki/Item:Q9443

data/fields/climbing/aid.json
elnappo: Added data items and wiki page redirects.

elnappo:
I added `strings` object to allow translations of `manyCombo` fields.

data/presets/climbing/route_bottom.json
elnappo: After taking another look at the shema, only idea I have is to use `defaultCheck` instead of `check`.

tordans:
Apparently **we have a check** that limits the number of fields. Never saw that and don't know why we have it…
```
> @openstreetmap/id-tagging-schema@7.2.0-dev dist
> node scripts/dist.ts
🏗   Validating and building dist files...
11 values in "fields" of "undefined" (climbing/route_bottom). Limit: 10. Please move lower-priority fields to "moreFields".
Error: Process completed with exit code 1.
```

elnappo:
fixed

data/presets/climbing/route_bottom.json
tordans: I think `defaultCheck` https://github.com/openstreetmap/id-tagging-schema/blob/main/SCHEMA.md#checkboxes would be the better choice. It will prevent users from adding `no` just because they see the list. We have this issue in iD again and again. AFAIK one can always manually set `no` in the Tags UI and the field will show that but not via the field.

The only other idea I have is that we try to split the preset into a generic "(unspecified)" preset and more precise Presets for Boulder and "With Rope". Would that even work? The schema does not seem to be made for this kind of split with it's multiple-choice checkbox form.

So maybe it is just what it is…

data/fields/climbing/rock.json
tordans: Looking at some numbers [with Postpass ](https://github.com/woodpeck/postpass#llm) – below – I don't see that this value is in a good place to be added to the schema right now. TBH it looks like a mess in tagging. `rock` and `climbing:rock` are the same IMO. But the fact that an over specific tag `climbing:rock` exists is bad tagging and causes confusion and errors as evident by the data below. I don't see any reason why the climbing data would not just use `rock` instead. (Data consumer can and should use both of course.). In an ideal world we would do a cleanup effort and migrate all easily migrated values to the `rock` schema and create a MR challenge for the rest.

|rock|climbing_rock|count|
|---|---|---|
|"10"|"limestone"|1|
|"18"|"limestone"|1|
|"amphibolite"|`null`|1|
|"andezit"|`null`|1|
|"andezit-agglomerátum"|`null`|8|
|"awash"|`null`|16|
|"basalt"|`null`|19|
|"Basalt"|`null`|34|
|"boulder"|`null`|1|
|"Boulder"|`null`|1|
|"Buntsandstein"|`null`|4|
|"concrete"|`null`|1|
|"conglomerate"|"conglomerate"|26|
|"conglomerate"|"limestone"|1|
|"conglomerate"|`null`|35|
|"Corbin Sandstone"|`null`|1|
|"damfill"|`null`|1|
|"diamictite;dolomicrite"|`null`|1|
|"erratic"|`null`|23|
|"exposed"|`null`|2|
|"fairy_chimney"|`null`|1|
|"foundation"|`null`|1|
|"glacial_pothole"|`null`|1|
|"Gneis"|"gneiss"|1|
|"gneiss"|"gneiss"|16|
|"gneiss"|"limestone"|14|
|"gneiss"|`null`|42|
|"Gneiss"|`null`|2|
|"gneiss ? Roche moutonnée"|`null`|1|
|"granite"|"granite"|116|
|"granite"|"limestone"|173|
|"granite"|`null`|416|
|"Granite"|`null`|1|
|"granit poli"|`null`|1|
|"granit très abrasif"|`null`|1|
|"green"|`null`|1|
|"Gritstone"|`null`|1|
|"guidepost"|`null`|1|
|"hoodoo"|`null`|38|
|"hornfels"|"granite"|1|
|"hornfels"|`null`|1|
|"lava"|`null`|19|
|"limestone"|"limestone"|1145|
|"limestone"|"sandstone"|1|
|"limestone"|`null`|1371|
|"Limestone"|`null`|5|
|"marlstone;sandstone"|`null`|1|
|"mound"|`null`|1|
|"mushroom_rock"|`null`|1|
|"offshorerock"|`null`|1|
|"Phonolite"|`null`|1|
|"pinnacle"|`null`|30|
|"Pískovec"|`null`|1|
|"porfir"|"porfir"|2|
|"porfir"|`null`|5|
|"Porphyr"|`null`|1|
|"porphyry"|"limestone"|2|
|"porphyry"|"porphyry"|2|
|"porphyry"|`null`|4|
|"Q119138890"|`null`|1|
|"quartz"|`null`|1|
|"quartzite"|`null`|2|
|"Quartzite"|"quartzite"|1|
|"rhyolite"|`null`|10|
|"rock tower"|`null`|1|
|"sandstone"|"limestone"|85|
|"sandstone"|"sandstone"|54|
|"sandstone"|`null`|234|
|"Sandstone"|`null`|6|
|"schist"|`null`|1|
|"schiste"|`null`|1|
|"shale"|`null`|1|
|"spire"|`null`|4|
|"submerged"|`null`|10|
|"submerged_dangerous"|`null`|37|
|"Trachyte"|`null`|1|
|"track"|`null`|1|
|"vieux granite"|`null`|1|
|"yes"|`null`|12|
|null|"andesita"|1|
|null|"andesite"|2|
|null|"arkose"|1|
|null|"artificial"|4|
|null|"artificial (cement)"|7|
|null|"artificial (concrete)"|7|
|null|"basalt"|67|
|null|"Calcaire"|1|
|null|"clay"|1|
|null|"concrete"|14|
|null|"conglomerate"|51|
|null|"crag"|1|
|null|"designated"|2|
|null|"dolerite"|1|
|null|"dolomite"|14|
|null|"dunite"|1|
|null|"eclogite;serpentine"|1|
|null|"Gneis"|14|
|null|"gneiss"|149|
|null|"Gneiss"|7|
|null|"granit"|2|
|null|"granite"|681|
|null|"Granite"|11|
|null|"granite;basalt"|2|
|null|"gritstone"|115|
|null|"Gritstone"|1|
|null|"ice"|1|
|null|"limestone"|2097|
|null|"limestone-conglomerat"|1|
|null|"limistone"|1|
|null|"marble"|4|
|null|"migmatite"|1|
|null|"Migmatite"|1|
|null|"monzonit"|1|
|null|"Nagelfluh"|1|
|null|"no"|2|
|null|"northbrae_rhyolite"|3|
|null|"ophiolite"|1|
|null|"paragneiss"|1|
|null|"phonolite"|1|
|null|"Phonolite"|1|
|null|"phyllite"|10|
|null|"pocketted ignimbrite"|7|
|null|"porhyry"|4|
|null|"porphyr"|53|
|null|"Porphyr"|5|
|null|"porphyry"|74|
|null|"psephite"|1|
|null|"puddingstone"|9|
|null|"quartzite"|153|
|null|"Quarz"|1|
|null|"rhyolite"|6|
|null|"Rhyolite"|2|
|null|"rubblestone"|1|
|null|"sandstone"|1608|
|null|"schist"|4|
|null|"serpentine"|6|
|null|"serpentinite"|3|
|null|"shist"|1|
|null|"slate"|23|
|null|"tower"|1|
|null|"trachite"|2|
|null|"trachyandesite"|1|
|null|"verrucano"|1|
|null|"yes"|2|


<details><summary>Postpass Query</summary>
<p>


```
% curl -G 'https://postpass.geofabrik.de/api/interpreter' \
  --data-urlencode 'options[geojson]=false' \
  --data-urlencode "q=
SELECT
  tags->>'rock' AS rock,
  tags->>'climbing:rock' AS climbing_rock,
  COUNT(*) AS count
FROM postpass_pointlinepolygon
WHERE tags ? 'rock'
   OR tags ? 'climbing:rock'
GROUP BY
  tags->>'rock',
  tags->>'climbing:rock'
ORDER BY
  tags->>'rock',
  tags->>'climbing:rock'
"


```

</p>
</details>

elnappo: Yeah that's true, it's a mess... I also only added the most important ons. I guess you mean `material` not `rock` [Key:material ](https://wiki.openstreetmap.org/wiki/Key:material)? I like your idea, I also had the urge to fix it. I can try to open a topic to migrate `climbing:rock` -> `material`.

data/presets/climbing/route_bottom.json
elnappo: Then lets start with that, if we find it to limiting we can change it easily in the future.

tordans:
Just saw this interesting schema detail on https://wiki.openstreetmap.org/wiki/Climbing#Climbing_styles

> yes / no / <amount of routes>

What is `<amount of routes>`? Is it a percent value? Or a meter value? And how do we handle those with the presets we have here?

I really think this should be more clear in the wiki!

It is very unfortunate that the schema was written in this way because there is not good way to represent a "yes"|"no"|"length value" field and I consider it very uncommon in OSM.

I looked into some data and they are quite common unfortunately. So we should document in here a few examples of how those show in iD with this new setup. But it looks like I am doing something wrong because the test URLs that I tried to generate dont show show the fields, eg https://pr-2793--ideditor-presets-preview.netlify.app/id/dist/#locale=en&disable_features=boundaries&map=20.00/37.57672/25.16739&background=Bing&id=n4029150857

|boulder|sport|speed|toprope|trad|multipitch|ice|mixed|dry|aid|deepwater|count|
|---|---|---|---|---|---|---|---|---|---|---|---|
|-|yes|-|-|-|-|-|-|-|-|-|6529|
|-|-|-|-|yes|-|-|-|-|-|-|3532|
|yes|-|-|-|-|-|-|-|-|-|-|3338|
|-|yes|-|-|yes|-|-|-|-|-|-|1060|
|-|(number)|-|-|-|-|-|-|-|-|-|503|
|-|-|-|-|(number)|-|-|-|-|-|-|317|
|-|-|-|-|-|yes|-|-|-|-|-|296|
|-|yes|-|-|-|yes|-|-|-|-|-|267|
|-|-|-|-|-|-|yes|-|-|-|-|230|
|-|yes|-|yes|-|-|-|-|-|-|-|208|
|-|-|-|yes|-|-|-|-|-|-|-|206|
|-|yes|-|-|no|-|-|-|-|-|-|170|
|yes|yes|-|-|-|-|-|-|-|-|-|142|
|-|yes|-|-|-|no|-|-|-|-|-|126|
|-|-|-|-|yes|yes|-|-|-|-|-|120|
|-|no|-|-|yes|-|-|-|-|-|-|115|
|yes|yes|-|yes|-|-|-|-|-|-|-|103|
|-|(number)|-|-|(number)|-|-|-|-|-|-|86|
|yes|no|-|no|-|-|-|-|-|-|-|79|
|-|-|-|yes|yes|-|-|-|-|-|-|74|
|yes|-|-|yes|-|-|-|-|-|-|-|70|
|(number)|-|-|-|-|-|-|-|-|-|-|58|
|no|-|-|-|-|-|-|-|-|-|-|57|
|yes|no|-|-|no|-|-|-|-|-|-|44|
|-|no|-|-|no|-|-|-|-|-|-|43|
|-|yes|-|-|yes|yes|-|-|-|-|-|42|
|-|no|-|-|-|-|-|-|-|-|-|41|
|yes|no|-|-|-|-|-|-|-|-|-|37|
|-|-|-|-|-|(number)|-|-|-|-|-|37|
|-|yes|-|yes|yes|-|-|-|-|-|-|35|
|yes|no|no|no|-|-|-|-|-|-|-|33|
|-|-|-|-|-|-|-|yes|-|-|-|32|
|-|no|-|yes|yes|-|-|-|-|-|-|30|
|-|yes|-|no|-|-|-|-|-|-|-|30|
|no|yes|-|-|-|-|-|-|-|-|-|30|
|-|-|-|-|(number)|-|-|(number)|-|-|-|28|
|-|(number)|-|-|-|(number)|-|-|-|-|-|27|
|yes|-|-|no|-|-|-|-|-|-|-|23|
|-|-|-|-|yes|-|-|yes|-|-|-|19|
|yes|yes|-|-|yes|-|-|-|-|-|-|18|
|yes|yes|-|-|yes|yes|-|-|-|-|-|18|
|-|-|-|-|-|-|-|-|-|-|yes|17|
|-|-|-|-|-|no|-|-|-|-|-|17|
|yes|-|-|-|yes|-|-|-|-|-|-|17|
|yes|yes|yes|yes|-|-|-|-|-|-|-|16|
|-|(number)|-|(number)|(number)|-|-|-|-|-|-|16|
|-|yes|-|-|no|no|-|-|-|-|-|16|
|-|yes|-|yes|-|yes|-|-|-|-|-|16|
|no|yes|-|yes|-|-|-|-|-|-|-|16|
|yes|yes|no|yes|-|-|-|-|-|-|-|15|
|-|yes|-|-|-|-|-|yes|-|-|-|15|
|yes|no|-|no|no|-|-|-|-|-|-|14|
|-|yes|-|-|no|yes|-|-|-|-|-|13|
|yes|-|-|-|-|-|-|-|-|-|yes|12|
|-|-|-|-|-|-|(number)|-|-|-|-|12|
|-|yes|-|-|yes|-|-|yes|-|-|-|11|
|-|(number)|-|-|(number)|-|-|(number)|-|-|-|11|
|-|(number)|-|-|-|yes|-|-|-|-|-|11|
|-|-|-|-|(number)|(number)|-|-|-|-|-|11|
|yes|(number)|-|-|-|-|-|-|-|-|-|11|
|-|yes|-|yes|-|no|-|-|-|-|-|10|
|-|yes|-|yes|no|no|-|-|-|-|-|10|
|no|-|-|-|no|-|-|-|-|-|-|10|
|-|yes|-|-|yes|no|-|-|-|-|-|9|
|-|yes|-|-|yes|-|-|no|-|-|-|9|
|-|yes|-|no|-|no|-|-|-|-|-|9|
|-|(number)|-|-|-|no|-|-|-|-|-|9|
|yes|no|-|no|no|no|no|no|-|-|no|9|
|-|-|-|no|-|-|-|-|-|-|-|9|
|yes|yes|-|yes|-|no|-|-|-|-|-|9|
|-|no|-|-|no|-|yes|no|-|-|-|8|
|-|(number)|-|(number)|-|-|-|-|-|-|-|8|
|-|(number)|-|yes|-|-|-|-|-|-|-|8|
|no|yes|-|no|-|-|-|-|-|-|-|8|
|-|yes|-|yes|no|-|-|-|-|-|-|8|
|no|(number)|-|-|-|-|-|-|-|-|-|7|
|no|yes|-|no|-|no|-|-|-|-|-|7|
|-|yes|-|-|-|-|-|no|-|-|-|7|

Note, some tags are missing in the cols here which why lines like `|-|(number)|-|-|-|-|-|-|-|-|-|503|` look so weird, the other `number` is in a col that is missing. At least according to some spot checks.

data/fields/climbing/aid.json
tordans: I will resolve this here but lets document the PR with our check list so we have a screenshot that show that the data item actually show up.

data/fields/climbing/rock.json
tordans: >I guess you mean material not rock [Key:material ](https://wiki.openstreetmap.org/wiki/Key:material)?

The table above is from…
```
tags->>'rock' AS rock,
tags->>'climbing:rock' AS climbing_rock,
```

Did not look into material on this schema, yet. But assume it will be stone "all the time"?

> I can try to open a topic to migrate `climbing:rock` -> `material`.

My understanding is, that `rock` is a sub group of `material` in this case. But not in the classic sub-group sense of `foo=bar+bar=baz` but more like `material=stone|rock|the like` + `rock=detailed rock value`.

elnappo:
Looks like climbing needs more help from you :) I'm happy to bring your ideas and improvements into the climbing community ;) 

For crags `climbing=crag` the number is used sometimes to show how many routes there are in that crag (location) but there is also `climbing:routes=<int>`... In a perfect world the number of routes is based on how many routes are in the relation. But not all crags are a relation.
For a single route `climbing=route_bottom` only `yes`/`no` is used.

elnappo:
You have an idea why the help text for `climbing:aid` and `climbing:dry` is missing in the test env? There is a data item for them https://wiki.openstreetmap.org/wiki/Item:Q23971 https://wiki.openstreetmap.org/wiki/Item:Q23972

data/fields/climbing/rock.json
elnappo: I can’t find any documentation about `rock` at all. `material` already has all the values we need and they are even documented (limestone, sandstone, granite, …) in the wiki. Wouldn’t that be enought?

`climbing:rock` + `sport` = 7298
`material` + `sport` = 2804
`rock` + `sport` = 3298

As you sad, a mess which should be cleaned up.

tordans: > I can’t find any documentation about `rock` at all. `material` already has all the values we need and they are even documented (limestone, sandstone, granite, …) in the wiki. Wouldn’t that be enought?

Oh, yeah, did not check that, this would be even better. We could add the question to also deprecate `rock` at the same time. A good next step would be to write a tagging voting RFC and discuss this as a draft in the forum for feedback. We can talk more about this in Matrix or Signal if you like. Cannot do that work but share what I know about it.

matkoniecz: I also would be happy to share what I know / link examples to past successes / fails.

I can also run migration itself if there would be a clear agreement. See https://wiki.openstreetmap.org/wiki/Mechanical_Edits/Mateusz_Konieczny_-_bot_account

data/fields/climbing/author.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "author",
+    "type": "text",
+    "label": "Author",
+    "placeholder": "Jane Doe, 1984"
```
matkoniecz, line 5: so, is it supposed to be author or author and year? (presumably, of creation of route?)

from the first look https://taginfo.openstreetmap.org/keys/author#values is fairly suspect, the most popular value being "unknown author", not even in English

matkoniecz: also, https://taginfo.openstreetmap.org/keys/author#overview comes mostly from imports

data/fields/climbing/trad.json
```diff
@@ -1,5 +1,5 @@
 {
     "key": "climbing:trad",
-    "type": "check",
-    "label": "Traditional climbing"
+    "type": "defaultCheck",
```
matkoniecz, line 3: wiki mentions `no` values as valid, so why change to `defaultCheck` ?

data/fields/climbing/sport.json
```diff
@@ -1,5 +1,5 @@
 {
     "key": "climbing:sport",
-    "type": "check",
-    "label": "Sport climbing (bolted climbing)"
+    "type": "defaultCheck",
```
matkoniecz, line 3: wiki mentions `no` values as valid, so why change to `defaultCheck` ?

matkoniecz (changes requested):
climbing:hazard:* tagging is NOT established yet

iD tagging schema is for established tags

I am not saying it is a bad tagging schema or that it should not be used

see in general https://taginfo.openstreetmap.org/search?q=climbing%3Ahazard%3A

or https://taginfo.openstreetmap.org/keys/climbing:hazard:missing_anchor

or https://taginfo.openstreetmap.org/keys/climbing:hazard:unstable_anchor

or https://taginfo.openstreetmap.org/keys/climbing:hazard:animal_nest

or https://taginfo.openstreetmap.org/keys/climbing:hazard:death_fall_zone

or https://taginfo.openstreetmap.org/keys/climbing:hazard:first_bolt_high

or https://taginfo.openstreetmap.org/keys/climbing:hazard:vegetation

or https://taginfo.openstreetmap.org/keys/climbing:hazard:rockfall_zone

----------------------

though tagging seems subjective from looking at https://wiki.openstreetmap.org/wiki/Climbing

----------------

note: I have made some comments as I was pinged for review, but it is not a full review and I feel uncomfortable reviewing this PR as I lack expert knowledge here

data/fields/climbing/rock.json
matkoniecz: I also have some minor cleanup projects, added some values of these tags to them. Maybe some translations and capitalization issues will start disappearing soon.

elnappo: Thanks for your offer! When this PR is done I will work on that and will come back to you :)

data/fields/climbing/author.json
elnappo: Yes, its "creation of route". These information are usually shared together with the route name.

data/fields/climbing/trad.json
elnappo: See @tordans comment https://github.com/openstreetmap/id-tagging-schema/pull/2793#discussion_r3960875426

elnappo:
I assume `climbing:hazard:*` will never be on more then a few hundreds objects. Fortunately, most of the crags don't have any hazards, which is a good thing. It's the minority of crags with issues like that. As this is kind of a "safety" feature, I would prefer to make the barrier to add these as low as possible. Thanks for your review, I think because appeared there because to made the first comment on this PR? 🤷‍♂️

data/fields/climbing/sport.json
tordans: Lets talk about this at https://github.com/openstreetmap/id-tagging-schema/pull/2793#discussion_r3980106091 . I find it makes PRs harder to review when we spread the discussion.

tordans:
@elnappo can you do a round of testing in the spirit of "Read on to get your PR merged faster…" (inside the PR body). Especially the Links to the Preview Version of this PR. This will help to move the discussion into a more concrete "how does it work and look in iD" phase.

---

Quick link for myself: https://pr-2793--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=20.00/47.07694/10.94879&disable_features=boundaries&background=fr.ign.bdortho&id=n13397242762 for https://openclimbing.org/relation/19993518#17.00/47.0769/10.9475

tordans:
> I assume `climbing:hazard:*` will never be on more then a few hundreds objects. Fortunately, most of the crags don't have any hazards, which is a good thing. It's the minority of crags with issues like that. As this is kind of a "safety" feature, I would prefer to make the barrier to add these as low as possible.

In the spirit to get this PR in a mergable shape, I suggest we extract the hazard tag into a follow up. I share the sentiment that the `climbing:hazard` schema is not well thought out yet and that in combination with low usage make leads the to postpone adding it right now. Lets make this a separate thread in a follow up PR. My question is then, which of those values are really for the `hazard` namespace and which just landed there over time and should rather be something else. Especially those values that also have a number as possible value are something that IMO belongs one level up. They describe a fact of the route, not a hazard. What I mean is: Every `6b` is a hazard for some people just as a "first anchor at 3 m" is a hazard for some people. So that should rather be `climbing:first_anchor_height=3m` or something like this…
And all the "anchor loose, broken, missing, rusted" seem out of place in `hazard` as well.
I am open to adding some of those those in the preset, but since we will also boost those values a lot once we do, I think we should give the community some time to first establish them some more before pushing something that is harder to cleanup later…

data/fields/climbing/author.json
tordans: I think https://wiki.openstreetmap.org/wiki/Key%3Aauthor looks OK and there is no strong overlap with other schema and it is the way this is known / names in climbing circles so I think this is ok as is.

(Side note: The [505 "Autoři neznámí"](https://taginfo.openstreetmap.org/tags/author=Auto%C5%99i%20nezn%C3%A1m%C3%AD) could also mean "it is know that this author is unknown or anonymous" in which case the value would make a bit more sense … if it where in english …)

data/fields/climbing/grade/hueco.json
tordans: I was referring to a different label.
Example:
> `"key": "climbing:grade:aid",    "label": "Grade (Aid)",` =>     `"key": "climbing:aid",     "label": "Aid Climbing"`

Both have _"Aid"_.

But
>    `"key": "climbing:grade:hueco",     "label": "Grade (Hueco / V-grade)",` =>     `"key": "climbing:boulder",    "label": "Bouldering"`

Is there a way to get the "Bouldering" in the "Grade (Hueco / V-grade)" string that makes sense?

tordans:
> You have an idea why the help text for `climbing:aid` and `climbing:dry` is missing in the test env? There is a data item for them https://wiki.openstreetmap.org/wiki/Item:Q23971 https://wiki.openstreetmap.org/wiki/Item:Q23972

I can only confirm it does not work =(
https://wiki.openstreetmap.org/w/api.php?action=wbgetentities&sites=wiki&titles=Key%3Aclimbing%3Adry&languages=en%7Cde%7Cen-us&languagefallback=1&origin=*&format=json

Could you check out this thread https://github.com/openstreetmap/id-tagging-schema/pull/1600#discussion_r2637755592
Maybe this step is missing for those items? And should that help, could you check if our docs are mentioning that correctly? 
We had an issue here in Jan about this https://github.com/openstreetmap/id-tagging-schema/pull/1921#discussion_r2749549347 but that either was resolved by that or by just waiting… which in this case did not work, yet.

data/presets/climbing/route_bottom.json
tordans: I looked at an example with the now set `defaultCheck` https://pr-2793--ideditor-presets-preview.netlify.app/id/dist/#locale=en&map=20.00/47.07694/10.94879&disable_features=boundaries&background=fr.ign.bdortho&id=n13397242762
I don't think this is better ATM. I think we should add translated `strings` here
* [Example for `defaultCheck`](https://github.com/openstreetmap/id-tagging-schema/blob/2375a6b/data/fields/crossing_raised.json#L11-L16)
* [Example for `check`](https://github.com/openstreetmap/id-tagging-schema/blob/2375a6b/data/fields/parcel_pickup.json#L5-L11)

And then we could name the `undefined` chase "Unknown or not relevant" to indicate that we don't want "no" values everywhere but that "null" is a valid and "desirable" value.

With this `undefined` "translation" I think we can go back to `check` to make the `no` a propper option again.

data/fields/climbing/trad.json
tordans: Lets keep this thread up there https://github.com/openstreetmap/id-tagging-schema/pull/2793#discussion_r3989551833 so we don't have to jump around too much.

tordans:
Sorry, I am getting lost a bit here due to a lack of practical knowledge on this topic's tagging practice. 

> For crags `climbing=crag` the number is used sometimes to show how many routes there are in that crag (location) but there is also `climbing:routes=<int>`... In a perfect world the number of routes is based on how many routes are in the relation. But not all crags are a relation. For a single route `climbing=route_bottom` only `yes`/`no` is used.

To me this sounds like the wiki need to split up this table to be more clear on that some combinations only use yes|no, and other use number … and what the number means in this situation.

Could you please share the postpass (or similar) results that show that the combination of tags that we use in this PR will all (mostly…) be `yes|no` so we can document and confirm that this PR does the right thing by focussing on a `yes|no`. And ideally with a few examples of the cases you mentioned above for `crag`, `:routes` and so on so we see that this PR works (see also [real world testing request](https://github.com/openstreetmap/id-tagging-schema/pull/2793#issuecomment-5634512346)).

elnappo:
Removed `climbing:hazard:*`.

I'm unable to set the sitelink for the two data items because of https://github.com/openstreetmap/operations/issues/1400#issuecomment-5644167925
> The site ID "enwiki" is unknown. Please use an existing site ID, such as "enwiki".

Regarding `climbing:mixed=yes/no` I was just duplicating what's already in the preset before for `climbing:sport`(was `"type": "defaultCheck"`). 
From taginfo:
`climbing:sport=yes/no` 10217
`climbing:sport=<other>` 803
`climbing:trad=yes/no`6069
`climbing:trad=<other>`583
`climbing:boulder=yes/no`4763
`climbing:boulder=<other>`110

Examples for `climbing=crag` & `climbing:routes=<int>` overpass `nwr["climbing"="crag"]["climbing:routes"]`
* https://www.openstreetmap.org/way/1539119435
* https://www.openstreetmap.org/way/841260415
* https://www.openstreetmap.org/way/1476546144
* https://www.openstreetmap.org/way/1356738993
* https://www.openstreetmap.org/node/292021739

Could/should we just add `"customValues": true,` to that?

matkoniecz (commented):
thanks for changes! At this point I have no useful feedback

matkoniecz (commented):
EDIT: github is githubbing again, presumably after too much of LLM

sorry for duplicate

data/fields/climbing/grade/hueco.json
elnappo: Only the "new" ones are that clear and simple. We have these existing labels in the preset already:
* "Grade (French)"
* "Grade (Saxon)"
* "Grade (UIAA)"
* "Grade (YDS)"

data/presets/climbing/route_bottom.json
zbycz: (hi👋) I would say the message "unknown or not relevant" is a little confusing in this case. A user willing to fill all the data in that form would perhaps try to not have any "unknowns". It is known to them, that it is not eg. ice climbing.

I understand that you probably wanted to express that null value is usually unknown, but could we perhaps say only "not relevant" or "doesn't apply"?

elnappo:
Climbing styles are now `"type": "check"` again.

data/presets/climbing/route_bottom.json
elnappo: @tordans are you fine with that?

PR #2823, opened 2026-09-06
Remove `replacement` from the hidden bus station preset preset

Geo-2695:
See https://github.com/openstreetmap/id-tagging-schema/issues/2409#issuecomment-4966519849

### DIFF
--- a/data/presets/amenity/_bus_station.json
+++ b/data/presets/amenity/_bus_station.json
@@ -17,6 +17,5 @@
         "amenity": "bus_station"
     },
     "name": "{public_transport/station_bus}",
-    "searchable": false,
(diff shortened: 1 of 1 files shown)


Geo-2695:
### Test
BUS

matkoniecz:
Are there any bus stations which are not public transport?

https://www.openstreetmap.org/node/9415412775 was linked in https://github.com/openstreetmap/id-tagging-schema/issues/2409#issuecomment-4966747985 but it does not really look like bus station on the first look (map only for now)

@1ec5 maybe by US standards this large bus stop is bus station?

matkoniecz:
@Geo-2695 Can you give example of any bus stations which is not a public transport?

PR #2828, opened 2026-09-07
fix: preserve preset names in validation errors

### LINKED ISSUE #2827: preset.name is `undefined` in error messages of build process
preset.name is `undefined` in `process.stderr.write('Unknown preset "' + preset.replacement + '" referenced as replacement of preset "' + presetID + '" (' + preset.name + ')\n');`

and many (all?) other places where error message is trying to use it

nightcityblade:

Validation runs after localized preset names are removed from the generated preset objects. Resolve each display name from the translation/reference maps so diagnostics show the real name instead of `undefined`.

### Related issues

Closes #2827

### Links and data

**Relevant OSM Wiki links:**
- N/A

**Relevant tag usage stats:**
> N/A — build tooling only.

N/A — this change only affects build-time error messages and has no preset preview impact.

### DIFF
--- a/scripts/lib/build.ts
+++ b/scripts/lib/build.ts
@@ -130,7 +130,7 @@ async function processData(_options: Partial<Options> | undefined, type: string)
 
   // Additional consistency checks
   validateCategoryPresets(categories, presets);
-  validatePresetFields(presets, fields);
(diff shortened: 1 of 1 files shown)


k-yle (approved):
as far as i can tell this work, thanks

PR #2829, opened 2026-09-07
Run development builds independently of distribution builds

### LINKED ISSUE #2821: buildDev is not writing to _currBuild
I am still trying to fully understand some aspects, but it looks very suspect that _currBuild is used as guard in 

https://github.com/openstreetmap/id-tagging-schema/blob/main/scripts/lib/build.ts#L45

but not written to unlike 

https://github.com/openstreetmap/id-tagging-schema/blob/main/scripts/lib/build.ts#L71

lokendra005:

A `buildDev()` call made while `buildDist()` is pending returns the dist promise without producing its requested output, even when it uses another directory. Remove the dist guard from the synchronous development build path.

### Related issues

Closes #2821

### Links and data

N/A — build tooling only.

- `npm run build && npm run dist && npm test`: 4 tests pass.
- `npm run lint`: passes.
- The added test fails on the original code because the requested development output is missing.

### DIFF
--- a/scripts/lib/__tests__/schema-builder.test.js
+++ b/scripts/lib/__tests__/schema-builder.test.js
@@ -53,6 +53,37 @@ describe('schema-builder', () => {
     expect(fs.existsSync(_workspace + '/dist')).toBe(false);
   });
 
+  it('runs buildDev while a different buildDist is pending', async () => {
+    writeSourceData({
+      'data/presets/natural.json': {
+        tags: { natural: '*' },
+        geometry: ['point'],
+        name: 'Natural Feature'
+      }
+    });
+    await Promise.all([
+      schemaBuilder.buildDist({
+        inDirectory: _workspace + '/data',
+        interimDirectory: _workspace + '/dist-interim',
+        outDirectory: _workspace + '/dist',
+        taginfoProjectInfo: {
+          name: 'IntrepiD',
+          description: 'iD editor, but adventurous.',
+          project_url: 'https://example.com/IntrepiD',
+          contact_name: 'J. Maintainer',
+          contact_email: 'maintainer@example.com'
+        }
+      }),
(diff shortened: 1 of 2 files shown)


scripts/lib/build.ts
k-yle, line 60: this line is also dead code and could be deleted

scripts/lib/__tests__/schema-builder.test.js
```diff
@@ -53,6 +53,37 @@ describe('schema-builder', () => {
     expect(fs.existsSync(_workspace + '/dist')).toBe(false);
   });
 
+  it('runs buildDev while a different buildDist is pending', async () => {
```
k-yle, line 56: thanks for putting effort into writing unit tests

but what's the purpose of a unit test for an impossible situation? 

i think deleting the one line of dead code would be better. this test implies that we have some sort of parralellised build process

k-yle (changes requested):

scripts/lib/__tests__/schema-builder.test.js
matkoniecz: from looking at https://github.com/openstreetmap/id-tagging-schema/pull/2832 and lack of reaction and style of this code it looks like fully automated LLM spam without human overview

matkoniecz (changes requested):
it is clearly LLM generated with minimal or no human overwiev

I am thinking about closing it on this grounds 

if there is human at all here - can you take look at https://github.com/openstreetmap/id-tagging-schema/pull/2832 and finish steps there?

matkoniecz:
see https://github.com/openstreetmap/id-tagging-schema/pull/2832#issuecomment-5661951008

closing as LLM spam generated without human oversight (if human is reading it: please look at https://github.com/openstreetmap/id-tagging-schema/pull/2832 )

PR #2830, opened 2026-09-07
Add optional religion fields to cemetery sectors

### LINKED ISSUE #2801: Add `religion` and `denomination` fields to cemetery sector preset
A large cemetery park is often devoted to whatever the main religion is in the area. Even in such cemeteries, it is possible to find some sectors which are designated for other religions.

It should be fine to have these fields in `fields` since this preset has only 2 `fields` and 0 `moreFields`. Or would that encourage some mappers to tag `religion=christian` on sector in `religion=christian` cemetery`?

lokendra005:

A cemetery section can have a different religion or denomination from the surrounding cemetery. Put both existing fields in **Add field**, as suggested in the issue, so they stay optional.

### Related issues

Closes #2801

### Links and data

The [cemetery-sector wiki page](https://wiki.openstreetmap.org/wiki/Tag:cemetery%3Dsector) lists both tags as useful combinations. [Taginfo](https://taginfo.openstreetmap.org/tags/cemetery=sector) reports 43,589 cemetery sectors as of September 7, 2026.

Build, distribution validation, unit tests and lint pass. The preset's existing default fields and tags are unchanged. Preview check pending deployment.

### DIFF
--- a/data/presets/cemetery/sector.json
+++ b/data/presets/cemetery/sector.json
@@ -4,6 +4,10 @@
         "name",
         "ref_sector"
     ],
+    "moreFields": [
+        "religion",
+        "denomination"
+    ],
     "geometry": [
         "point",
         "area"


matkoniecz (approved):

Dimitar5555:
Did a quick test. When you add religion field and enter some value, the denomination field is not automatically shown. Instead you have to manually add it.

Is there some way to solve this issue?

matkoniecz:
IIRC denomination field was changed to appear only when religion field is present.

So it can be safely added to `fields` while `religion` is in more fields.

> When you add religion field and enter some value, the denomination field is not automatically shown. Instead you have to manually add it.

this is result of both being in `moreFields`

data/presets/cemetery/sector.json
```diff
@@ -4,6 +4,10 @@
         "name",
         "ref_sector"
     ],
+    "moreFields": [
+        "religion",
+        "denomination"
+    ],
```
matkoniecz, line 10:
```suggestion
        "ref_sector",
        "denomination"
    ],
    "moreFields": [
        "religion"
    ],
```

matkoniecz: this should result in `denomination` appearing as soon as `religion` is filled

matkoniecz:
@Dimitar5555 thanks for this very useful review!

Issue #1031, opened 2023-10-15
Support highway=crossing + bicycle=yes combo (aka. cycle crossings)


Zaczero:
### OSM Tag(s)

highway=crossing bicycle=yes

### How would you like this tag to see supported?

As a Preset

### Label

Cycle Crossing

### Link to OSM Wiki page

https://wiki.openstreetmap.org/wiki/Tag:highway=crossing

### Status of the Tag

Unspecified / Unknown

### Usage of the tag

188,266

### Further Information

Combo usage count: https://overpass-turbo.eu/s/1BZA

I would be willing to work on that issue but I need some general guidance/hints. For example, I don't really understand what the "reference" setting means.

yaomtc:
bicycle=designated might be more appropriate for cycleways than bicycle=yes

1ec5:
> For example, I don't really understand what the "reference" setting means.

The `reference` setting determines the data item or wiki page that the ℹ️ button in iD (or the equivalent in other editors) opens up when clicked.

> bicycle=designated might be more appropriate for cycleway crossings than bicycle=yes

`bicycle=designated` would be consistent with the tags for a cyclist and pedestrian crossing when mapped as a way:

https://github.com/openstreetmap/id-tagging-schema/blob/0cd54b5493e2446ff8bcdc029e4a6fac04c35016/data/presets/highway/cycleway/crossing/bicycle_foot.json#L29-L32

matkoniecz:
> bicycle=designated

maybe, but it would require first changing existing tagging practice: https://taginfo.openstreetmap.org/tags/highway=crossing#combinations

mvl22:
We (CycleStreets) would very much like to see this supported - there are currently a lot of cases of dual foot/cycle crossings where `bicycle=yes` is being incorrectly omitted.

matkoniecz:
> Label
>
> Cycle Crossing

What about combined foot + bicycle crossing? Is there tagging distinguishing them? Labeling them as "Cycle Crossing" seems also confusing

1ec5:
A forum discussion [last September](https://community.openstreetmap.org/t/routing-at-highway-crossing-with-foot-bicycle-horse-no/135081) led to consensus that a `highway=crossing` node must not have any negative access tag and ideally should not have any positive access tag either. Instead the crossing way, if any, should have the access tags.

matkoniecz:
so I propose to close it as rejected

mvl22:
[...]

PR #2845, opened 2026-09-09
Crossing With Pedestrian Signals -> Crossing With Traffic Signals

matkoniecz:

less misleading for cyclist-only crossings

### Related issues

encountered while looking at https://github.com/openstreetmap/id-tagging-schema/issues/1031

### Links and data

**Relevant OSM Wiki links:**
- …

**Relevant tag usage stats:**
> …

**This is how it works:**

```
## Test-Documentation

<!-- FYI: What we will check:

### Search

### Info-`i`

 -->

### Wording

- [ ] American English
```

</details>

### DIFF
--- a/data/presets/highway/crossing/traffic_signals.json
+++ b/data/presets/highway/crossing/traffic_signals.json
@@ -23,7 +23,7 @@
         "key": "crossing",
         "value": "traffic_signals"
     },
-    "name": "Crossing With Pedestrian Signals",
+    "name": "Crossing With Traffic Signals",
     "terms": [
         "bicycle crossing (lights)",
         "crosswalk (lights)",


matkoniecz:
pinging @1ec5 for language review because crossings are cursed with horrible complexity

1ec5:
Pedestrian signals and cyclist signals are two kinds of traffic signals. If it just says “traffic signals”, there’s a good chance that a user will think it refers to vehicular traffic signals, the kind that says stop and go at an intersection. I think this is acceptable, since the thing that I’ve proposed as `crossing:signals=shared`[^shared] is normally tagged as `crossing=traffic_signals` anyways. But it is unfortunate that we can’t call them pedestrian signals, since that’s what most people would call the thing we’re trying to refer to.

The unwieldy name of this preset really argues in favor of dropping these overly specific presets in favor of making users fill out `crossing:markings=*`, `crossing:signals=*`, and `crossing:signed=*` fields manually.

[^shared]: If this distinction between pedestrian and cyclist signals is so important, I might consider changing [the proposal](https://wiki.openstreetmap.org/wiki/Proposal:Crossing_signalization) to call for `crossing:signals=go`/`walk`/`cycle` instead.

matkoniecz:
> The unwieldy name of this preset really argues in favor of dropping these overly specific presets in favor of making users fill out crossing:markings=*, crossing:signals=*, and crossing:signed=* fields manually.

definitely, the problem is caused by `crossing=` field support. I really would love to see this awful tagging schema being deprecated.

1ec5:
The proposed change is consistent with the label in the field:

https://github.com/openstreetmap/id-tagging-schema/blob/2acef1a0bf06a6eb25f6aec38ccea8bf6aa946e5/data/fields/crossing.json#L12

Maybe it wouldn’t be such a bad idea to remove the more specific presets in favor of this field and the one for `crossing:markings=*`, similar to the existing workflow for the railroad crossing preset. Then introducing a field for `crossing:signals=*` wouldn’t be as big of a change.

matkoniecz:
The problem here is not technical, the problem is whether iD should just start considering `crossing=` to be deprecated and start using `crossing:signals=*` as a replacement.

Personally, I think it is a much better tagging schema but I am not sure do we have consensus from community on that and I do not want to force tagging just because I prefer it and because it is better (in my opinion).

matkoniecz:
@1ec5 understood your comments as approving (or at least not protesting) merge of this PR

let me know if I am wrong

1ec5:
To clarify, I think the change trades one source of confusion for another. It does solve a problem that the current name is inconsistent with the Type field. To me, a better fix would be to rename that option in the Type field to say “pedestrian signals” too. But I concede that the fallout would be minor, since `crossing=traffic_signals` doesn’t distinguish the three kinds of signals anyways.

To fix both issues, I suggested eliminating this preset and the other global `crossing=*`-based presets. These presets are too specialized, regardless of tags. We would have the same problem even if Crossing With Traffic Signals were based on `crossing:signals=yes` instead. Removing the presets would be purely a UI change, not a tagging change. But it doesn’t have to happen in this PR.

tyrasd:
> To fix both issues, I suggested eliminating this preset and the other global crossing=*-based presets.

I would agree that defragmenting the crossing presets would be the best solution. But the fact that the current documented tagging using the [`crossing`](https://wiki.openstreetmap.org/wiki/Key:crossing) key (which conflates both the _signals_ and _markings_ properties in one tag) makes it pretty difficult or impossible to model correctly otherwise.

It would be great if the tagging proposal for `crossing:signals` would be moved to the next steps after the draft phase. Until that is somehow concluded, I don't see anything actionable that we can do on the id-tagging-schema side of things to improve this further.

tyrasd (approved):

matkoniecz:
> Until that is somehow concluded, I don't see anything actionable

some clear agreement with community that crossing= is in fact terrible and should be replaced also would be enough (sadly it is often not noticed by people how broken these tags are)

1ec5:
@tyrasd To reiterate, I’m not talking about deprecating `crossing=*` here. That’s a separate discussion. My suggestion here is to eliminate the global Marked Crossing, Unmarked Crossing, and Crossing with Traffic Signals presets, forcing the user to set `crossing=*` using the Crossing field.

tordans:
@1ec5 I think we talk about the same thing here…

> My suggestion here is to eliminate the global Marked Crossing, Unmarked Crossing, and Crossing with Traffic Signals presets, forcing the user to set `crossing=*` using the Crossing field.

The issue is, that this change is not possible without loosing relevant(?) part of the UX that we have right now.
I had Grok+Opus summarize our options to get an idea on how this could look like right now:

> # Merging the crossing presets
> 
> ## Summary
> 
> - **Goal:** mappers pick "Crossing", then choose the type (marked, unmarked, traffic signals) inside the form. The form should still offer the fields each type offers today.
> - **Feasible:** showing the right fields per type. The schema's `prerequisiteTag` can hide a field until the type has a given value.
> - **Not feasible with one preset:** a wiki link per type, tags added automatically for a type, and search words like "zebra crossing" that create that type.
> - **[PR 1044](https://github.com/openstreetmap/id-tagging-schema/pull/1044):** hides the type presets from search but keeps them for existing data. It also removes the type dropdown. Its branch predates the current template layout, so it would be rewritten, not merged.
> 
> Scope: node presets in [`data/presets/highway/crossing/`](data/presets/highway/crossing/) and [`data/presets/highway/crossing.json`](data/presets/highway/crossing.json). The crossing way presets (`footway`, `cycleway`, `path`) repeat the same split and are out of scope.
> 
> ## Today
> 
> Search offers four crossing node presets.
> 
> - **Crossing** ([`crossing.json`](data/presets/highway/crossing.json)) sets only `highway=crossing`. Its form shows the type dropdown, markings, island, raised, and tactile paving.
> - **Marked Crossing** ([`uncontrolled.json`](data/presets/highway/crossing/uncontrolled.json)) sets `crossing=uncontrolled`. Its markings field defaults to `crossing:markings=yes`.
> - **Unmarked Crossing** ([`unmarked.json`](data/presets/highway/crossing/unmarked.json)) sets `crossing=unmarked` and `crossing:markings=no`. The form has no markings field.
> - **Crossing With Traffic Signals** ([`traffic_signals.json`](data/presets/highway/crossing/traffic_signals.json)) sets `crossing=traffic_signals`. It adds button, sound, and vibration fields. Arrow, countdown, and minimap are optional fields. Markings have no default.
> 
> Three more presets are hidden from search. They are only used when existing data already has the tag.
> 
> - `crossing=marked` ([`_marked.json`](data/presets/highway/crossing/_marked.json)) is shown as "Marked Crossing". It lists `crossing:markings=yes` as an added tag.
> - `crossing=zebra` ([`_zebra.json`](data/presets/highway/crossing/_zebra.json)) is shown as "Marked Crossing". It adds no markings tag.
> - `crossing=informal` ([`_informal.json`](data/presets/highway/crossing/_informal.json)) matches even without `highway=crossing`. It shows only the type field. The other crossing fields are optional.
> 
> Each type preset links its info button to its own wiki page, for example `Tag:crossing=traffic_signals`.
> 
> The type dropdown ([`fields/crossing.json`](data/fields/crossing.json)) lists `uncontrolled`, `traffic_signals`, and `unmarked`. The other three values can still appear as taginfo suggestions.
> 
> Fields shared by all types except informal: island, raised, tactile paving, and the optional kerb and flashing lights. Country-specific marking lists (DE-AT-CH, PL, BG) come from [`@templates/crossing/markings`](data/presets/@templates/crossing/markings.json).
> 
> ## What one preset can do
> 
> Keep only `crossing.json`, with the type dropdown first. Use `prerequisiteTag` on the fields that depend on the type:
> 
> - Signal fields (button, sound, vibration, and the optional arrow, countdown, minimap) only when `crossing=traffic_signals`.
> - Markings fields for every type except `unmarked` and `informal`.
> 
> The dropdown can also show help text and an icon per type (`strings.options` descriptions and `icons`). Hidden fields drop out; the rest keep the order in the preset. Country-specific marking lists keep working.
> 
> ## What one preset loses
> 
> These features belong to a preset in the schema. A dropdown value cannot carry them.
> 
> - **Wiki link per type.** A preset has a single `reference`. The info button stays on `Tag:highway=crossing` whatever type is chosen. Dropdown options have help text but no link.

Site note: We still have the wiki links on the field, just the preset link becomes less precise.

> - **Tags added for a type.** Added tags (`addTags`) and field defaults apply when a preset is chosen. They do not apply when the dropdown changes. Lost:
>   - `crossing:markings=no` for unmarked
>   - `crossing:markings=yes` for marked
>   - the markings default `yes` for uncontrolled
> 
>   The validator hint for a missing added tag disappears too.
> - **Search that sets the type.** "pedestrian traffic lights" and "zebra crossing" can stay as search terms on Crossing. They then open Crossing without a type.
> - **Name and icon per type.** The header always reads "Crossing". Type icons only appear inside the dropdown.
> - **Informal crossings without `highway=crossing`.** They no longer match the merged preset.
> - **Hiding tags that are already there.** `prerequisiteTag` never hides a field whose tag is set. The unmarked preset hides markings today. A merged preset would show an existing `crossing:markings` on an unmarked crossing.
> - **Type-specific form after a change.** Today, changing the type switches to another preset with other fields and another wiki link. In one preset only the conditional fields change. Signal tags stay on the object after switching away from traffic signals.
> 
> Still editable in the merged preset: type, markings (including country lists), island, raised, tactile paving, kerb, flashing lights, and signal fields for traffic-signal crossings.
> 
> To confirm in iD before building option 2: whether a field default runs once the field's prerequisite becomes true. This plan assumes it does not.
> 
> ## PR 1044
> 
> [PR 1044](https://github.com/openstreetmap/id-tagging-schema/pull/1044) is a closed draft from 2023. It predates the template refactor in [#1201](https://github.com/openstreetmap/id-tagging-schema/pull/1201).
> 
> What it changes:
> 
> - Marked, Unmarked, and Traffic Signals crossings become hidden from search, and their search terms are deleted. The presets remain, so existing objects keep their name, fields, wiki link, and validator hints.
> - Crossing loses the type dropdown and gains a `crossing:signals` checkbox. That field does not exist in the current files; the follow-up [#1203](https://github.com/openstreetmap/id-tagging-schema/pull/1203) was closed.
> - Search keeps Crossing, Pedestrian Crossing, Cycle Crossing, and Cycle & Foot Crossing.
> 
> Why the branch cannot be reused: it edits files that now use `@templates`, moves the `bicycle_foot` presets, and adds a `path=crossing` preset that already exists as [`path/_crossing.json`](data/presets/highway/path/_crossing.json).
> 
> How it fits the goal: it shortens the preset list. It does not let mappers choose the type inside the form, since the dropdown is removed. New crossings get no `crossing=*` tag.
> 
> ## Options
> 
> **Option 1: hide type presets from search (PR 1044 approach, rebuilt).**
> - Set `searchable: false` on `uncontrolled`, `unmarked`, and `traffic_signals`.
> - Move their search terms to `crossing.json`.
> - Keep the type dropdown on Crossing. Choosing a type makes iD switch to the hidden preset, with its fields, name, and wiki link. Its added tags and markings default do not run on that switch.
> - Pro: small change; wiki links, names, and validator hints stay for every type.
> - Con: several preset files remain, and the form changes when the type changes.
> 
> **Option 2: one preset.**
> - Delete the six type presets.
> - Add `prerequisiteTag` to the type-dependent fields.
> - Pro: one form, one file to maintain.
> - Con: every loss in "What one preset loses" applies.
> 
> Neither option lets a search result set the type, apply the markings default, and link the right wiki page in one step. Only a searchable preset per type does that, which is today's setup.

matkoniecz:
> To reiterate, I’m not talking about deprecating crossing=* here. That’s a separate discussion. My suggestion here is to eliminate the global Marked Crossing, Unmarked Crossing, and Crossing with Traffic Signals presets, forcing the user to set crossing=* using the Crossing field.

that sounds like soft deprecation to me

PR #2847, opened 2026-09-10
Fix schema to allow empty strings as roles

matkoniecz:
contrary to what the [documentation](https://github.com/openstreetmap/id-tagging-schema/blob/main/SCHEMA.md#relation) says

> [role]. An empty string is allowed.

the schema defined a minimum length of 1 for these strings, which was probably not intentional.

### DIFF
--- a/schemas/preset.json
+++ b/schemas/preset.json
@@ -198,7 +198,6 @@
                     "type": "object",
                     "additionalProperties": {
                         "type": "string",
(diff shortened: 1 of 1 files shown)


tyrasd (approved):
Yes, this was probably left over from an earlier iteration of the schema where role labels were defined inline in the members array, I'd assume. Thanks for spotting!

PR #2848, opened 2026-09-10
Fix examples for `matchTags` in schema for relation members

matkoniecz:
as values are supposed to be strings

maybe it should be caught by build checks? maybe we have more similar cases?


tyrasd (approved):

tyrasd:
> maybe we have more similar cases?

AFAICS this is the only `exampes` we include in the schemas, so we should be fine for now. It's a bit 

> maybe it should be caught by build checks?

I didn't find an easy way to do this out of the box. As these bugs will be easy to spot once we're actually using a new feature of the schema, I think it's not necessary to spend too much time on this.

PR #2851, opened 2026-09-10
Enforce string pattern of `stringsCrossReference` property in json-schema

matkoniecz:


matkoniecz:
Compare failures on https://github.com/openstreetmap/id-tagging-schema/pull/2854 and https://github.com/openstreetmap/id-tagging-schema/pull/2855

One is quite confusing, another one works much better

tyrasd (approved):

schemas/field.json
```diff
@@ -214,7 +214,8 @@
         "stringsCrossReference": {
             "description": "A field can reference strings of another by using that field's identifier contained in brackets, like {field}.",
             "type": "string",
-            "minLength": 1
+            "minLength": 1,
```
tyrasd:
```suggestion
            "minLength": 3,
```

as these will always have two brackets + a non-empty field id

PR #2853, opened 2026-09-10
Remove duplicate links to code of conduct / privacy policy

matkoniecz:
leftover from time when it was a a separate project

now it is a duplicate of https://github.com/openstreetmap/id-tagging-schema/blob/main/CONTRIBUTING.md


tyrasd (approved):
True, not needed anymore since the merging of the two repos. Good catch!

PR #2859, opened 2026-09-10
locationSetCrossReference can reference both fields and presets

matkoniecz:

current documentation was omitting possibility it can reference also fields

SCHEMA.md docs are already covering that

and yes, it can really reference both - in presets and in fields. See

```

    // fields can reference the locationSet from other fields or presets
    if (field.locationSetCrossReference) {
      const [type, ...foreignId] = field.locationSetCrossReference
        .slice(1, -1)
        .split('/');

      const referenced = (
        type === 'presets' ? presets : type === 'fields' ? fields : {}
      )[foreignId.join('/')];
```

in https://github.com/openstreetmap/id-tagging-schema/blob/main/scripts/lib/references.ts#L151C1-L160C30

and

```
      const referenced = (
        type === 'presets' ? presets : type === 'fields' ? fields : {}
      )[foreignId.join('/')];
```

in https://github.com/openstreetmap/id-tagging-schema/blob/main/scripts/lib/references.ts#L101C1-L103C30

### DIFF
--- a/schemas/field.json
+++ b/schemas/field.json
@@ -409,7 +409,7 @@
             "type": "string",
             "minLength": 1,
             "pattern": "^\\{.+\\}$",
-            "description": "An string referencing another preset or field which has a locationSet"
+            "description": "A string referencing another preset or field which has a locationSet"
         },
         "urlFormat": {
             "description": "Permalink URL for `identifier` fields. Must contain a {value} placeholder",
--- a/schemas/field.json
+++ b/schemas/field.json
@@ -409,7 +409,7 @@
             "type": "string",
             "minLength": 1,
             "pattern": "^\\{.+\\}$",
-            "description": "An string referencing another preset which has a locationSet"
+            "description": "An string referencing another preset or field which has a locationSet"
         },
         "urlFormat": {
             "description": "Permalink URL for `identifier` fields. Must contain a {value} placeholder",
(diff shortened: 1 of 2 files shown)


schemas/field.json
```diff
@@ -409,7 +409,7 @@
             "type": "string",
             "minLength": 1,
             "pattern": "^\\{.+\\}$",
-            "description": "An string referencing another preset which has a locationSet"
+            "description": "An string referencing another preset or field which has a locationSet"
```
tyrasd:
```suggestion
            "description": "A string referencing another preset or field which has a locationSet"
```

tyrasd (commented):
maybe we should also adjust the `pattern` accordingly?

```
"pattern": "^\\{(presets|fields)\\/.+\\}$",
```

schemas/field.json
matkoniecz: ironically it was changed already upstream

k-yle (approved):
current changes LGTM, if you're planning to apply https://github.com/openstreetmap/id-tagging-schema/pull/2859#pullrequestreview-5178525751 then it should be done to [all 3 cases where this syntax is allowed](https://grep.app/search?f.repo=openstreetmap%2Fschema-builder&q=%5Btype%2C+...foreignId%5D).

matkoniecz:
probably will do it as a separate PR to avoid it growing and then getting stuck?

yell at me if I chunk PRs/issues into too small pieces but I prefer 150 separate issues/PR/commits over some of them merging multiple distinct things and getting blocked over one of them

but at times I overdid it, even according to myself

Issue #2673, opened 2026-08-06
Exploration: Extract shared tag interpretation data from iD into id-tagging-schema


tordans:
After we talked about https://github.com/openstreetmap/id-tagging-schema/issues/2623 in the [Community Chat](https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2026#Community_Chat_August_5,_12:00_-_13:00_(UTC)) I wondered what other parts we might look into to extract.

Here is the LLM list of those topics as reference.
I don't think we need to do all that now and we can close this issue soon. But I wanted to document it as a reference for further discussions.

> # Tag-based config in iD → extraction candidates for id-tagging-schema
> 
> Most of the extractable surface sits in `modules/osm/tags.ts`. Presets, fields, `deprecated.json`, and `discarded.json` already live in id-tagging-schema. Field *types* are declared there; the *behavior* below still lives in iD.
> 
> Suitability scale: **High** / **Medium** / **Low** / **Keep in app** — driven by whether the data is list/map-shaped (JSON-friendly) vs regex/algorithm-heavy, and whether other OSM tools need the same interpretation.
> 
> ---
> 
> ## Already extracted (baseline)
> 
> | Topic | In schema | Still in iD |
> |---|---|---|
> | Presets, fields, categories, defaults | `data/presets`, `fields`, … | Matching engine (`presets/index.js`) |
> | Deprecated / discarded tags | `deprecated.json`, `discarded.json` | `osm/deprecated.ts`, `actions/discard_tags.ts` |
> | Geometry allowed on presets | `geometry: []` on presets | Area inference algorithm |
> | Field types (`onewayCheck`, `access`, `directionalCombo`, …) | Field JSON | UI widgets under `modules/ui/fields/` |
> 
> ---
> 
> ## Strong extraction candidates
> 
> ### 1. Closed-way / area interpretation
> **Where:** `modules/osm/tags.ts` (`osmLifecyclePrefixes`, `osmAreaKeysExceptions`, `osmTagSuggestingArea`); `presets/index.js` (`areaKeys()`); `way.ts` (`isArea`); `validations/mismatched_geometry.js`
> 
> **Shape:** Nested key→value maps + small prefix list. AreaKeys themselves are *derived* from presets at runtime.
> 
[...]

Issue #2623, opened 2026-07-30
Export lifecycle prefixes (and other data for interpreting closed ways / areas)


brawer:
Currently, the iD editor has a set of [osmLifecyclePrefixes](https://github.com/openstreetmap/iD/blob/3940995296b4f767451dcba2cd25edbebcd237f4/modules/osm/tags.ts#L31-L44) in its codebase. Could this be exported in id-tagging-schema, so that other projects would interpret OSM tags in the same way as iD?

Context: I'm working on a Rust pipeline that needs to figure out whether the geometry of a closed way is an area or a polyline. Therefore, I'm looking for data to replicate the logic of [osmTagSuggestingArea()](https://github.com/openstreetmap/iD/blob/3940995296b4f767451dcba2cd25edbebcd237f4/modules/osm/tags.ts#L31-L122) in my Rust code. At the moment, the script to generate my Rust function has to extract `osmLifecyclePrefixes` and `osmAreaKeysExceptions` from the TypeScript source code of the iD editor, which is a little fragile. I’d rather read this data in a structured way from the id-tagging-schema project. At some point, there existed a decicated project for interpreting area keys at https://github.com/ideditor/id-area-keys, but its last release was in December 2023, so this seems unmaintained.

tordans:
IMO we should start a new or rekindle the an existing package to make this logic shareable. I did some research a while back and we talked about this in an iD Community Chat a while back. There are a few projects that need this info – from the top of my head its OSMCha, iD, osm2-geojson-ultra. 
Would you be up to consolidate this info in a new/old package? We could discuss it on the next iD Community Chat for example https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2026#Community_Chat_August_5,_12:00_-_13:00_(UTC)

brawer:
For what it’s worth, for now I’m taking the data with a combination of reading id-tagging-schema, plus some regular expressions over the iD sources to extract its list of lifecycle prefixes. However, I don’t really like [the generated code](https://github.com/alltheplaces/osm-diffs/blob/main/src/pipeline/osm/id_tagging_schema.rs) or [my Python script that produces it](https://github.com/alltheplaces/osm-diffs/blob/main/scripts/generate_id_tagging_schema.py) because I’m not sure if I’m doing the right thing.

[...]

PR #2860, opened 2026-09-10
Fix example and documentation of `geometry` property in SCHEMA.md

matkoniecz:
relation was not mentioned as a valid geometry enum value

area was not mentioning as applicable to multipolygons

line was incorrectly claimed (implied?) to be applicable to not closed ways only

usage example was using wrong syntax

### DIFF
--- a/SCHEMA.md
+++ b/SCHEMA.md
@@ -50,7 +50,7 @@ A basic preset is of the form:
         "shop": "farm"
     },
     // The geometry types for which this preset is valid.
-    // options are point, area, line, and vertex.
+    // options are point, area, line, vertex and relation.
     // vertices are points that are parts of lines, like the nodes in a road
     // lines are unclosed ways, and areas are closed ways
     "geometry": [
@@ -454,7 +454,9 @@ associated with building features (but only if drawn as a closed area).
     "key": "building",
     "type": "combo",
     "default": "yes",
-    "geometry": "area",
+    "geometry": [
+        "area"
+    ],
     "label": "Building"
 }
 ```
--- a/SCHEMA.md
+++ b/SCHEMA.md
@@ -52,7 +52,10 @@ A basic preset is of the form:
     // The geometry types for which this preset is valid.
     // options are point, area, line, vertex and relation.
     // vertices are points that are parts of lines, like the nodes in a road
-    // lines are unclosed ways, and areas are closed ways
+    // lines are ways, except closed ones with tags indicating areas
+    // areas can be closed ways if tagged with something indicating them to be an area
+    // areas can be also be multipolygon relations
+    // relation type is for remaining relations
     "geometry": [
         "point", "area"
     ]
--- a/SCHEMA.md
+++ b/SCHEMA.md
@@ -50,12 +50,15 @@ A basic preset is of the form:
         "shop": "farm"
     },
     // The geometry types for which this preset is valid.
-    // options are point, area, line, vertex and relation.
-    // vertices are points that are parts of lines, like the nodes in a road
-    // lines are ways, except closed ones with tags indicating areas
-    // areas can be closed ways if tagged with something indicating them to be an area
-    // areas can be also be multipolygon relations
-    // relation type is for remaining relations
+    // Options are point, vertex, area, line, and relation.
+    // * `vertex` are points that are parts of lines, like the nodes in a road
+    // * `point` are all other nodes like POIs or place nodes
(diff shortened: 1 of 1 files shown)


SCHEMA.md
```diff
@@ -50,9 +50,12 @@ A basic preset is of the form:
         "shop": "farm"
     },
     // The geometry types for which this preset is valid.
-    // options are point, area, line, and vertex.
+    // options are point, area, line, vertex and relation.
     // vertices are points that are parts of lines, like the nodes in a road
-    // lines are unclosed ways, and areas are closed ways
+    // lines are ways, except closed ones with tags indicating areas
+    // areas can be closed ways if tagged with something indicating them to be an area
```
matkoniecz: should we link https://wiki.openstreetmap.org/wiki/Area#Tags_implying_area_status ?

tyrasd: good idea 👍

SCHEMA.md
```diff
@@ -454,7 +457,9 @@ associated with building features (but only if drawn as a closed area).
     "key": "building",
     "type": "combo",
     "default": "yes",
-    "geometry": "area",
```
tyrasd, line 457: good catch!

tyrasd (approved):

SCHEMA.md
tyrasd: PS: FWIW, at least for iD, the old way to define `area` was not actually wrong per se, as iD uses the preset's `geometry` type to define what it considers an area (if it's a closed way). So it was a kind of a self-fulfilling condition, at least for iD.

Anyway, I think we should be as implementation-independent here, so the new definitions are much better.

tyrasd:
let me know what you think about f0ef2c4f90b0a9f245966d4157bcbd92060690e4

SCHEMA.md
matkoniecz: I think that old one was outright wrong, as for example roundabout may have `highway=` closed way without it becoming an area

but I missed this self-fulfilling condition and implication of iD using presets to decide what is an area, and maybe it can be fir somehow

matkoniecz:
I like that change

not entirely sure how to balance between giving enough info and not doing lecture in comments, but I think that it is now good enough

SCHEMA.md
```diff
@@ -50,9 +50,15 @@ A basic preset is of the form:
...
     // The geometry types for which this preset is valid.
-    // options are point, area, line, and vertex.
-    // vertices are points that are parts of lines, like the nodes in a road
-    // lines are unclosed ways, and areas are closed ways
+    // Options are point, vertex, area, line, and relation.
+    // * `vertex` are points that are parts of lines, like the nodes in a road
+    // * `point` are all other nodes like POIs or place nodes
+    // * `line` are ways, except closed ones with tags indicating an area
+    // * `area` are closed ways if tagged with a tag indicating them to be an area
+    //   see https://wiki.openstreetmap.org/wiki/Area#Tags_implying_area_status
```
tordans, line 58:
```suggestion
```

I find it confusing to link to that page given that we have our own rules on what an area is. Once https://github.com/openstreetmap/id-tagging-schema/issues/2623#issuecomment-5200993246 was done, we could link to that file, though…

Ping https://github.com/openstreetmap/id-tagging-schema/issues/2673

tordans (approved):

SCHEMA.md
matkoniecz: on the other hand how iD rules work is a bit scattered and that page has link to them, and they can be now updated in one place...

tordans: I don't think 
>iD - sets them based on iD tagging schema info and exception lists ( [1][2][3] )

is really useful for someone reading this here and since we are in the schema, the other apps and their handling is not relevant.

But if you want we can keep and merge…

matkoniecz: I am fine also with it being kicked out

matkoniecz: it was definitely not reason for that PR

SCHEMA.md
matkoniecz: I admit it was found by LLM (downloaded this file and schema definition, posted it into LLM with instruction "find bugs")

matkoniecz: (all PRs etc were made by me after review with clanker not allowed to produce any part of them. I ignored or left for later very large part of LLM complaints that were false/unimportant/confused/unsolvable etc)

(and bunch of changes, like changes to comments above were result of me spotting nearby issues)

tyrasd (approved):

PR #2861, opened 2026-09-10
Define default value of `increment` property in field schema as 1

matkoniecz:
see documentation in [schema.md](https://github.com/openstreetmap/id-tagging-schema/blob/main/SCHEMA.md#increment).

> For number & integer fields, the amount the stepper control increases or decreases the value. The default is 1.

### DIFF
--- a/schemas/field.json
+++ b/schemas/field.json
@@ -242,6 +242,7 @@
         "increment": {
             "description": "The amount the stepper control should add or subtract (number fields only)",
             "minimum": 1,
+            "default": 1,
             "type": "integer"
         },
         "prerequisiteTag": {


matkoniecz:
alternative: change docs to "iD defaults to 1" but that would be silly, right

(is this property even used by schema? is it read by iD? TODO: check)

tyrasd (approved):
looks good.

>  is it read by iD? 

yes, try it with `direction.json` which has `"increment": 5`

matkoniecz:
this change should make https://github.com/openstreetmap/iD/blob/350ed0d2da4b317dd98c77ddf977ccb2b66183eb/modules/presets/field.ts#L61 unnecessary but I guess it can stay there

tyrasd:
The default value is not written to the `dist/fields.json` output as far as I can see. It's just the default value the code editors will pre-fill the property with. So the code on client-side is still needed.

PR #2863, opened 2026-09-11
Clarify that fields can stand for `fields` exclusively or also for `moreFields`

matkoniecz:
better comments in references.ts file, as I am trying to understand what it is doing (again)

Currently comments are using field sometimes to refer to [`fields`](https://github.com/openstreetmap/id-tagging-schema/blob/main/SCHEMA.md#fieldsmorefields) (rather than `moreFields`) - and sometimes to both `fields` and `moreFields` fields

Which is quite confusing

this attempts to reduce confusion

### DIFF
--- a/scripts/lib/references.ts
+++ b/scripts/lib/references.ts
@@ -70,7 +70,7 @@ export function dereferenceUntranslatedContent(presets: AllPresets, fields: AllF
             return true;
           }
 
-          // replace the reference with what parent defined. decrement i to reprocess this array index.
+          // replace the reference with what the referenced preset defined. decrement i to reprocess this array index.
           // this is necessary as it can also be a reference
           preset[prop].splice(i--, 1, ...referencedPreset[prop].filter(shouldInherit));
         }
--- a/scripts/lib/references.ts
+++ b/scripts/lib/references.ts
@@ -27,8 +27,8 @@ export function dereferenceUntranslatedContent(presets: AllPresets, fields: AllF
             );
           }
 
-          // preset (A) references the fields / moreFields of preset (B), but (B) has them not defined
-          // We silently and intentionally skip this, as it allows presets
+          // preset (A) references the fields / moreFields of preset (B), but (B) does not
+          // define them. We silently and intentionally skip this, as it allows presets
           // to specify fields and moreFields inheritance even while parent has no such field yet.
           // Otherwise defining for example new moreFields would require checking all children presets
           // whether inheritance should be added there.
--- a/scripts/lib/references.ts
+++ b/scripts/lib/references.ts
@@ -27,8 +27,8 @@ export function dereferenceUntranslatedContent(presets: AllPresets, fields: AllF
             );
           }
 
-          // preset (A) references the fields of preset (B), but (B) has no
-          // fields. We silently and intentionally skip this, as it allows presets
+          // preset (A) references the fields / moreFields of preset (B), but (B) has them not defined
+          // We silently and intentionally skip this, as it allows presets
           // to specify fields and moreFields inheritance even while parent has no such field yet.
           // Otherwise defining for example new moreFields would require checking all children presets
           // whether inheritance should be added there.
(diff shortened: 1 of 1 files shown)


scripts/lib/references.ts
matkoniecz, line 49: BTW, what is happening here? I guess I should step through with a debugger...

matkoniecz, line 52: OK, I understand this one - but maybe it still deserves comment?

matkoniecz, line 66: maybe I misread it, but this would stop inheritance of entry in `fields` if preset defined it in `moreFields`, right?

I want to clarify "preset already has a field for this field" comment, then new comment would be something like 

> preset already has defined this field in fields or moreFields, this allows to explicitly define field in moreFields and stop inheritance of it in fields

is it any more clear?

matkoniecz: If I read it correctly it causes fields with types structureRadio, access, restrictions, manyCombo  to be always inherited, without further checking

I guess that structureRadio and access has exclusivity on defining this tags, so it will be never skipped by preset also defining those? Or in other way, even if preset defines those we still want to have structureRadio interface

restrictions does not even operate on tags

not sure yet why manyCombo is affected here, especially as it gets special rule later?

tyrasd: The comment for this the one is the same as the one for the whole function, isn't it?

```js
// Skip fields and moreFields for the keys which define the preset.
```

matkoniecz: yes - I though about moving it or pointing out that is where main reason for function happens

tyrasd: yeah, we only want to filter out "simple" fields, which really are redundantly defining the tags of the preset again. E.g. for Driveway preset, we don't want to show the _Type_ field of the Service Road preset. But for a Pizza Restaurant, we still want to show the Cuisine field, as it will allow to add specify additional values/properties of the POI.

The `if (!field.key) return true;` is basically a shorthand to filter out all of those more complex field types, which we never consider in this method anyway. I think it's a bit of a left-over of how this method was implemented over time and became more and more complex (see e.g. https://github.com/openstreetmap/iD/commit/63caa02ceac3afb5381f54afab2e4dc7e10910d8 for an early version of this code).

In principle, it would probably be more easy to read the code if the logic was more explicitly looking at an allow-list of field types. I would guess that writing it as `if (field.type !== 'typeCombo') return true;` would probably cover the majority of cases already..?

tyrasd: I would actually leave this behaviour somewhat undefined. I think it's not something we actively document as valid (i.e. what should happen id preset A has a field as `moreFields` while referencing fields of preset B which has it as a regular field). AFICT, the particular code here is meant to just filtering out potentially duplicate fields (when defined both in the preset itself and the referenced one), and the situation you describe is merely a side-effect of the current implementation, be it intentional or not. 

We might want to change the logic slightly at a later point, or not.

PS: Do we even have the mentioned case in the data somewhere?

scripts/lib/references.ts
```diff
@@ -70,7 +70,7 @@ export function dereferenceUntranslatedContent(presets: AllPresets, fields: AllF
             return true;
           }
 
-          // replace the reference with every field. decrement i to reprocess this array index.
+          // replace the reference with what parent defined. decrement i to reprocess this array index.
```
tyrasd: _parent_ is maybe not the right word, as one can reference any other preset, not just a parent of a sub-preset?

```suggestion
          // replace the reference with every field defined in the referenced preset. decrement i to reprocess this array index.
```

matkoniecz: I agree with parent comment

But what I wanted to avoid here is ambiguity of `field`

because what can be happening here is that we copy in entries in `fields` ("every field" in `fields`) or in `moreFields`  ("every field" in `moreFields`) 

also, we claim "every field" and then we actually filter fields using `shouldInherit` - and whole point of complicated processing here is that not every field is inherited

tyrasd: yeah, true, I already forgot the PR title. :sweat_smile: 

```suggestion
          // replace the reference with what the referenced preset defined. decrement i to reprocess this array index.
```

scripts/lib/references.ts
matkoniecz: either way I would leave it for separate PR

matkoniecz: hmm, it is in the same area but kind of separate and may be controversial - I will leave it for a separate PR

matkoniecz: I will look, this can be (and IMHO should be) changed as a separate PR

will add to my TODO pile

tyrasd (approved):

PR #2864, opened 2026-09-11
Try to give hints about how to start creating a new preset

matkoniecz:
may be obvious, but I think it is worth including


tyrasd (approved):
I would have hoped this is obvious and would not need to be said explicitly, but it does also not hurt to be a little verbose, I guess.

PR #2866, opened 2026-09-11
Replace (deprecated) `substr` with `substring` method

matkoniecz:
in this case it should be an exact equivalent

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

still, I probably still should test it

### DIFF
--- a/scripts/lib/build.ts
+++ b/scripts/lib/build.ts
@@ -1083,8 +1083,8 @@ function translationsToYAML(translations: SourceStrings) {
     if (a === '#' + b) return -1;
     if (b === '#' + a) return 1;
     if (a[0] !== b[0]) {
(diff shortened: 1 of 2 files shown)


tyrasd:
There is one more `substr()` in `translations.ts`. Can you also include that one, please?

matkoniecz:
good catch, added now

not entirely sure is this one is simple enough to skip testing, probably should be tested anyway

tyrasd (approved):
> simple enough to skip testing

IMHO it is, but FWIW:

```js
"1234".substr(2) === "34"
"1234".substring(2) === "34" 
```

PR #2488, opened 2026-07-14
Remove `replacement` from the hidden ferry terminal preset

Geo-2695:
### Description, Motivation & Context

When mappers attempt to add ferry terminals without the `public_transport=station` tag in the future, they will no longer receive prompts indicating incomplete tagging.

### Related issues

Close #2183

PR #2878, opened 2026-09-12
Remove suggestion to add `public_transport=station` to all `railway=station`

Geo-2695:
stop misleading railway enthusiasts editing the map. For example, in China, more than two-thirds of train stations have no public passenger service.
This means over 3500 train stations will throw errors.
Similar to #2309 #2488 #2823

### DIFF
--- a/data/presets/railway/_station.json
+++ b/data/presets/railway/_station.json
@@ -1,7 +1,11 @@
 {
     "icon": "temaki-train",
     "fields": [
-        "{public_transport/station}"
+        "name",
+        "network",
+        "operator",
+        "address",
+        "building_area"
     ],
     "moreFields": [
         "{public_transport/station}"
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -1357,10 +1357,6 @@
     "replace": {"railway": "platform", "public_transport": "platform"}
   },
   {
-    "old": {"railway": "station"},
-    "replace": {"railway": "station", "public_transport": "station"}
-  },
-  {
     "old": {"railway": "tram_stop"},
     "replace": {"railway": "tram_stop", "public_transport": "stop_position", "tram": "yes"}
   },


Geo-2695:
### **Test**
STOP

data/presets/railway/_station.json
```diff
@@ -1,7 +1,11 @@
 {
     "icon": "temaki-train",
     "fields": [
-        "{public_transport/station}"
+        "name",
```
matkoniecz, line 4: I would not remove this inheritance, I expect that many will be tagged just as `railway=station` wile providing some public transport service

and practical effect of this change are minuscule right now anyway

Geo-2695: I originally wanted to remove this field so that when a mapper removes the `public_transport` tag, they can see that the fields have changed. This is already in the `moreField` property for compatibility.
I guess mappers will still primarily select the visible public_transport preset, unless they manually remove the tag

matkoniecz: I am more thinking about cases like https://www.openstreetmap.org/node/555381643 that are public transport stations and just not tagged yet with so called ptv2 tags

Geo-2695: I’m mainly concerned that some users might mindlessly apply suggested edits, adding wrong tags to these non‑passenger stations and messing up station tagging data. As far as I know, stations of this kind greatly outnumber passenger‑serving ones. Take the examples below: 

[https://www.openstreetmap.org/node/9178344027](https://www.openstreetmap.org/node/9178344027)
[https://www.openstreetmap.org/node/9178344026](https://www.openstreetmap.org/node/9178344026)
[https://www.openstreetmap.org/node/2349464753](https://www.openstreetmap.org/node/2349464753)
[https://www.openstreetmap.org/node/9169258495](https://www.openstreetmap.org/node/9169258495)
[https://www.openstreetmap.org/node/6659058858](https://www.openstreetmap.org/node/6659058858)
[https://www.openstreetmap.org/node/9178344025](https://www.openstreetmap.org/node/9178344025)
[https://www.openstreetmap.org/node/9178344024](https://www.openstreetmap.org/node/9178344024)
[https://www.openstreetmap.org/node/2349495350](https://www.openstreetmap.org/node/2349495350)

there are as many as 8 non‑passenger stations between two cities that are not far apart. I believe incorrectly adding the `public_transport` tag to non‑passenger train stations is a far bigger issue than passenger stations missing this tag.

matkoniecz: > I’m mainly concerned that some users might mindlessly apply suggested edits

I agree with that part

that is why I am suggesting to amend this PR, rather than proposing rejecting it altogether

the proposed change is to keep the same field set while still removing "upgrade"

(I suspect that every single "upgrade" in iD presets should be removed)

matkoniecz (approved):

matkoniecz:
Thanks!

PR #2901, opened 2026-09-16
Support full-width comma in translation terms

### LINKED ISSUE #2898: consider the full-width comma (，) when splitting terms/aliases
see https://github.com/openstreetmap/id-tagging-schema/pull/2852#issuecomment-5654508754

Someone who's writing `zh` translations with the pinyin keyboard layout will end up inserting [`，` (full-width comma)](https://www.compart.com/en/unicode/U+FF0C) when pressing <kbd>,</kbd> on their keyboard.  As a result, there are some broken terms: 

https://github.com/openstreetmap/id-tagging-schema/blob/2f3f3b9daab7e1c8298d4040f3c6d24bbc5a708e/dist/translations/zh-CN.json#L24378

I'm pretty sure this is safe. Asked a chinese-speaking friend who agreed. 

---

CJK languages also have [a second comma (`、`)](https://www.compart.com/en/unicode/U+3001) which is used for lists. We should not split on this comma, because it would break some translations: 

https://github.com/openstreetmap/id-tagging-schema/blob/2f3f3b9daab7e1c8298d4040f3c6d24bbc5a708e/dist/translations/zh.json#L1990

SaifuddinM23:
### Description

Adds support for the full-width comma (`，`) when splitting translation terms.

### Related issues

Closes #2898

### Testing

- `npx vitest run scripts/lib/__tests__/translations.test.js`
- `npm run lint`
- Verified that the ideographic comma (`、`) is not treated as a separator.


matkoniecz (changes requested):
if you use LLM you must clearly mention it at the beginning of a submitted PR

matkoniecz:
closing as undisclosed LLM-generated PR

undisclosed LLM-generated PR are obnoxious, do not do them at all

SaifuddinM23:
> closing as undisclosed LLM-generated PR
> 
> undisclosed LLM-generated PR are obnoxious, do not do them at all

 Thanks for pointing this out. I apologize for not disclosing my LLM use in the PR. I understand the requirement and will follow it going forward.

If possible, could I submit a new PR with the proper disclosure?

matkoniecz:
yes, it is would be fine

at least for now LLM are not banned outright

just ensure that human looked through all submitted content, understand it and it is rewritten where needed

(I have slight doubts do we need tests here, though maybe it is worth doing? And documenting "does not split field terms by ideographic comma" has value... But whoever submits PR should check that and think a bit on it, situation where maintainer is first human looking at generated code is not sustainable)

SaifuddinM23:
> yes, it is would be fine
> 
> at least for now LLM are not banned outright
> 
> just ensure that human looked through all submitted content, understand it and it is rewritten where needed
> 
> (I have slight doubts do we need tests here, though maybe it is worth doing? And documenting "does not split field terms by ideographic comma" has value... But whoever submits PR should check that and think a bit on it, situation where maintainer is first human looking at generated code is not sustainable)

Resubmitted as #2904 with the required LLM disclosure. I also reviewed the implementation and tests, checked the  code paths, and made sure I understood the changes which I had submitted. Thanks for the feedback.

PR #2903, opened 2026-09-16
fix `*:both:*:both` keys appearing in the taginfo project file

### LINKED ISSUE #2900: parking:both:orientation:both listed as supported on taginfo
**Describe the bug**

https://taginfo.openstreetmap.org/keys/parking%3Aboth%3Aorientation:both is listing it as supported

**To Reproduce**

visit https://taginfo.openstreetmap.org/keys/parking%3Aboth%3Aorientation:both

**Expected behavior**

it is not listed as supported and it is not supported

**Screenshots**

**Additional context**

likely it is bug specifically of taginfo tag lister

see wiki at https://wiki.openstreetmap.org/wiki/Street_parking#Parking_orientation

related code is at https://github.com/openstreetmap/id-tagging-schema/blob/2f3f3b9daab7e1c8298d4040f3c6d24bbc5a708e/data/fields/parking/side/orientation.json#L2

k-yle:
closes #2900 

now it matches the behaviour used by iD itself: 

https://github.com/openstreetmap/iD/blame/f4cad2b7b84097330a9d00ac9263bd444951af29/modules/ui/fields/directional_combo.js#L135-L137


dist/taginfo.json
```diff
@@ -22250,32 +22250,32 @@
             "description": "🄵🅅 Parking Orientation: Meets the Street at a Straight Angle (~90°)"
         },
         {
-            "key": "parking:both:orientation",
+            "key": "parking:orientation",
```
matkoniecz, line 22253: from my testing this tag is in fact not supported in iD

```
highway=tertiary
parking:left=street_side
parking:orientation=parallel
parking:right=lane
```

vs

```
highway=tertiary
parking:left=street_side
parking:both:orientation=parallel
parking:right=lane
```

is my testing bad? iD buggy? Something else?

matkoniecz (changes requested):

dist/taginfo.json
k-yle: image

it does work, but if the field is defined in `moreFields`, then it won't be shown by default. which is an unrelated bug on the iD side

matkoniecz: > which is an unrelated bug on the iD side

is it reported already on iD side?

matkoniecz (approved):

dist/taginfo.json
k-yle: don't think so

matkoniecz: thanks, I will try to figure out what this bug is exactly about and report it

matkoniecz:
And thanks again for one more fix!

dist/taginfo.json
matkoniecz: https://github.com/openstreetmap/iD/pull/12906 seems to be a fix

PR #2904, opened 2026-09-16
Support full-width comma in translation terms

### LINKED ISSUE #2898: consider the full-width comma (，) when splitting terms/aliases

SaifuddinM23:
### LLM disclosure

This contribution was developed with assistance from ChatGPT. I reviewed and understand the submitted change.

### Description

Adds support for the full-width comma (`，`) when splitting translation terms.
The existing separators `,` and `،` are preserved, while the ideographic comma (`、`) is not treated as a separator.

### Related issue
Closes #2898
### Testing

- `npm test`
- `npm run lint`
- `git diff --check`


SaifuddinM23:
@matkoniecz I have made the changes as you suggested . Can you have a look on it ?

scripts/lib/translations.ts
```diff
@@ -6,7 +6,7 @@ import { transifexApi, type Collection } from '@transifex/api';
 import { dereferencedTranslatableContent } from './references.ts';
 import type { Options, References, ResourceInfo, SourceStrings, TStrings } from './types.def.ts';
 
-const COMMA_REGEX = /[,،]/
+const COMMA_REGEX = /[,،，]/
```
tordans, line 9: In my code I would build this array from `const`s where each explains this char. Those are hard do understand without context. Could ref the issue here too as comment on the const. (Just the id ad done in the iD codebase often)

matkoniecz: maybe worth doing as a followup PR?

matkoniecz (approved):

PR #2913, opened 2026-09-18
Add fields to the hidden `railway=halt` preset and unify its geometry

Geo-2695:
When tagging some service halts, they do not have `public_transport` tag, which causes the relevant fields to fail to display properly (especially `operator`). This issue has been fixed in this change

~need to wait for https://github.com/openstreetmap/iD/pull/12900 merged~


Geo-2695:
### Test
| Before | After |
| ----- | ----- |
|haltori|ha|

data/presets/railway/_halt.json
```diff
@@ -5,7 +5,10 @@
         "vertex"
     ],
     "fields": [
-        "{railway}"
+        "name",
```
matkoniecz, line 8: the same as elsewhere - why remove inheritance?

Geo-2695: This field is unnecessary here. It is essentially a field specific to the `Railway Feature` preset, used to define the value of the `railway` key (`railway=xxx`). Since its value here is fixed to `halt`, it is not needed.

Geo-2695: See
https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/_railway.json

https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/railway.json

matkoniecz: actually, this field would be skipped so nothing would be inherited in effect

matkoniecz: though in this case inheriting fields from public_transport/station_train_halt.json would likely make sense, right? Rather than defining separate ones?

PR #2914, opened 2026-09-19
Use more fiiting icon for river rapids ride (from National Park Service icon set, via Pinhead)

Geo-2695:
The original icon looks like a ferry


Geo-2695:
yu

matkoniecz (approved):

matkoniecz:
> looks like a ferry

given its maki-ferry code it seems appropriate :)

matkoniecz:
Thanks for finding this better icon!

Issue #1511, opened 2024-03-27
Preset and rendering for boundary=protected_area


road-runner-osm:
### Description

boundary=protected_area was introduced to indicate nature conservation areas. It is rendered on Mapnik, but in iD (as in JOSM) it is not defined or displayed by a preset. 

The relationship between boundary=protected_area and leisure_nature=reserve is unclear and is also not clear when reading the wiki pages. In practice, both are often used together. Unfortunately the explanations differ also depending on the language of the wiki page. See also [https://wiki.openstreetmap.org/wiki/Talk:Tag:leisure%3Dnature_reserve#leisure=nature_reserve_vs._boundary=protected_area](https://wiki.openstreetmap.org/wiki/Talk:Tag:leisure%3Dnature_reserve#leisure=nature_reserve_vs._boundary=protected_area)

However, if a preset would be created, boundary=protected_area could be better used (for protected areas, as intended). In iD it could be named "Conservation Area" or "Protected Area".

Issue #12, opened 2020-11-10
missing boundary=protected_area in area and relation


abdeldjalil09:
https://wiki.openstreetmap.org/wiki/Tag%3Aboundary%3Dprotected_area
https://wiki.openstreetmap.org/wiki/Key:protect_class
https://wiki.openstreetmap.org/wiki/Key:protection_title

PR #2917, opened 2026-09-19
Also add `boundary=protected_area` for areas formed by closed lines

fibonacci-matrix:

According to the wiki, protected_area can be used for both `area `and `relation`. Currently, there are **83,161** such lines in the database.

### Related issues

https://github.com/openstreetmap/id-tagging-schema/issues/12
https://github.com/openstreetmap/id-tagging-schema/issues/1511

### Links and data

**Relevant OSM Wiki links:**
- https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dprotected_area

**Relevant tag usage stats:**
> https://taginfo.openstreetmap.org/tags/boundary=protected_area

### Wording

- [x] American English


matkoniecz:
before/after on some area would be useful

fibonacci-matrix:
## Test-Documentation

https://pr-2917--ideditor-presets-preview.netlify.app/id/dist/#locale=en&id=w458905870&map=13.62/42.8157/12.1410

Before



After



### Search










### Info-`i`

matkoniecz (approved):

matkoniecz:
As multipolygons are also areas

matkoniecz (changes requested):
Wait, is it impossible to simply add an extra geometry value to the existing preset?

Kovoschiz:
`boundary=` on closed members is frowned upon. It confuses applications from the archaic method.

fibonacci-matrix:
> `boundary=` on closed members is frowned upon. It confuses applications from the archaic method.

The wiki declares this as a valid tagging method, and there are also `83K `existing uses of it.

> Wait, is it impossible to simply add an extra geometry value to the existing preset?

The relation template contains an additional `type=boundary` tag, which should not be added to a regular "area". Something similar is already used for presets for `route=ferry`.

https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/route/ferry.json
https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/type/route/ferry.json

At the moment, my only concern is the ambiguity that we would be able to add `boundary=protected_area` with both `type=boundary` and `type=multipolygon`, and I'm not sure how OK that is. Taginfo shows that both types are popular: 34K for `type=boundary` and 13K for `type=multipolygon`.

https://taginfo.openstreetmap.org/tags/boundary=protected_area#combinations

To be completely honest, I don't like the existence of a separate `type=boundary` at all. It seems to be needed mainly to allow additional roles in a multipolygon. This is widely used only with administrative and religious boundaries, and is hardly ever used for protected areas.

matkoniecz:
> It confuses applications from the archaic method.

Not sure what you mean here

> is frowned upon

If it is really unwanted global or regional bot edit may be run to retag it

But iD tagging schema is not supposed to unilaterally deprecate stuff

k-yle (changes requested):
we definitely need a solution to this problem. But I'm not a fan of this approach. 

- adding `"geometry": ["area"]` will cause iD to encourage mappers to use `type=multipolygon` + `boundary=protected_area`, which is wrong. The proper solution is unfortuantely blocked by #1538
- it breaks iD's existing rendering of protected_area

This needs to be implemented on the iD side. #1538 provides the data for iD to determine whether the split, merge, and create operations should use `type=multipolygon` or `type=boundary`

Kovoschiz:
@fibonacci-matrix What are you referring to? The area symbol in the template box includes `=multipolygon` and `=boundary` , not specific to closed lines. "Avoid using the tag on closed ways, or repeating it on boundary relation members, as this can lead to ambiguities and cause unexpected behavior in data consumers.[1]" https://wiki.openstreetmap.org/wiki/Key:boundary https://community.openstreetmap.org/t/boundary-tagging-on-boundary-relation-member-ways/111345/7

